PR #19, opened 2014-05-15
Adding benchmarks LoginUserTest and LoginLogoutUserTest

alexkli:
Adding benchmarks LoginUserTest and LoginLogoutUserTest that use a normal user instead of the special admin user the existing LoginTest and LoginLogoutTest are using, and which often has shortcut implementations since no groups etc. have to be resolved for him.

I kept the existing LoginTest and LoginLogoutTest as is using admin, to ensure comparisons with older benchmark runs still hold true.


tripodsan:
I merged in this pull request. you can close it now.

alexkli:
Merci!

PR #54, opened 2016-08-24
OAK-4696 log sync result and commit only in case of result != NOP

kwin:


oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/ExternalLoginModule.java
```diff
@@ -351,11 +352,13 @@ private void syncUser(@Nonnull ExternalUser user) throws SyncException {
             try {
                 DebugTimer timer = new DebugTimer();
                 context = syncHandler.createContext(idp, userManager, new ValueFactoryImpl(root, NamePathMapper.DEFAULT));
-                context.sync(user);
+                SyncResult syncResult =  context.sync(user);
                 timer.mark("sync");
-                root.commit();
-                timer.mark("commit");
-                debug("syncUser({}) {}", user.getId(), timer.getString());
+                if (syncResult.getStatus() != SyncResult.Status.NOP) {
```
anchela, line 357: i think it would be better to have this replace by
                if (root.hasPendingChanges())
because there might be other sync-result status that are expected to not perform any changes (e.g. FOREIGN, NO_SUCH_AUTHORIZABLE, NO_SUCH_IDENTITY, MISSING). also, it would feel safer to me to make sure all changes get committed irrespective of the status code. in particular as the sync-context is not guaranteed to be the one shipped with Oak. that may sound overly defensive but from my experience that would be the kind of issues that would be really hard to track... rather be on the safe side.

oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/ExternalLoginModule.java
```diff
@@ -351,11 +352,13 @@ private void syncUser(@Nonnull ExternalUser user) throws SyncException {
...
+                    root.commit();
+                    timer.mark("commit");
+                }
+                debug("syncUser({}) {}, result: {}", user.getId(), timer.getString(), syncResult.getStatus().toString());
```
anchela, line 361: i'd rather use "status" than "result" here

kwin:
This has been applied with slight modifications.

PR #55, opened 2016-09-30
OAK-4871 -  Multiplexing NodeStore

rombert:
For a variety of use cases it is useful to multiplex multiple node stores behind a single interface.

Together with @trekawek  I have been working on a limited-purpose MultiplexingNodeStore which aims to multiplex multiple stores with the limitation that only one store may be written to, while others are read-only.

The implementation has been validated to be functional, but not yet tested for behaviour under heavy load, performance, etc.

I believe that we can still incorporate this early to make maintenance simpler and ensure that the high-level design is correct.


rombert:
This is already part of Oak under the "CompositeNodeStore" name - see https://jackrabbit.apache.org/oak/docs/nodestore/compositens.html . Closing.

PR #58, opened 2017-03-09
Oak 4933: Add support for AzureDataStore

rhudea:
Adds support for a new DataStore based on Azure Blob Storage.


oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
```diff
@@ -0,0 +1,784 @@
...
+
+            blob.downloadAttributes();
+            if (blob.getProperties().getLength() != len) {
+                throw new DataStoreException("Length Collision. identifier=" + key +
+                                             " new length=" + len +
+                                             " old length=" + blob.getProperties().getLength());
+            }
+            LOG.trace("Blob already exists. identifier={} lastModified={}", key, blob.getProperties().getLastModified().getTime());
+            blob.startCopy(blob);
+            //TODO: better way of updating lastModified (use custom metadata?)
```
mattvryan, line 219: I'm not sure this is needed.  The blob store API in Oak 1.4 exposed a method to "touch" a file, or in other words, update the last modified time of the file.  IIRC this API changed in Oak 1.6 and so this functionality is no longer required in order to complete the API.

Is my recollection of the API update in Oak 1.6 correct?  Is there another reason why we would need a way to modify the lastModified time manually?  I know Azure maintains a read-only property on each blob for the lastModified time, so if we only need to be able to read the lastModified time that property should be sufficient and we shouldn't need to jump through hoops to update it ourselves.

If all my assumptions and recollections are accurate, I'd argue we don't need to worry about updating the lastModified time at all.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
```diff
@@ -0,0 +1,784 @@
...
+            CloudBlockBlob blob = getAzureContainer().getBlockBlobReference(key);
+            if (blob.exists()) {
+                blob.downloadAttributes();
+                AzureBlobStoreDataRecord record = new AzureBlobStoreDataRecord(
+                    this,
+                    connectionString,
+                    containerName,
+                    new DataIdentifier(getIdentifierName(blob.getName())),
+                    blob.getProperties().getLastModified().getTime(),
+                    blob.getProperties().getLength());
```
mattvryan, line 293: In this implementation the blob object is being retrieved from Azure on line 284 above, but then the blob itself is not passed into the constructor of the data record; instead, the blob object must again be retrieved from Azure in the getStream() method of the data record object.

I just want to make sure this is intentional and that the performance of this implementation was tested, since potentially two calls will be made to retrieve the block blob reference.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
```diff
@@ -0,0 +1,784 @@
...
+        if (null == identifier) throw new NullPointerException("identifier");
+
+        String key = getKeyName(identifier);
+        long start = System.currentTimeMillis();
+        ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
+        try {
+            Thread.currentThread().setContextClassLoader(getClass().getClassLoader());
+
+            boolean result = getAzureContainer().getBlockBlobReference(key).deleteIfExists();
+            LOG.debug("Blob deleted. existed={} identifier={} duration={}", result, key, (System.currentTimeMillis() - start));
```
mattvryan: Suggest perhaps two different log messages here, depending on the value of "result".  If the blob being referenced does not exist, then deleteIfExists() will return a value of false, and then the log message wording of "Blob deleted. existed=false" which would be logged, and I think that is a confusing message for someone not familiar with the Azure API.

I'd suggest keep the same message if result is true, and if result is false, log instead "Blob delete requested but blob does not exist (perhaps already deleted)" along with the diagnostic info like the current log message has, or something like that.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
```diff
@@ -0,0 +1,784 @@
...
+        long start = System.currentTimeMillis();
+        final List<DataRecord> records = Lists.newArrayList();
+        ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
+        try {
+            Thread.currentThread().setContextClassLoader(getClass().getClassLoader());
+
+            CloudBlobDirectory metaDir = getAzureContainer().getDirectoryReference(META_DIR_NAME);
+            for (ListBlobItem item : metaDir.listBlobs(prefix)) {
+                if (item instanceof CloudBlob) {
+                    CloudBlob blob = (CloudBlob) item;
```
mattvryan, line 521: In getMetadataRecord() the code checks that the blob exists before moving on, as it should.  I'm wondering, do we need the same check here?  Would listBlobs() return items that have a reference but don't actually exist?  Was this use case tested?

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
```diff
@@ -0,0 +1,784 @@
...
+    @Override
+    public boolean deleteMetadataRecord(String name) {
+        long start = System.currentTimeMillis();
+        ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
+        try {
+            Thread.currentThread().setContextClassLoader(getClass().getClassLoader());
+
+            CloudBlockBlob blob = getAzureContainer().getBlockBlobReference(addMetaKeyPrefix(name));
+            boolean result = blob.deleteIfExists();
+            LOG.debug("Metadata record deleted. existed={} metadataName={} duration={}", result, name, (System.currentTimeMillis() - start));
```
mattvryan: Same comment here as for deleteRecord() earlier, suggesting that if deleteIfExists() returns false this message is confusing to those unfamiliar with the workings of the Azure API, and perhaps two messages would be better.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
```diff
@@ -0,0 +1,784 @@
...
+        ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
+        try {
+            Thread.currentThread().setContextClassLoader(getClass().getClassLoader());
+
+            CloudBlobDirectory metaDir = getAzureContainer().getDirectoryReference(META_DIR_NAME);
+            int total = 0;
+            for (ListBlobItem item : metaDir.listBlobs(prefix)) {
+                if (item instanceof CloudBlob) {
+                    ((CloudBlob)item).deleteIfExists();
+                    total ++;
```
mattvryan: The variable "total" here is actually a count of all CloudBlob objects enumerated, and not a count of records deleted.  This is true because the deleteIfExists() call is used.  So if the current item does not exist, nothing will be deleted on that iteration through the loop, but the value of "total" will be incremented anyway.

Once this loop completes, a log message will be displayed indicating the number of records deleted, and the value of "total" is used to display that number.  But it is possible that the number may indicate more records were actually deleted if one or more blob items in the directory were deleted before but a reference still existed.

I don't know if that can actually happen or not; it depends on whether listBlobs() will return items that are actually deleted.  This was discussed in an earlier comment.

If it is possible that deleteIfExists() could return false, we should test the result of deleteIfExists() and only increment total if the result is true.  If this is not a possible case, a comment explaining this would help.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStore.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.apache.jackrabbit.oak.plugins.blob.AbstractSharedCachingDataStore;
+import org.apache.jackrabbit.oak.spi.blob.AbstractSharedBackend;
+import org.apache.jackrabbit.oak.spi.blob.SharedBackend;
+
+import java.util.Properties;
+
+public class AzureDataStore extends AbstractSharedCachingDataStore {
+
+    private int minRecordLength = 16*1024;
```
mattvryan, line 30: Should this be a constant value in the parent class, or perhaps an overridable getter in the parent class?  I think this is the same value used by S3DataStore.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/Utils.java
```diff
@@ -0,0 +1,143 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.InetSocketAddress;
+import java.net.Proxy;
+import java.net.SocketAddress;
+import java.net.URISyntaxException;
+import java.security.InvalidKeyException;
+import java.util.Properties;
+
+public final class Utils {
```
mattvryan, line 45: Isn't there a similar class to this for S3DataStore?  I'm wondering how much alike they are, and if some commonalities should be combined into a single parent class or interface.

mattvryan (commented):
Some ideas for consideration.

oak-blob-cloud-azure/src/test/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreTest.java
```diff
@@ -0,0 +1,710 @@
...
+        int ctr = 0;
+        while (iter.hasNext()) { iter.next(); ++ctr; }
+        return ctr;
+    }
+
+    @Test
+    public void testCreateAndDeleteBlobHappyPath() throws DataStoreException, IOException {
+        final DataRecord uploadedRecord = ds.addRecord(new ByteArrayInputStream(testBuffer));
+        DataIdentifier identifier = uploadedRecord.getIdentifier();
+        //DataIdentifier identifier = new DataIdentifier("a94a8fe5ccb19ba61c4c0873d391e987982fbbd3");
```
mattvryan: This comment was kept in my initial prototype for local testing purposes.  It should be deleted now.

mattvryan (commented):
One comment to delete.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
rhudea: It is intentional since the thread safety of the Azure APIs is not guaranteed (https://github.com/Azure/azure-storage-java/issues/38) . And since the fetching of all dataRecords can be made on a different thread than the one reading one, for now, we're playing the safe card.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
rhudea: The blob already contains all the needed attributes (lastModified, length), so it should exists at the moment of the listBlobs call. However, between this call and getMetadataRecord call, it might be gone.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
rhudea: Since it might be possible to have total != listBlobs.length (since calls are async), the code was update to increase total only if delete succeeds.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
rhudea: Message updated

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/Utils.java
rhudea: Not that much (just 2 functions: readConfig and deleteIfPossible). I'm thinking that this should be better handled as a different issue, which should move the S3 implementation out of  oak-blob-cloud to oak-blob-cloud-aws (as you previously suggested).

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStore.java
rhudea: The same implementation is done in others DataStores as well. This change should be done in a different issue, as this one already bigger that I would have wanted to be.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
rhudea: The code is similar to the new S3Backend, which I've assumed does the right thing. I'll have to follow up on this one since I'm not sure whether lastModified still needs to be kept updated .

mattvryan: I think it is worth checking into.

I'm not overly familiar with the history of S3Backend, but it was written prior to the API change for Oak 1.6, so at some point in time it needed to handle the ability to update the lastModified time.  If S3Backend is still doing this it might simply be because the code wasn't changed yet, and not because copying the blob is still required to hack the update of the lastModified time.

amit-jain:
I just had a quick look at the pull request and my first reaction is that oak-blob-cloud module is meant to be a generic module for all cloud datastores and it would make sense to put the code in there instead of creating a separate module oak-blob-cloud-azure.  Then we could reuse or create some abstractions that could work for both and in future if we need support for jclouds that can also be added there.

rhudea:
From some offline discussion, I've got the impression that is the agreed idea, that we would have separate modules that can be released / patched independently for both Azure/S3/other and the oak-blob-cloud would eventually be a module where we could put the abstraction for all other cloud modules.

It would look like:
- oak-blob-cloud – abstraction/common classes 
- oak-blob-cloud-aws – AWS specific classes
- oak-blob-cloud-azure – Azure specific classes

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
```diff
@@ -0,0 +1,784 @@
...
+            LOG.debug("Error writing blob. identifier={}", key, e);
+            throw new DataStoreException(String.format("Cannot copy blob. identifier=%s", key), e);
+        } finally {
+            if (null != contextClassLoader) {
+                Thread.currentThread().setContextClassLoader(contextClassLoader);
+            }
+        }
+    }
+
+    public static boolean waitForCopy(CloudBlob blob) throws StorageException, InterruptedException {
```
amit-jain: Shouldn't this method should be private?

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/Utils.java
```diff
@@ -0,0 +1,143 @@
...
+
+    public static final String DEFAULT_CONFIG_FILE = "azure.properties";
+
+    private static final String DELETE_CONFIG_SUFFIX = ";burn";
```
amit-jain: I think this can be removed (the S3 code has some legacy code which is not used).

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureConstants.java
```diff
@@ -0,0 +1,53 @@
...
+     */
+    public static final String PROXY_HOST = "proxyHost";
+
+    /**
+     *  Constant to set proxy port.
+     */
+    public static final String PROXY_PORT = "proxyPort";
+
+    private AzureConstants() { }
+}
```
amit-jain, line 64: It would be better to use the same user facing constants as used in S3 here and add a translation layer to convert back to Azure specific operations. This way we don't have to change if we unify then in future and keeps change to a minimum.

oak-blob-cloud-azure/pom.xml
```diff
@@ -0,0 +1,174 @@
...
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <configuration>
+                    <instructions>
+                        <Export-Package>
+                            org.apache.jackrabbit.oak.blob.cloud.azure.blobstorage
+                        </Export-Package>
+                        <DynamicImport-Package>sun.io</DynamicImport-Package>
+                        <Embed-Dependency>
+                            azure-storage,
```
amit-jain, line 49: Any reason the dependency is embedded? Though I think the license is Apache 2.0, not sure if its a good idea to embed it here. S3 sdk jars are not embedded here. What is the size of the embeds?

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
```diff
@@ -0,0 +1,784 @@
...
+            throw new NullPointerException("prefix");
+        }
+        long start = System.currentTimeMillis();
+        final List<DataRecord> records = Lists.newArrayList();
+        ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
+        try {
+            Thread.currentThread().setContextClassLoader(getClass().getClassLoader());
+
+            CloudBlobDirectory metaDir = getAzureContainer().getDirectoryReference(META_DIR_NAME);
+            for (ListBlobItem item : metaDir.listBlobs(prefix)) {
```
amit-jain, line 519: What is the default for number of items returned in 1 call? Or does it return all?
In S3 its 1000 so it is safe to just have 1 call but if the default if low then we would need to use RecordsIterator to create the list.

oak-blob-cloud-azure/src/test/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureCachingDataStoreTest.java
```diff
@@ -0,0 +1,116 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.Properties;
+import java.util.Random;
+
+/**
+ * Test {@link AbstractSharedCachingDataStore} with AzureDataStore.
+ * It requires to pass aws config file via system property or system properties by prefixing with 'ds.'.
```
amit-jain: minor doc change required (is aws config)

oak-blob-cloud-azure/src/test/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreTest.java
```diff
@@ -0,0 +1,710 @@
...
+            IOUtils.closeQuietly(in);
+        }
+        return encodeHexString(digest.digest());
+    }
+
+    private static int countIteratorEntries(final Iterator<?> iter) {
+        int ctr = 0;
+        while (iter.hasNext()) { iter.next(); ++ctr; }
+        return ctr;
+    }
```
amit-jain: Guava's Iterators.size can be used instead.

oak-blob-cloud-azure/src/test/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreTest.java
```diff
@@ -0,0 +1,710 @@
...
+import java.util.Map;
+import java.util.Properties;
+import java.util.Random;
+import java.util.Set;
+
+import javax.jcr.RepositoryException;
+
+/**
+ * Test {@link AzureDataStore} with AzureDataStore and local cache on.
+ * It requires to pass aws config file via system property or system properties by prefixing with 'ds.'.
```
amit-jain, line 75: Minor doc change to change from aws

oak-blob-cloud-azure/src/test/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureCachingDataStoreTest.java
```diff
@@ -0,0 +1,116 @@
+/*
```
amit-jain: The super class adds methods for local testing. If the kind of tests that are covered there are covered in other areas it might be good to have it remain as a local test which will add some coverage for the AzureDatastore class.

oak-blob-cloud-azure/src/test/resources/azure.properties
```diff
@@ -0,0 +1,38 @@
...
+
+# Concurrent number of simultaneous requests per operation (default 1)
+concurrentRequestsPerOperation=3
+
+# Number of retries per request (default 3)
+maxRequestRetry=10
+
+# proxy configurations (optional)
+proxyHost=
+proxyPort=
```
amit-jain, line 40: See my comment for constants. It will be good where ever we can to use the same user specific property

oak-blob-cloud-azure/pom.xml
amit-jain: We should obviously check with the community, but if they are not too big then it might be ok.

rhudea: The azure libraries are not available as OSGI bundle. The size of the jar is ~700KB, meaning that the total size of the package would be around ~700KB, as opposed to the S3 one which is ~30MB due to the big AWS OSGI jar bundle.

Eventually, once azure libraries are available as OSGI bundles, we're going to change this.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
rhudea: Azure has a 5000 default limit in this case.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureConstants.java
rhudea: I'm not convinced about this since it might get confusing since they don't overlap in all cases:

s3Bucket -> azureContainer ?
maxConnections -> no equivalent on azure, since in Azure you can specify the number of maximum thread per each request, but not the size of a thread pool.

My logic was to match the properties with the documentation of the cloud, since usually people will look up to that to better understand where this is used and what are their limitations.

I don't think that using AWS terminology in Azure context would be beneficial to users.

amit-jain: Obviously we can't use properties which have a s3 prefixed and properties which don't make sense for Azure.  OTOH it might be easier for users to understand keys which can be mapped easily between the two. Most of the end users would be referring to documentation provided by us. 
What I am trying to do is have a rather minimal change of properties if we add a different blob store or a different implementation. Do you think that is not worth the effort here?

rhudea: I guess we can use the following mappings:

- Same for S3 and Azure:
**accessKey** - used as accessKey for S3 and as accountName in Azure
**secretKey** - used as secretKey for S3 and as accountKey in Azure
**maxErrorRetry** - same for S3 and Azure
**socketTimeout** - socket timeout for S3 and timeout interval for Azure
**proxyHost** - same for S3 and Azure
**proxyPort** - same for S3 and Azure
**maxConnections** - maximum connections to S3 / in Azure: Concurrent number of simultaneous requests per operation

- Only in S3
**s3Bucket** 
rest of other s3 only properties that are not used in Azure

- Only in Azure
**azureContainer**

oak-blob-cloud-azure/pom.xml
amit-jain: 700 KB sounds quite reasonable and we may not need to change this.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureConstants.java
amit-jain: Yes this is what I had in mind and sounds reasonable. We can add documentation in the properties file as well as Oak documentation. 
But maybe also 
* rename azureContainer -> container, wdyt? We can add a mapping in S3 also which understands container so that we have parity.

rhudea: Yep, renamed.

PR #62, opened 2017-06-23
OAK-6388 - Support for Azure SAS that can be used for a finer grained access to …

akalfas:
…the storage container


akalfas:
This is for fixing https://issues.apache.org/jira/browse/OAK-6388

PR #70, opened 2017-09-29
OAK-6575 Provide a secure external URL to a DataStore binary.

ieb:
This pull request should be squashed into a single commit for the Oak 1.6 Branch.
It has been tested against current Sling Trunk.
It contains a patch to expose S3 Binaries via CloudFront Signed URLs.
See https://issues.apache.org/jira/browse/OAK-6575 for more info.


ieb:
This should not be merged at this time.

PR #71, opened 2017-10-26
Updates AbstractDataStoreService for use with data store factories.

mattvryan:
The primary change is to separate out the data store registration
action and to make the function that does this static.  This will
allow the code to be shared with other data store creators, like
factories.  This is similar to the pattern used in
SegmentNodeStoreService and SegmentNodeStoreFactory.

Part of this change includes the addition of a Closer.  Many
items in the service that had been class variables which were
cleaned up in the deactivate() method are now added to the
Closer which is closed in the deactivate method.  Since a Closer
is passed to the now-static registration method, this means
that the resources set up in that method that used to be cleaned
up during deactivate() will now be handled by whatever Closer
is passed to the static method.


mattvryan:
PR updated today to include AbstractDataStoreFactory and FileDataStoreFactory implementation.  This also required the addition of DataStoreProvider in oak-blob.

PR #74, opened 2017-11-03
CompositeDataStoreService

mattvryan:
This PR includes https://github.com/apache/jackrabbit-oak/pull/71 and adds CompositeDataStore and some supporting infrastructure.  This simply takes care of registering the service and adding the DataStoreProvider instances as they are created, if they match the roles used in configuration for the CompositeDataStore.


oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+            DataStoreProvider ds = delegate.getDataStore();
+            delegates.add(ds);
+            return true;
+        }
+        return false;
+    }
+
+    boolean removeDelegate(final DataStoreProvider ds) {
+        return delegates.remove(ds);
+    }
```
amit-jain, line 179: It would make sense to have either the CompositeDataStoreDelegate or DataStoreProvider as parameters for add/removeDelegate methods

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+
+    Iterator<DataStoreProvider> getDelegateIterator() {
+        return delegates.iterator();
+    }
+
+    @Override
+    public DataRecord getRecordIfStored(DataIdentifier identifier) throws DataStoreException {
+        return null;
+    }
```
amit-jain, line 248: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+
+    @Override
+    public DataRecord getRecord(DataIdentifier identifier) throws DataStoreException {
```
amit-jain, line 251: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+        return null;
+    }
+
+    @Override
+    public DataRecord getRecordFromReference(String reference) throws DataStoreException {
```
amit-jain, line 260: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+        return null;
+    }
+
+    @Override
+    public DataRecord addRecord(InputStream stream) throws DataStoreException {
```
amit-jain, line 306: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+
+    }
+
+    @Override
+    public int deleteAllOlderThan(long min) throws DataStoreException {
+        return 0;
+    }
+
+    @Override
+    public Iterator<DataIdentifier> getAllIdentifiers() throws DataStoreException {
```
amit-jain: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+        return null;
+    }
+
+    @Override
+    public void init(String homeDir) throws RepositoryException {
+
+    }
+
+    @Override
+    public int getMinRecordLength() {
```
amit-jain, line 419: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+        return 0;
+    }
+
+    @Override
+    public void close() throws DataStoreException {
```
amit-jain, line 429: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+
+    }
+
+    @Override
+    public void clearInUse() {
+
+    }
+
+    @Override
+    public void deleteRecord(DataIdentifier identifier) throws DataStoreException {
```
amit-jain, line 465: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+
+    }
+
+    @Override
+    public void addMetadataRecord(InputStream stream, String name) throws DataStoreException {
```
amit-jain, line 516: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+
+    }
+
+    @Override
+    public void addMetadataRecord(File f, String name) throws DataStoreException {
```
amit-jain: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+
+    }
+
+    @Override
+    public DataRecord getMetadataRecord(String name) {
```
amit-jain, line 574: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+        return null;
+    }
+
+    @Override
+    public List<DataRecord> getAllMetadataRecords(String prefix) {
```
amit-jain, line 593: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+        return null;
+    }
+
+    @Override
+    public boolean deleteMetadataRecord(String name) {
```
amit-jain, line 606: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+        return false;
+    }
+
+    @Override
+    public void deleteAllMetadataRecords(String prefix) {
```
amit-jain, line 619: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+
+    }
+
+    @Override
+    public Iterator<DataRecord> getAllRecords() throws DataStoreException {
```
amit-jain, line 630: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+        return null;
+    }
+
+    @Override
+    public DataRecord getRecordForId(DataIdentifier id) throws DataStoreException {
```
amit-jain: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+        return null;
+    }
+
+    @Override
+    public Type getType() {
```
amit-jain: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+        return null;
+    }
+
+    @Override
+    public DataRecord addRecord(InputStream input, BlobOptions options) throws DataStoreException {
```
amit-jain: Need to implement

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStoreService.java
```diff
@@ -0,0 +1,180 @@
...
+            }
+        }
+
+        registerCompositeDataStore();
+
+        return dataStore;
+    }
+
+    private void registerCompositeDataStore() {
+        if (delegateDataStores.isEmpty()) {
```
amit-jain: IMO this is too simplistic a condition i.e it should be based on all the roles configured being satisfied with corresponding delegate datastores. Otherwise we could have a situation where for e.g. a read-only DataStore gets configured without the writable DataStore but still the CompositeDataStore is initialized but now the application would not store any blob.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStoreDelegate.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public Map<String, ?> getConfig() {
+        return config;
+    }
+
+    public static CompositeDataStoreDelegateBuilder builder(final DataStoreProvider ds) {
+        return new CompositeDataStoreDelegateBuilder(ds);
+    }
+
+    static class CompositeDataStoreDelegateBuilder {
```
amit-jain: Unless we have complex conditionals to be satisfied and based on that CompositeDataStore (s) constructed we can simplify things here without the builder.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,181 @@
...
+
+    private Properties properties = new Properties();
+    private Set<DataStoreProvider> delegates = Sets.newConcurrentHashSet();
+
+    CompositeDataStore(final Properties properties) {
+        this.properties = properties;
+    }
+
+    boolean addDelegate(final CompositeDataStoreDelegate delegate) {
+        String delegateRole = delegate.getRole();
```
amit-jain: Do these 'roles' have any specific purpose beyond OSGi registration? If yes then that is not clear to me. Otherwise, if its just osgi specific then it would be better to move this logic.

mattvryan:
This change is now also dependent on https://github.com/apache/jackrabbit-oak/pull/80.

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3DataStoreFactory.java
```diff
@@ -0,0 +1,42 @@
...
+@Component(policy = ConfigurationPolicy.REQUIRE,
+        name = "org.apache.jackrabbit.oak.plugins.blob.datastore.S3DataStoreFactory",
+        configurationFactory = true,
+        label = "Apache Jackrabbit Oak S3DataStore Factory",
+        description = "Factory allowing configuration of multiple instances of S3DataStores, " +
+                "for use with CompositeDataStore.")
+public class S3DataStoreFactory extends AbstractDataStoreFactory {
+    @Override
+    protected DataStore createDataStore(ComponentContext context, Map<String, Object> config) {
+        DataStore ds = S3DataStoreService.createS3DataStore(context, config, getStatisticsProvider(), getDescription(), closer);
```
amit-jain, line 39: Important piece missing is the _init()_ of the DataStore. Calling just the createS3DataStore()/createDataStore() and missing parts of the registerDataStore() in the AbstractDataStoreService would fail to initialize the DataStore itself.
Not sure you have tested with the S3DataStore delegate but I don't think that works at present.

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/datastore/AbstractDataStoreService.java
```diff
@@ -60,70 +67,103 @@
...
+
+        Closeables closeables = new Closeables(closer);
+
         boolean encodeLengthInId = PropertiesUtil.toBoolean(config.get(PROP_ENCODE_LENGTH), true);
         int cacheSizeInMB = PropertiesUtil.toInteger(config.get(PROP_CACHE_SIZE), DataStoreBlobStore.DEFAULT_CACHE_SIZE);
 
         String homeDir = lookup(context, PROP_HOME);
         if (config.containsKey(PATH) && !Strings.isNullOrEmpty((String) config.get(PATH))) {
             log.info("Initializing the DataStore with path [{}]", config.get(PATH));
+            homeDir = (String) config.get(PATH);
```
amit-jain: This is wrong and completely changes the location of the datastore folder. This is a byproduct of not properly calling _init()_ on the delegate DataStores

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/datastore/AbstractDataStoreService.java
```diff
@@ -60,70 +67,103 @@
...
         else if (homeDir != null) {
             log.info("Initializing the DataStore with homeDir [{}]", homeDir);
         }
         PropertiesUtil.populate(ds, config, false);
         ds.init(homeDir);
 
-        BlobStoreStats stats = new BlobStoreStats(getStatisticsProvider());
-        this.dataStore = new DataStoreBlobStore(ds, encodeLengthInId, cacheSizeInMB);
-        this.dataStore.setBlobStatsCollector(stats);
+        DataStoreBlobStore dataStore = new DataStoreBlobStore(ds, encodeLengthInId, cacheSizeInMB);
```
amit-jain, line 145: Line 120 onwards till the _init()_ has to be called for any DataStore (even for delegates) for proper initialization.  S3/Azure would not work without it (FDS may have worked because the path directory is not needed beforehand and would be created lazily on first addition of a record/blob).

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/datastore/AbstractDataStoreService.java
```diff
@@ -60,70 +67,103 @@
...
+        closeables.add(dataStore);
+        BlobStoreStats stats = new BlobStoreStats(statisticsProvider);
+        dataStore.setBlobStatsCollector(stats);
+
```
amit-jain, line 149: TODO - We don't have the ability now to track data store stats for each of the delegate which should be added in some form. Either the BlobStoreStats should be enhanced to map stats for each delegate or provide each DataStore the ability to do that.

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/datastore/FileDataStoreFactory.java
```diff
@@ -0,0 +1,50 @@
...
+        name = "org.apache.jackrabbit.oak.plugins.blob.datastore.FileDataStoreFactory",
+        configurationFactory = true,
+        label = "Apache Jackrabbit Oak FileDataStore Factory",
+        description = "Factory allowing configuration of multiple instances of FileDataStores, " +
+                "for use with CompositeDataStore.")
+public class FileDataStoreFactory extends AbstractDataStoreFactory {
+    @Override
+    protected DataStore createDataStore(ComponentContext context, Map<String, Object> config) {
+        DataStore ds = FileDataStoreService.createFileDataStore(context, config, getDescription(), closer);
+        if (ds instanceof FileDataStore && null != config.get("path")) {
```
amit-jain, line 40: Important piece missing is the **init()** of the DataStore. Calling just the createFileDataStore()/createDataStore() and missing parts of the registerDataStore() in the AbstractDataStoreService would fail to initialize the DataStore itself. Also, some important properties won't get injected into the DataStore for e.g. the _minRecordLength_.

Only the _path_ here is insufficient.

amit-jain (commented):
Partial review of the patch with only the initialization pieces reviewed.
Azure changes also required similar to S3.

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3DataStoreFactory.java
mattvryan: During the _init()_ for CompositeDataStore, _init()_ is called for each delegate also (see [CompositeDataStore.java.init()](https://github.com/mattvryan/jackrabbit-oak/blob/9001fc9e30ae36d5d1526012eedee0e917184efb/oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java#L159)).
Is that approach okay, or should I change it so that each factory calls _init()_ instead?

I did test with a S3DataStore delegate and it does work.  But I can change the approach if this one is not correct.  Note that AzureDataStore support hasn't been added yet.

amit-jain: Ok that's why I missed it but what is still not being done is the injection of properties via the PropertyUtils. This is needed for some props for the DataStore including the caching layer.
It would be better to initialize them (properties injection and calling init()) in the service or factory.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/DelegateHandler.java
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * Remove a {@link DelegateDataStore} from this handler.  This means the
+     * specified delegate will no longer be considered a delegate by this delegate
+     * handler.
+     *
+     * @param provider
+     * @return True if this action caused a {@link DataStoreProvider} to be removed,
+     * False if no matching {@link DataStoreProvider} was found.
+     */
+    boolean removeDelegateDataStore(final DataStoreProvider provider);
```
amit-jain: The confusing and somewhat interchangeable usage of DataStoreProvider and the DelegateDataStore should be avoided. I understand that the DataStoreProvider's purpose is in the AbstractDataStoreFactory to register the required DataStores and is required in the blob-plugins bundle.
So I see 2 options:
* Limit the usage of DataStoreProvider to the blob-plugins bundle and only use DelegateDataStore to register in the CompositeDataStore
* Merge the 2 and move the builders etc out of the DelegateDataStore (See the other change suggested in their)

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/DelegateHandler.java
```diff
@@ -0,0 +1,104 @@
...
+     * Get a data store iterator for all writable delegates managed by the
+     * {@link CompositeDataStore} that this handler believes contain a matching record
+     * for the provided identifier.  The iterator may contain no data stores if
+     * the handler believes the record doesn't exist in any writable delegate
+     * data stores.
+     *
+     * @param identifier A {@link DataIdentifier} for filtering the data stores
+     * @return Iterator to writable data stores
+     */
+    Iterator<DataStore> getWritableDelegatesIterator(final DataIdentifier identifier);
```
amit-jain, line 74: Aren't the getWritable* methods tied to the roles the delegates and the specific topology of the CompositeDataStore (Read-write/Read-only)? According to me the methods in the interface should be generic enough to allow other use cases if required. So, methods along the lines of passing a role and then getting back the role specific DataStore iterators would make more sense.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/IntelligentDelegateHandler.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.jackrabbit.core.data.DataStore;
+import org.apache.jackrabbit.oak.spi.blob.DataStoreProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+import java.util.Iterator;
+import java.util.List;
+
+@Component
```
amit-jain: I don't think it is appropriate to use the felix annotations here and inject dependencies. Any dependencies should be injected into the relevant service - CompositeDataStoreService and initialize the object from there. This way it can be used in non-OSGi environments also.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/DelegateMinRecordLengthSelector.java
```diff
@@ -0,0 +1,24 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.jackrabbit.oak.blob.composite;
+
+public interface DelegateMinRecordLengthSelector {
```
amit-jain: Is there a need to have multiple strategies for the minRecordLength? I mean would anything else even make sense.

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/AbstractSharedCachingDataStore.java
```diff
@@ -204,7 +203,13 @@ public DataRecord getRecordIfStored(DataIdentifier dataIdentifier)
                 return new FileCacheDataRecord(this, backend, dataIdentifier, rec.getLength(),
                     rec.getLastModified());
             } catch (Exception e) {
-                LOG.error("Error retrieving record [{}]", dataIdentifier, e);
+                if (this instanceof CompositeDataStoreAware &&
```
amit-jain, line 206: Please remove this check.
This specific abstract class in not an instance of CompositeDataStoreAware and this check seems quite odd to be here. Also, this would mask any errors like unavailability of backends like S3/Azure.

oak-blob/src/main/java/org/apache/jackrabbit/oak/spi/blob/DataStoreReference.java
```diff
@@ -0,0 +1,68 @@
+/*
```
amit-jain: Unused class to be removed.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,675 @@
...
+    static final String ROLES = "roles";
+
+    private Properties properties = new Properties();
+    private Set<DataStore> initialiedDataStores = Sets.newConcurrentHashSet();
+    private Set<String> roles = Sets.newConcurrentHashSet();
+    private boolean isInitialized = false;
+
+    private LoadingCache<DataIdentifier, DataStore> blobIdMap;
+
+    @Reference
```
amit-jain, line 75: This should be left as a plain object and any dependencies should be injected through the relevant OSGi service CompositeDataStoreService.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStore.java
```diff
@@ -0,0 +1,675 @@
...
+    }
+
+    CompositeDataStore(final Properties properties, final Collection<String> roles) {
+        this.properties = properties;
+        this.roles = ImmutableSet.copyOf(roles);
+
+        blobIdMap = setupIdMapper();
+    }
+
+    private LoadingCache<DataIdentifier, DataStore> setupIdMapper() {
```
amit-jain, line 101: Change to non-loading cache i.e. change from LoadingCache to Cache.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/DelegateDataStore.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public String getRole() {
+        return ds.getRole();
+    }
+
+    public Map<String, ?> getConfig() {
+        return config;
+    }
+
+    public boolean isReadOnly() {
```
amit-jain: Shouldn't there be a role-specific method here rather than a specific method for a specific implementation.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStoreService.java
```diff
@@ -0,0 +1,218 @@
...
+                return;
+            }
+
+            Set<String> uniqueRoles = CompositeDataStore.getRolesFromConfig(properties);
+            if (0 == uniqueRoles.size()) {
+                log.error("No roles configured for Composite Data Store");
+                return;
+            }
+
+            if (delegateDataStores.size() != uniqueRoles.size()) {
```
amit-jain: The check rather should be that all the required role-specific DataStore(s) have been initialized. This will ensure early and clear failures when configured incorrectly.

amit-jain (commented):
Review of the classes structure

mattvryan:
@amit-jain I started going through your concerns, beginning with moving the invocation of `init()` from `CompositeDataStore.init()` to the data store factory (it ended up in `AbstractDataStoreFactory` instead).  That change is in the latest update.  Will you please take a look and see if it addresses your concerns?  Meanwhile I'll continue on to other issues you mentioned.

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/datastore/AbstractDataStoreService.java
mattvryan: Yes, that it does.  I assumed it was supposed to be changed based on the log message above.  Why are we even checking for the PATH configuration value, and why are we logging that we are initializing the data store with that value, if we actually are not doing that?  If we don't change the value here we are basically logging a misleading message.

amit-jain: Well it was not being changed earlier because it would change the location and the logging was added to precisely remove the confusion of where  exactly the path is initialized. Do take a look at the FileDataStore.init(). 
We cannot change it because it would cause existing systems to not find the DataStore.

amit-jain:
While we certainly need to initialize a part of the registerDataStore() method. I don't think the changes at present suffice:
* Don't need the delegate BlobStores wrapped as DataStoreBlobStore.
* Haven't given a detailed thought but we *may* need to register the stats for each delegate if that's possible (Maybe the registration description should be changed)
* There is a check currently to ignore subsequent registrations of the BlobStore. So, we will have to stop the individual BlobStore registrations so that only the CompositeDataStore is registered.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStoreService.java
mattvryan: I'm not sure what you have in mind.

One way I think this could be done easily would be to capture the role in addition to the delegate in `addDelegateDataStore()`.  If each role was put into a `Set` then the size of the added roles could be compared to the `uniqueRoles` and if they were the same then we would know that all configured roles had been added.

If you think we need to do something more involved I suppose we could also query the context for registered services, using `DataStoreProvider.class` as the service and using each role as the filter.

Do either of those match what you had in mind or are you thinking of something different?

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/DelegateDataStore.java
mattvryan: Roles were intended to be free-form text, based on feedback from @trekawek (see https://markmail.org/message/k6gl6v3rmq357tj2).

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/DelegateHandler.java
mattvryan: I think what you are asserting here is based on a previous comment where you were asking whether there should be role-specific methods.  My understanding based on previous feedback was that roles should be free-form text.  In other words, I don't have a tightly defined "read-only" role.  Rather, I can name the role whatever I want, and then in the factory configuration for the delegate I tell it what role it has and specify any other configuration in there.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/DelegateMinRecordLengthSelector.java
mattvryan: Maybe not, I don't know.

It's hard to know what minRecordLength even means with CompositeDataStore.  It is unclear to me what this is actually used for within Oak itself.

If Oak actually calls (or if it were to call, in the future) the `minRecordLength()` method of the registered data store in order to determine if it should try to write a blob to the data store or to the node store, then it definitely matters.  In fact it becomes somewhat of a big mess with a `CompositeDataStore` because whether a blob can be written to the `CompositeDataStore` depends entirely on what delegate the blob would be written to and what the value of `minRecordLength` is for that delegate.

For the currently implemented use case for `CompositeDataStore` it doesn't matter but it could in the future if other supported use cases are added.  Suppose we were to support a scenario where a `CompositeDataStore` could have two delegates, both writable, one for everything under the `/images` path and another delegate for everything else.  Suppose the delegate containing the images is configured with `minRecordLength=16384` and the other delegate containing everything else is configured with `minRecordLength=8192`.  Now suppose a binary being written is of size 12000.  What value of `minRecordLength` should the composite data store return?

Since I wasn't sure which approach was best I made it an interface so we can provide different strategies.  If you don't think it is necessary we can eliminate it.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/IntelligentDelegateHandler.java
mattvryan: Ok I removed `@Component` on this class as well as on `GuaranteeedMinRecordLengthSelector`.  Is more required beyond just removing the class annotation?

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/AbstractSharedCachingDataStore.java
mattvryan: The abstract class may not be an instance of `CompositeDataStoreAware` but the concrete class may be (e.g. `S3DataStore`).
The issue here is the `LOG.error()` call when the record is not found.  When you have only a single data store configured and you try to retrieve a record from the configured data store, it is an error if the record cannot be found.

When you are using a composite data store, each delegate may be looking for the record if the location if the record is unknown.  Thus it is possible that one delegate may look for the record and not find it, and then a second delegate looks for it and does find it.  Logging this as an error is misleading because it causes an admin to think the system is failing or the data store may be corrupt, which could be the case with a normal data store but is not necessarily the case with a composite data store.

So this change was made to address an actual issue I saw during testing with error messages showing up in the logs even though the system was working.

oak-blob/src/main/java/org/apache/jackrabbit/oak/spi/blob/DataStoreReference.java
mattvryan: Done

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/DelegateMinRecordLengthSelector.java
amit-jain: `minRecordLength` decides upto what length the data can be encoded itself in the ID returned with no actual blob being stored in the datastore.
With the `CompositeDataStore` the problem is that it is decorated by the DataStoreBlobStore and thus needs *a* value. If the encoding logic was in the `CompositeDataStore` it would have been easier to have the individual delegates take that decision.
With this context in mind I don't think the there is scope for any other value than the minimum and thus eliminate the strategies. We should simply add that in the documentation that the `minRecordLength` should be configured the same otherwise the min value takes affect.

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/AbstractSharedCachingDataStore.java
amit-jain: Yes, that should be addressed by the encoding we have discussed in the past and is true for all the delegates as well not just the AbstractSharedCachingDataStore implementations for e.g. FileDataStore/OakFileDataStore would also throw these errors. We cannot mask this error as that would lead to debugging real problems.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/DelegateDataStore.java
amit-jain: Exactly what I am alluding to. The methods should be adjusted to not be for specific roles. the {{isReadOnly}} here is specific to the 'read-only' role and will not be relevant for other roles configured say 2 different writable delegates for different parts of the repository tree. If required the method should change to return that free-form text role.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/CompositeDataStoreService.java
amit-jain: No what I had in mind was a simple check which determines that the all the roles configured have their respective delegates configured. This will just be a comparison of all the roles collected from the configured delegates (DelegateDataStore provides a method to return the role iirc) and the configured roles.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/DelegateHandler.java
amit-jain: Yes see my comment for that also. Essentially what I am asking is rather than have methods which are tied to specific role like `isReadOnly` or `getWritable` the methods should be more generic like: 
* `getRole(role)`
* `isRole(role)`
* `getDelegatesForRole(role)`
* `getDelegatesIteratorForRole(role)`

I understand that we don't have any use cases for others except read-only/write but if we are to support that in future then we'll have to change lot of things particularly the APIs as well.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/IntelligentDelegateHandler.java
amit-jain: Well yes if there has to be a configurable selector which I don't think is required (see next comment on this). Currently, `@Reference` is used to inject the selector implementation. If required then there will have to be a way to wire the different selectors even if we only provide implementation of one initially.

So, this is not just about the `@Component` annotations. There are `@Reference` annotations also being used which indicates that OSGi is being relied upon to inject the various dependencies. How is the `IntelligentDelegateHandler` being wired. Also, as a side note we should probably change the name to it being Simple/Role as right now it is not doing the basic things specially if we also remove the `minRecordLength` selectors etc. which are not needed.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/DelegateHandler.java
mattvryan: But doesn't that make an assumption that the role has some meaning with respect to whether it is read-only or not?  The phrasing "rather than have methods which are tied to a specific role" is making me think we aren't talking about the same thing, because there are no such roles.

For example, in my testing I simulate an environment that has two roles, "production" and "staging".  Those roles have no meaning to the code, they only mean something after you look at the configuration.  They could have just as easily been called "role1" and "role2", or "Alice" and "Bob".  The configuration indicates whether a data store is read-only, not the role.

In my mind `getWritableDelegatesIterator()` and `getAllDelegatesIterator()` map  more closely to the data store itself.  If you call `addRecord()` on the composite data store, then you'd only want writable delegates.  Whereas if you call `getRecord()` you'd want all delegates, not just the writable ones.  Read-only and read-write maps pretty well to the data store API.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/DelegateMinRecordLengthSelector.java
mattvryan: Ok - I'll move it just to `CompositeDataStore` then, that will be much simpler.

oak-blob-composite/src/main/java/org/apache/jackrabbit/oak/blob/composite/DelegateHandler.java
amit-jain: Aha Ok! I understand now. Ignore the above and my deleted comment

mattvryan:
@amit-jain We can separate delegate initialization from registration easily enough, meaning in the factory we only initialize delegates but don't register them, but from the service we do both.  Then we can skip wrapping delegates as a `DataStoreBlobStore` since that is only needed for the registration, which we would skip for factories.  The delegates would get registered as a `DataStoreProvider` instead.

However the API to add a statistics provider is in the DataStoreBlobStore class, not the data store API, so we couldn't add stats to a delegate unless it is wrapped as a `DataStoreBlobStore`.

PR #75, opened 2017-12-17
OAK-7070: reverted changes made to excerpt handling by OAK-6750

buuhuu:
Support for rep:excerpt(.) and rep:excerpt(propertyName) as query columns not yet covered.


buuhuu:
merged in svn@r1818645

PR #76, opened 2017-12-17
OAK-6597: allow lucene to store all of the fulltext fields in the nodeScopeIndex

buuhuu:
… with one exception: :nodename.


buuhuu:
This requires #75 to be merged, so the test is failing.

PR #80, opened 2018-02-21
Changes MarkSweepGarbageCollector to support CompositeDataStore.

mattvryan:
This change adds a new metadata file, sweepComplete, and changes
the garbage collection logic slightly.  Before this change,
GC on shared data stores followed the basic logic that all
repositories must first perform the mark phase, then any
repository could perform the sweep phase.  Once one repository
completes the sweep phase it deletes its corresponding
references file from the data store metadata area.  Subsequent
repositories sharing the data store will not sweep because the
references that should be deleted have already been deleted
during the first sweep phase.

This does not work with CompositeDataStore because the repository
with a CompositeDataStore may be sharing one data store and not
sharing another, and deletes may span both data stores.  If the
composite sweeps second, the other repository will finish sweeping
and then delete the references file.  The composite then cannot
complete its sweep phase because there is a mismatch between
the number of references files and repository files in the shared
data store.  The problem exists in reverse if the composite sweeps
first, in that case for the other repository.

To address this issue this change was made.  When a repository
finishes sweeping, instead of removing its references file it
first indicates that the sweep is complete by creating the
sweepComplete file.  The references files are only removed if a
sweepComplete file exists for every repository.  Otherwise,
the references are allowed to remain so another repository can
complete its sweep phase.  This allows the CompositeDataStore
to complete the sweep phase and delete changes it has made
to its own non-shared data store(s) as well as any that should
be made to the shared data store(s).

### DIFF
--- a/oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
+++ b/oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
@@ -441,17 +441,8 @@ public class MarkSweepGarbageCollector implements BlobGarbageCollector {
         if(count != deleted) {
             LOG.warn("Deleted only [{}] blobs entries from the [{}] candidates identified. This may happen if blob " 
                          + "modified time is > "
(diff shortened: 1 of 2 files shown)


mattvryan:
After discussion in OAK-7083 we don't need this change.  Closing

PR #81, opened 2018-03-11
OAK-7209: Fix race condition between snapshot and GC.

csaboka:
This is a proposed fix for the issue reported in OAK-7209: if a snapshot happens during GC, the removed files may not be removed from the tracker.
The change doesn't cover the use case where two Oak instances sharing the same blob store both try to delete files from the store.

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentStore.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentStore.java
@@ -1112,7 +1112,7 @@ public class RDBDocumentStore implements DocumentStore {
         }
     }
 
-    private static String dumpIndexData(DatabaseMetaData met, ResultSetMetaData rmet, String tableName) {
+    private static String dumpIndexData(DatabaseMetaData met, ResultSetMetaData rmet, String tableName, Set<String> indexedColumns) {
 
         ResultSet rs = null;
         try {
@@ -1134,6 +1134,13 @@ public class RDBDocumentStore implements DocumentStore {
                 // retry ucase
                 rs = met.getIndexInfo(null, null, tableName.toUpperCase(Locale.ENGLISH), false, true);
                 indices = getIndexInformation(rs, rmetSchemaName);
+            }
+            if (indexedColumns != null) {
+                for (Map<String, Object> map : indices.values()) {
+                    if (map.containsKey("columns")) {
+                        indexedColumns.addAll((Set<String>) (map.get("columns")));
+                    }
+                }
             }
             return dumpIndexData(indices);
         } catch (SQLException ex) {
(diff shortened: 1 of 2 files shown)


oak-blob-plugins/src/test/java/org/apache/jackrabbit/oak/plugins/blob/datastore/BlobIdTrackerTest.java
```diff
@@ -167,7 +167,8 @@ public void snapshotExecuteAfterRemove() throws Exception {
         ScheduledFuture<?> scheduledFuture =
             scheduler.schedule(job, 0, TimeUnit.MILLISECONDS);
         scheduledFuture.get();
-        assertEquals("Extra elements after add", initAdd, retrieve(tracker));
```
amit-jain: This is still a valid case as would happen in most GC cases where there's no snapshot in after the global merge and before delete.  So, I'll create a new variation of test to test the intervening snapshot case.

csaboka: I may be missing something, but I think it's safe to remove this line because it checks the same thing addSnapshot() already tests: the list of records is OK after adding items and making a snapshot. The remove() call happens after the removed assertion, so it couldn't affect the assertion.

You are right that it's worthwhile to test add() and remove() following each other directly without an intervening snapshot, but there wasn't such a test in the code before this change. It only had a test case for a sequence of add-snapshot-retrieve-delete, which I've changed to test for add-snapshot-delete.

amit-jain: The additional thing this method does is checks the snapshot *after* the remove. It was added for that specific test case and as would mimic where the GC wouldn't have an intervening snapshot.

csaboka: I don't understand how an assertion before the remove call can check anything about the state after the remove, but I'll defer to your judgement on this. I'm not as familiar with the project as you are after all.

amit-jain: The retrieve mimicks the get() called made when collecting the blobs in the sweep phase of the GC.  I tried the tests to be as close as possible to the actual operations and I'll try to add some comments so that its a little clear. I'd like to keep that and add a new method mimicking the new behavior you described which would just not have a retrieve call before. I hope that's ok.

csaboka: Oh, it makes sense now. Let me know if I can help with anything else to have this accepted into trunk.

PR #84, opened 2018-04-19
OAK-7401: Changes kept in memory when update limit is hit in commit hook

mreutegg:


mreutegg:
Changes were committed to trunk.

PR #85, opened 2018-04-19
OAK-7427: Inconsistent method name.

Kui-Liu:
Change the method name "serialize" to "format".

The method is named as "serialize", but the method seems like to format the data.
The method name "format" should be more clear than "serialize".


mduerig:
@francescomari , @dulceanu I don't think we should change this method's name. While it returns a string, it still serialises the passed references.

ShangwenWang:
@mduerig Dear developer,
In my opinion, the function of this method is to serialize the passed references. However, only naming it as "serialize" cannot describe it clearly.
Given that the class name (GetReferencesResponseEncoder) provides lots of semantic information, it would be better to change this method's name into "serializeReferences" for providing enough guidance.
How do you feel about this idea?

PR #86, opened 2018-04-26
OAK-6087: Avoid reads from MongoDB primary

mreutegg:


mreutegg:
The changes proposed in this pull request have been committed to Oak trunk.

PR #87, opened 2018-05-07
OAK-4818: Handle version nodes correctly in Node.getNodes().

csaboka:
Version and version history nodes should implement the Version and
VersionHistory interfaces respectively no matter how they are accessed,
but Node.getNodes() always returned plain Node instances.


csaboka:
Hi!

Is there any intention of ever merging this into trunk? As far as I can tell trunk still has the issue.

If it helps, I can redo my fix on top of the current trunk, so you don't need to deal with almost five years' worth of conflicts.

PR #89, opened 2018-06-21
Filter org.apache.jackrabbit.oak.plugins.value.jcr from BND view.

mattvryan:
This means that when changes are made to BinaryImpl.java,
ValueFactoryImpl.java, or ValueImpl.java, those changes would not
be consider API changes and would not require a version bump
when evaluated by the Maven bnd plugin.


mreutegg:
The proposed changes in this PR have been committed via https://issues.apache.org/jira/browse/OAK-7601

PR #101, opened 2018-10-17
OAK-7839 - calling MeterStats#mark whenever an index update fails

tteofili:


tteofili:
merged patch from this PR in r1844315 on SVN.

PR #103, opened 2018-10-23
S3 Bucket iterator stops too early

wimsymons:
Fixed major bug in the S3 bucket iterator. When the queue is empty due to the fact that we get a full page of records starting with the META/ key, the iterator stops while there is still data available in the bucket. This causes problems with datastore GC, and datastore consistency checks (and possibly even more).

### DIFF
--- a/oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
+++ b/oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
@@ -993,6 +993,11 @@ public class S3Backend extends AbstractSharedBackend {
                 loadBatch();
             }
 
+            while (queue.isEmpty() && prevObjectListing.getNextMarker() != null) {
+                LOG.debug("Queue is empty, but there is more data in the S3 bucket");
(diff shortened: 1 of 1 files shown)


wimsymons:
The fix I provided is for the latest release, but older releases are affected as well. We are running AEM 6.3 with Oak Core 1.6.6 and the problem is the same there. I will create another PR for the 1.6 branch as well.

wimsymons:
A little explainer. But based on a batch size of 2 instead of 1000.

Suppose your S3 keys look as follows:
- 1
- 2
- 3
- 4
- META/1
- META/2
- 5
- 6

loadBatch would first load [1, 2], filter out no META/ keys and pass [1, 2] to the caller. 
Next time, loadBatch would load [3, 4], filter out no META/ keys and pass [3, 4] to the caller. 
Than, loadBatch would load [META/1, META/2], filter out the META/ keys and pass [] to the caller. 
After that, traversing the bucket would stop, because the returned list is empty, even if there are many more batches to load. 
The fix checks if the returned list is empty and there are more batches available, it would load a new batch until there is data in the batch or there is no more batch available.

wimsymons:
Updated the commit message after creating the JIRA issue

PR #104, opened 2018-10-23
S3 Bucket iterator stops too early

wimsymons:
Fixed major bug in the S3 bucket iterator. When the queue is empty due to the fact that we get a full page of records starting with the META/ key, the iterator stops while there is still data available in the bucket. This causes problems with datastore GC, and datastore consistency checks (and possibly even more).

This is the same fix as in trunk provided in #103


wimsymons:
See #103 for an explainer.

wimsymons:
Updated the commit message after creating the JIRA issue

PR #111, opened 2018-11-29
OAK-7925: OakUpgrade for RDBMK: user name and password must match

sbluhm:


thomasmueller:
Tests are not changed. I recommend the patch in OAK-7925 instead.

PR #122, opened 2019-03-13
Oak 8114

nit0906:


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/util/IndexDefinitionBuilder.java
```diff
@@ -567,21 +576,71 @@ private static Tree getOrCreateChild(Tree tree, String name){
...
+        public boolean propertyDeleted(PropertyState before) {
+            if(ignorablePropertiesList.contains(before.getName())) {
+                return true;
+            }
+            return super.propertyDeleted(before);
+        }
+
+        @Override
+        public boolean childNodeAdded(String name, NodeState after) {
+            if(name.equals(FulltextIndexConstants.PROP_FACETS)) {
```
catholicon, line 622: While I don't have a good idea atm, but I don't quite like the assumption that any change under `facets` nodes only requires a refresh. Yes, it's true for currently supported properties but in the future if we add something else here that requires reindexing then it might slip away and won't be caught. I'd probably be happier if we make sure that adding `facets/foo` reindexes while adding `facets/topChildren` only refreshes.

nit0906: @catholicon  - makes sense . I have made the changes . Basically removed override for childNodeChanged and added handling in childNodeAdded and childNodeDeleted to set reindex to true in case there are any new properties apart from the current three . In case the new properties are also such that they wouldn't req re indexing - then of course we will have to make changes accordingly .

catholicon: Thanks for the change. Just a minor bit, maybe we could move the life of properties under facets at top along with other properties (or, I'd do that myself)

catholicon:
Thanks @nit0906. While I haven't looked at the test, but just the sheer number of it and roughly reading them gave me quite comfortable feeling with the change. I'd wait for @thomasmueller to have a look and if all's ok, I'd commit this tomorrow.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/util/IndexDefinitionBuilder.java
```diff
@@ -567,21 +576,78 @@ private static Tree getOrCreateChild(Tree tree, String name){
     }
 
     static class SelectiveEqualsDiff extends EqualsDiff {
+
+        List<String> ignorablePropertiesList = ImmutableList.of(FulltextIndexConstants.PROP_NAME,FulltextIndexConstants.PROP_WEIGHT,FIELD_BOOST,
```
catholicon: I'd remove facet related ones from here while moving facet related properties list here (or it'd great if you push that too :)

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/util/IndexDefinitionBuilder.java
nit0906: @catholicon  - done . Moved the declaration on top . I can also handle the facet properties change by overriding childNodeChanged method (IMO either way seems fine - I just picked one - but let me know if you think I need to change)

I also did another change - I had added [name] property for index node as ignorable but I realized this will conflict with changes in name property on property def node (which will req reindex if changed) - so I have removed the handling for name property.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/util/IndexDefinitionBuilder.java
nit0906: changes done.

PR #123, opened 2019-03-19
OAK-8138 | Switching statistical facet mode to secure in case hit cou…

nit0906:
…nt is less than sample size


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
```diff
@@ -365,7 +365,7 @@ private boolean loadDocs() {
                             long f = PERF_LOGGER.start();
                             if (facetProvider == null) {
                                 facetProvider = new LuceneFacetProvider(
-                                        FacetHelper.getFacets(searcher, query, plan, indexNode.getDefinition().getSecureFacetConfiguration())
+                                        FacetHelper.getFacets(searcher, query, plan, indexNode.getDefinition().getSecureFacetConfiguration(), docs.totalHits)
```
catholicon: This `docs` object is first batch of results that fills up cursor. While `totalHits` might be accurate (it most likely is but I'd need to double check it), but I would prefer if we don't need to pass this on most likely because this requires an external agent to pass on an information that is actually implementation requirement and the external agent has a potential to mess up some internal assumption.

I understand with current code structure (switch deciding which type of facet counter to use quite early) it's easier to pass on this information. But, maybe, what we can do is to let `StatisticalSortedSetDocValuesFacetCounts#getTopChildren0` start doing its work and if it realizes that hit count is small enough then it can delegate to a new instance of `SortedSetDocValuesFacetCounts`. Wdyt?

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/util/FacetHelper.java
```diff
@@ -62,6 +62,14 @@ public static FacetsConfig getFacetsConfig(NodeBuilder definition) {
 
     public static Facets getFacets(IndexSearcher searcher, Query query, QueryIndex.IndexPlan plan,
                                    SecureFacetConfiguration secureFacetConfiguration) throws IOException {
+        // This method is just a fail safe in case it is being called via the old signature in any code that consumes oak.
```
catholicon: I don't think we need to worry about these - this class should not be exposed outside oak-lucene and it should be ok to change method signature any way we want.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/SecureFacetTest.java
```diff
@@ -172,6 +172,51 @@ private void createLargeDataset() throws RepositoryException {
         assertNotEquals("Acl-ed and actual counts mustn't be same", actualLabelCount, actualAclLabelCount);
     }
 
+    private void createSmallDataset() throws RepositoryException{
```
catholicon, line 175: I haven't read the test, but please do a quick check of existing tests that they are working with "larger than threshold" data set...

nit0906: Ran all the tests - they were all green .

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
nit0906: totalHits was accurate - that I checked - for instance - in case the query had 400 valid results but docs had only 50 top docs - then docs.size was equal to 50 but docs.totalHits was 400 ..I validated this with a couple of different use cases.

But yes - I agree with you , that it might be better if we don't pass total hits here (even I was a bit skeptical about this and wanted to get the approach reviewed). Let me check if we can work on a different approach as you suggested .

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/SecureFacetTest.java
catholicon: no, i didn't mean if they pass or not... with this issue fixed, lower result size would still give "right" results expected from a statistical pov... I think we should make sure that existing tests create/get-results which are bigger that sample size threshold.

catholicon: ok... ignore the noise I created... just read the tests again and found that `createLargeDataset` already is creating num_labels * sample_size nodes (and including all of them in results)... I just wasn't sure if that was ensure already or not.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/SecureFacetTest.java
```diff
@@ -172,6 +172,51 @@ private void createLargeDataset() throws RepositoryException {
...
+        Random rGen = new Random(42);
+        int[] labelCount = new int[NUM_LABELS];
+        int[] aclLabelCount = new int[NUM_LABELS];
+        int[] aclPar1LabelCount = new int[NUM_LABELS];
+
+        Node par = allow(getOrCreateByPath("/parent", "oak:Unstructured", session));
+
+        for (int i = 0; i < 4; i++) {
+            Node subPar = par.addNode("par" + i);
+            for (int j = 0; j < NUM_LEAF_NODES/10; j++) {
```
catholicon:
```suggestion
            for (int j = 0; j < NUM_LEAF_NODES/(2 * NUM_LABELS); j++) {
```

just to kind of keep num labels fixed... btw, I think we should also replace `4` with `NUM_LABELS` both here in and in `createLargeDataset`

PR #128, opened 2019-05-13
OAK-8309: SLOW_QUERY_COUNT don't get updated for each slow query

tihom88:


oak-core/src/main/java/org/apache/jackrabbit/oak/query/RuntimeNodeTraversalException.java
```diff
@@ -0,0 +1,8 @@
+package org.apache.jackrabbit.oak.query;
+
+public class RuntimeNodeTraversalException extends RuntimeException {
```
thomasmueller: Could you extend UnsupportedOperationException? That would make it compatible with the previously thrown exception, UnsupportedOperationException. (Some code is / was catching UnsupportedOperationException.)

oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/SelectorImpl.java
```diff
@@ -541,6 +558,21 @@ private boolean nextInternal() {
         return false;
     }
 
+    private void totalQueryStats (QueryEngineSettings queryEngineSettings) {
+        if (updateTotalQueryHistogram) {
```
thomasmueller, line 562: (here is a space between "totalQueryStats" and "(" which is not necessary)

oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/SelectorImpl.java
```diff
@@ -541,6 +558,21 @@ private boolean nextInternal() {
...
+            updateTotalQueryHistogram = false;
+            HistogramStats histogramStats = queryEngineSettings.getStatisticsProvider().getHistogram(SLOW_QUERY_PERCENTILE_METRICS_NAME, StatsOptions.METRICS_ONLY);
+            histogramStats.update(TOTAL_QUERY_HISTOGRAM);
+        }
+    }
+
+    private void addSlowQueryStats(QueryEngineSettings queryEngineSettings){
```
thomasmueller: (here is no space between ")" and "{"... By the way a simpley way to ensure consistent formatting is to use the IDE auto-format feature.)

thomasmueller (commented):
Other than the inline comments, the patch looks perfect! Thanks a lot!

oak-core/src/main/java/org/apache/jackrabbit/oak/query/RuntimeNodeTraversalException.java
tihom88: Done

PR #129, opened 2019-05-19
OAK-8328 | Config errors in index defintion causes index update to ge…

nit0906:
…t stuck in a loop


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
```diff
@@ -276,9 +276,19 @@ private void collectIndexEditors(NodeBuilder definitions,
...
+                Editor editor = null;
+                try {
+                    editor = rootState.provider.getIndexEditor(type, definition, rootState.root,
+                            rootState.newCallback(indexPath, shouldReindex, getEstimatedCount(definition)));
+                } catch (IllegalStateException e) {
+                    // This will be caught here in case there is any config related error in the index definition
+                    // where multiple values are assigned to a property that is supposed to be single valued
+                    // We log an error message here and continue - this way the bad index defintion is ignored and doesn't block the async index update
+
+                    log.error("Unable to get Index Editor for index at {} . Please correct the index definition " +
```
thomasmueller, line 288: I would also log the stack trace of the root cause, that is in this case append ", e" to the parameters: log.error(..., e)

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/util/ConfigUtil.java
```diff
@@ -36,34 +36,71 @@
  */
 public class ConfigUtil {
 
+    private static final String ILLEGAL_STATE_EXCEPTION_ERROR_MESSAGE = "Multiple Values provided for property %s in Index Defintion . Single value was expected";
```
thomasmueller: There are some typos, and some words are uppercase. Could you fix it please.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/util/ConfigUtil.java
```diff
@@ -36,34 +36,71 @@
...
+
     public static boolean getOptionalValue(NodeState definition, String propName, boolean defaultVal){
-        PropertyState ps = definition.getProperty(propName);
-        return ps == null ? defaultVal : ps.getValue(Type.BOOLEAN);
+        try{
+            PropertyState ps = definition.getProperty(propName);
+            return ps == null ? defaultVal : ps.getValue(Type.BOOLEAN);
+        }catch(IllegalStateException e ){
```
thomasmueller: Please add spaces after "}", before "{", around "()", and remove the superfluous space after e.

thomasmueller: What I usually do is use the IDE formatter, but only for the selected text. (Eclipse allows this, and I'm almost sure IntelliJ as well.)

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
```diff
@@ -276,9 +276,19 @@ private void collectIndexEditors(NodeBuilder definitions,
...
                     continue;
                 }
-
-                Editor editor = rootState.provider.getIndexEditor(type, definition, rootState.root,
-                        rootState.newCallback(indexPath, shouldReindex, getEstimatedCount(definition)));
+                Editor editor = null;
+                try {
+                    editor = rootState.provider.getIndexEditor(type, definition, rootState.root,
+                            rootState.newCallback(indexPath, shouldReindex, getEstimatedCount(definition)));
+                } catch (IllegalStateException e) {
```
thomasmueller, line 283: Would it be possible to log the problem where the IllegalStateException is thrown, that is in the IndexDefinition.java file? Maybe in the builder... I just think that the IndexUpdate class is more far away than necessary from there the exception is thrown. I think it is possible to move it closer to index definition building. Still it might re-throw an exception. It's just that in the future, there might be other reasons why IllegalStateException that are not related to a wrong index definition itself (maybe for a remote index).

nit0906: @thomasmueller  - Sure - I will log the problem in IndexDefinition . I will rethrow the exception from there - and in the catch in the IndexUpdate  and just skip the iteration for this index (without logging the error) Or maybe log error at both the places - IndexUpdate as well as IndexDefintion

PR #131, opened 2019-05-27
OAK-8344 | Exposing local index directory size as a metric

nit0906:


thomasmueller (commented):
It looks good to me.

catholicon:
lgtm to me as well. One minor tidbit, I wonder if it's worth to have schedule period configurable by environment variable and/or jvm system property.

thomasmueller:
@catholicon  yes it might make sense... I committed the patch now, without adding such a setting for now.

PR #132, opened 2019-06-18
OAK-7254: Indexes with excludedPaths, or includedPaths should not be …

tihom88:
…picked for queries without path


oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndexPlanner.java
```diff
@@ -76,7 +77,7 @@
      */
     public static final String ATTR_FACET_FIELDS = "oak.facet.fields";
 
-    private static final String FLAG_ENTRY_COUNT = "oak.fulltext.useActualEntryCount";
+    private static final String FLAG_ENTRY_COUNT = "oak.fulltext.useActualEntryCount";;
```
thomasmueller: Typo at the end.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
```diff
@@ -86,6 +86,24 @@
...
+    boolean isEnablePathRestrictions();
+
+    /**
+     * Enable/Disable evaluating path restriction while figuring out which index should be used
+     *
+     * @param enablePathRestrictions enablePathRestrictions
+     */
+    @Description("Enable/Disable evaluating path restriction while figuring out which index should be used")
+    void setEnablePathRestrictions(
+            @Description("Whether path restrictions be evaluated during index plan evaluation")
```
thomasmueller: This description is seen in the JMX UI right? What about "Whether path restrictions of indexes (excludedPaths / includedPaths) are taken into account during query execution, for Lucene indexes. When enabled, only indexes are considered if the index path restriction is compatible with the query path restrictions. When disabled, only the queryPaths of the index is taken into account."

thomasmueller: The above can also be used for documentation (in oak-doc), together with some examples. Don't worry about the documentation right now, we can do that later.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
```diff
@@ -86,6 +86,24 @@
 
     void setFastQuerySize(boolean fastQuerySize);
 
+    /**
+     * Whether Path restrictions are enabled while figuring out index plan
+     *
+     * @return true if enabled
+     */
+    boolean isEnablePathRestrictions();
```
thomasmueller: Maybe it's a bit clearer when we use "get / setStrictPathRestrictions"?

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettingsService.java
```diff
@@ -68,6 +68,14 @@
                         "Note: even if enabled, getSize may still return -1 if the index used does not support the feature." 
         )
         boolean fastQuerySize() default false;
+
+        @AttributeDefinition(
+                name = "Enable Path restrictions for indexes to be used",
+                description = "Enable path restriction while finding indexplan e.g. if we have path restrictions on index definitions," +
+                 "we will not use index if our query can have results from the excluded paths"
```
thomasmueller: The description from above can be used.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
thomasmueller: Right now this is a boolean setting: if disabled, ignore the path condition, if enabled, use it. But we need to do it two stages. The first stage is, only log a warning if indexes are used where the query path condition doesn't match the index path restriction. So 3 stages:

- disabled (current behavior)
- warn (log a warning only, but still use those indexes)
- enabled (in the (more) far future)

The "enabled" behavior is implemented in the cost estimation phase (like now). The "warn" behaviour would need to be implemented when _executing_ a query.

thomasmueller: The question is, should we have an "enum" instead of a boolean flag? Or two boolean flags. I would opt for an "enum" setting: "get / setStrictPathRestrictions": DISABLED, WARN, ENABLED.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/StrictPathRestriction.java
```diff
@@ -0,0 +1,17 @@
+package org.apache.jackrabbit.oak.api;
```
thomasmueller, line 17: I think the license header is missing.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/StrictPathRestriction.java
```diff
@@ -0,0 +1,17 @@
...
+
+public enum StrictPathRestriction {
+    ENABLE("enable"),
+    WARN("warn"),
+    DISABLE("disable");
+    private String stringValue;
```
thomasmueller: Do you use the stringValue somewhere? I couldn't find... If not, could you remove it.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/StrictPathRestriction.java
```diff
@@ -0,0 +1,17 @@
...
+
+    StrictPathRestriction(String pathRestriction) {
+        stringValue = pathRestriction;
+    }
+
+    public static StrictPathRestriction stringToEnum(String strictPathRestrictionInString) {
+        return StrictPathRestriction.valueOf(strictPathRestrictionInString.toUpperCase());
+    }
```
thomasmueller, line 29: I understand lowercase is a bit easier to read... but sometimes it is uppercase (setStrictPathRestriction below), sometimes lowercase (here). I think it's better if it's consistent (either upper or lowercase). I would be fine if it's consistently uppercase, if that simplifies the code (I guess it does). In which case this method wouldn't be needed.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
```diff
@@ -86,6 +86,38 @@
...
+     */
+    String getStrictPathRestriction();
+
+    /**
+     *  Whether path restrictions of indexes (excludedPaths / includedPaths) are taken into account during query execution,
+     *  for Lucene indexes. When enabled, only indexes are considered if the index path restriction is compatible with the
+     *  query path restrictions. When disabled, only the queryPaths of the index is taken into account.
+     *
+     * @param pathRestriction Set path restriction: Expected value is either of ENABLE/DISABLE/WARN
+     *                        ENABLE: enable path restriction- Index won't be used if index definition path restrictions are not compatible with query's path restriction
```
thomasmueller, line 102: The same info is available in "@Description" below... even thought that won't result in Javadocs, I guess it's better to just keep the documentation in "@Description", and simplify the Javadocs. That way, we don't repeat the info (and so only have to update one case).

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/aggregate/AggregateIndexPlan.java
```diff
@@ -34,6 +34,7 @@
     
     private Filter filter;
     private boolean pathWithoutPlan;
+    private boolean logWarning = false;
```
thomasmueller: "= false" isn't needed here, as that's the default.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettingsService.java
```diff
@@ -68,6 +68,16 @@
...
         boolean fastQuerySize() default false;
+
+        @AttributeDefinition(
+                name = "Enable Strict Path restrictions for indexes to be used",
+                description = "Whether path restrictions of indexes (excludedPaths / includedPaths) are taken into" +
+                        "account during query execution, for Lucene indexes. When enabled, only indexes are considered if" +
+                        "the index path restriction is compatible with the query path restrictions. When disabled, only" +
+                        "the queryPaths of the index is taken into account."
+        )
+        String getStrictPathRestrictionsForIndexes() default DISABLED_STRICT_PATH_RESTRICTION;
```
thomasmueller, line 79: Interesting... I thought default methods need a body... learned something!

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettingsService.java
```diff
@@ -82,6 +92,7 @@
     static final String QUERY_FAIL_TRAVERSAL = "queryFailTraversal";
     
     static final String QUERY_FAST_QUERY_SIZE = "fastQuerySize";
+    static final String DISABLED_STRICT_PATH_RESTRICTION = "DISABLE";
```
thomasmueller, line 95: Maybe better use StrictPathRestriction.DISABLE? I guess you can import StrictPathRestriction here right?

oak-core/src/main/java/org/apache/jackrabbit/oak/query/plan/SelectorExecutionPlan.java
```diff
@@ -32,6 +32,7 @@
     private final double estimatedCost;
     private final QueryIndex index;
     private final IndexPlan plan;
+    private boolean logWarning = false;
```
thomasmueller: Same here, "= false" isn't needed.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneStrictPathRestrictionEnableTest.java
```diff
@@ -0,0 +1,246 @@
...
+    public void shutdownExecutor() {
+        executorService.shutdown();
+    }
+
+    @Test
+    public void pathIncludeWithPathRestrictionsEnabled() throws Exception {
+
+        Tree idx = createIndex("test1", of("propa", "propb"));
+        idx.setProperty(createProperty(PROP_INCLUDED_PATHS, of("/test/a"), Type.STRINGS));
+        //Do not provide type information
```
thomasmueller: Not sure what this comment means... I see you moved the code, that's fine, but I guess you can remove it.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneStrictPathRestrictionEnableTest.java
```diff
@@ -0,0 +1,246 @@
...
+        assertThat(explain("select [jcr:path] from [nt:base] where [propa] = 10 and isDescendantNode('/test/a')"), containsString("lucene:test1"));
+        assertQuery("select [jcr:path] from [nt:base] where [propa] = 10 and isDescendantNode('/test/a')", asList("/test/a/b"));
+
+    }
+
+    @Test
+    public void pathExcludeWithPathRestrictionsEnabled() throws Exception {
+        Tree idx = createIndex("test1", of("propa", "propb"));
+        idx.setProperty(createProperty(PROP_EXCLUDED_PATHS, of("/test/a"), Type.STRINGS));
+        //Do not provide type information
```
thomasmueller: Same here.

oak-query-spi/src/main/java/org/apache/jackrabbit/oak/spi/query/QueryIndex.java
```diff
@@ -348,6 +348,10 @@ default String getIndexName(Filter filter, NodeState rootState) {
          * @return if it is deprecated
          */
         boolean isDeprecated();
+        
+        default boolean logWarning(){
```
thomasmueller: Please add a space here, before "{".

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettingsService.java
tihom88: This value is being used in interface above with this as default value and default value needs to be a constant so I had to use a String instead of StrictPathRestriction.DISABLE

thomasmueller: OK, I didn't know... It's fine then!

oak-api/src/main/java/org/apache/jackrabbit/oak/api/StrictPathRestriction.java
tihom88: yes removing

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
tihom88: removed @description as it is also available on parameter

PR #139, opened 2019-08-06
OAK-8529 - Eagerly deregister MBeans on SecurityProviderRegistration …

rombert:
…deactivation


rombert:
Note sure why this is still open, the issue has already been resolved - see https://issues.apache.org/jira/browse/OAK-8529 .

PR #149, opened 2019-09-10
OAK-8603: counter index fix to allow re-indexing of composite stores

fabriziofortino:
It includes:

* Patch in [OAK-8603](https://issues.apache.org/jira/browse/OAK-8603) but instead of catching the  `UnsupportedOperationException` checks if the mount is not read-only.
* Patch in [OAK-8579](https://issues.apache.org/jira/browse/OAK-8579) to allow index creation in read-only repo first.


fabriziofortino:
fixed with [OAK-8603](https://issues.apache.org/jira/browse/OAK-8603).

PR #150, opened 2019-09-12
OAK-8494 - Support AWS Key Managed Service (SSE-KMS)

ankitaagar:


oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3RequestDecorator.java
```diff
@@ -20,34 +20,61 @@
...
-            this.dataEncryption = dataEncryption.valueOf(props.getProperty(S3Constants.S3_ENCRYPTION));
+        String encryptionType = props.getProperty(S3Constants.S3_ENCRYPTION);
+        if (encryptionType != null) {
+            this.dataEncryption = dataEncryption.valueOf(encryptionType);
+
+            if (encryptionType.equals(S3Constants.S3_ENCRYPTION_SSE_KMS)) {
+                String keyId = props.getProperty(S3Constants.S3_SSE_KMS_KEYID);
+                if(!StringUtils.isNullOrEmpty(keyId)){
+                    sseParams = new SSEAwsKeyManagementParams(keyId);
+                    sseParams.withAwsKmsKeyId(keyId);
```
amit-jain, line 49: Why is the need to initialize with key id then again set it. The whole block can be accomplished with
```
sseParams = new SSEAwsKeyManagementParams();
if (!StringUtils.isNullOrEmpty(keyId)) {
    sseParams.withAwsKmsKeyId(keyId);
}     
```

PR #151, opened 2019-09-18
OAK-8633: Add warn logs if we add/update a string property larger tha…

tihom88:
…n 100KB


oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/NodeImpl.java
```diff
@@ -1379,6 +1380,11 @@ private Property internalSetProperty(
         final String oakName = getOakPathOrThrow(checkNotNull(jcrName));
         final PropertyState state = createSingleState(
                 oakName, value, Type.fromTag(value.getType(), false));
+        long maxStringPropertyLength = sessionContext.getRepository().getDescriptorValue(RepositoryImpl.MAX_STRING_PROPERTY_SIZE).getLong();
```
thomasmueller, line 1383: internalSetProperty is called very, very often. It is performance critical.

The method call getDescriptorValue is relatively slow. Instead, the value should be stored in the session context or the repository as a field, so it is faster to retrieve the value.

Also, the value is only needed for properties of type "string". So this line can be moved inside the "if (type == "string")".

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/NodeImpl.java
```diff
@@ -1414,7 +1420,13 @@ private Property internalSetProperty(
         if (values.length > MV_PROPERTY_WARN_THRESHOLD) {
             LOG.warn("Large multi valued property [{}/{}] detected ({} values).",dlg.getPath(), jcrName, values.length);
         }
-
+        long maxStringPropertyLength = sessionContext.getRepository().getDescriptorValue(RepositoryImpl.MAX_STRING_PROPERTY_SIZE).getLong();
```
thomasmueller, line 1423: Same as above.

PR #152, opened 2019-09-23
OAK-8633: Add warn logs if we add/update a string property larger than 100KB

tihom88:


catholicon:
I think we should simply allow config via command line params - that'd significantly reduce the plumbing to move configured value to intended location.

catholicon (changes requested):
I think we should simply allow config via command line params - that'd significantly reduce the plumbing to move configured value to intended location.

tihom88:
@catholicon yes we can enable these via command line params  and that will reduce plumbing but in my opinion we should only have those params via command line which have major impact over application. The changes presented here are not that that big to have these enabled via commandline params. Also with osgi config framework, we can persist these and we don't have to add command line params each time  we restart application.

tihom88:
@catholicon  suggested changes are incorporated in PR #155 .
Closing this PR.

PR #153, opened 2019-09-23
OAK-8642: Add warn logs while indexing if string property is larger than 100KB

tihom88:


oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -224,13 +238,19 @@ private boolean indexProperty(String path,
             addBinary(doc, null, binaryValues);
             dirty = true;
         } else {
+            if (property.getType() == Type.STRING || property.getType() == Type.STRINGS) {
```
catholicon: It could be the case that the large property that shows up here might not be indexed as property or fulltext. I think this block should be extracted out in some private method and called individually from code indexing property and/or fulltext (that can lead to double warns one each for prop and fulltext but that might be ok I think).

catholicon:
I think we should simply allow config via command line params - that'd significantly reduce the plumbing to move configured value to DocumentMaker. I'd suggest the same should be done in #152

catholicon (changes requested):

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
tihom88: Created a new method and logging on both fulltext as well as property indexing.

tihom88:
Raised another PR [#154 ](https://github.com/apache/jackrabbit-oak/pull/154) with review comment updates.

PR #154, opened 2019-09-26
OAK-8642: Add warn logs while indexing if string property is larger than 100KB (Indexing)_

tihom88:


oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMakerLargeStringPropertiesLogTest.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    @After
+    public void loggingAppenderStop() {
+        listAppender.stop();
+        System.setProperty("oak.repository.property.logWarnStringSizeThreshold", "");
+    }
+
+    private void setThresholdLimit(String threshold) {
+        System.setProperty("oak.repository.property.logWarnStringSizeThreshold", threshold);
```
catholicon: ~You might want to use `TemporarySystemProperty` ... the rule would reset sys props to their initial state after test.~ (EDIT: misplaced comment)

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMakerLargeStringPropertiesLogTest.java
```diff
@@ -0,0 +1,189 @@
...
+        LoggerContext context = (LoggerContext) LoggerFactory.getILoggerFactory();
+        listAppender = new ListAppender<>();
+        listAppender.start();
+        context.getLogger(nodeImplLogger).addAppender(listAppender);
+    }
+
+    @After
+    public void loggingAppenderStop() {
+        listAppender.stop();
+        System.setProperty("oak.repository.property.logWarnStringSizeThreshold", "");
```
catholicon: You might want to use `TemporarySystemProperty` ... the rule would reset sys props to their initial state after test.

Similarly for logger, it might be easier (better??) to use `LogCustomizer` maybe

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -57,6 +58,8 @@
 public abstract class FulltextDocumentMaker<D> implements DocumentMaker<D> {
 
     private final Logger log = LoggerFactory.getLogger(getClass());
+    private static final String WARN_LOG_STRING_SIZE_THRESHOLD_KEY = "oak.repository.property.logWarnStringSizeThreshold";
```
catholicon: Maybe make this field package private so that the test can simply use the variable instead of hard coding the value!?

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -57,6 +58,8 @@
...
+    private int DEFAULT_WARN_LOG_STRING_SIZE_THRESHOLD_VALUE = 102400;
```
catholicon: Shouldn't this be static final too?

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -105,6 +108,16 @@ public FulltextDocumentMaker(@Nullable FulltextBinaryTextExtractor textExtractor
 
     protected abstract void indexNodeName(D doc, String value);
 
+    protected void logLargeStringProperties(String propertyName, String value) {
+        int logWarnStringSizeThreshold = DEFAULT_WARN_LOG_STRING_SIZE_THRESHOLD_VALUE;
+        if (System.getProperty(WARN_LOG_STRING_SIZE_THRESHOLD_KEY) != null && !System.getProperty(WARN_LOG_STRING_SIZE_THRESHOLD_KEY).isEmpty()) {
+            logWarnStringSizeThreshold = Integer.parseInt(System.getProperty(WARN_LOG_STRING_SIZE_THRESHOLD_KEY));
+        }
```
catholicon: This should probably be done in constructor to some `private final int logWarnStringSIzeThreshold` field.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMaker.java
```diff
@@ -123,6 +123,9 @@ protected boolean indexTypedProperty(Document doc, PropertyState property, Strin
             if (includePropertyValue(property, i, pd)){
                 doc.add(f);
                 fieldAdded = true;
+                if (tag == Type.STRING.tag()) {
+                    logLargeStringProperties(property.getName(), property.getValue(Type.STRING, i));
+                }
```
catholicon: I'd prefer if we move this in `FulltextDocumentMaker` itself (comment over there too)

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -230,7 +243,7 @@ private boolean indexProperty(String path,
 
             if (pd.fulltextEnabled() && includeTypeForFullText) {
                 for (String value : property.getValue(Type.STRINGS)) {
-
+                    logLargeStringProperties(property.getName(), value);
```
catholicon, line 251: I wanted to add this comment at line 241 but apparently I can't do that.

I think we should split `dirty |= addTypedFields(doc, property, pname, pd);` and see if `addTypedFields` returns true. If so, log here for large property indexing after checking that tag was `STRING`.

catholicon (changes requested):

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMakerLargeStringPropertiesLogTest.java
tihom88: Now using TemporarySystemProperty.
Going with logappender itself as I am also checking log message and logappender can compare message formats.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
tihom88: Changed access modifier to public.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
tihom88: Yes. Changed.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
tihom88: Done

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMaker.java
tihom88: Here we are iterating and if we do logging in FulltextDocument. I will have to reiterate again on string array. So leaving it as is.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
tihom88: Answered above.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMaker.java
catholicon: Well, I would still prefer that we refractor a bit and keep logging in oak search. But, if you don't want to do that, then at the very least, we should ensure that we put this log in elasticsearch module as well.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMakerLargeStringPropertiesLogTest.java
catholicon: Hmm... I wonder if it might be more useful to improve `LogCustomizer` to allow to compare message format. But, I'd leave that to you.
(Orthogonal to this conversation, looking at `LogCustomizer`, it seems it doesn't support to be used as a `Rule`. It would be interesting to expose a Rule which exposes a `LogCustomizerBuilder` type methods and takes care of calling `finished` when test ends)

catholicon:
@tihom88 I wasn't feeling happy about adding another generic param `F` to `FulltextDocumentMaker`. I've uploaded a patch to oak issue which seems to extract much of the logic in `FulltextDocumentMaker` (I also modified system param parsing). Can you see if that change looks ok to you?

tihom88:
Change looks good.

catholicon:
> Change looks good.

Ok, can you please commit it then? Also, did you think about `LogCustomizer` change. Maybe, at least, open an improvement for that I guess.

tihom88:
Opened an improvement Jira (OAK-8675) for LogCustomizer

PR #155, opened 2019-09-29
OAK-8633: Add warn logs if we add/update a string property larger than 100KB (Node)

tihom88:


oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/NodeImpl.java
```diff
@@ -159,6 +161,16 @@
...
     public NodeImpl(T dlg, SessionContext sessionContext) {
         super(dlg, sessionContext);
+        logWarnStringSizeThreshold = getLogWarnStringSizeThreshold();
+    }
+
+    private int getLogWarnStringSizeThreshold() {
+    int threshold = OakJcrConstants.DEFAULT_WARN_LOG_STRING_SIZE_THRESHOLD_VALUE;
+        if (System.getProperty(OakJcrConstants.WARN_LOG_STRING_SIZE_THRESHOLD_KEY) != null
+                && !System.getProperty(OakJcrConstants.WARN_LOG_STRING_SIZE_THRESHOLD_KEY).isEmpty()) {
+            threshold = Integer.parseInt(System.getProperty(OakJcrConstants.WARN_LOG_STRING_SIZE_THRESHOLD_KEY));
```
catholicon: `Integer#parseInt` can throw `NumberFormatException` which should be caught an handled. But, better yet, it would cleaner to simply use `Integer#getInt(String, int)`.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/OakJcrConstants.java
```diff
@@ -0,0 +1,6 @@
+package org.apache.jackrabbit.oak.jcr.session;
+
+public interface OakJcrConstants {
+    int DEFAULT_WARN_LOG_STRING_SIZE_THRESHOLD_VALUE = 102400;
```
catholicon, line 4: I don't recall our earlier discussion, but I though we were saying that indexing end (OAK-8642) would log for 100K while jcr side would be a bit more tolerable. But maybe, I'm mis-remembering.

catholicon (changes requested):

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/OakJcrConstants.java
tihom88: AFAIR, two different values were not suggested. If you think we should have different values, let me know what should be default value for jcr.

catholicon: Nah... It's just what I recalled. I'm on both ways.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/NodeImpl.java
tihom88: Done

tihom88:
Changes merged in trunk. Closing this PR

PR #160, opened 2019-11-04
OAK-8734 | Sorted query with multiple path restrictions and sorting p…

nit0906:
…erforms very slowly if one of the passed paths is a traversal

### DIFF
--- a/oak-core/src/test/java/org/apache/jackrabbit/oak/query/UnionQueryTest.java
+++ b/oak-core/src/test/java/org/apache/jackrabbit/oak/query/UnionQueryTest.java
@@ -17,6 +17,11 @@
 
 package org.apache.jackrabbit.oak.query;
 
+import static org.apache.jackrabbit.JcrConstants.JCR_PRIMARYTYPE;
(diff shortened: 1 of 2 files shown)


oak-core/src/test/java/org/apache/jackrabbit/oak/query/UnionQueryTest.java
```diff
@@ -74,6 +77,52 @@ public void after() throws Exception {
         root.commit();
     }
 
+    @Test
+    public void testMergeSortedVsConcat() throws  Exception {
```
thomasmueller, line 108: While this is a good test in general, and we should anyway keep it... Does this new test fail if the changes in the UnionQueryImpl above are _not_ done, that is, with the old code? (I didn't test, but I think the test doesn't fail...).

I think it would be possible to add a test that fails with the old code in UnionQueryImpl, but I guess I need to see myself (using a custom Query subclass in the test).

thomasmueller (changes requested):

PR #161, opened 2019-11-12
Clarified a pitfall with --merge-paths

sbluhm:
Clarified a pitfall with a merge operation, should a node be renamed.

### DIFF
--- a/oak-doc/src/site/markdown/migration.md
+++ b/oak-doc/src/site/markdown/migration.md
@@ -152,6 +152,8 @@ By default, the source repository replaces the destination repository (if there'
 
 Please notice that in the last example it's necessary to narrow the migration scope using `--include-paths` parameter.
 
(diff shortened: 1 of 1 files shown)


sbluhm:
@stefan-egli , any chance of merging this PR? It has been in my queue for a while.

jsedding:
@sbluhm I believe this pitfall is only true for referenceable nodes. Other nodes don't have a UUID property, and thus are not subject to the same uniqueness constraint. Maybe you can verify and include this detail as well?

sbluhm:
@jsedding , thank you for your pointer. This indeed applies only to UUIDs. I just verified it. I have revised the statement and re-submitted.

stefan-egli:
@jsedding do you think we should incorporate this PR?

jsedding:
@stefan-egli yes, I think this is helpful information and should be applied!

@sbluhm sorry this fell between the cracks!

PR #163, opened 2019-11-26
OAK-8780 - Remove the synchronized block on the FSBackend

ankitaagar:


oak-blob-plugins/src/test/java/org/apache/jackrabbit/oak/plugins/blob/datastore/FSBackendTest.java
```diff
@@ -0,0 +1,228 @@
...
+	            File tmpFile = fileFactory.createTransientFile("upload", null, tmp);
+	            MessageDigest digest = MessageDigest.getInstance(DIGEST);
+	            OutputStream output = new DigestOutputStream(new FileOutputStream(tmpFile), digest);
+	            try {
+	                 IOUtils.copyLarge(in, output);
+	            } finally {
+	                output.close();
+	            }
+	            DataIdentifier identifier = new DataIdentifier(encodeHexString(digest.digest()));
+	            backend.write(identifier, tmpFile);
```
amit-jain: The code after this should be executed concurrently. If the processing to get the id also happens within the thread we are not really testing he backend.write concurrently

oak-blob-plugins/src/test/java/org/apache/jackrabbit/oak/plugins/blob/datastore/FSBackendTest.java
```diff
@@ -0,0 +1,228 @@
...
+	        } catch (Exception e) {
+	            LOG.error("error:", e);
+	            fail(e.getMessage());
+	        }
+	    }
+
+	    @After
+	    public void tearDown() {
+	    	try {
+				ds.close();
```
amit-jain: Please correct indentation/formatting

amit-jain (changes requested):
Also, same file concurrent tests are missing

PR #165, opened 2019-11-29
fix doc link

mrozati:


mreutegg:
In the meantime this has been fixed with some other change.

PR #166, opened 2019-12-02
OAK-8780 - Remove the synchronized block on the FSBackend

ankitaagar:


oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/AbstractSharedCachingDataStore.java
```diff
@@ -24,10 +24,14 @@
 import java.io.IOException;
```
amit-jain: Why is there a need to change here. A temp file is already being created here. It won't add any benefit to create a another temp file in the local file system anyways.

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/datastore/FSBackend.java
```diff
@@ -105,20 +105,28 @@ public InputStream read(DataIdentifier identifier) throws DataStoreException {
...
-                    setLastModified(dest, now + ACCESS_TIME_RESOLUTION);
-                }
-            } else {
-                try {
-                    FileUtils.copyFile(file, dest);
-                } catch (IOException ie) {
-                    LOG.error("failed to copy [{}] to [{}]", file.getAbsolutePath(),
-                        dest.getAbsolutePath());
-                    throw new DataStoreException("Not able to write file [" + identifier + "]", ie);
+        if (dest.exists()) {
```
amit-jain: What is required is to copy the file already created in the caching layer (which is local fs) to a temp file in NFS (which the FSBackend would typically represent) and then do everything in the synchronized block.

oak-blob-plugins/src/test/java/org/apache/jackrabbit/oak/plugins/blob/datastore/FSBackendTest.java
```diff
@@ -0,0 +1,252 @@
+package org.apache.jackrabbit.oak.plugins.blob.datastore;
```
amit-jain: - Please add Apache license header.
- I think better convert this to IT i.e. add suffix IT. Would require -PintegrationTesting on mvn command line.

amit-jain (changes requested):
Summary of changes requires:
- Revert changes to AbstractSharedCachingDataStore - For explanation see specific comments.
- Copy to a new temporary file in FSBackend (i.e. which would be NFS), then rename/move in a synchronized block.
- Convert test into IT because it uses lot of threads

amit-jain:
@ankitaagar The changes in UploadStagingCache are not related to the issue. Please file another issue for the changes.

PR #170, opened 2019-12-18
OAK-8788: Add index creation and index completion time to indexes

tihom88:


oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+    private List<IndexName> getVersionedIndexNameObjects(Set<String> allIndexes) {
+        List<IndexName> indexNameObjectList = new ArrayList<>();
+        for (String indexNameString : allIndexes) {
+            indexNameObjectList.add(IndexName.parse(indexNameString));
+        }
+        return indexNameObjectList;
+    }
+
+    private Options parseCommandLineParams(String... args) throws Exception {
+        OptionParser parser = new OptionParser();
```
fabriziofortino: I think the command line parsing logic should be moved in `PurgeOldIndexVersionCommand`. This would make more sense from a logical point of view and would be in line with how the other commands are implemented.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class PurgeOldIndexVersion {
+    private static final Logger log = LoggerFactory.getLogger(PurgeOldIndexVersion.class);
```
fabriziofortino:
```suggestion
    private static final Logger LOG = LoggerFactory.getLogger(PurgeOldIndexVersion.class);
```

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+    private boolean isReadWriteRepository = false;
```
fabriziofortino: This is only used in `execute()`. I would remove it from here.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+    private long threshold;
+    private List<String> indexPaths;
+    private long DEFAULT_THRESHOLD = 432000000l; // 5 days in millis
```
fabriziofortino:
```suggestion
    private final static long DEFAULT_THRESHOLD = 432_000_000L; // 5 days in millis
```

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+        return indexNameObjectList;
+    }
+
+    private Options parseCommandLineParams(String... args) throws Exception {
+        OptionParser parser = new OptionParser();
+        OptionSpec<Long> thresholdOption = parser.accepts("threshold")
+                .withOptionalArg().ofType(Long.class).defaultsTo(DEFAULT_THRESHOLD);
+        OptionSpec<String> indexPathsOption = parser.accepts("index-paths", "Comma separated list of index paths for which the " +
+                "selected operations need to be performed")
+                .withOptionalArg().ofType(String.class).withValuesSeparatedBy(",").defaultsTo("/oak:index");
```
fabriziofortino: For consistency, I would move the default value (`/oak:index`) in a constant.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+    private boolean isReadWriteRepository = false;
+    private long threshold;
+    private List<String> indexPaths;
+    private long DEFAULT_THRESHOLD = 432000000l; // 5 days in millis
+
+    public void execute(String... args) throws Exception {
+        Options opts = parseCommandLineParams(args);
+        this.isReadWriteRepository = opts.getCommonOpts().isReadWrite();
+        if (!isReadWriteRepository) {
+            log.info("Repository connected in read-only mode. Use '--read-write' for write operations");
```
fabriziofortino: should we go ahead here since the repo is connected in read-only mode?

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+        String slashTrimmedIndexPath = PurgeOldVersionUtils.trimSlash(indexPath);
+        if (slashTrimmedIndexPath.startsWith(slashTrimmedIndexPathOption)) {
+            if (slashTrimmedIndexPathOption.length() == slashTrimmedIndexPath.length()
+                    || (slashTrimmedIndexPathOption.length() < slashTrimmedIndexPath.length()
+                    && (slashTrimmedIndexPath.charAt(slashTrimmedIndexPathOption.length()) == '/'
+                    || slashTrimmedIndexPath.charAt(slashTrimmedIndexPathOption.length()) == '-'))) {
+                return true;
+            } else {
+                return false;
+            }
```
fabriziofortino: the if-else statement could be slightly simplified 

```java
return slashTrimmedIndexPathOption.length() == slashTrimmedIndexPath.length() ||
        (slashTrimmedIndexPathOption.length() < slashTrimmedIndexPath.length()
                && (slashTrimmedIndexPath.charAt(slashTrimmedIndexPathOption.length()) == '/'
                || slashTrimmedIndexPath.charAt(slashTrimmedIndexPathOption.length()) == '-')
        );
```

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+            if (indexPaths == null) {
+                indexPaths = new HashSet<>();
+            }
+            indexPaths.add(PurgeOldVersionUtils.trimSlash(path));
+            segregatedIndexes.put(baseIndexPath, indexPaths);
+        }
+        return segregatedIndexes;
+    }
+
+    private Iterable<String> getIndexPathList(NodeStore store) throws CommitFailedException, IOException {
```
fabriziofortino: CFE and IOE are never thrown in the method

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+        IndexPathService indexPathService = new IndexPathServiceImpl(store);
+        Iterable<String> indexPaths = indexPathService.getIndexPaths();
+        return indexPaths;
```
fabriziofortino: this could be simplified in one line `return new IndexPathServiceImpl(store).getIndexPaths();`

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+            String indexName = splitPathArray[splitPathArray.length - 1];
+            indexVersionNames.add(indexName);
+        }
+        return indexVersionNames;
+    }
+
+    private List<IndexName> getIndexNameObjectList(Set<String> versionedIndexPaths) {
+        Set<String> allIndexNameVersions = getIndexVersionNames(versionedIndexPaths);
+        List<IndexName> indexNameObjectList = getVersionedIndexNameObjects(allIndexNameVersions);
+        return indexNameObjectList;
```
fabriziofortino, line 146: nit: remove unnecessary local variable `indexNameObjectList `

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+            for (String indexPath : indexPathList) {
+                if (isIndexPathOptionValid(indexPathOption, indexPath)) {
+                    filteredIndexPaths.add(PurgeOldVersionUtils.trimSlash(indexPath));
+                }
+            }
+        }
+        return filteredIndexPaths;
+    }
+
+    StringBuilder stringJoiner(String[] stringArray, int arrayIndexToJoin, String delimiter) {
```
fabriziofortino:
```suggestion
    private StringBuilder stringJoiner(String[] stringArray, int arrayIndexToJoin, String delimiter) {
```

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+
+    private List<IndexName> getVersionedIndexNameObjects(Set<String> allIndexes) {
+        List<IndexName> indexNameObjectList = new ArrayList<>();
+        for (String indexNameString : allIndexes) {
+            indexNameObjectList.add(IndexName.parse(indexNameString));
+        }
+        return indexNameObjectList;
+    }
+
+    private Options parseCommandLineParams(String... args) throws Exception {
```
fabriziofortino:
```suggestion
    private Options parseCommandLineParams(String... args) throws IOException {
```

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+                                      List<IndexVersionOperation> toDeleteIndexNameObjectList, String trimmedIndexPath) throws CommitFailedException {
+        String indexPathParent = "/" + trimmedIndexPath.substring(0, trimmedIndexPath.lastIndexOf('/'));
+        for (IndexVersionOperation toDeleteIndexNameObject : toDeleteIndexNameObjectList) {
+            NodeState root = store.getRoot();
+            NodeBuilder rootBuilder = root.builder();
+            NodeBuilder nodeBuilder = PurgeOldVersionUtils.getNode(rootBuilder, indexPathParent + "/" + toDeleteIndexNameObject.getIndexName().getNodeName());
+            if (nodeBuilder.exists()) {
+                if (toDeleteIndexNameObject.getOperation() == IndexVersionOperation.Operation.DELETE_HIDDEN_AND_DISABLE) {
+                    nodeBuilder.setProperty("type", "disabled", Type.STRING);
+                    PurgeOldVersionUtils.recursiveDeleteHiddenChildNodes(store, trimmedIndexPath.substring(0, trimmedIndexPath.lastIndexOf('/')) + "/" + toDeleteIndexNameObject.getIndexName().getNodeName());
```
fabriziofortino: if one recursive purge fails here (`CommitFailedException`), the other indexes that need to be deleted won't. Wondering if we should LOG the error here and try to delete the remaining ones.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+
+    public static List<IndexVersionOperation> applyOperation(NodeState indexDefParentNode,
+                                                             List<IndexName> indexNameObjectList, long threshold) {
+        int activeProductVersion = -1;
+        indexNameObjectList = getSortedVersionedIndexNameObjects(indexNameObjectList);
+        removeDisabledCustomIndexes(indexDefParentNode, indexNameObjectList);
+        List<IndexVersionOperation> indexVersionOperationList = new LinkedList<>();
+        for (int i = 0; i < indexNameObjectList.size(); i++) {
```
fabriziofortino: I would use the enhanced for and avoid the number of calls to `indexNameObjectList.get(i)`

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -0,0 +1,168 @@
...
+                    indexVersionOperationList.add(indexVersionOperation);
+                } else {
+                    IndexVersionOperation indexVersionOperation = new IndexVersionOperation(indexNameObjectList.get(i));
+                    indexVersionOperation.setOperation(Operation.DELETE);
+                    indexVersionOperationList.add(indexVersionOperation);
+                }
+            }
+        }
+        if (!isValidIndexVersionOperationList(indexVersionOperationList)) {
+            indexVersionOperationList = Collections.EMPTY_LIST;
```
fabriziofortino:
```suggestion
            indexVersionOperationList = Collections.emptyList();
```

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -0,0 +1,168 @@
...
+            }
+        });
+        return indexNameObjectList;
+    }
+
+    private static boolean isValidIndexVersionOperationList(List<IndexVersionOperation> indexVersionOperations) {
+        boolean isValid = false;
+        IndexVersionOperation lastNoopOperationIndexVersion = null;
+        IndexVersionOperation indexWithDeleteHiddenOp = null;
+        for (int i = 0; i < indexVersionOperations.size(); i++) {
```
fabriziofortino: I would use the enhanced for and avoid the number of calls to `indexVersionOperations.get(i)`

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -0,0 +1,168 @@
...
+    public IndexVersionOperation(IndexName indexName) {
+        this.indexName = indexName;
+        this.operation = Operation.NOOP;
+    }
+
+
+    public static List<IndexVersionOperation> applyOperation(NodeState indexDefParentNode,
+                                                             List<IndexName> indexNameObjectList, long threshold) {
+        int activeProductVersion = -1;
+        indexNameObjectList = getSortedVersionedIndexNameObjects(indexNameObjectList);
```
fabriziofortino: avoid reassigning values to incoming params. Better to use a temporary variable instead.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/ReindexOperations.java
```diff
@@ -58,9 +63,12 @@ public IndexDefinition apply(boolean useStateFromBuilder) {
         NodeState defnState = useStateFromBuilder ? definitionBuilder.getNodeState() : definitionBuilder.getBaseState();
         if (!IndexDefinition.isDisableStoredIndexDefinition()) {
             definitionBuilder.setChildNode(INDEX_DEFINITION_NODE, NodeStateCloner.cloneVisibleState(defnState));
+            if (definitionBuilder.getChildNode(STATUS_NODE).exists()) {
+                definitionBuilder.getChildNode(STATUS_NODE).removeProperty(REINDEX_COMPLETION_TIMESTAMP);
+                log.info(REINDEX_COMPLETION_TIMESTAMP + " property removed for index at {}", this.indexPath);
```
fabriziofortino:
```suggestion
                log.info("{} property removed for index at {}", REINDEX_COMPLETION_TIMESTAMP, this.indexPath);
```

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/ReindexOperations.java
```diff
@@ -22,14 +22,19 @@
...
 import static org.apache.jackrabbit.oak.plugins.index.search.IndexDefinition.INDEX_DEFINITION_NODE;
+import static org.apache.jackrabbit.oak.plugins.index.search.IndexDefinition.REINDEX_COMPLETION_TIMESTAMP;
+import static org.apache.jackrabbit.oak.plugins.index.search.IndexDefinition.STATUS_NODE;
 import static org.apache.jackrabbit.oak.plugins.index.search.spi.editor.FulltextIndexEditorContext.configureUniqueId;
 
 /**
  * Reindexing operations
  */
 public class ReindexOperations {
+    private static final Logger log = LoggerFactory.getLogger(ReindexOperations.class);
```
fabriziofortino:
```suggestion
    private static final Logger LOG = LoggerFactory.getLogger(ReindexOperations.class);
```

fabriziofortino (changes requested):

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
thomasmueller: I would write 5 * 24 * 60 * 60 * 1000L, that way it is clear how the number was calculated. Or TimeUnit.DAYS.toMillis(5)

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
thomasmueller: Yes I would continue, it is useful to do a dry-run.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+                String trimmedIndexParentPath = trimmedBaseIndexPath.substring(0, trimmedBaseIndexPath.lastIndexOf('/'));
+                List<IndexName> indexNameObjectList = getIndexNameObjectList(entry.getValue());
+                NodeState indexDefParentNode = NodeStateUtils.getNode(nodeStore.getRoot(),
+                        "/" + trimmedIndexParentPath);
+                List<IndexVersionOperation> toDeleteIndexNameObjectList =
+                        IndexVersionOperation.applyOperation(indexDefParentNode, indexNameObjectList, this.threshold);
+                if (isReadWriteRepository) {
+                    purgeOldIndexVersion(nodeStore, toDeleteIndexNameObjectList, trimmedBaseIndexPath);
+                } else {
+                    log.info("Repository is opened in read-write mode: IndexOperations" +
```
thomasmueller: I think the log statement is incorrect... It applies to "if not read write", so that would mean read-only... so you should log "Repository is opened in read-only mode, operations would be:..."

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
thomasmueller: I think two lines would also be good:
  IndexPathService indexPathService = new IndexPathServiceImpl(store);
  return indexPathService.getIndexPaths();

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
thomasmueller: I agree, parsing the command line options should go to the ...Command class.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/ReindexOperations.java
thomasmueller: I wouldn't mind the original REINDEX_COMPLETION_TIMESTAMP + ... here (it's more a matter of taste here I think).

oak-run/src/test/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersionTest.java
```diff
@@ -0,0 +1,325 @@
...
+                storeDir.getAbsolutePath(),
+                "--read-write",
+                "--threshold=1"
+        };
+
+        command.execute(args);
+        fixture = new RepositoryFixture(storeDir);
+        fixture.close();
+
+        Assert.assertFalse("Index:" + "fooIndex-2" + " deleted", fixture.getNodeStore().getRoot().getChildNode("oak:index").getChildNode("fooIndex-2").exists());
```
thomasmueller, line 97: It would probably be good to use multiple lines here (format the lines).

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldVersionUtils.java
```diff
@@ -0,0 +1,76 @@
...
+            startIndex++;
+        }
+        if (str.charAt(endIndex) == '/') {
+            endIndex--;
+        }
+        return str.substring(startIndex, endIndex + 1);
+    }
+
+    public static long getMillisFromString(String strDate) {
+        long millis = ISO8601.parse(strDate).getTimeInMillis();
```
thomasmueller: I would say "return ISO8601.parse(strDate).getTimeInMillis();" would be enough.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+        this.isReadWriteRepository = opts.getCommonOpts().isReadWrite();
+        if (!isReadWriteRepository) {
+            log.info("Repository connected in read-only mode. Use '--read-write' for write operations");
+        }
+        try (NodeStoreFixture fixture = NodeStoreFixtureProvider.create(opts)) {
+            NodeStore nodeStore = fixture.getStore();
+            Set<String> indexPathSet = filterIndexPaths(getIndexPathList(nodeStore));
+            Map<String, Set<String>> segregateIndexes = segregateIndexes(indexPathSet);
+            for (Map.Entry<String, Set<String>> entry : segregateIndexes.entrySet()) {
+                String trimmedBaseIndexPath = PurgeOldVersionUtils.trimSlash(entry.getKey());
```
thomasmueller: Would it be possible to use oak-commons PathUtils.getParentPath instead?

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldVersionUtils.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.jackrabbit.oak.spi.state.NodeStateUtils;
+import org.apache.jackrabbit.oak.spi.state.NodeStore;
+import org.apache.jackrabbit.util.ISO8601;
+import org.jetbrains.annotations.NotNull;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class PurgeOldVersionUtils {
+
+    public static String trimSlash(String str) {
```
thomasmueller: It's not clear from the name whether you trim at the beginning, or at the end, or both. Javadocs would help. Also, unit tests for this method are needed. Maybe oak-commons PathUtils can be used, so this method is no longer needed?

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+        return stringJoiner(fragmentedPath,
+                fragmentedPath.length - 2, "/")
+                .append(baseIndexName).toString();
+    }
+
+    private Set<String> getIndexVersionNames(Set<String> versionedIndexPaths) {
+        Set<String> indexVersionNames = new HashSet<>();
+        for (String indexPath : versionedIndexPaths) {
+            String trimmedIndexPath = PurgeOldVersionUtils.trimSlash(indexPath);
+            String[] splitPathArray = trimmedIndexPath.split("/");
```
thomasmueller: oak-commons PathUtils.elements should be used here.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+                    nodeBuilder.setProperty("type", "disabled", Type.STRING);
+                    PurgeOldVersionUtils.recursiveDeleteHiddenChildNodes(store, trimmedIndexPath.substring(0, trimmedIndexPath.lastIndexOf('/')) + "/" + toDeleteIndexNameObject.getIndexName().getNodeName());
+                } else if (toDeleteIndexNameObject.getOperation() == IndexVersionOperation.Operation.DELETE) {
+                    nodeBuilder.remove();
+                }
+                EditorHook hook = new EditorHook(
+                        new IndexUpdateProvider(new PropertyIndexEditorProvider()));
+                store.merge(rootBuilder, hook, CommitInfo.EMPTY);
+            } else {
+                log.error("nodebuilder null for path " + indexPathParent + "/" + toDeleteIndexNameObject.getIndexName().getNodeName());
```
thomasmueller: I think it's a bit weird to log an error, but not throw an exception. I would throw an exception, catch it at the caller, and then log an error there. To avoid having a lot of code indented, I would use:

    if (!nodeBuilder.exists()) {
        throw ...
    }
    ... code for normal case ...

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+        for (int index = 0; index <= arrayIndexToJoin; index++) {
+            stringBuilder.append(stringArray[index]);
+            stringBuilder.append(delimiter);
+        }
+        return stringBuilder;
+    }
+
+    private String getBaseIndexPath(String path) {
+        String trimmedPath = PurgeOldVersionUtils.trimSlash(path);
+        String[] fragmentedPath = trimmedPath.split("/");
```
thomasmueller: oak-commons PathUtils should be used.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+        String indexName = fragmentedPath[fragmentedPath.length - 1];
+        String baseIndexName = PurgeOldVersionUtils.getBaseIndexName(indexName);
+        return stringJoiner(fragmentedPath,
+                fragmentedPath.length - 2, "/")
```
thomasmueller: PathUtils could be used here as well (getParentPath, and concat).

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
thomasmueller: I would prefer not having this method and use PathUtils instead... if we need it, then maybe move to a utility method, add javadocs, make it static, and add unit tests.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+        return segregatedIndexes;
+    }
+
+    private Iterable<String> getIndexPathList(NodeStore store) throws CommitFailedException, IOException {
+        IndexPathService indexPathService = new IndexPathServiceImpl(store);
+        Iterable<String> indexPaths = indexPathService.getIndexPaths();
+        return indexPaths;
+    }
+
+    private boolean isIndexPathOptionValid(String indexPathOption, String indexPath) {
```
thomasmueller: I think this should be static, and unit tests should be added. Javadocs are needed.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+        String slashTrimmedIndexPathOption = PurgeOldVersionUtils.trimSlash(indexPathOption);
+        String slashTrimmedIndexPath = PurgeOldVersionUtils.trimSlash(indexPath);
+        if (slashTrimmedIndexPath.startsWith(slashTrimmedIndexPathOption)) {
+            if (slashTrimmedIndexPathOption.length() == slashTrimmedIndexPath.length()
```
thomasmueller: I would do one comparison at a time, for example:

    if slashTrimmedIndexPathOption.equals(slashTrimmedIndexPath)) {
        return true;
    }
    ...

That way, you get less and less cases as you continue.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,226 @@
...
+                    || (slashTrimmedIndexPathOption.length() < slashTrimmedIndexPath.length()
+                    && (slashTrimmedIndexPath.charAt(slashTrimmedIndexPathOption.length()) == '/'
+                    || slashTrimmedIndexPath.charAt(slashTrimmedIndexPathOption.length()) == '-'))) {
```
thomasmueller: I think the logic here is hard to understand... Maybe use PathUtils instead? E.g. PathUtils.isAncestor

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldVersionUtils.java
```diff
@@ -0,0 +1,76 @@
...
+        return str.substring(startIndex, endIndex + 1);
+    }
+
+    public static long getMillisFromString(String strDate) {
+        long millis = ISO8601.parse(strDate).getTimeInMillis();
+        return millis;
+    }
+
+    public static String getBaseIndexName(String versionedIndexName) {
+        String indexBaseName = versionedIndexName.split("-")[0];
```
thomasmueller: That would split at the first "-"... shouldn't it split at a specific place? What if the index path contains "-"? I think the logic should be moved to IndexName.

thomasmueller (changes requested):
I think it's a good patch! However, I would take the time to go over the possible improvements, hopefully it will simplify the code and make it easier for others to read the code.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
tihom88: Changed to PathUtils.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldVersionUtils.java
tihom88: Done , now using IndexName method getBaseName

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
tihom88: Changed logic, instead of checking path now using baseName from IndexName

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -0,0 +1,197 @@
...
+
+    public Operation getOperation() {
+        return this.operation;
+    }
+
+    public IndexName getIndexName() {
+        return this.indexName;
+    }
+
+    private Operation operation;
```
thomasmueller: nit-pick: I usually keep all fields at the beginning of the class (above functions)

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -0,0 +1,197 @@
...
+
+    public IndexVersionOperation(IndexName indexName) {
+        this.indexName = indexName;
+        this.operation = Operation.NOOP;
+    }
```
thomasmueller, line 63: nit-pick: I use the following order:
- public static final fields
- private static final fields 
- public static fields (usually none)
- private static fields
- public fields (usually none)
- private fields
- constructors
- member functions
- static functions
- inner classes
But that's just me.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -0,0 +1,197 @@
...
+
+    /**
+     * @param indexDefParentNode       NodeState of parent of baseIndex
+     * @param indexNameObjectListParam This is a list of IndexName Objects with same baseIndexName on which operations will be applied.
```
thomasmueller: Would "indexNameObjectList" also be OK?

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -0,0 +1,197 @@
...
+     * @param thresholdInMillis        after which a fully functional index is eligible for purge operations
```
thomasmueller: What about purgeThresholdMillis?

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -0,0 +1,197 @@
...
+     * @return This method returns an IndexVersionOperation list after applying the operations to be performed
```
thomasmueller: The signature of the method already tells us that it returns a list of IndexVersionOperation... What is not clear to me, does the method actually perform the operations (A), or does it just add the operations to the list, without actually performing them (B)? To me it looks like (B). But in this case, it would make sense to rename the function to for example "generateOperations". 


    @return the list of operations it performed

or

    @return the list of operations that should be performed

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -0,0 +1,197 @@
...
+                isValid = true;
+            }
+        }
+        if (!isValid) {
+            LOG.info("IndexVersionOperation List is not valid for index {}", lastNoopOperationIndexVersion.getIndexName().getNodeName());
+        }
+        return isValid;
+    }
+
+    private static void removeDisabledCustomIndexes(NodeState indexDefParentNode,
```
thomasmueller: Initially I thought this method is actually removing indexes from the repository, but looking at the method body it seems it only removes them from the list. To make that clear, what about renaming to "removeDisabledCustomIndexesFromList"?

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -0,0 +1,197 @@
...
+     * @return This method returns an IndexVersionOperation list after applying the operations to be performed
+     */
+    public static List<IndexVersionOperation> applyOperation(NodeState indexDefParentNode,
+                                                             List<IndexName> indexNameObjectListParam, long thresholdInMillis) {
+        int activeProductVersion = -1;
+        List<IndexName> indexNameObjectList = new ArrayList<>(indexNameObjectListParam);
+        indexNameObjectList = getSortedVersionedIndexNameObjects(indexNameObjectList);
+        removeDisabledCustomIndexes(indexDefParentNode, indexNameObjectList);
+        List<IndexVersionOperation> indexVersionOperationList = new LinkedList<>();
+        for (int i = 0; i < indexNameObjectList.size(); i++) {
```
thomasmueller: I would use either:

    for (IndexName in : indexNameObjectList) {

or

    for (int i = 0; i < indexNameObjectList.size(); i++) {
        IndexName in = indexNameObjectList.get(i);

The first version is a bit shorter; if possible I would use it. The second version allows you to do more stuff like removing / adding entries from / to the list, change i (repeat), or so.

In both cases, the advantage is that you can use "in" instead of "indexNameObjectList.get(i)" later in the loop.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -0,0 +1,197 @@
...
+        if (!isValidIndexVersionOperationList(indexVersionOperationList)) {
+            indexVersionOperationList = Collections.emptyList();
+        }
+        return indexVersionOperationList;
+    }
+
+    /*
+        returns IndexNameObjects in descending order of version i.e from latest to oldest
+    */
+    private static List<IndexName> getSortedVersionedIndexNameObjects(List<IndexName> indexNameObjectList) {
```
thomasmueller: I would rename to "sortedVersionedIndexNameObjects" because the method is modifying the list that was passed. (Which is fine). The method name "getSortedVersionedIndexNameObjects" to me means that it returns a different (cloned) list, and doesn't modify the list that was passed.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -0,0 +1,197 @@
...
+            }
+        });
+        return indexNameObjectList;
+    }
+
+    /**
+     * @param indexVersionOperations
+     * @return true if the IndexVersionOperation list passes following criteria.
+     * For merging indexes we need baseIndex and latest custom index.
+     * So we first validate that if there are custom indexes than OOTB index with same product must be marked with DELETE_HIDDEN_AND_DISABLE
```
thomasmueller, line 138: This is a quite important function, and I think it is correct... But would it make sense to add documentation, or even better test cases, with examples of what is valid and what is invalid, and why it's valid / invalid? If test case, the method would no longer be private, which I think is fine (no need to make it public... default access is fine).

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -0,0 +1,197 @@
...
+        }
+        return isValid;
+    }
+
+    private static void removeDisabledCustomIndexes(NodeState indexDefParentNode,
+                                                    List<IndexName> indexNameObjectList) {
+        for (int i = 0; i < indexNameObjectList.size(); i++) {
+            NodeState indexNode = indexDefParentNode
+                    .getChildNode(PathUtils.getName(indexNameObjectList.get(i).getNodeName()));
+            if (indexNode.getProperty("type") != null && indexNode.getProperty("type").getValue(Type.STRING).equals("disabled")) {
```
thomasmueller: It is fine like it is, I would keep it. But just a general remark about methods that can return null: you can often simplify the logic by reversing the comparison. I'm not sure if it would work here:

    if ("disabled".equals(indexNode.getProperty("type").getValue(Type.STRING)) {

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -0,0 +1,170 @@
...
+     */
+    private List<String> sanitiseUserIndexPaths(List<String> userIndexPaths) {
+        List<String> sanitisedUserIndexPaths = new ArrayList<>();
+        for (String userIndexPath : userIndexPaths) {
+            if (PathUtils.getName(userIndexPath).equals(PurgeOldVersionUtils.OAK_INDEX)) {
+                sanitisedUserIndexPaths.add(userIndexPath);
+            } else if (PathUtils.getName(PathUtils.getParentPath(userIndexPath)).equals(PurgeOldVersionUtils.OAK_INDEX)) {
+                sanitisedUserIndexPaths.add(IndexName.parse(userIndexPath).getBaseName());
+            } else {
+                throw new RuntimeException(userIndexPath + " indexpath is not valid");
```
thomasmueller: nit-pick: I would probably use IllegalArgumentException here

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
thomasmueller: Hm, seems to be necessary here... (maybe the code changed in the meantime)

thomasmueller (changes requested):
I think it's very good now. There are still a few things I like to see changed now... And I'm sorry for the nit-picking, if you like you can ignore that part.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldVersionUtils.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * @param nodeBuilder
+     * @param path        Path of node whose nodeBuilder object should be returned.
+     * @return nodeBuilder object of node at @param{path}
+     */
+    public static NodeBuilder getNode(@NotNull NodeBuilder nodeBuilder, @NotNull String path) {
+        for (String name : PathUtils.elements(checkNotNull(path))) {
+            nodeBuilder = nodeBuilder.getChildNode(checkNotNull(name));
```
fabriziofortino: I tend to not re-assign the method's parameters. You could create a temp variable here.

On a separate note, this type of logic refers to a [**fold**](https://en.wikipedia.org/wiki/Fold_(higher-order_function)) in the functional paradigm. This usually helps to write this logic in a more idiomatic way. In java this is possible but it requires a bit more "ceremony" compared to other languages:

```java
return StreamSupport.stream(PathUtils.elements(checkNotNull(path)).spliterator(), false)
    .reduce(nodeBuilder,
        (nb, name) -> nb.getChildNode(checkNotNull(name)),
        (nb, nb2) -> nb); // combinator function, not needed in sequential reductions but in java needs to be specified :-(
```

Here you can find a nice explanation of what happens underneath https://stackoverflow.com/a/33971436/1026224

fabriziofortino (approved):
LGTM just a minor possible improvement

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldVersionUtils.java
thomasmueller: In my view, it's a matter of preference whether to use the functional style or the procedural. I personally prefer the procedural style, at least here. And here, I don't think re-assigning the parameter is a problem. I would say, do what you prefer.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
tihom88: Changed

fabriziofortino (approved):

PR #171, opened 2020-01-10
OAK-8845 | Function based ordering should used index and not be sorte…

nit0906:
…d in Query Engine later on


fabriziofortino (approved):
LGTM 👍

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/FunctionIndexTest.java
```diff
@@ -240,6 +241,114 @@ public void upperCase() throws Exception {
...
+        List<String> paths = Lists.newArrayList();
+        for (int idx = 0; idx < 10; idx++) {
+            paths.add("/test/n" + idx);
+            if(idx%2 ==0) continue;
+            Tree a = test.addChild("n"+idx);
+            a.setProperty("jcr:primaryType", "nt:unstructured", Type.NAME);
+            a.setProperty("foo", "bar"+idx);
+
+        }
+        for(int idx =0 ; idx<10 ; idx++) {
```
thomasmueller: Formatting (spaces) are inconsistent, could you run the formatter please?

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -626,8 +626,8 @@ private boolean canSortByIndex() {
                         OrderingImpl o = orderings[i];
                         DynamicOperandImpl op = o.getOperand();
                         if (!(op instanceof PropertyValueImpl)) {
-                            // ordered by a function: currently not supported
-                            canSortByIndex = false;
+                            // ordered by a function - supported by index so we simply break here and set canSortByIndex = true
+                            // so that QueryImpl does not sort the entries again
                             break;
```
thomasmueller: I think breaking here is incorrect, in case of "order by a, b" and an index that only sorts by property a. Also, what if "order by a desc", and the index sorts ascending?

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
nit0906: @thomasmueller ...agree with the order a,b point ..will make the changes to handle that 

But with the desc/asc thing ..we don't define the in our indexes if it should sort asc or desc..we just add ordered = true on the prop..so shouldn't the index always  handle both the cases ? Or maybe I am missing something ?

nit0906: @thomasmueller  - even for a,b if the index doesn't support indexing on property b - this is effectively handled in this condition  if (list != null && list.size() == orderings.length) above the change I made. But somehow this only works in case of normal properties and not the function ones - looking into why that's the case. Seems like for normal properties plan.getSortOrder doesn't return any property that hasn't got oredered = true but in case of function props this doesn't seem to be working

thomasmueller: Hm, I might be wrong with the multi-column case... about asc / desc, I think Lucene can only sort ascending, not descending, but I might be wrong there too. I probably need to debug this.

thomasmueller: About the multi-column case: let's say the query has "order by lower(a), b, c" and the index can sort by "lower(a), x, y". Your code would stop (break) after the first column, and not check the second and third column. Right? Well currently it's more a theoretical problem, because I don't think the Lucene version we use can sort by more than one column at the same time... But for a remote index (Elasticsearch), it might be possible.

thomasmueller: I think the proper code would look something like this. This is not tested; I'm not sure if e.getPropertyName() will return the function name as expected.

    for (int i = 0; i < list.size(); i++) {
        OrderEntry e = list.get(i);
        OrderingImpl o = orderings[i];
        DynamicOperandImpl op = o.getOperand();
        if (op instanceof PropertyValueImpl) {
            String pn = ((PropertyValueImpl) op).getPropertyName();
            if (!pn.equals(e.getPropertyName())) {
                // ordered by another property
                canSortByIndex = false;
                break;
            }
        } else {
            // ordered by a function
            String fn = op.getFunction(selectors.get(0));
            if (!fn.equals(e.getPropertyName())) {
                // ordered by a different function
                canSortByIndex = false;
                break;
            }
        }
        // we only have one selector, so no need to check that
        // TODO support joins
        if (o.isDescending() != (e.getOrder() == Order.DESCENDING)) {
            // ordered ascending versus descending
            canSortByIndex = false;
            break;
        }
    }

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
nit0906: @thomasmueller  - It seems to me as if this check is redundant 
```
if (!pn.equals(e.getPropertyName())) {
            // ordered by another property
            canSortByIndex = false;
            break;
        }
```

I haven't been able to create a test that would actually get the flow here. It seems to me that this check is for the case if the query is something like order by a,b whereas the index is ordered on let's say a,c - in that case this should kick in and set canSortByIndex = false. But in that case 
`if (list != null && list.size() == orderings.length)`
This would return false and the name condition check wouldn't even come into picture. I am debugging a bit further to verify this.

thomasmueller: > I haven't been able to create a test that would actually get the flow here

Yes, because I don't think we currently support indexes that order by more than one property. But we might support this in the future.

> It seems to me that this check is for the case if the query is something like order by a,b whereas the index is ordered on let's say a,c

Yes, exactly.

> in that case this should kick in and set canSortByIndex = false.

Yes.

> But in that case if (list != null && list.size() == orderings.length)  This would return false 

No, it would return true: both list.size() and ordering.length would be 2:
* List<OrderEntry> list = plan.getSortOrder(); // from the index
* orderings // field in the QueryImpl, from the query

thomasmueller: Note that the index might not need to support ordering by multiple properties: maybe the index "knows" that some property is always the same value. In this case the index could say "yes I can order by this property". In theory the query could be "order by a, a" (even thought that's weird, no need to support it). Right now it's a theoretical case, and existing indexes don't support ordering by multiple columns. So I don't think a test case for this is needed right now. But it makes sense to have the right logic in the code.

nit0906: > No, it would return true: both list.size() and ordering.length would be 2:
> 
>     List list = plan.getSortOrder(); // from the index
>     orderings // field in the QueryImpl, from the query
> 

Actually plan.getSortOrder() in this case would return 1.
If the query is something like order by a,b whereas the index is ordered on let's say a,c -> then the ordering.size would be 2 (from the query). But plan.getSortOrder() would basically be the number of properties in the index that can sort on the given query (which would be 1 in this case i.e. a ).

I have written a few test cases around this that help in debugging and understanding this workflow. Also, using them it appears that indexes do support multi column ordering. I will push in those changes and let you have a look.

nit0906: @thomasmueller  - I pushed in a temp change commenting out the check around property name that I thought was redundant. I have added a few test cases in org.apache.jackrabbit.oak.plugins.index.lucene.FunctionIndexTest - testOrdering4,testOrdering5,testOrdering6 (have added comments specifying what they do) - with those I was able to debug to indicate that indexes do support multi column ordering and  (list != null && list.size() == orderings.length) would handle any case in which query tries to order on any property that does not supports ordering in the index.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/FunctionIndexTest.java
nit0906: Done

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -625,19 +625,15 @@ private boolean canSortByIndex() {
...
                         DynamicOperandImpl op = o.getOperand();
-                        if (!(op instanceof PropertyValueImpl)) {
-                            // ordered by a function: currently not supported
-                            canSortByIndex = false;
-                            break;
-                        }
                         // we only have one selector, so no need to check that
                         // TODO support joins
-                        String pn = ((PropertyValueImpl) op).getPropertyName();
+                        String pn = op.getOrderEntryPropertyName(selectors.get(0));
```
thomasmueller: getOrderEntryPropertyName can in theory return null, then you get a NullPointerException in QueryImpl. I know, it can only happen in theory, but avoiding a potential NPE would be good. I think it should be relatively easy, for example throw a different exception with an appropriate message, if getOrderEntryPropertyName returns null.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/FullTextSearchScoreImpl.java
```diff
@@ -118,4 +118,9 @@ public OrderEntry getOrderEntry(SelectorImpl s, OrderingImpl o) {
         return null;
     }
 
+    @Override
+    public String getOrderEntryPropertyName(SelectorImpl s) {
+        return null;
```
thomasmueller: Order by score is possible... I think it is always ignored right now, but that's in implementation. If it always returns null, then the logic in QueryImpl would throw an exception (NPE right now). I think that's not correct. Instead, it's probably better to return the name of the score pseudo-property here.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/NodeLocalNameImpl.java
```diff
@@ -149,4 +149,13 @@ public OrderEntry getOrderEntry(SelectorImpl s, OrderingImpl o) {
             OrderEntry.Order.DESCENDING : OrderEntry.Order.ASCENDING);
     }
 
+    @Override
+    public String getOrderEntryPropertyName(SelectorImpl s) {
+        String fn = getFunction(s);
+        if (fn != null) {
+            return  QueryConstants.FUNCTION_RESTRICTION_PREFIX + fn;
+        }
```
thomasmueller: We have some of repeated code, as this method is always the same. Shouldn't we rather write the method in the base class (if there is one)? To avoid repeating code.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndexPlanner.java
```diff
@@ -884,7 +884,7 @@ private static boolean isEqualityRestriction(PropertyRestriction pr) {
                 orderEntries.add(IndexDefinition.NATIVE_SORT_ORDER);
             }
             for (PropertyDefinition functionIndex : rule.getFunctionRestrictions()) {
-                if (o.getPropertyName().equals(functionIndex.function)) {
+                if (o.getPropertyName().equals(functionIndex.function) && functionIndex.ordered) {
```
thomasmueller: Ah, is this a bugfix?

I would probably reverse the order of the condition:
    functionIndex.ordered && o.getPropertyName().equals(functionIndex.function)
because testing a boolean is faster than comparison, but also because of readability of the code... first check if the property is ordered in the index, and then test if it matches.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
nit0906: @thomasmueller  - Instead of throwing an exception - maybe we can simply set canSortByIndex = false in QueryImpl in case getOrderEntryPropertyName returns null (that way the query would still return result , sorted by QueryEngine). WDYT ?

oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/FullTextSearchScoreImpl.java
nit0906: Actually getOrderEntryPropertyName basically is needed to compare the property name set for the Ordering any particular implementation of DynamicOperand - In case of FullTextSearchScoreImpl - no Order is set , so I thought maybe it would be better to return null from getOrderEntryPropertyName  too.

But I can return score pseudo-property here .

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndexPlanner.java
nit0906: Yes - this was a bugfix. Agreed - will make the ordering change

thomasmueller (approved):
LGTM

PR #174, opened 2020-01-27
OAK-8711 - Queries with facets should not use traversal

averma21:


nit0906 (changes requested):
@averma21  - It would be better to add an extended test as well in oak-lucene where we have an index created that can serve the facet query but the cost for traverse comes out to be less. Such a test case should fail without your change and pass with the change.

averma21:
@nit0906 `QueryResult#getRows()` throws exception even without my changes since `TraversingCursor` doesn't work with facets. So what should be the test flow?

nit0906:
> @nit0906 `QueryResult#getRows()` throws exception even without my changes since `TraversingCursor` doesn't work with facets. So what should be the test flow?

Yes - so we need to create a test to check that if there is a index present to serve the query then TraversingCursor should not be picked up.

So the test flow should be something where without your fix TraversingCursor is picked up and ends up in exception (and hence the test fails) but with your fix the actual index will pick up and serve the facet query.

averma21:
Actually I am not very sure what the intent of that test would be. Since we are now returning infinity as a cost for traversal in case of facets query, I think traversal would never be picked up in that case. To me it seems that writing the other test would mean that we are checking that a higher cost index is not picked up if a lower cost index is present and I hope that logic should already be tested elsewhere.

thomasmueller:
The patch does contain a test case that (to me) seems to address the comment from @nit0906 . Or maybe I misunderstood the comment?

Other than that, the patch looks good to me.

nit0906:
The test sort of handles what I was asking for - basically a condition where traversal would have a lower cost than an index but the expected flow would be that index gets picked up instead of traversal (because traversal in this case would throw an exception.) So I think we are good.

PR #175, opened 2020-02-03
OAK-8892: Add javadoc to package-info files

averma21:


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/util/fv/package-info.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This package contains tokenization related classes.
```
thomasmueller: I believe "FV" stands for "feature vector", and I think that should be added in the package documentation. And probably a link to https://en.wikipedia.org/wiki/Feature_(machine_learning). The documentation about this in Oak is here: http://jackrabbit.apache.org/oak/docs/query/lucene.html#similar-fv - maybe you want to read this as well and update the Javadocs a bit.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/package-info.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This package contains index creation and editing related classes and interfaces.
```
thomasmueller: "editing related" is not the right term I think... The word "edit" comes from the oak "Editor", which is the mechanism used to observe changes (like a "trigger" in a relational database).

I would write something like "Contains classes that deal with updating indexes, including how to build indexing backend documents (e.g. Lucene documents)".

thomasmueller (changes requested):

thomasmueller (approved):

PR #177, opened 2020-02-05
OAK-8659: deprecate lucene custom scorer

fabriziofortino:
https://issues.apache.org/jira/browse/OAK-8659


averma21:
Would it make sense to deprecate the package also through annotation in package-info file?

fabriziofortino:
@averma21 it does. Good catch!

thomasmueller (approved):

fabriziofortino:
Merged https://issues.apache.org/jira/browse/OAK-8659

PR #180, opened 2020-02-10
OAK-8898: On querying, IndexReader failed with AlreadyClosedException

tihom88:


oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/update/TimedRefreshPolicy.java
```diff
@@ -53,16 +54,20 @@ public void refreshOnWriteIfRequired(Runnable refreshCallback) {
...
-            long currentTime = clock.getTime();
-            if (currentTime - lastRefreshTime > refreshDelta
-                    && dirty.compareAndSet(true, false)){
-                lastRefreshTime = currentTime;
-                refreshCallback.run();
+        synchronized (lock) {
+            if (dirty.get()) {
+                long currentTime = clock.getTime();
+                if (currentTime - lastRefreshTime > refreshDelta
+                        && dirty.compareAndSet(true, false)) {
```
averma21: Could you please explain what is the intention here. Do we want that dirty stays `false` while `refreshCallback` is running?

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/update/TimedRefreshPolicy.java
```diff
@@ -53,16 +54,20 @@ public void refreshOnWriteIfRequired(Runnable refreshCallback) {
...
     }
 
     private void refreshIfRequired(Runnable refreshCallback) {
-        if (dirty.get()){
-            long currentTime = clock.getTime();
-            if (currentTime - lastRefreshTime > refreshDelta
-                    && dirty.compareAndSet(true, false)){
-                lastRefreshTime = currentTime;
-                refreshCallback.run();
+        synchronized (lock) {
```
averma21: Could you please explain what is the intention behind synchronizing on `lock` . Do we want that dirty stays `false` while `refreshCallback` is running?

thomasmueller: I think it's not needed to synchronize updated(). It should be enough to synchronize just the call to refreshCallback.run() in my view.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexNodeManager.java
```diff
@@ -272,7 +271,9 @@ private void closeReaders(Iterable<LuceneIndexReader> readers) {
     }
 
     private void releaseHolder(SearcherHolder holder) {
-        decrementSearcherUsageCount(holder.searcher);
+        synchronized (holder) {
```
thomasmueller: Shouldn't we also synchronize closeReaders(), together with releaseHolder(..)?

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/update/TimedRefreshPolicy.java
```diff
@@ -21,6 +21,7 @@
 
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
```
averma21: This `AtomicLong` import is not needed now.

tihom88:
Looks like we were looking at this in a wrong way. What we want is to change it in such a way so that release holder in not called twice on a holder. So removed all locks and instead of blocking thread now using an atomic boolean to make threads bypass decrementing references of index reader.

PR #184, opened 2020-02-17
OAK-8902 | Getting list of blob ids under a specific path

nit0906:


oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java
```diff
@@ -142,6 +144,10 @@ static int checkDataStore(String... args) {
             // Optional argument to specify tracking
             OptionSpecBuilder verbose = parser.accepts("verbose", "Output backend formatted ids/paths");
 
+            // Optional argument to specify root path under which tracking if to be done. Defaults to "/" if not specified
+            ArgumentAcceptingOptionSpec verboseRootPath = parser.accepts("verboseRootPath", "Root path to output backend formatted ids/paths")
```
thomasmueller: This line is a bit long.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java
```diff
@@ -117,7 +119,7 @@ static int checkDataStore(String... args) {
         String helpStr =
             "datastorecheck [--id] [--ref] [--consistency] [--store <path>|<mongo_uri>] "
                 + "[--s3ds <s3ds_config>|--fds <fds_config>|--azureblobds <azureblobds_config>|--nods]"
-                + " [--dump <path>] [--repoHome <repo_home>] [--track] [--verbose]";
+                + " [--dump <path>] [--repoHome <repo_home>] [--track] [--verbose] [--verboseRootPath <verbose_root_path>]";
```
thomasmueller: A bit long, could you split the line?

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java
```diff
@@ -242,16 +248,26 @@ static int checkDataStore(String... args) {
...
 
             if (options.has(refOp) || options.has(consistencyOp)) {
-                if (options.has(verbose) &&
+
+                // Find blob ids by traversal for verbose mode + Segment Store or if verboseRootPath option
+                // is present (find blob references under a specific root path.)
+                if ((options.has(verbose) &&
                     (nodeStore instanceof SegmentNodeStore ||
-                        nodeStore instanceof org.apache.jackrabbit.oak.segment.SegmentNodeStore)) {
+                        nodeStore instanceof org.apache.jackrabbit.oak.segment.SegmentNodeStore)) || options.has(verboseRootPath)) {
```
thomasmueller: Line is a bit long

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java
```diff
@@ -242,16 +248,26 @@ static int checkDataStore(String... args) {
...
                     NodeTraverser traverser = new NodeTraverser(nodeStore, dsType);
                     closer.register(traverser);
-                    traverser.traverse();
+
+                    if (options.has(verboseRootPath)) {
+                        List<String> rootPathList = options.valuesOf(verboseRootPath);
+                        traverser.traverse((String[]) rootPathList.toArray(new String[rootPathList.size()]));
```
thomasmueller: I think this case isn't needed.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java
```diff
@@ -486,22 +502,31 @@ private void binaryProperties(NodeState state, String path, BufferedWriter write
             for (PropertyState p : state.getProperties()) {
                 String propPath = PathUtils.concat(path, p.getName());
                 try {
+                    String id ;
                     if (p.getType() == Type.BINARY) {
-                        count.incrementAndGet();
+                        id = p.getValue(Type.BINARY).getContentIdentity();
+                        // Ignore inline encoded binaries in document mk and null references in segment mk
+                        if (id == null || id.startsWith(INLINE_BINARY_SUFFIX)) continue;
```
thomasmueller: I think this is a Blob... would it be possible to use isInlined() instead of startsWith? That way, INLINE_BINARY_SUFFIX is not needed.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java
```diff
@@ -526,7 +551,19 @@ public void traverse() throws IOException {
 
             try {
                 writer = Files.newWriter(references, Charsets.UTF_8);
-                traverseChildren(nodeStore.getRoot(), "/", writer, count);
+                if (paths.length == 0) {
+                    traverseChildren(nodeStore.getRoot(), PATH_DELIM, writer, count);
+                } else {
+                    for (String path: paths ) {
+                        String[] nodeList = path.split(PATH_DELIM);
```
thomasmueller: Could you use org.apache.jackrabbit.oak.commons PathUtils.elements instead?

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java
```diff
@@ -526,7 +551,19 @@ public void traverse() throws IOException {
...
+                        NodeState state = nodeStore.getRoot();
+                        for (String node: nodeList) {
+                            if (node.isEmpty()) continue;
```
thomasmueller: I don't think that's needed... That would mean the path is something like "a//b", right? Instead of ignoring the empty entry, I would throw an exception, or do what PathUtils does normally.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java
```diff
@@ -97,6 +97,8 @@
     private static final String AZUREDS = "--azureblobds";
     private static final String DASH = "-";
     private static final String HASH = "#";
+    private static final String INLINE_BINARY_SUFFIX = "0x";
```
thomasmueller: Shouldn't be needed (see below)

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java
```diff
@@ -97,6 +97,8 @@
...
+    private static final String PATH_DELIM = "/";
```
thomasmueller: Shouldn't be needed (see below)

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java
thomasmueller: Does this mean there are two instanceof checks?

    (nodeStore instanceof SegmentNodeStore ||
    nodeStore instanceof org.apache.jackrabbit.oak.segment.SegmentNodeStore

And why do we do an instanceof check at all? Shouldn't DocumentNodeStore be supported as well?

thomasmueller (changes requested):

thomasmueller:
Could you create a PR for https://github.com/oak-indexing/jackrabbit-oak instead of https://github.com/apache/jackrabbit-oak ? That way, we would see comments from codacy as well, similar to  https://app.codacy.com/gh/oak-indexing/jackrabbit-oak/pullRequest?prid=5022787

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java
nit0906: Actually "/oak:index".split("/") would return ["","oak:index"] -> so this ignores the first empty entry there. But yes , I think PathUtils.elements could be used here (I didn't know about that.)

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java
nit0906: This check was already there actually, and this now is needed to check a specific case, where if --verboseRootPath is not present but only --verbose is present - then in case of documentMk we don't need the traversal approach there, since the blobReferenceCollector also gives us the paths to the nodes and hence the check for segmentMK, but if we have --verboseRootPath (finding blobIds under a specific path) , then It needs to do traversal for document mk as well.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java
nit0906: options.has(verboseRootPath) - this check ? Yes , makes sense, options.valuesOf(verboseRootPath) returns an empty list in case that option is not specified - I was thinking that this might throw an NPE or something

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java
thomasmueller: I understand. You are right!

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java
thomasmueller: Ah this is old SegmentStore and new SegmentStore.

PR #188, opened 2020-02-19
Oak 8902

nit0906:


amit-jain:
For the current patch I have a fundamental question, when requiring to retrieve references do you require to do the whole consistency check and the gc command to be working? 

My biggest concern is that the added option (–verboseRootPath) is applicable to both the the commands i.e. {{--consistency-check}} & {{--collect-garbage}}. But if run with --collect-garbage it will end up completely wiping the DataStore save the blobs referenced from the path given. While I think consistency check option is ok.
So, there are 2 changes required:
* Limit the new option to only when specified along with consistency-check 
* Need addition of tests to cover the cases (at least 3 to cover consistency check pass/fail and one to disallow new option with collect-garbage)

PR #193, opened 2020-03-03
OAK-8934: Indexing: filter entries with a regular expression

averma21:


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/DefaultIndexWriter.java
```diff
@@ -80,10 +83,22 @@ public DefaultIndexWriter(LuceneIndexDefinition definition, NodeBuilder definiti
 
     @Override
     public void updateDocument(String path, Iterable<? extends IndexableField> doc) throws IOException {
+        Iterator<? extends IndexableField> f = doc.iterator();
```
averma21, line 85: The change to this method is causing 7 tests to fail.
```
[ERROR]   DefaultIndexReaderFactoryTest.indexDir:90 expected:<1> but was:<0>
[ERROR]   DefaultIndexReaderFactoryTest.indexDirWithBlobStore:128 expected:<1> but was:<0>
[ERROR]   DefaultIndexReaderFactoryTest.multipleReaders:174 expected:<2> but was:<1>
[ERROR]   MultiplexingIndexWriterTest.deleteIncludingMount:240 expected:<1> but was:<0>
[ERROR]   MultiplexingIndexWriterTest.deletes:209 expected:<2> but was:<0>
[ERROR]   MultiplexingIndexWriterTest.writesInDefaultMount:117 expected:<1> but was:<0>
[ERROR]   MultiplexingIndexWriterTest.writesInDefaultMountBlobStore:165 expected:<1> but was:<0>
```
These tests expect even the document only containing `:path` field to be added to index. I am not sure if that completely makes sense and what is the use case.
In addition to checking `containsOnlyPath` if i also check `definition.getPropertyRegex() != null` in lines 90 and 97 (i.e. before returning in case of reindex and before deleting the document in the other case), the tests pass.

thomasmueller (approved):

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/DefaultIndexWriter.java
```diff
@@ -30,10 +31,12 @@
 import org.apache.jackrabbit.oak.plugins.index.lucene.LuceneIndexDefinition;
 import org.apache.jackrabbit.oak.plugins.index.lucene.directory.DirectoryFactory;
 import org.apache.jackrabbit.oak.plugins.index.lucene.util.SuggestHelper;
+import org.apache.jackrabbit.oak.plugins.index.search.FieldNames;
 import org.apache.jackrabbit.oak.plugins.index.search.IndexDefinition;
 import org.apache.jackrabbit.oak.spi.state.NodeBuilder;
 import org.apache.jackrabbit.util.ISO8601;
 import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.document.Document;
```
fabriziofortino: unused import
```suggestion
```

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/DefaultIndexWriter.java
thomasmueller: > These tests expect even the document only containing :path field to be added to index. I am not sure if that completely makes sense and what is the use case.

I don't know... However if feels a bit risky to change the behavior. What if we keep the current behavior (even if I agree it is weird), and open a new ticket to improve this area? So that documents that only contain :path are not indexed. You can then immediately add that part of your patch there.

Or, as you write, add the check definition.getPropertyRegex() != null in lines 90 and 97. That's also an option. (Even thought, decoupling the two aspects is a bit cleaner I think).

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -128,6 +129,7 @@ public D makeDocument(NodeState state, boolean isUpdate, List<PropertyState> pro
 
         D document = initDoc();
         boolean dirty = false;
+        Pattern propertyRegex = definition.getPropertyRegex();
```
fabriziofortino:
```suggestion
```

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -187,7 +189,10 @@ public D makeDocument(NodeState state, boolean isUpdate, List<PropertyState> pro
         }
 
         if (indexingRule.isFulltextEnabled()) {
-            indexFulltextValue(document, name);
+            boolean shouldAdd = propertyRegex == null || propertyRegex.matcher(name).find();
```
fabriziofortino, line 192: I would move the `propertyRegex` declaration inside the condition where it is used
```suggestion
            Pattern propertyRegex = definition.getPropertyRegex();
            boolean shouldAdd = propertyRegex == null || propertyRegex.matcher(name).find();
```

fabriziofortino (approved):
LGTM. Just a couple of minors

thomasmueller (approved):

PR #196, opened 2020-03-11
[OAK-8948] - Handle null entry in 'excludePaths' Set.

rafiya:


tihom88:
@rafiya ,
I think we can have null check at https://github.com/apache/jackrabbit-oak/blob/db2bff1033a7336069a39eff61563eac91006670/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/observation/ObservationManagerImpl.java#L394

rafiya:
@tihom88  review comments accomodated. Can you review?

PR #198, opened 2020-03-16
OAK-8898 On querying, IndexReader failed with AlreadyClosedException

tihom88:


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/FacetTestHelper.java
```diff
@@ -0,0 +1,38 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(FacetTestHelper.class);
+
+    public static void sleep(int timeInMillis) {
+        try {
+            if (timeInMillis > 0) {
+                LOG.info("Sleep time set to:" + timeInMillis + " ms");
+                Thread.sleep(timeInMillis);
+            }
+        } catch (InterruptedException e) {
+            e.printStackTrace();
```
thomasmueller: I would use LOG.error here

thomasmueller (changes requested):
Looks good! (Except for one tiny change)

PR #212, opened 2020-05-20
Feature/oak 9068 aws improvements

synox:


oak-segment-aws/pom.xml
```diff
@@ -175,13 +175,13 @@
             <groupId>org.apache.jackrabbit</groupId>
             <artifactId>oak-segment-tar</artifactId>
             <version>${project.version}</version>
-            <scope>provided</scope>
+            <scope>compile</scope>
```
dulceanu: @synox, I don't think we need to change the scope of these two dependencies, for both oak-segment-aws and oak-segment-azure modules.

dulceanu (changes requested):
LGTM, apart from the change in scope of those dependencies.

synox:
Thanks, I will update and create patch.

PR #203, opened 2020-04-20
OAK-9026 Performance Improvement: Unnecessary boxing and unboxing

larsgrefer:

PR #213, opened 2020-05-20
OAK-9069 Remove unnecessary (un)boxing in oak-auth-external

larsgrefer:


reschke (commented):
All these affect only tests. Why do you care?

larsgrefer:
I've just split #203 up by module as requested here: https://issues.apache.org/jira/browse/OAK-9026?focusedCommentId=17091262&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17091262

But I agree that these changes aren't that important :smile:

anchela:
@reschke while it's for sure not super critical, i still find it valuable to care about the quality of test cases. i will look at it and have to say that i appreciate every contribution to the apache open source projects. starting with something easy is IMO a very good way to become familiar with a code base as a very first step. so, thanks @larsgrefer  for taking the time to report this and split it up into subtasks.

PR #231, opened 2020-06-01
GRANITE-30089

Vinod-holani:
Getting {String or binary data would be truncated... } error as it is not handled cause error code for the same is updated recently in MSSQL

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentStore.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentStore.java
@@ -2128,7 +2128,8 @@ public class RDBDocumentStore implements DocumentStore {
 
     private static void continueIfStringOverflow(SQLException ex) throws SQLException {
         String state = ex.getSQLState();
-        if ("22001".equals(state) /* everybody */|| ("72000".equals(state) && 1489 == ex.getErrorCode()) /* Oracle */) {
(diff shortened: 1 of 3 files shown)


reschke (commented):
It would be good to understand why we don't have a failing test for this (and add that first).

Vinod-holani:
> It would be good to understand why we don't have a failing test for this (and add that first).

We do have test case for this. But by default that test class is disabled. I enabled them and saw test case was failing but silently so I have updated the test class with fix and also rethrowing caught exception. For disabled test case See [0]
[0] : https://github.com/apache/jackrabbit-oak/blob/trunk/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentStorePerformanceTest.java#L43

Shouldn't we enable this test class by default also? WDYT?

reschke:
That one is disabled because it's a long-running lowlevel test.

What I'm looking for is a test on the DocumentStore API (or even a higher layer) that reproduces the issue. In other words, the code path that handles the expected SQLException doesn't seem to be covered by a regular test.

Vinod-holani:
While updating major driver version or validating on new database version. All tests should be executed. Also this issue happens only under special case where column limit crosses in short duration with heavy updates. So this would come under performance validation which seems ok.

reschke:
> While updating major driver version or validating on new database version. All tests should be executed.

It should be possible to construct a unit test with the exact sequence of UpdateOps that reproduces this case.

Vinod-holani:
@reschke could you have a look on PR again??

reschke (approved):
If this reproduces the issue, great.

(Other than that, indentation looks slightly funny)

mreutegg:

https://issues.apache.org/jira/browse/OAK-9096

PR #238, opened 2020-07-02
OAK-9131 : introducing scanners for references to nt:frozenNode

stefan-egli:


oak-run/src/main/java/org/apache/jackrabbit/oak/run/FrozenNodeRefsByScanningCommand.java
```diff
@@ -0,0 +1,293 @@
...
+ * <p/>
+ * Test Example for creating a reference:
+ * <pre>
+ * java -mx4g -jar oak-run-*.jar frozennoderefsbyscanning mongodb://localhost/&lt;dbname&gt; -user=admin -password=admin -testCreateRefPath=&lt;mypath&gt; -testCreateRefProp=&lt;mypropertyname&gt; -testCreateRefUuid=&lt;myuuid&gt; -read-write=true
+ * </pre>
+ */
+public class FrozenNodeRefsByScanningCommand implements Command {
+
+    static {
+        // disable any query limits as our query is going to be a fulls can, and we are aware of it
```
mreutegg:
```suggestion
        // disable any query limits as our query is going to be a full scan, and we are aware of it
```

oak-run/src/main/java/org/apache/jackrabbit/oak/run/FrozenNodeRefsByScanningCommand.java
```diff
@@ -0,0 +1,293 @@
...
+
+        //		// disable the WARN of the TraversingCursor
+        //		LoggerContext c = (LoggerContext) LoggerFactory.getILoggerFactory();
+        //		Logger logger = c.getLogger("org.apache.jackrabbit.oak.plugins.index.Cursors$TraversingCursor");
+        //		logger.setLevel(Level.ERROR);
+    }
+
+    public static final String NAME = "frozennoderefsbyscanning";
+
+    private final String summary = "Scans and lists all references to nt:frozenNode";
```
mreutegg:
```suggestion
    private final String summary = "Scans repository and lists all references to nt:frozenNode";
```

oak-run/src/main/java/org/apache/jackrabbit/oak/run/FrozenNodeRefsByScanningCommand.java
```diff
@@ -0,0 +1,293 @@
...
+        String createRefUuid = testCreateRefUuidOption.value(options);
+
+        Closer closer = Utils.createCloserWithShutdownHook();
+        closer.register(nodeStoreFixture);
+        try {
+
+            System.out.println("Logging in...");
+            Session session = openSession(nodeStore, "crx.default", user, password);
+
+            if (createRefPath != null && createRefProp != null && createRefUuid != null) {
```
mreutegg: I would move this part into a dedicated test class.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/FrozenNodeRefsByScanningCommand.java
```diff
@@ -0,0 +1,293 @@
...
+                // this is where we ultimately have to continue out in any case - as only an nt:frozenNode
+                // is what we're interested in.
+                return;
+            }
+
+            String uuid = propValue;
+            String referrerPath = n.getPath();
+            String referrerProperty = p.getName();
+            FrozenNodeRef ref = new FrozenNodeRef(referrerPath, referrerProperty, PropertyType.nameFromValue(p.getType()), uuid, path);
+            list.add(ref);
```
mreutegg: Won't this consume too much memory when there are many references? The only use of the list seems to be to decide on the return of `uuidscan()`.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/FrozenNodeRefsUsingIndexCommand.java
```diff
@@ -0,0 +1,159 @@
...
+            store.merge(b, EmptyHook.INSTANCE, CommitInfo.EMPTY);
+
+            NodeState oakIndex = root.getChildNode("oak:index");
+            NodeState refIndex = oakIndex.getChildNode("reference");
+            NodeState uuidIndex = oakIndex.getChildNode("uuid");
+            NodeState uuids = uuidIndex.getChildNode(":index");
+            System.out.println("Scanning ... refindex = " + refIndex);
+            NodeState references = refIndex.getChildNode(":references");
+
+            List<FrozenNodeRef> list = new LinkedList<FrozenNodeRef>();
```
mreutegg: Same concern about memory as above.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/FrozenNodeRefsUsingIndexCommand.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+
+    private static boolean hasNtFrozenNodeReferences(NodeStoreOptions nopts) throws IOException {
+        Closer closer = Utils.createCloserWithShutdownHook();
+        try {
+            NodeStore store = Utils.bootstrapNodeStore(nopts, closer);
+            NodeState root = store.getRoot();
+
+            NodeBuilder b = root.builder();
```
mreutegg: Move to test class?

mreutegg (commented):
In general, can those two commands run in read-only mode? This would make them safe to run on data that is in use. E.g. a production system.

Could you please also update the README.md?

oak-run/src/main/java/org/apache/jackrabbit/oak/run/FrozenNodeRefsByScanningCommand.java
stefan-egli: thx! fixed

oak-run/src/main/java/org/apache/jackrabbit/oak/run/FrozenNodeRefsByScanningCommand.java
stefan-egli: thx! fixed

oak-run/src/main/java/org/apache/jackrabbit/oak/run/FrozenNodeRefsByScanningCommand.java
stefan-egli: Agree, much nicer. Done now at https://github.com/apache/jackrabbit-oak/pull/238/commits/fb8b330c4f248078bb2dce62cfb19ae90a50030c

oak-run/src/main/java/org/apache/jackrabbit/oak/run/FrozenNodeRefsByScanningCommand.java
stefan-egli: Good point. The list was more for reporting each find - but ultimately there's no other use case for this atm, and the reporting itself can done as it happens. So I've now changed this to only carry a `count` at https://github.com/apache/jackrabbit-oak/pull/238/commits/fb8b330c4f248078bb2dce62cfb19ae90a50030c

oak-run/src/main/java/org/apache/jackrabbit/oak/run/FrozenNodeRefsUsingIndexCommand.java
stefan-egli: Agree - changed to a `count` at https://github.com/apache/jackrabbit-oak/pull/238/commits/fb8b330c4f248078bb2dce62cfb19ae90a50030c

oak-run/src/main/java/org/apache/jackrabbit/oak/run/FrozenNodeRefsUsingIndexCommand.java
stefan-egli: Oups, this one is particularly ugly - removed this part entirely (at https://github.com/apache/jackrabbit-oak/pull/238/commits/fb8b330c4f248078bb2dce62cfb19ae90a50030c )

stefan-egli:
@mreutegg thx for the review - I've done the changes in [this commit](https://github.com/apache/jackrabbit-oak/pull/238/commits/fb8b330c4f248078bb2dce62cfb19ae90a50030c)

> In general, can those two commands run in read-only mode? This would make them safe to run on data that is in use. E.g. a production system.

Yes, and with the suggested move of the test code to a test class this is now also more clear: only the test class uses the `-read-write=true` parameter - so even more reason to move test code to a test class indeed.
> Could you please also update the README.md?

Yes, will add next, thx for the pointer

stefan-egli:
> > Could you please also update the README.md?
>
> Yes, will add next, thx for the pointer

Now done in [this commit](https://github.com/apache/jackrabbit-oak/pull/238/commits/7a0e090c200b4081416393d0a8bcbb643dd87945)

oak-run/README.md
```diff
@@ -136,6 +138,42 @@ browsing of an existing oak repository.
 
     $ java -jar oak-run-*.jar explore /path/to/oak/repository [skip-size-check]
 
+frozennoderefsbyscanning
+------------------------
+
+This command executes a potentially expensive (!) travering query searching for
```
mreutegg:
```suggestion
This command executes a potentially expensive (!) traversing query searching for
```

mreutegg (changes requested):

mreutegg:
> > In general, can those two commands run in read-only mode? This would make them safe to run on data that is in use. E.g. a production system.
> 
> Yes, and with the suggested move of the test code to a test class this is now also more clear: only the test class uses the `-read-write=true` parameter - so even more reason to move test code to a test class indeed.

Hmm, it still looks like the scan command starts a repository in read-write mode.

oak-run/README.md
stefan-egli: committed, thx

stefan-egli:
> Hmm, it still looks like the scan command starts a repository in read-write mode.

Indeed, thx for the heads-up!

Fixed it now:
* the `frozennoderefsbyscanning` variant already opened in readOnly mode, unless `-read-write=true` was set. Now explicitly disabled that option and therefore always using readOnly.
* the `frozennoderefsusingindex` variant was indeed opening in readWrite (!) mode. As it is using a completely different way of accessing the NodeStore, that utils variant didn't yet support readOnly mode - so I've now added this, and used it for this case.

stefan-egli:
Merged to trunk in rev [1879938](http://svn.apache.org/viewvc?rev=1879938&view=rev)

PR #241, opened 2020-07-10
OAK-9139: Log message on frozen node lookup by identifier

mreutegg:
Logger implementation hidden behind a feature toggle. The logger has flood protection and prints at most one message per second.

### DIFF
--- a/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/version/FrozenNodeLogger.java
+++ b/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/version/FrozenNodeLogger.java
@@ -49,7 +49,7 @@ public class FrozenNodeLogger implements Closeable {
 
     private static final Logger log = LoggerFactory.getLogger(FrozenNodeLogger.class);
 
-    private static final long LOG_INTERVAL = 1000L;
+    private static final long LOG_INTERVAL = Long.getLong("oak.frozenNodeLogger.minInterval", 1000);
 
     private static long NO_MESSAGE_UNTIL = 0;
 
--- a/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/version/FrozenNodeLogger.java
+++ b/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/version/FrozenNodeLogger.java
@@ -1,0 +1,95 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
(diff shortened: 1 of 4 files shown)


oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/version/FrozenNodeLogger.java
```diff
@@ -0,0 +1,95 @@
...
+ * INFO. Log messages are rate limited to one per second. If multiple frozen
+ * nodes are looked up by identifier with the period of one second, then only
+ * the first lookup is logged; the other lookups are not logged.
+ * Enabling DEBUG level for this class reveals the stack of the calling thread.
+ */
+public class FrozenNodeLogger implements Closeable {
+
+    private static final Logger log = LoggerFactory.getLogger(FrozenNodeLogger.class);
+
+    private static final long LOG_INTERVAL = 1000L;
```
stefan-egli: I was wondering if there might ever be a need for this to be adjusted? (eg convert this to a System Property? but might also be overkill)

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/version/FrozenNodeLogger.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Override
+    public void close() {
+        feature.close();
+    }
+
+    protected void logFrozenNode(Tree tree) {
+        log.info("Frozen node {} looked up by id", tree.getPath());
+        if (log.isDebugEnabled()) {
+            log.debug("Call stack", new Exception());
```
stefan-egli: nitpicking: replace "Call stack" with sth more descriptive? (might potentially simplify future `grep` effort)

stefan-egli (approved):
Lgtm. I guess the most important thing really is that this is the only place where a lookup to nt:frozenNode is possible. This I don't know but assume it is (there would be many call possibilities to Root.getTree or also idManager.getTree has more but I assume you ruled them out - as well as any alternative approaches).

Other than that I was contemplating whether it could be useful to show all lookups (bypassing the 1sec filter) - perhaps via the trace level, or by making the 1sec configurable - but might also be overkill not sure. We could always add it if we see need in the future.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/version/FrozenNodeLogger.java
```diff
@@ -0,0 +1,95 @@
...
+    private static long NO_MESSAGE_UNTIL = 0;
+
+    private final Clock clock;
+
+    private final Feature feature;
+
+    public FrozenNodeLogger(@NotNull Clock clock,
+                            @NotNull Whiteboard whiteboard) {
+        this.clock = checkNotNull(clock);
+        this.feature = newFeature("OAK-9139_log_frozen_node_lookup", whiteboard);
```
stefan-egli, line 63: Question: is the goal that this feature is by default enabled? I guess that would be useful.

mreutegg:
> I guess the most important thing really is that this is the only place where a lookup to nt:frozenNode is possible.

Yes, I checked and all lookups are routed through this location. There are a few exceptions that talk directly to the IdentifierManager, but those lookups are for specific nodes that cannot be nt:frozenNode (e.g. a nt:version node).

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/version/FrozenNodeLogger.java
mreutegg: Good point. I'll introduce a system property.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/version/FrozenNodeLogger.java
mreutegg: My idea was to have it disabled by default. It does add a little bit of overhead, even though I did not see a significant change running the UUIDLookupTest benchmark.

PR #242, opened 2020-07-21
OAK-9145: Swap order of default filters

dlh3:
WordDelimiterFilter is invoked with the GENERATE_WORD_PARTS flag, which splits camelCase/PascalCase into multiple terms.  But since the LowerCaseFilter is applied first, the mixed-case is lost and the terms can't be split by WordDelimiterFilter.  Reversing filters so that behaviour occurs as expected.

I would have liked to add unit tests, but there didn't seem to be any around this class.  I *may* come back to add some.


dlh3:
I have made several attempts to add unit tests for this, but I'm really struggling to figure out how to add a fulltext test which uses an index with OakAnalyzer.  It would be awesome if somebody else in the community helped out with the unit tests.

mreutegg:
JIRA issue was resolved WONTFIX.

PR #243, opened 2020-07-22
OAK-9149 : Use batch calls in backgroundSplit

stefan-egli:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
```diff
@@ -3160,6 +3187,22 @@ private void signalClusterStateChange() {
...
 
+    /**
+     * FOR TESTING ONLY :
+     * stops the backgroundUpdateThread (by overwriting its
+     * isDisposed flag) and optionally waits for the thread to
+     * terminate.
+     * @param timeoutMillis optional amount of millis to wait for the thread to terminate at max
+     * @return true if thread is no longer running
+     */
+    boolean stopBackgroundUpdateThread(long timeoutMillis) throws InterruptedException {
```
Joscorbe, line 3201: I tried to think about a more elegant way to do this, maybe using Reflection in the test, but after discussing with @stefan-egli this is probably the clearer way to do this.
I think it would be interesting to have a second opinion here.

Joscorbe (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
stefan-egli: Agreed it's not elegant, but without changing too much the advantage of this is that it's explicit and I think maintainable.

stefan-egli:
Thx for the review @Joscorbe !

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
fabriziofortino: you could mark this method with the `org.jetbrains.annotations.TestOnly` annotation.

stefan-egli: Nice, thx @fabriziofortino, I've added this now.

jsedding: Since @Joscorbe asked for a second opinion. I could imagine passing a thread-factory (could be a BiFunction<BackgroundUpdateOperation, String>) into a new `DocumentNodeStore` constructor. That way the tests could pass in a suitable factory and keep a reference to the thread and `BackgroundUpdateOperation` to be able to accomplish the desired behaviour. IMHO that would be more elegant, but that, as always, is in the eye of the beholder.

stefan-egli: @jsedding interesting approach indeed. I think this might be something we could look into. I'm wondering if it could be done in a separate ticket though, as I think it could involve quite a bit of changes and would meanwhile block progress on this? The concern I see with a thread-factory is that for gracefully stopping the `BackgroundUpdateOperation` (or the `NodeStoreTask` actually) one needs to hook into the `NodeStoreTask.run` loop - and the current patch does that by overwriting the `isDisposed` in `forceStop()`. The thread-factory variant would still need to do that as well, one way or another, I guess. So something like the suggested `forceStop()` would still be necessary I think. But the "ugly" `stopBackgroundUpdateThread` would be gone, agreed. 

Should I create a follow-up ticket for this (with the risk of that being prioritized lower though, agreed)?

stefan-egli:
Merged to SVN in [1880432](http://svn.apache.org/viewvc?rev=1880432&view=rev)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
stefan-egli: Created OAK-9159 to follow-up.

mreutegg: AFAICS the test simply wants to take control of when a background update happens. That means, the alternative is setting the asyncDelay of the DocumentNodeStore to zero in the test. See proposed changes in https://github.com/apache/jackrabbit-oak/pull/244

PR #245, opened 2020-08-04
Oak 9149 : reverting svn rev 1880432 

stefan-egli:


stefan-egli:
* reverted in trunk in [1880565](http://svn.apache.org/viewvc?rev=1880565&view=rev)

PR #247, opened 2020-09-01
OAK-9176 : sweep2 PR for review

stefan-egli:
This represents work-inprogress as the final PR for OAK-9176 is not yet finished. In particular the current sweep2 doesn't have any throttling, so that is missing.

This PR is however already open for review. The missing changes might be added here or a new/final PR might be done at a later stage.


stefan-egli:
Added a 50% throttling mechanism

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/ClusterViewDocument.java
```diff
@@ -483,6 +483,15 @@ static boolean matches(Integer[] expected, Set<Integer> actual) {
         return new HashSet<Integer>(Arrays.asList(inactiveIds));
     }
 
+    /** Returns the set of active, recovering and inactive ids of this cluster view **/
+    Set<Integer> getAllIds() {
```
mreutegg: This method is never used. Remove?

The current set of clusterIds is also available in DocumentNodeStore#clusterNodes as keys.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/ClusterViewDocument.java
```diff
@@ -376,7 +376,7 @@ static String arrayToCsv(Integer[] arr) {
      * internal reader of an existing clusterView document from the settings
      * collection
      **/
-    private static ClusterViewDocument doRead(DocumentNodeStore documentNodeStore) {
+    static ClusterViewDocument doRead(DocumentNodeStore documentNodeStore) {
```
mreutegg: We can avoid opening this up by looking at the data in the `clusterNodes` collection using ClusterNodeInfoDocument. See comment in Sweep2Helper.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Sweep2Helper.java
```diff
@@ -0,0 +1,178 @@
...
+            return 0;
+        }
+    
+        if (clusterViewDoc.getActiveIds().contains(lockClusterId)) {
+            // then another instance is busy sweep2-ing, which is fine.
+            // but we should continue monitoring until that other instance is done
+            LOG.debug("acquireSweep2LockIfNecessary : another instance (id {}) is (still) busy doing a sweep2.",
+                    lockClusterId);
+            return 0;
+        }
```
mreutegg: I suggest replacing this with:
```suggestion
        if (ClusterNodeInfoDocument.all(ns.getDocumentStore()).stream()
                .anyMatch(info -> info.getClusterId() == lockClusterId && info.isActive())) {
            // then another instance is busy sweep2-ing, which is fine.
            // but we should continue monitoring until that other instance is done
            LOG.debug("acquireSweep2LockIfNecessary : another instance (id {}) is (still) busy doing a sweep2.",
                    lockClusterId);
            return 0;
        }
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Sweep2StatusDocument.java
```diff
@@ -0,0 +1,180 @@
+package org.apache.jackrabbit.oak.plugins.document;
```
mreutegg, line 19: License is missing.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -106,6 +108,26 @@ public boolean apply(@Nullable String input) {
         }
     };
 
+    /**
+     * A predicate for property and _deleted names.
```
mreutegg: Needs update. Was copy/pasted from existing predicate.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -106,6 +108,26 @@ public boolean apply(@Nullable String input) {
...
+     */
+    public static final Predicate<String> PROPERTY_OR_DELETED_OR_COMMITROOT_OR_REVISIONS = new Predicate<String>() {
+        @Override
+        public boolean apply(@Nullable String input) {
+            return Utils.isPropertyName(input) || isDeletedEntry(input) || isCommitRootEntry(input) || isRevisionsEntry(input);
+        }
+    };
+
+    /**
+     * A predicate for property and _deleted names.
```
mreutegg: Needs update. Was copy/pasted from existing predicate.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -983,7 +1005,9 @@ public static void alignWithExternalRevisions(@NotNull NodeDocument rootDoc,
     public static void joinQuietly(Thread... threads) {
         for (Thread t : threads) {
             try {
-                t.join();
+                if (t != null) {
+                    t.join();
+                }
```
mreutegg, line 1010: My preference is to revert this change and instead always create the backgroundSweep2Thread. See other comment in DocumentNodeStore.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
```diff
@@ -718,6 +756,15 @@ public int getMemory() {
 
             backgroundUpdateThread.start();
             backgroundSweepThread.start();
+
+            if (sweep2Lock >= 0) {
+                // sweep2 is necessary - so start a sweep2 background task
+                backgroundSweep2Thread = new Thread(
+                        new BackgroundSweep2Operation(this, isDisposed, sweep2Lock),
+                        "DocumentNodeStore background sweep2 thread " + threadNamePostfix);
+                backgroundSweep2Thread.setDaemon(true);
```
mreutegg, line 791: Would rather move this above where all other background threads are created but only start the thread here if needed.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentSweeper2.java
```diff
@@ -0,0 +1,357 @@
+/*
```
mreutegg: General comment about this class: there is a lot of code duplication.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentSweeper2.java
```diff
@@ -0,0 +1,357 @@
...
+     * revision on the root document. This method returns {@code null} if no
+     * update was possible.
+     *
+     * @param documents the documents to sweep
+     * @param listener the listener to receive required sweep update operations.
+     * @return the new sweep revision or {@code null} if no updates were done.
+     * @throws DocumentStoreException if reading from the store or writing to
+     *          the store failed.
+     */
+    @Nullable
```
mreutegg: Nullable does not apply to void return type.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentSweeper2.java
```diff
@@ -0,0 +1,357 @@
...
+     * Performs a sweep and reports the required updates to the given sweep
+     * listener. The returned revision is the new sweep revision for the
+     * clusterId associated with the revision context used to create this
+     * sweeper. The caller is responsible for storing the returned sweep
+     * revision on the root document. This method returns {@code null} if no
+     * update was possible.
+     *
+     * @param documents the documents to sweep
+     * @param listener the listener to receive required sweep update operations.
+     * @return the new sweep revision or {@code null} if no updates were done.
```
mreutegg: This method does not return anything.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
```diff
@@ -2493,6 +2541,114 @@ private void maybeRefreshHeadRevision() {
...
+        // we just went through a full, successful sweep2 and we want to record it
+        // that way - irrespective of any interference with the status
+        // -> hence the 'force' aspect of releasing here
+        if (!Sweep2StatusDocument.forceReleaseSweep2LockAndMarkSwept(store, clusterId)) {
+            LOG.error("backgroundSweep2 : sweep2 finished but we failed to update the sweep2 status accordingly");
+        }
+        return true;
+    }
+
+    private int forceBackgroundSweep2(Revision startRev) throws DocumentStoreException {
```
mreutegg: Why is there a Revision argument? It is always `new Revision(0, 0, clusterId)`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentSweeper2.java
```diff
@@ -0,0 +1,357 @@
...
+        for (String property : filter(doc.keySet(), COMMITROOT_OR_REVISIONS)) {
+            Map<Revision, String> valueMap = doc.getLocalMap(property);
+            for (Map.Entry<Revision, String> entry : valueMap.entrySet()) {
+                Revision rev = entry.getKey();
+
+                // consider any clusterId for sweep2
+
+                Revision cRev = getCommitRevision(doc, rev);
+                if (cRev == null) {
+                    uncommitted(doc, property, rev, op);
```
mreutegg: This will remove valid committed changes!

NodeDocumentSweeper2 runs on a CommitValueResolver without sweep revisions. Starting with 1.8 and introduction of the sweep revision, the revision garbage collector is allowed to remove split documents with commit information older than the sweep revision. The changes referring to the removed commit information must still be considered committed because of to the sweep revision.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/ClusterViewDocument.java
stefan-egli: +1

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/ClusterViewDocument.java
stefan-egli: +1, it was intermittently used during development and not cleaned up.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentSweeper2.java
stefan-egli: Agree, and I was aware of this. However, I did this on purpose as indicated in the class comment. The alternative would have been to reuse NodeDocumentSweeper - with the risk or cost that any future change on that class would have to be carefully crafted to avoid any side-effects on NodeDocumentSweeper2 behaviour. I considered this future risk as higher versus the code-duplication. Wdyt?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
stefan-egli: That would also require `sweep2Lock` to be passed to the `BackgroundSweep2Operation` via some other means though, so not sure that's worth it?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
stefan-egli: See related [comment below](https://github.com/apache/jackrabbit-oak/pull/247#discussion_r481248717)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentSweeper2.java
stefan-egli: Very good and important find. This class was not mean to replicate actual sweep[1] behaviour but only apply missing `_bc` - definitely it should not uncommit anything. Will add a test-case to verify this would have uncommited and fix accordingly.

stefan-egli: Removed the uncommit-cleanup logic and added a test case to cover this.

stefan-egli:
Incorportated review feedback plus cleaned up sweep2 functionality plus added more tests.

stefan-egli:
[Added an ignored test ](https://github.com/apache/jackrabbit-oak/pull/247/commits/0b31072ec9b538910857c084d4ed08aa7d9f9221) to illustrate the fact that currently sweep2 goes through all clusterIds if it thinks it needs to do a sweep2 - it doesn't distinguish between clusterIds that have or have not done a normal sweep(1)

stefan-egli:
[fixed partial-sweep2](https://github.com/apache/jackrabbit-oak/pull/247/commits/3536bbb8ee8de097d700b4b303eafd0cd429d50f): now sweep2 only handles clusterIds in "_sweepRev" (instead of doing it blindly for all clusterIds). A full sweep2 is never happening though, as in that case the sweep2-aspect is delegated to sweep(1) and sweep2 is skipped.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
mreutegg: Ok, I didn't notice the `sweep2Lock`. Then let's keep it that way.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Sweep2Helper.java
```diff
@@ -0,0 +1,166 @@
...
+        // 3) Oak >= 1.8 which was previously doing an Oak <= 1.6 upgrade:
+        //    -> A (full) sweep2 is needed. This is the main case of OAK-9176.
+        Map<Revision, String> bcValueMap = rootNodeDoc.getValueMap("_bc");
+        Map<Revision, String> valueMap = rootNodeDoc.getValueMap(NodeDocument.REVISIONS);
+        for (Map.Entry<Revision, String> entry : valueMap.entrySet()) {
+            Revision rev = entry.getKey();
+    
+            // consider all clusterIds..
+            String rawCommitValue = entry.getValue();
+            String cv = rawCommitValue;
```
mreutegg: `rawCommitValue` is redundant.
```suggestion
            String cv = entry.getValue();
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Sweep2Helper.java
```diff
@@ -0,0 +1,166 @@
...
+        // in this case we have a post (>=) 1.8 repository
+        // which might or might not have previously been a pre (<) 1.8
+        // and we need to distinguish those 2 cases - which, to repeat, are:
+        // 2) Oak >= 1.8 which never did an Oak <= 1.6 upgrade:
+        //    -> no sweep2 is needed as OAK-9176 doesn't apply (the repository
+        //       never ran <= 1.6)
+        // 3) Oak >= 1.8 which was previously doing an Oak <= 1.6 upgrade:
+        //    -> A (full) sweep2 is needed. This is the main case of OAK-9176.
+        Map<Revision, String> bcValueMap = rootNodeDoc.getValueMap("_bc");
+        Map<Revision, String> valueMap = rootNodeDoc.getValueMap(NodeDocument.REVISIONS);
```
mreutegg, line 56: Going through these two maps can be rather expensive. The map not just reflects the `_revisions` and `_bc` entries on the root document (`0:/`), but also all the previous (aka split) documents associated with the root document. Worst case, on a repository that never ran <= 1.6, this will go through all `_revisions` entries on the root and its split documents to find out no sweep is necessary. If this is intentional, then there might be a faster way.
IIUC the method tries to find out if there is a branch commit that was created with Oak <= 1.6. Not just on the root document itself but also by looking at its entire history. Is this correct?
As an alternative Sweep2Helper could query all split documents with `_sdType` DEFAULT(10). On Oak >= 1.8 the majority of split documents for the root are of type DEFAULT_NO_BRANCH(70) and those can be ignored. Before 1.8, most were created as DEFAULT(10).
I'm also a bit concerned about the time it takes to perform this check because it blocks startup. If the check cannot be implemented cheaply, it may be better to just run the sweep and then mark the repository accordingly.

mreutegg: Hmm, maybe it is not that bad after all. With Oak >= 1.8 most split document for the root are DEFAULT_NO_BRANCH(70) but then they can be garbage collection. That is, the number shouldn't be too high and even on an upgraded repository the loop over the revisions map will soon see revisions created before 1.8.
As an improvement: what about performing the check if a sweep2 is necessary after the sweep2 lock has been acquired? What I basically want is a non-blocking startup and perform as much as possible in the background thread.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Sweep2StatusDocument.java
```diff
@@ -0,0 +1,198 @@
...
+            // (force) mark it as done
+            UpdateOp updateOp = new UpdateOp(SWEEP2_STATUS_ID, false);
+            updateOp.set(STATUS_PROPERTY, STATUS_VALUE_SWEPT);
+            updateOp.set(MOD_COUNT_PROPERTY, existing.getModCount() + 1);
+            updateOp.set(SWEPT_BY_PROPERTY, clusterId);
+            if (existing.keySet().contains(LOCK_PROPERTY)) {
+                updateOp.remove(LOCK_PROPERTY);
+            }
+            ArrayList<UpdateOp> updateOps = new ArrayList<UpdateOp>();
+            updateOps.add(updateOp);
```
mreutegg: The updateOps List is never used.
```suggestion
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Sweep2StatusDocument.java
```diff
@@ -0,0 +1,198 @@
...
+            if (documentStore.findAndUpdate(Collection.SETTINGS, updateOp) == null) {
+                LOG.info("forceReleaseLockAndMarkSwept: another instance just wanted to mark sweep2 as done a few moments ago too.");
+                existing = documentStore.find(Collection.SETTINGS, SWEEP2_STATUS_ID,
+                        -1 /* -1; avoid caching */);
```
mreutegg: `existing` is not used after this assignment.
```suggestion
```

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Sweep2Helper.java
stefan-egli: > As an improvement: what about performing the check if a sweep2 is necessary after the sweep2 lock has been acquired?

+1, not only would it block the startup less, it would also ensure that if multiple instances start concurrently that only one does do the `isSweep2Necessary`

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/FailingDocumentStore.java
```diff
@@ -43,6 +45,8 @@
 
     private Type exceptionType = Type.GENERIC;
 
+    private List<Collection<? extends Document>> collectionIncludeList;
```
mreutegg, line 48: Should be reset to null in `Fail.never()`.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/MissingBcSweeper2.java
```diff
@@ -0,0 +1,281 @@
...
+     */
+    void sweep2(@NotNull Iterable<NodeDocument> documents,
+                   @NotNull NodeDocumentSweepListener listener)
+            throws DocumentStoreException {
+        performSweep2(documents, checkNotNull(listener));
+    }
+
+    //----------------------------< internal >----------------------------------
+
+    @Nullable
```
mreutegg: Method does not have a return value.
```suggestion
```

mreutegg (changes requested):

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/FailingDocumentStore.java
stefan-egli: done [here](https://github.com/apache/jackrabbit-oak/pull/247/commits/20edcac5f269b088710cff473c842308df0bae06)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/MissingBcSweeper2.java
```diff
@@ -0,0 +1,281 @@
...
+        LOG.debug("Document sweep2 finished");
+    }
+
+    private Iterable<Map.Entry<Path, UpdateOp>> sweepOperations(
+            final Iterable<NodeDocument> docs) {
+        return filter(transform(docs,
+                new Function<NodeDocument, Map.Entry<Path, UpdateOp>>() {
+
+            int yieldCnt = 0;
+            long lastYield = System.currentTimeMillis();
```
mreutegg: I would rather use the `Clock` from the `RevisionContext`. It will make it easier to control the time in a unit test.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/MissingBcSweeper2.java
```diff
@@ -0,0 +1,280 @@
...
+        for (String property : filter(doc.keySet(), COMMITROOT_OR_REVISIONS)) {
+            Map<Revision, String> valueMap = doc.getLocalMap(property);
+            for (Map.Entry<Revision, String> entry : valueMap.entrySet()) {
+                Revision rev = entry.getKey();
+
+                if (!includedClusterIds.isEmpty() && !includedClusterIds.contains(rev.getClusterId())) {
+                    // if the normal sweep(1) happened for some clusterIds already,
+                    // then sweep2 is only done for those.
+                    // In this case includedClusterIds contains that list.
+                    // If that list is empty, it's done for all.
```
mreutegg: This is confusing and counter-intuitive to me. In which situation would the list be empty and all clusterIds should we swept?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentSweeper2.java
mreutegg: Agreed.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Sweep2Helper.java
stefan-egli: Implemented this with https://github.com/apache/jackrabbit-oak/pull/247/commits/1a8c1dca73bef9e8a1e5e95c80f25ab2f2230f7a

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/MissingBcSweeper2.java
stefan-egli: Agreed, done now in https://github.com/apache/jackrabbit-oak/pull/247/commits/86e3a7d309fad88c349d725b4a7f8434c766dbeb

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/MissingBcSweeper2.java
stefan-egli: Agree, it is confusing. I have clarified the intention in [this commit](https://github.com/apache/jackrabbit-oak/pull/247/commits/f43e4617fba99ff6e2128cf67e83b27e5dd64c40) : basically the empty list case is only used for testing (since a lot of tests were written before that logic was introduced it was easiest to do it that way)

stefan-egli:
@mreutegg thx for the 2nd review round. I've addressed all findings and pushed resolutions to the PR-branch.

mreutegg (commented):
Looks good to me now.

I have only one minor suggestion for a potential improvement. The sweep2 thread can currently block the DocumentNodeStore shutdown. The AtomicBoolean DocumentNodeStore#isDisposed could be used as a poison pill in MissingBcSweeper2.

stefan-egli:
> I have only one minor suggestion for a potential improvement. The sweep2 thread can currently block the DocumentNodeStore shutdown. The AtomicBoolean DocumentNodeStore#isDisposed could be used as a poison pill in MissingBcSweeper2.

Good idea. Added in [this commit](https://github.com/apache/jackrabbit-oak/pull/247/commits/48ab434ad284d0a4c7c7299da4692ec9d73e502b) (including a test case to cover this scenario)

stefan-egli:
Also [added](https://github.com/apache/jackrabbit-oak/pull/247/commits/94f2b3bd00afa8e5b12210cb38c8ee54009d05d8) a System Property "oak.documentMK.disableSweep2" (default: false) that allows to skip sweep2. This will mark the sweep2 status in the settings collection as "swept"

stefan-egli:
Just added some more log.infos about sweep2 status changed [here](https://github.com/apache/jackrabbit-oak/pull/247/commits/56bf43b9334b6a25af909368df9678b314788077)

Vinod-holani:
@stefan-egli we should stop the backgroundsweep2 thread, once sweep2 is done, other then this LGTM.

stefan-egli:
@Vinod-holani, I just checked and it should actually stop the thread [here](https://github.com/apache/jackrabbit-oak/pull/247/files#diff-8275f1ceacb45f9a971993016d545158R3547)?

(if the above link doesn't auto-expand and link to the code, here's the excerpt : it calls `done()` once `backgroundSweep2()` considers its job done:
```       
protected void execute(@NotNull DocumentNodeStore nodeStore) {
            if (retryCount > 0) {
                LOG.info("BackgroundSweep2Operation.execute: retrying sweep2. retryCount=" + retryCount);
            }
            if (nodeStore.backgroundSweep2(sweep2Lock)) {
                done();
            }
```

Joscorbe (approved):
Looks good to me

mreutegg:

PR #249, opened 2020-09-08
OAK-9200

Vinod-holani:
Fixing Oak BlobAccessProvider reference initialization in UserConfigurationImpl causing performance issues


anchela:
@Vinod-holani , had a quick look at the change, because it affects a class in the security area. i cannot comment on the usage of blobaccessprovider in the userconfiguration (not really familiar with the blob-handling in oak). but i have a general comment: i would very much appreciate, if your patch would come with tests illustrating the problem you are aiming to fix. it should fail without the patch and obviously pass with the patch applied.
thanks for updating your PR with tests.

Vinod-holani:
Locally test cases are fine, and the failing test case doesn't seem relevant with the patch, so closing this PR and creating new one to retrigger travis build.

mreutegg:
Changes have been merged to SVN.

PR #252, opened 2020-09-24
OAK-9230: CachingCommitValueResolver with negative cache

mreutegg:


mreutegg:
Slightly modified version has been committed to svn.

PR #256, opened 2020-09-28
OAK-9239 Described support for azure-segment-tar in oak-benchmarks/RE…

synox:
…ADME.md


synox:
fixes https://issues.apache.org/jira/browse/OAK-9239

PR #258, opened 2020-10-27
OAK-9265 : Recovery utility function of LastRevRecoveryAgent

Vinod-holani:
should not process previous docs

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
@@ -310,12 +310,6 @@ public class LastRevRecoveryAgent {
         long lastLog = startOfScan;
 
         for (NodeDocument doc : suspects) {
-        	
-        	//Should skip previous/split documents
-        	if(doc.isSplitDocument()) {
-        		continue;
-        	}
-        	
             totalCount++;
             lastCount++;
 
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
@@ -311,8 +311,8 @@ public class LastRevRecoveryAgent {
 
         for (NodeDocument doc : suspects) {
         	
-        	//Should skip previous documents
-        	if(!doc.getPath().isAbsolute()) {
+        	//Should skip previous/split documents
+        	if(doc.isSplitDocument()) {
         		continue;
         	}
         	
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
@@ -42,7 +42,6 @@ import com.google.common.base.Supplier;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
 
-import org.apache.jackrabbit.oak.commons.PathUtils;
 import org.apache.jackrabbit.oak.commons.TimeDurationFormatter;
 import org.apache.jackrabbit.oak.plugins.document.bundlor.DocumentBundlor;
 import org.apache.jackrabbit.oak.plugins.document.util.MapFactory;
@@ -313,7 +312,7 @@ public class LastRevRecoveryAgent {
         for (NodeDocument doc : suspects) {
         	
         	//Should skip previous documents
-        	if(!doc.getPath().toString().startsWith(PathUtils.ROOT_PATH)) {
+        	if(!doc.getPath().isAbsolute()) {
         		continue;
         	}
         	
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
@@ -42,6 +42,7 @@ import com.google.common.base.Supplier;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
 
+import org.apache.jackrabbit.oak.commons.PathUtils;
 import org.apache.jackrabbit.oak.commons.TimeDurationFormatter;
 import org.apache.jackrabbit.oak.plugins.document.bundlor.DocumentBundlor;
 import org.apache.jackrabbit.oak.plugins.document.util.MapFactory;
@@ -310,6 +311,12 @@ public class LastRevRecoveryAgent {
         long lastLog = startOfScan;
 
         for (NodeDocument doc : suspects) {
+        	
+        	//Should skip previous documents
+        	if(!doc.getPath().toString().startsWith(PathUtils.ROOT_PATH)) {
+        		continue;
+        	}
+        	
             totalCount++;
             lastCount++;
 
--- a/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BasicDocumentStoreTest.java
+++ b/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BasicDocumentStoreTest.java
@@ -36,7 +36,6 @@ import java.util.Set;
 import java.util.UUID;
 
 import org.apache.jackrabbit.oak.plugins.document.util.Utils;
-import org.junit.After;
 import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -53,18 +52,6 @@ public class BasicDocumentStoreTest extends AbstractDocumentStoreTest {
 
     public BasicDocumentStoreTest(DocumentStoreFixture dsf) {
         super(dsf);
-    }
-
-    @After
-    public void tearDown() {
-        Revision.resetClockToDefault();
-        markDocumentsForCleanup();
-    }
-
-    private void markDocumentsForCleanup() {
-        for (NodeDocument doc : Utils.getAllDocuments(ds)) {
-            removeMe.add(doc.getId());
-        }
     }
 
     @Test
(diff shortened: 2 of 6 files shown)


mreutegg:
Alternatively the implementation(s) of MissingLastRevSeeker could be changed to filter our split documents. But apply the filtering later also works.

Would it be possible to have a test that reproduces the problem?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
```diff
@@ -310,6 +310,12 @@ public void sweepUpdate(Map<Path, UpdateOp> updates)
         long lastLog = startOfScan;
 
         for (NodeDocument doc : suspects) {
+        	
+        	//Should skip previous documents
+        	if(!doc.getPath().isAbsolute()) {
```
mreutegg: I would rather make a decision on NodeDocument.getSplitDocType(). WDYT?

Vinod-holani: Yes that is better, I will use NodeDocument.isSplitDocument() to check previous/split document

Vinod-holani:
> Would it be possible to have a test that reproduces the problem?

I already thought about this but recover function only returns no. of inconsistent documents, in logging the scanned document no. will be different but not sure should we try this. Could you suggest something for test case? 

And in current code infinite loop issue won't occur due to Path class implementation, so original issue won't be reproducible.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
mreutegg: This is even better. +1

mreutegg:
Hmm, it is indeed difficult to come up with a test that fails with the current code. What about a test that runs just fine in trunk, but would fail in the maintenance branches without the Path class? Though, I don't see how to write such a test either. The infinite loop would simply cause the test to run forever.

Vinod-holani:
@mreutegg @reschke could you please review this PR?

reschke:
sorry, currently not working on Oak...

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
```diff
@@ -216,4 +220,26 @@ public void getClusterNodeInfo() {
...
+            b1 = dns.getRoot().builder();
+            b1.child(nodeName).setProperty("prop",i);
+            dns.merge(b1, EmptyHook.INSTANCE, CommitInfo.EMPTY);
+        }
+        dns.runBackgroundOperations();
+        int docs = 0;
+        //seeker should return only non split documents
+        for(Document doc : seeker.getCandidates(0)) {
+        	docs++;
+        }
```
mreutegg: Can be simplified with:
```suggestion
        //seeker should return only non split documents
        int docs = Iterables.size(seeker.getCandidates(0));
```

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BasicDocumentStoreTest.java
```diff
@@ -54,6 +55,18 @@ public BasicDocumentStoreTest(DocumentStoreFixture dsf) {
...
+        Revision.resetClockToDefault();
+        markDocumentsForCleanup();
+    }
+
+    private void markDocumentsForCleanup() {
+        for (NodeDocument doc : Utils.getAllDocuments(ds)) {
+            removeMe.add(doc.getId());
+        }
+    }
+
```
mreutegg: Why is this needed in BasicDocumentStoreTest? AFAICS none of the tests uses a Clock. I think this should rather go to `MissingLastRevSeekerTest` and clean up documents created by the new `getNonSplitDocs()` test.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
```diff
@@ -216,4 +220,26 @@ public void getClusterNodeInfo() {
...
+        assertEquals(2, docs);
+    }
```
mreutegg, line 256: The DocumentNodeStore is not disposed.
```suggestion
        dns.dispose();
    }
```

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
```diff
@@ -216,4 +220,26 @@ public void getClusterNodeInfo() {
...
+        NodeBuilder b1 = dns.getRoot().builder();
+        b1.child(nodeName);
+        dns.merge(b1, EmptyHook.INSTANCE, CommitInfo.EMPTY);
+        //Modify and commit changes on this node 100 times to create a split document
+        for (int i = 0; i < NUM_REVS_THRESHOLD; i++) {
+            b1 = dns.getRoot().builder();
+            b1.child(nodeName).setProperty("prop",i);
+            dns.merge(b1, EmptyHook.INSTANCE, CommitInfo.EMPTY);
+        }
+        dns.runBackgroundOperations();
```
mreutegg, line 250: When triggering background operations manually, the background threads should be disabled on the DocumentNodeStore with `DocumentNodeStoreBuilder.setAsyncDelay(0)`.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
```diff
@@ -216,4 +220,26 @@ public void getClusterNodeInfo() {
     private ClusterNodeInfoDocument getClusterNodeInfo(int clusterId) {
         return seeker.getClusterNodeInfo(clusterId);
     }
+    
+    @Test
+    public void getNonSplitDocs() throws Exception {
+        String nodeName = this.getClass().getName() + "-foo";
+        DocumentNodeStore dns = getBuilder().setDocumentStore(store).getNodeStore();
```
mreutegg: The DocumentNodeStore should use the clock setup by this test. Otherwise it can be unstable because `Revision` and `ClusterNodeInfo` are using the virtual clock. Unfortunately it is also necessary to wrap the store and prevent disposal of the underlying store.
```suggestion
        DocumentNodeStore dns = getBuilder().clock(clock).setAsyncDelay(0)
                .setDocumentStore(new DocumentStoreWrapper(store) {
                    @Override
                    public void dispose() {
                        // do not close underlying store, otherwise cleanup
                        // cannot remove documents after the test
                    }
                }).getNodeStore();
```

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
mreutegg, line 250: At this point the test should remember the documents for the cleanup.
```suggestion
        dns.runBackgroundOperations();
        // remember all documents for cleanup
        for (NodeDocument doc : Utils.getAllDocuments(ds)) {
            removeMe.add(doc.getId());
        }
```

mreutegg (changes requested):

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BasicDocumentStoreTest.java
Vinod-holani: There was one node created by this class which was not getting cleaned up and causing the newly created test to fail. Thats why I added this cleanup over here. I will dig deeper in this.

Vinod-holani: Basically this [0] is the culprit function creating a node but not able to delete it. Need to either fix this or can add after cleanup function as already done.

[0] : https://github.com/apache/jackrabbit-oak/blob/trunk/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BasicDocumentStoreTest.java#L1144

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
Vinod-holani: Does this wrapping is required if we are not disposing the nodestore?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BasicDocumentStoreTest.java
Vinod-holani: Fixed the culprit test case from BasicDocumentStoreTest class

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
Vinod-holani: Done

Vinod-holani: Done

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
Vinod-holani: Done

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
Vinod-holani: As per above review comment we should not do this right?

mreutegg: Actually, the test _should_ dispose the `DocumentNodeStore` when it is not needed anymore. Some tests use a `DocumentMKBuilderProvider` rule that takes care of disposing the node store automatically after the test, but this test creates the node store manually.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
mreutegg: Disposal is needed, otherwise some background threads continue to run and occupy a clusterId. E.g. the lease update thread.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
Vinod-holani: Ok so we do need to dispose the DocumentNodeStore but not underlying DocumentStore for this test case. I was confused with these two comments.

PR #260, opened 2020-11-04
OAK-9149 : phased, batched backgroundSplit

stefan-egli:
2nd iteration of OAK-9149 building on
* https://github.com/apache/jackrabbit-oak/pull/243
* and https://github.com/apache/jackrabbit-oak/pull/244

which was redone based on the following review finding:
* https://issues.apache.org/jira/browse/OAK-9149?focusedCommentId=17170693&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17170693

This version now uses a _phased_ and _batched_ backgroundSplit.
* _phased_ : it combines all updateOp except the one on the main document into the first phase. These updateOps can be executed in 1 batch and a partial execution in case of an error/exception doesn't do any harm. Each individual updateOP will be redone if it needs to. The more important part is that the update on the main document is done in a second step/phase. That is therefore what's now implemented. If this 2nd phase is executed only partially, then again that's no problem as that just means that for some the split didn't finish properly and needs to be redone.
** Note that if one of the phases is only executed partially, while it is fine from a consistency point of view, it still has the potential to leave garbage. A comment about this is added in a `// TODO` in `backgroundSplit()`.
* _batched_ : as mentioned above, all phase 1 updateOps are combined up to the configured batch size (`oak.documentMK.createOrUpdateBatchSize`, default 1000) and executed in one go towards the DocumentStore (ie using the bulk version of the 'createOrUpdate' command).


oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MeasuringDocumentStore.java
```diff
@@ -0,0 +1,286 @@
...
+
+import org.apache.jackrabbit.oak.cache.CacheStats;
+import org.apache.jackrabbit.oak.plugins.document.cache.CacheInvalidationStats;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * DocumentStore wrapper which can measure time spent in any call.
+ */
+public class MeasuringDocumentStore implements DocumentStore {
```
mreutegg: There is already TimingDocumentStoreWrapper with a very similar feature. Would it make sense to use the existing class and adjust if needed?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentBatchSplitTest.java
```diff
@@ -0,0 +1,244 @@
...
+    @After
+    public void tearDown() throws Exception {
+        if (mk != null) {
+            mk.dispose();
+            mk = null;
+        }
+        fixture.dispose();
+        if (fixture.getName().equals("MongoDB")) {
+            MongoUtils.dropCollections(MongoUtils.DB);
+        }
```
mreutegg: This shouldn't be necessary. Disposing the MongoDB fixture already drops the collections.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MeasuringDocumentStore.java
stefan-egli: Yes, I did see that one but somehow as it was more complex and production code I considered it less intrusive to add a dedicated, simpler test code one. 

But revisiting it, I suppose what's missing is really only `resetMeasurement` and `getAndResetMeasurement` - so that could be added to TimingDocumentStoreWrapper indeed ..

stefan-egli: Added in [this commit](https://github.com/apache/jackrabbit-oak/pull/260/commits/a88af85d12fb4a56afb98771f2b3f986428ba3d3)

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentBatchSplitTest.java
```diff
@@ -0,0 +1,245 @@
...
+        return fixtures;
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        if (mk != null) {
+            mk.dispose();
+            mk = null;
+        }
+        fixture.dispose();
```
mreutegg, line 85: The log level should also be reset to default. Otherwise tests running after defaultLogLevelBatchSplit() will also run on DEBUG.
```suggestion
        fixture.dispose();
        // reset log level to default
        enableLevel("org", null);        
```

mreutegg (changes requested):

mreutegg:
The log level is the only remaining suggestion I have. Otherwise looks good to me.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentBatchSplitTest.java
stefan-egli: absolutely, nice find!

stefan-egli:
Merged with http://svn.apache.org/viewvc?rev=1883591&view=rev

PR #261, opened 2020-12-03
OAK-9290: Respect nt:frozenNode definition on upgrade and migration

mreutegg:


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/FilteringNodeState.java
```diff
@@ -60,6 +66,8 @@
 
     public static final Set<String> NONE = ImmutableSet.of();
 
+    private static final Set<String> JCR_SYSTEM_PATH_PREFIX = ImmutableSet.of("/", "/" + JcrConstants.JCR_SYSTEM);
```
stefan-egli: IIUC "/" makes "/jcr:system" superfluous - why is "/" in this set, for tests only?

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/migration/FilteringNodeStateTest.java
```diff
@@ -294,12 +301,25 @@ public void shouldDelegateHasProperty() {
...
-        final NodeState decorated = wrap("/", rootNodeState, null, of("/content/foo"), null, null);
+        final NodeState decorated = wrap("/", rootNodeState, null, of("/content/foo"), null, null, true);
         assertTrue("/content should exist and be visible", getNodeState(decorated, "/content").exists());
         assertFalse("/content/foo should be hidden", getNodeState(decorated, "/content/foo").exists());
         assertFalse("/nonexisting should not exist", getNodeState(decorated, "/nonexisting").exists());
     }
 
+    @Test
+    public void withoutReferenceableFrozenNode() {
+        final NodeState decorated = wrap("/", rootNodeState, null, of("/content/foo"), null, null, false);
```
stefan-egli, line 311: It looks like this is the only test which uses `referenceableFrozenNodes=false` - while I assume `false` is now (with OAK-9134) the default. So I'm wondering if we shouldn't have more test cases that test `false`? (Ok, afterwards noticed that `VersionCopierTest` has more `false` tests - so that might be sufficient I guess)

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/migration/version/VersionCopierTest.java
```diff
@@ -0,0 +1,167 @@
...
+        String path = "/foo";
+        NodeStore source = createVersionFor(path, createStore(true));
+        NodeStore target = createStore(true);
+
+        copyContent(source, target, path);
+        copyVersionStorage(source, target);
+        assertVersionReferenceable(target.getRoot(), path);
+    }
+
+    @Ignore("VersionCopier does not generate referenceable target nt:frozenNode from source frozen nodes that are not referenceable")
```
stefan-egli, line 89: It seems if the assertion was inverted (to become `assertVersionNotReferenceable`) this test would succeed and could be included. Would that make sense?

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/migration/FilteringNodeStateTest.java
```diff
@@ -43,6 +46,7 @@
 import static org.apache.jackrabbit.oak.plugins.migration.NodeStateTestUtils.createNodeStoreWithContent;
 import static org.apache.jackrabbit.oak.plugins.migration.NodeStateTestUtils.getNodeState;
 import static org.apache.jackrabbit.oak.plugins.tree.TreeConstants.OAK_CHILD_ORDER;
+import static org.apache.jackrabbit.oak.spi.state.AbstractNodeState.getBoolean;
```
stefan-egli: unused import?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/NodeStateCopier.java
```diff
@@ -308,6 +316,8 @@ private static NodeBuilder getChildNodeBuilder(@NotNull final NodeBuilder root,
 
         private Set<String> mergePaths = emptySet();
 
+        private boolean referenceableFrozenNodes = true;
```
stefan-egli, line 319: Was thinking if this should rather be `false` here - as that's the new default. But haven't checked all implications..

stefan-egli (commented):
@mreutegg, I had a first quick review and added some comments. I might want to go through it again a 2nd time more thoroughly anyway, but thought I want to provide feedback early on already.

stefan-egli:
@mreutegg I had a 2nd look but no further comments, so other than the previous ones it lgtm.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/FilteringNodeState.java
mreutegg: This is indeed incorrect. I was trying to use isDescendantOfAnyPath() when in fact, I needed something like isDescendantOrSelf().

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/migration/version/VersionCopierTest.java
mreutegg: The initial idea was more like test driven design. Ideally, it would be possible to copy frozen nodes that are not referenceable to a target store that has referenceable frozen nodes and the operation would generate a jcr:uuid property for each frozen node on the target store. This is currently not the case and the test is ignored. It something like a TODO, but I don't consider it a priority.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/migration/FilteringNodeStateTest.java
mreutegg: Thanks, will fix.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/NodeStateCopier.java
mreutegg: That's exactly the reason why I set the default to true. I was worried there may be code out there using this class without us knowing. Setting it to false by default would then change the behaviour.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/FilteringNodeState.java
mreutegg: I simplified this and removed any path check logic. NodeState wrapping is now done whenever the jcr:uuid property of frozen nodes must be hidden.

mreutegg:

PR #267, opened 2021-02-03
Change the method name "serialize" to "serializeReferences"

ShangwenWang:
The function of this method is to serialize the passed references.  However, only naming it as "serialize" cannot describe it clearly.
Given that the class name (_GetReferencesResponseEncoder_) provides lots of semantic information, it would be better to change this method's name into "_serializeReferences_" for providing enough guidance.


ShangwenWang:
Does anyone know why build errors happen?

PR #269, opened 2021-02-09
OAK-9134 : backport to 1.22 from trunk

stefan-egli:
with one major difference: the default of 'oak.referenceableFrozenNode' being true in 1.22 for backwards compatibility reason - while in trunk it is false to promote the new behaviour there


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/nodetype/write/NodeTypeRegistry.java
```diff
@@ -89,9 +93,34 @@ public static void register(Root root, InputStream input, String systemId) {
...
+            final boolean referenceableFrozenNode;
+            if (System.getProperty(referenceableFrozenNodeProperty) == null) {
+                // the default for referenceableFrozenNode is true in the 1.22 branch.
+                // this is in contrast to it being false in newer versions.
+                // the reason for choosing true as the default is to maintain higher
+                // backwards compatibility and minimize an otherwise high impact in this branch.
+                referenceableFrozenNode = true;
+            } else {
+                referenceableFrozenNode = Boolean.getBoolean(referenceableFrozenNodeProperty);
+            }
```
mreutegg: To reduce divergence, what do you think about backporting OAK-9010 and then use a `SystemPropertySupplier`? The only difference then would be the default value when calling the factory method.

stefan-egli: +1, good point on reducing divergence

mreutegg:
Other than above, looks good to me.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/nodetype/write/NodeTypeRegistry.java
stefan-egli: turns out SystemPropertySupplier is in oak-store-document so not available here .. unless we move that class somewhere more generic ..

stefan-egli: Created a PR suggesting such a move in https://issues.apache.org/jira/browse/OAK-9352

stefan-egli:
Adjusted to use the SystemPropertySupplier - see corresponding trunk PR https://github.com/apache/jackrabbit-oak/pull/278
Currently that fails however, with
```
caused by: Unable to resolve org.apache.jackrabbit.oak-core [32](R 32.0): 
  missing requirement [org.apache.jackrabbit.oak-core [32](R 32.0)] 
  osgi.wiring.package; (osgi.wiring.package=org.apache.jackrabbit.oak.commons.properties)
```

stefan-egli:
* merged to svn 1.22 branch in rev [1887142](http://svn.apache.org/viewvc?rev=1887142&view=rev)

PR #272, opened 2021-02-15
OAK-9158: Performance issue due to AbstractDocumentNodeState#equals

mreutegg:
Patch provided by Alexander Lüders
Added test to reproduce the problem


stefan-egli:
+1, lgtm

PR #274, opened 2021-02-16
OAK-9305: Create checkpoint with given revision

mreutegg:
Initial patch provided by José Andrés Cordero Benítez
Restructured code and added tests

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreMBeanImpl.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreMBeanImpl.java
@@ -284,7 +284,7 @@ final class DocumentNodeStoreMBeanImpl extends AnnotatedStandardMBean implements
 
             String cp = nodeStore.getCheckpoints().create(lifetime, info, rev).toString();
             log.info("Created checkpoint [{}] with lifetime {} for Revision {}", cp, lifetime, revision);
-            return cp;
+            return String.format("Created checkpoint [%s] with lifetime %d for Revision %s", cp, lifetime, revision);
         } else {
             throw new IllegalArgumentException(String.format("Cannot create a checkpoint for revision %s. " +
                     "Revision timestamp is %d and oldest timestamp to keep is %d",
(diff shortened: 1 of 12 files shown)


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreMBeanImpl.java
```diff
@@ -264,4 +268,28 @@ public String cleanIndividualCache(String name) {
...
+            oldestTimestamp = Math.min(oldestTimestamp, oldestCheckpoint.getTimestamp());
+        }
+        if (force || oldestTimestamp < rev.getTimestamp()) {
+            Map<String, String> info = new HashMap<>();
+            info.put(COMPOSITE_INFO + "created", Long.toString(rev.getTimestamp()));
+            info.put(COMPOSITE_INFO + "expires", Long.toString(Utils.sum(rev.getTimestamp() + lifetime)));
+
+            String cp = nodeStore.getCheckpoints().create(lifetime, info, rev).toString();
+            log.info("Created checkpoint [{}] with lifetime {} for Revision {}", cp, lifetime, revision);
+            return cp;
```
Joscorbe: I was thinking about returning here the string inside the log.info, instead of the "cp" variable that just returns the revision string (which is the first argument you send to this method, so not really useful).
But that would be less machine readable, and I'm not sure if that would be needed.

Joscorbe (approved):
It looks good to me, I just added a suggestion to improve the result of the method to be more human friendly.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreMBeanImpl.java
Joscorbe: Something like:
return String.format("Created checkpoint [{}] with lifetime {} for Revision {}", cp, lifetime, revision);

But changing this would mean at least 2 tests would need to be adjusted:
createCheckpointForceWithRevisionTooOld()
createCheckpointBeforeExistingCheckpoint()

mreutegg: Thanks for the suggestion. I merged your PR.

PR #283, opened 2021-04-06
OAK-9392: Improve resilience when primary becomes unavailable

mreutegg:
Introduces a new configuration that sets a socket timeout for operations on the clusterNodes collection.

Default is up for discussion. Initially I set it to 30 seconds.

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceConfiguration.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceConfiguration.java
@@ -77,7 +77,7 @@ final class DocumentNodeStoreServiceConfiguration {
      * Name of framework property to configure socket timeout for lease update
      * operations on MongoDB.
      */
-    private static final String FWK_PROP_LEASE_SO_TIMEOUT = "oak.mongo.leaseSocketTimeout";
+    private static final String FWK_PROP_MONGO_LEASE_SO_TIMEOUT = "oak.mongo.leaseSocketTimeout";
 
     /**
      * Name of the framework property to configure the update limit.
@@ -101,7 +101,7 @@ final class DocumentNodeStoreServiceConfiguration {
             .put(PROP_URI, FWK_PROP_URI)
             .put(PROP_HOME, PROP_HOME)
             .put(PROP_SO_KEEP_ALIVE, FWK_PROP_SO_KEEP_ALIVE)
-            .put(PROP_LEASE_SO_TIMEOUT, FWK_PROP_LEASE_SO_TIMEOUT)
+            .put(PROP_LEASE_SO_TIMEOUT, FWK_PROP_MONGO_LEASE_SO_TIMEOUT)
             .put(PROP_UPDATE_LIMIT, FWK_PROP_UPDATE_LIMIT)
             .build();
 
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceConfiguration.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceConfiguration.java
@@ -74,6 +74,12 @@ final class DocumentNodeStoreServiceConfiguration {
     private static final String FWK_PROP_SO_KEEP_ALIVE = "oak.mongo.socketKeepAlive";
 
     /**
+     * Name of framework property to configure socket timeout for lease update
+     * operations on MongoDB.
+     */
+    private static final String FWK_PROP_LEASE_SO_TIMEOUT = "oak.mongo.leaseSocketTimeout";
+
+    /**
      * Name of the framework property to configure the update limit.
      */
     private static final String FWK_PROP_UPDATE_LIMIT = "update.limit";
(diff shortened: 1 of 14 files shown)


reschke (commented):
can we change the description so it also (potentially) applies to RDB?

mreutegg:
I consider this issue MongoDocumentStore specific, based on how the MongoDB Java driver interacts with the replica-set. I don't know whether a similar problem exists for RDB, but it would probably depend on the JDBC driver.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoSessionFactory.java
```diff
@@ -37,22 +37,23 @@
 
     private final ClientSessionOptions options;
 
+    private final MongoClock clock;
```
stefan-egli, line 40: I might not fully understand this, but MongoClock is somewhat confusing me. IIUC it is only used by MongoSessionFactory and serves as a replacement for the `clusterTime` and `operationTime` fields that were previously used (and now become obsolete). In that case, could the scope of `MongoClock` not actually be limited to be a private/inner class of MongoSessionFactory - or is the MongoClock indeed being used outside of this class?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBClient.java
```diff
@@ -0,0 +1,145 @@
...
+import org.slf4j.LoggerFactory;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.apache.jackrabbit.oak.plugins.document.util.MongoConnection.readConcernLevel;
+
+/**
+ * Simple struct that contains {@code MongoClient}, {@code MongoDatabase} and
+ * {@code MongoStatus}.
+ */
+final class MongoDBClient {
```
stefan-egli: I was wondering if there would be an alternative name for `MongoDBClient`. Reason is that there is code that does `client.getClient()` which is a bit difficult to read, as you need to know that there are effectively 2 layers or 2 concepts of client (the MongoDBClient and the MongoClient).. IIUC this is like a `Holder` or an `Info` (but the latter is also a bit underappreciating the functionality of this class).. so, not sure what's a better name, but at least this name was initially not directly intuitive (as it collides with MongoClient)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceConfiguration.java
```diff
@@ -73,6 +73,12 @@
      */
     private static final String FWK_PROP_SO_KEEP_ALIVE = "oak.mongo.socketKeepAlive";
 
+    /**
+     * Name of framework property to configure socket timeout for lease update
+     * operations on MongoDB.
+     */
+    private static final String FWK_PROP_LEASE_SO_TIMEOUT = "oak.mongo.leaseSocketTimeout";
```
stefan-egli: To @reschke 's comment re naming/generalising : what I noticed is that the property key has `mongo` in it (to make it clear, it is only for mongo) - but this field does not. Perhaps it could be renamed to sth like `FWK_PROP_MONGO_LEASE_SO_TIMEOUT` (and the same would then also apply to other places/methods that use leaseSoTimeout and not specifically call it out as belonging to mongo

stefan-egli (commented):
Added some naming and design comments but otherwise lgtm

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoSessionFactory.java
mreutegg: The only usage is test, which is the reason I pulled it out of MongoSessionFactory. Otherwise, it is somewhat of an implementation details.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBClient.java
mreutegg: Agreed, will think of something better.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceConfiguration.java
mreutegg: Not sure I understand. Do you suggest turning the configuration `leaseSocketTimeout` into a generic one that also applies to RDB or rather making it fully MongoDB specific?

stefan-egli: I meant making it fully MongoDB specific

reschke: Ack; I misread the original ticket; primary/secondary config does not apply to RDB.

mreutegg: Done with https://github.com/apache/jackrabbit-oak/pull/283/commits/da55aa3195fe63a249b19ebc86e3e7509cb92439

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBClient.java
mreutegg: Renamed `MongoDBClient` to `MongoDBConnection` in https://github.com/apache/jackrabbit-oak/pull/283/commits/10404cbdd52341ba517e57a0e671537fa8a8853d

PR #287, opened 2021-04-21
OAK-9416 add bridge method to get Oak ContentSession from JCR Session

kwin:


kwin:
Due to concerns raised in JIRA I close this PR.

PR #288, opened 2021-04-21
OAK-9415 expose attached principals in JackrabbitSession

kwin:
Alternative to #287.

### DIFF
--- a/oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/JackrabbitSession.java
+++ b/oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/JackrabbitSession.java
@@ -23,15 +23,21 @@ import javax.jcr.Item;
 import javax.jcr.Node;
 import javax.jcr.Property;
 import javax.jcr.Session;
+
+import java.security.Principal;
+import java.util.Set;
+
 import javax.jcr.AccessDeniedException;
 import javax.jcr.RepositoryException;
 import javax.jcr.UnsupportedRepositoryOperationException;
 
 import org.jetbrains.annotations.NotNull;
+import org.osgi.annotation.versioning.ProviderType;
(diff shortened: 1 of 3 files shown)


oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/JackrabbitSession.java
```diff
@@ -23,15 +23,21 @@
...
 import javax.jcr.RepositoryException;
 import javax.jcr.UnsupportedRepositoryOperationException;
 
 import org.jetbrains.annotations.NotNull;
+import org.osgi.annotation.versioning.ProviderType;
 
 /**
  * Jackrabbit specific extension of the JCR {@link javax.jcr.Session} interface.
  */
+@ProviderType
```
kwin, line 40: This was not marked as provider type before. The default is consumer type which requires a major version change for each added method.

oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/JackrabbitSession.java
```diff
@@ -252,4 +258,12 @@
      */
     Node getNodeOrNull(final String absPath) throws RepositoryException;
 
+    /**
+     * Returns the set of principals associated with this session.
+     *
+     * @return the associated principals (may be the empty set if no principal are associated).
+     * @throws RepositoryException If an error occurs.
+     * @since 1.40.0
```
kwin, line 266: The since refers to the bundle version I guess, but this looks weird as the other method refer to Jackrabbit 2 versioning 2.x.. while this refers to Oak versioning. That leads to the case that 2.11.1 is older than 1.40.0...

kwin:
Superseded by #291.

PR #290, opened 2021-04-24
OAK-9417 : ClassCastException in ConsolidatedListenerMBeanImpl

cziegeler:


anchela (approved):
@cziegeler , looks good to me.... only the usage of white space is not totally consistent with other oak code..... but that's a minor detail

mreutegg:

PR #291, opened 2021-04-28
OAK-9415 expose the bound principals of a session as attribute

kwin:


oak-doc/src/site/markdown/differences.md
```diff
@@ -316,6 +318,19 @@ Node Name Length Limit
 
 With the document storage backend (MongoDB, RDBMS), there is currently 
 a limit of 150 UTF-8 bytes on the length of the node names.
-See also OAK-2644.
+See also [OAK-2644](https://issues.apache.org/jira/browse/OAK-2644).
+
+Session Attributes
+------------------
+
+Oak exposes the following attributes via [`Session.getAttribute(...)`][1] and [`Session.getAttributeNames()`][2].
```
anchela: that's not entirely correct.... it also exposes credentials attributes as intended by the JCR specification. maybe rather state something like: in addition to attributes defined on credentials.......

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionImpl.java
```diff
@@ -248,11 +249,15 @@ public String getUserID() {
     public String[] getAttributeNames() {
         Set<String> names = newTreeSet(sessionContext.getAttributes().keySet());
         Collections.addAll(names, sd.getAuthInfo().getAttributeNames());
+        names.add(RepositoryImpl.BOUND_PRINCIPALS);
         return names.toArray(new String[names.size()]);
     }
 
     @Override
     public Object getAttribute(String name) {
+        if (RepositoryImpl.BOUND_PRINCIPALS.equals(name)) {
```
anchela, line 258: i would rather move that to the end of the method.... as the last piece to check.... in other words: if someone came up with the crazy idea to set "oak.bound_principals" attribute on his credentials, it probably should take precedence to fulfill the API contract defined by the JCR specification, shouldn't it?

anchela (changes requested):

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionImpl.java
kwin: That is IMHO dangerous, as the bound principals can be used for checking if certain code paths should be allowed (e.g. in the context of FileVault). If someone can forge these attributes by just creating a (otherwise restricted) session with those parameters, this could easily be abused for privilege escalation. IMHO those values should not be allowed to be overwritten by consumers.

oak-doc/src/site/markdown/differences.md
kwin: Fixed in https://github.com/apache/jackrabbit-oak/pull/291/commits/3d7a7e57c2ca1a117acc0b65c45c2f7739a966e1.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionImpl.java
anchela: fair point..... so let's keep it the way it is.

anchela (commented):
@kwin , i would suggest to also add a dedicated test to RepositoryTest.class for the new attribute.

anchela (approved):
@kwin thanks for the tests.

kwin:
Merged in https://github.com/apache/jackrabbit-oak/commit/cde907088a5c460178fd714cb4edc141cd45e23d.

PR #292, opened 2021-05-14
OAK-9432: oak-run checkpoints command should use invisible clusterId

mreutegg:


stefan-egli:
+1, agree it should be invisible

mreutegg:

PR #294, opened 2021-05-20
OAK-9435: Speed up oak-run datastore --dump-ref

mreutegg:


oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCommand.java
```diff
@@ -462,11 +472,36 @@ private void binaryProperties(NodeState state, String path, ReferenceCollector c
...
             binaryProperties(state, path, collector);
-            for (ChildNodeEntry c : state.getChildNodeEntries()) {
+            for (ChildNodeEntry c : getChildNodeEntries(state)) {
                 traverseChildren(c.getNodeState(), PathUtils.concat(path, c.getName()), collector);
             }
         }
 
+        private Iterable<? extends ChildNodeEntry> getChildNodeEntries(NodeState state) {
+            if (useDirListing) {
+                PropertyState dirListing = state.getProperty(PROP_DIRLISTING);
```
thomasmueller: I wonder if this property "dirListing" is already set? Or is this something we need to add?

nit0906:
@mreutegg - would be helpful to update this section (https://github.com/apache/jackrabbit-oak/tree/trunk/oak-run#oak-datastore) in the README file for the additional option added

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCommand.java
```diff
@@ -428,15 +433,20 @@ private static void logCliArgs(String[] args) {
         private final NodeStore nodeStore;
         private final String[] paths;
         private final String[] inclusionRegex;
+        private boolean useDirListing;
 
         public NodeTraverserReferenceRetriever(NodeStore nodeStore) {
-            this(nodeStore, null, null);
+            this(nodeStore, null, null, false);
```
thomasmueller, line 439: Why not simply enable useDirListing by default?

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCommand.java
mreutegg: This is already in use by oak-lucene: https://github.com/apache/jackrabbit-oak/blob/jackrabbit-oak-1.38.0/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/directory/OakDirectory.java#L68 and is the reason to introduce it. See also the description in OAK-9435.

mreutegg: Hmm, I wasn't aware the field is public. I will update the PR and use the existing constant.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCommand.java
mreutegg: Because it may change behaviour. There may be arbitrary content with such a property, which then impacts how the tree is traversed. With the new option you explicitly allow the command to take this short cut and kind of give your consent that you are aware of the consequences.

mreutegg:
> @mreutegg - would be helpful to update this section (https://github.com/apache/jackrabbit-oak/tree/trunk/oak-run#oak-datastore) in the README file for the additional option added

Good point. Will do.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCommand.java
thomasmueller: You are right, I understand.

thomasmueller (approved):

mreutegg:

PR #295, opened 2021-05-27
OAK-9447: Support for Mongo Java Driver 4.8

Joscorbe:


oak-parent/pom.xml
```diff
@@ -54,7 +54,7 @@
     <segment.db>SegmentMK</segment.db>
     <lucene.version>4.7.1</lucene.version>
     <solr.version>8.6.3</solr.version>
-    <mongo.driver.version>3.12.7</mongo.driver.version>
+    <mongo.driver.version>4.2.1</mongo.driver.version>
```
mreutegg: Can you please update to 4.2.3? That seems to be the most recently available.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java
```diff
@@ -300,4 +308,65 @@ public static ReadConcernLevel readConcernLevel(ReadConcern readConcern) {
...
+     *
+     * @param client the mongo client.
+     * @return {@code true} if part of Replica Set, {@code false} otherwise.
+     */
+    public static boolean isReplicaSet(@NotNull MongoClient client) {
+        MongoClusterListener listener = getOakClusterListener(client);
+        if (listener != null) {
+            return listener.isReplicaSet();
+        }
+        System.out.println("Method isReplicaSet called for a MongoClient without any OakClusterListener!");
```
mreutegg: Please do not use System.out. Writing to the logger is sufficient.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java
```diff
@@ -300,4 +308,65 @@ public static ReadConcernLevel readConcernLevel(ReadConcern readConcern) {
...
+     *
+     * @param client the mongo client.
+     * @return {@ServerAddress} of the cluster. {@null} if not connected or no listener was available.
+     */
+    public static ServerAddress getAddress(@NotNull MongoClient client) {
+        MongoClusterListener listener = getOakClusterListener(client);
+        if (listener != null) {
+            return listener.getServerAddress();
+        }
+        System.out.println("Method getAddress called for a MongoClient without any OakClusterListener!");
```
mreutegg: Same as above.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java
```diff
@@ -300,4 +308,65 @@ public static ReadConcernLevel readConcernLevel(ReadConcern readConcern) {
...
+    }
+
+    private static MongoClusterListener getOakClusterListener(@NotNull MongoClient client) {
+        for (ClusterListener clusterListener : client.getMongoClientOptions().getClusterListeners()) {
+            if (clusterListener instanceof MongoClusterListener) {
+                MongoClusterListener replClusterListener = (MongoClusterListener) clusterListener;
+                return replClusterListener;
+            }
+        }
+        System.out.println("No OakClusterListener found for this MongoClient!");
```
mreutegg: Same as above.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java
```diff
@@ -300,4 +308,65 @@ public static ReadConcernLevel readConcernLevel(ReadConcern readConcern) {
...
+        LOG.warn("No OakClusterListener found for this MongoClient!");
```
mreutegg: In this case, I wouldn't log a warning. Returning null when there is no OakClusterListener is valid. I would leave it up to the caller to decide whether a WARN message is appropriate, depending on the context.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java
```diff
@@ -27,18 +27,27 @@
...
+import org.slf4j.LoggerFactory;
 
 /**
  * The {@code MongoConnection} abstracts connection to the {@code MongoDB}.
  */
 public class MongoConnection {
 
+    public static final String MONGODB_PREFIX = "mongodb://";
+
+    private static final Logger LOG = LoggerFactory.getLogger(MongoUtils.class);
```
mreutegg: This is the wrong class for the logger. It should get the logger for `MongoConnection`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoUtils.java
```diff
@@ -31,18 +29,19 @@
...
 
+import java.util.Set;
+
 import static com.google.common.base.Preconditions.checkArgument;
 
 /**
  * Provides static utility methods for MongoDB.
  */
-class MongoUtils {
+public class MongoUtils {
```
mreutegg: This is unnecessary because its only incorrect use is to get a logger for this class in `MongoConnection`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentNodeStoreBuilderBase.java
```diff
@@ -116,25 +114,6 @@ public T setMongoDB(@NotNull MongoClient client,
...
     }
 
-    /**
-     * Enables or disables the socket keep-alive option for MongoDB. The default
-     * is enabled.
-     *
-     * @param enable whether to enable or disable it.
-     * @return this
-     */
-    public T setSocketKeepAlive(boolean enable) {
```
mreutegg, line 108: To ease upgrades I'd suggest deprecating the methods first and then remove it in a subsequent Oak release.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoClusterListener.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public boolean isReplicaSet() {
+        try {
+            latch.await(15, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {}
+        return replicaSet;
+    }
+
+    @Override
+    public void clusterOpening(ClusterOpeningEvent event) {
```
mreutegg: The default implementation is already a NOP. Overriding the method is not necessary.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoClusterListener.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public void clusterClosed(ClusterClosedEvent event) {
```
mreutegg: Same as above.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoClusterListener.java
```diff
@@ -0,0 +1,88 @@
...
+    private boolean replicaSet = false;
+    private ServerAddress serverAddress;
+    private ServerAddress primaryAddress;
+
+    public MongoClusterListener() {
+        latch = new CountDownLatch(1);
+    }
+
+    public ServerAddress getServerAddress() {
+        try {
```
mreutegg: Can you please move the latch.await and exception handling into a separate method and reuse?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoClusterListener.java
```diff
@@ -0,0 +1,88 @@
...
+    private final CountDownLatch latch;
+    private boolean replicaSet = false;
+    private ServerAddress serverAddress;
+    private ServerAddress primaryAddress;
+
+    public MongoClusterListener() {
+        latch = new CountDownLatch(1);
+    }
+
+    public ServerAddress getServerAddress() {
```
mreutegg, line 46: Can you please add a `@Nullable` annotation for the return?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoClusterListener.java
```diff
@@ -0,0 +1,88 @@
...
+        try {
+            latch.await(15, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {}
+        return serverAddress;
+    }
+
+    public ServerAddress getPrimaryAddress() {
```
mreutegg, line 52: Can you please add a `@Nullable` annotation for the return?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MongoDbTest.java
```diff
@@ -201,7 +201,8 @@ public void updateDocument() {
 
     private static BasicDBObject explain(MongoCollection<BasicDBObject> collection,
                                          Bson query) {
-        return collection.find(query).modifiers(new BasicDBObject("$explain", true)).first();
+        //TODO: Determine if this test is still needed. Modifiers is removed in driver 4, so this has to be reworked
+        return collection.find(query)/*.modifiers(new BasicDBObject("$explain", true))*/.first();
```
mreutegg: While those tests are all ignored, this can still be fixed. I think simply calling `collection.find(query).explain(BasicDBObject.class)` will do the trick.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoTestClient.java
```diff
@@ -19,9 +19,11 @@
 import java.util.concurrent.atomic.AtomicReference;
 
 import com.mongodb.MongoClient;
+import com.mongodb.MongoClientOptions;
```
mreutegg: This import is not used.

mreutegg (changes requested):

mreutegg:
@Joscorbe, this PR currently has conflicts. Can you please rebase your changes and resolve the conflicts?

oak-parent/pom.xml
Joscorbe: Yes, upgraded to 4.2.3

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoUtils.java
Joscorbe: After re-checking this I think I have to keep the "public", as there is a test in a different package that needs to access the MongoUtils anyway.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java
```diff
@@ -27,18 +27,27 @@
...
 import com.mongodb.ReadConcern;
 import com.mongodb.ReadConcernLevel;
+import com.mongodb.ServerAddress;
 import com.mongodb.WriteConcern;
 import com.mongodb.client.MongoDatabase;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.mongodb.event.ClusterListener;
+import org.apache.jackrabbit.oak.plugins.document.mongo.MongoUtils;
```
mreutegg: This import is unused and can be removed.
```suggestion
```

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoUtils.java
mreutegg: I'm still not able to identify those tests. I'm able to build `oak-store-document` module successfully when I revert the changes on this class, that is, making it package private again.

mreutegg (changes requested):

mreutegg:
I think the import package in `oak-store-document/pom.xml` must be adjusted as well. It currently limits the version scope to something less than 4.x.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoUtils.java
```diff
@@ -31,12 +29,13 @@
 import com.mongodb.client.FindIterable;
 import com.mongodb.client.MongoCollection;
 import com.mongodb.client.model.IndexOptions;
-
 import org.apache.jackrabbit.oak.plugins.document.DocumentStoreException.Type;
 import org.bson.Document;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.Set;
```
mreutegg: Can you please revert the changes on this file entirely? There's only rearrangement of import statements left.

mreutegg (dismissed):

mreutegg:
@Joscorbe it seems to me only a few changes would be necessary to avoid use of the legacy driver, but I may be missing some less trivial changes. https://www.mongodb.com/docs/drivers/java/sync/current/legacy/#api-changes

WDYT?

Joscorbe:
@mreutegg Yes, legacy driver is the one I'm using for these changes, you can see it here: 
https://github.com/apache/jackrabbit-oak/pull/295/files#diff-61d831053d713cff9598e416a9aafe8a69a1ee87c8bec8226d805f388e2173f8

All of the changes I had to apply are mainly because of the breaking changes listed here since version 3.12:
https://www.mongodb.com/docs/drivers/java/sync/current/upgrade/

We were still using some deprecated methods that were removed, so I had to change them even to be able to build the project.

Joscorbe:
Hi @mreutegg, yes, the plan is to move to sync driver after the failing tests in this build are fixed. Apart from the documentation you have showed, I have this support case where they add some points about the transition from legacy to sync:

https://www.mongodb.com/community/forums/t/conversion-from-mongodb-driver-legacy-to-mongodb-driver-sync/125411

Joscorbe:
All checks have passed at this point, I will continue now the migration to sync driver in this same PR.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
```diff
@@ -305,13 +303,12 @@ private void registerNodeStore() throws IOException {
                                 "journalCache={}, blobCacheSize (MB)={}, maxReplicationLagInSecs={}",
                         mongoURI.getHosts(), db, config.cache(), persistentCache,
                         journalCache, config.blobCacheSize(), config.maxReplicationLagInSecs());
-                log.info("Mongo Connection details {}", MongoConnection.toString(mongoURI.getOptions()));
+                log.info("Mongo Connection details {}", MongoConnection.toString(mongoURI));
```
Joscorbe, line 306: I have added a new method MongoConnection.toString that accepts a ConnectionString as a parameter.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java
```diff
@@ -142,41 +153,53 @@ public void close() {
...
      *      otherwise.
      */
     public static boolean hasWriteConcern(@NotNull String uri) {
-        MongoClientOptions.Builder builder = MongoClientOptions.builder();
-        builder.writeConcern(WC_UNKNOWN);
-        WriteConcern wc = new MongoClientURI(checkNotNull(uri), builder)
-                .getOptions().getWriteConcern();
-        return !WC_UNKNOWN.equals(wc);
+        WriteConcern wc = new ConnectionString(checkNotNull(uri)).getWriteConcern();
+        return wc != null && !WC_UNKNOWN.equals(wc);
```
Joscorbe, line 202: WriteConcern can be null now

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java
```diff
@@ -186,9 +209,8 @@ public static boolean hasWriteConcern(@NotNull String uri) {
      *      otherwise.
      */
     public static boolean hasReadConcern(@NotNull String uri) {
-        ReadConcern rc = new MongoClientURI(checkNotNull(uri))
-                .getOptions().getReadConcern();
-        return readConcernLevel(rc) != null;
+        ReadConcern rc = new ConnectionString(checkNotNull(uri)).getReadConcern();
+        return rc != null && readConcernLevel(rc) != null;
```
Joscorbe, line 213: ReadConcern can be null now

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoTestClient.java
```diff
@@ -16,31 +16,136 @@
...
-
+import com.mongodb.client.MongoIterable;
+import com.mongodb.connection.ClusterDescription;
+import org.apache.jackrabbit.oak.plugins.document.util.MongoConnection;
+import org.bson.Document;
+import org.bson.conversions.Bson;
 import org.jetbrains.annotations.NotNull;
 
-class MongoTestClient extends MongoClient {
+class MongoTestClient implements MongoClient {
```
Joscorbe, line 37: This is now an interface and I have found no better way to achieve the same result.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/ReadPreferenceIT.java
```diff
@@ -116,10 +116,11 @@ public void testReadWriteMode() {
         mongoDS.setReadWriteMode(rwMode(ReadPreference.primary()));
         assertEquals(ReadPreference.primary(), mongoDS.getConfiguredReadPreference(NODES));
 
-        mongoDS.setReadWriteMode("readPreference=secondary&w=2&safe=true&j=true");
+        // Command failed with error 2 (BadValue): 'cannot use 'w' > 1 when a host is not replicated'
+        mongoDS.setReadWriteMode("readPreference=secondary&w=1&safe=true&j=true");
```
Joscorbe, line 120: I think this change doesn't affect in what the test wants to check.
Unless we start the cluster as a replica set, Mongo doesn't let you specify a 'w' value bigger than 1.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/ReadPreferenceIT.java
```diff
@@ -116,10 +116,11 @@ public void testReadWriteMode() {
...
 
         assertEquals(ReadPreference.secondary(), mongoDS.getDBCollection(NODES).getReadPreference());
-        assertEquals(2, mongoDS.getDBCollection(NODES).getWriteConcern().getW());
+        assertEquals(1, mongoDS.getDBCollection(NODES).getWriteConcern().getW());
```
Joscorbe, line 123: Same as before

oak-store-document/pom.xml
```diff
@@ -44,8 +44,8 @@
           <instructions>
             <Import-Package>
               <!-- OAK-7182 -->${guava.osgi.import},
-              com.mongodb*;version="[3.8, 4)";resolution:=optional,
-              org.bson*;version="[3.8, 4)";resolution:=optional,
+              com.mongodb*;version="[3.8, 5)";resolution:=optional,
+              org.bson*;version="[3.8, 5)";resolution:=optional,
```
mreutegg: I think the lower bound needs adjustment as well, or does oak-store-document with the changes in this PR still work with version 3.8 of the MongoDB Java driver?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentNodeStoreBuilderBase.java
```diff
@@ -104,18 +103,11 @@ public T setMongoDB(@NotNull MongoClient client,
      * @param enable whether to enable or disable it.
```
mreutegg: Can you please update JavaDoc and mention this is deprecated and calling this method does not have any effect? Socket keep-alive is always enabled.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java
```diff
@@ -142,41 +153,53 @@ public void close() {
...
-    public static MongoClientOptions.Builder getDefaultBuilder() {
-        return new MongoClientOptions.Builder()
-                .description("MongoConnection for Oak DocumentMK")
-                .maxWaitTime(DEFAULT_MAX_WAIT_TIME)
-                .heartbeatFrequency(DEFAULT_HEARTBEAT_FREQUENCY_MS)
-                .threadsAllowedToBlockForConnectionMultiplier(100);
+    public static MongoClientSettings.Builder getDefaultBuilder() {
+        return MongoClientSettings.builder()
+                .applicationName("Oak DocumentMK")
+                .applyToConnectionPoolSettings(builder -> {
```
mreutegg: Curly brackets are unnecessary here. Same for the two other that follow.

mreutegg (changes requested):

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceTest.java
```diff
@@ -203,42 +189,6 @@ public void persistentCacheExclude() throws Exception{
...
-        MockOsgi.setConfigForPid(context.bundleContext(), PID, config);
-
-        MockOsgi.activate(service, context.bundleContext());
-
-        DocumentNodeStore store = context.getService(DocumentNodeStore.class);
-        MongoDocumentStore mds = getMongoDocumentStore(store);
-        MongoClient client = MongoDocumentStoreTestHelper.getClient(mds);
-        assertFalse(client.getMongoClientOptions().isSocketKeepAlive());
-    }
-
```
mreutegg: Can you please re-introduce these tests again and use another configuration that is still available to exercise the configuration preset feature?

mreutegg (changes requested):

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoTestClient.java
mreutegg: The only alternative that comes to my mind is using `java.lang.reflect.Proxy`, but implementing the methods is also fine with me.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongodProcessFactory.java
```diff
@@ -145,7 +147,8 @@ private void initRS(String rs, int[] ports) {
         }
         Document config = new Document("_id", rs);
         config.append("members", members);
-        try (MongoClient c = new MongoClient(localhost(), ports[0])) {
+        ConnectionString mongoURI = new ConnectionString(localhost());
+        try (MongoClient c = MongoClients.create(localhost())) {
```
mreutegg: This does not look right to me. `mongoURI` is never used and IIUC the MongoClient created here will try to connect to a MongoDB instance on the default port, which isn't necessarily the case.

mreutegg (dismissed):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1924,15 +1924,15 @@ public void setReadWriteMode(String readWriteMode) {
             if(!readWriteMode.startsWith("mongodb://")){
                 rwModeUri = String.format("mongodb://localhost/?%s", readWriteMode);
             }
-            MongoClientURI uri = new MongoClientURI(rwModeUri);
-            ReadPreference readPref = uri.getOptions().getReadPreference();
+            ConnectionString uri = new ConnectionString(rwModeUri);
+            ReadPreference readPref = uri.getReadPreference();
```
mreutegg, line 1951: As you noted in another comment `ReadPreference` as well as `WriteConcern` further down can be `null` after the driver update. There need to be null checks.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoClusterListener.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+public class MongoClusterListener implements ClusterListener {
+
+    // Sometimes we need to wait a few seconds in case the connection was just created, the listener
+    // didn't have time to receive the description from the cluster. This latch is used to check
+    // if the connection was properly initialized.
+    private final CountDownLatch LATCH;
+    private final int LATCH_AWAIT_TIMEOUT = 15;
```
rishabhdaim: This could be a compile time constant.

rishabhdaim:
```suggestion
    private static final int LATCH_AWAIT_TIMEOUT = 15;
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoClusterListener.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+public class MongoClusterListener implements ClusterListener {
+
+    // Sometimes we need to wait a few seconds in case the connection was just created, the listener
+    // didn't have time to receive the description from the cluster. This latch is used to check
+    // if the connection was properly initialized.
+    private final CountDownLatch LATCH;
```
rishabhdaim:
```suggestion
    private final CountDownLatch latch;
```
Only compile time constants should be in upper case.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoClusterListener.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public boolean isReplicaSet() {
+        waitForLatch();
+        return replicaSet;
+    }
+
+    @Override
+    public void clusterDescriptionChanged(final ClusterDescriptionChangedEvent event) {
+        for (ServerDescription sd : event.getNewDescription().getServerDescriptions()) {
```
rishabhdaim:
```suggestion
        Optional<ServerDescription> ol = Iterables.tryFind(event.getNewDescription().getServerDescriptions(), s -> s.getState() == ServerConnectionState.CONNECTED);
        if (ol.isPresent()) {
            ServerDescription sd = ol.get();
            serverAddress = sd.getAddress();
            primaryAddress = new ServerAddress(sd.getPrimary());
            if (sd.isReplicaSetMember()) {
                // Can't assign directly the result of the function because in some cases the cluster
                // type is UNKNOWN, mainly when the cluster is changing it's PRIMARY.
                replicaSet = true;
            }
            latch.countDown();
        } else {
            // log here that no connected state is present
        }
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoClusterListener.java
```diff
@@ -0,0 +1,82 @@
...
+    private ServerAddress serverAddress;
+    private ServerAddress primaryAddress;
+
+    public MongoClusterListener() {
+        LATCH = new CountDownLatch(1);
+    }
+
+    @Nullable
+    public ServerAddress getServerAddress() {
+        waitForLatch();
```
rishabhdaim, line 47: What would happen in case `waitForLatch` timed out and returned false?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBConnection.java
```diff
@@ -61,16 +64,19 @@ final class MongoDBConnection {
...
                                                   @NotNull MongoClock clock,
-                                                  int socketTimeout,
-                                                  boolean socketKeepAlive) {
+                                                  int socketTimeout) {
         CompositeServerMonitorListener serverMonitorListener = new CompositeServerMonitorListener();
-        MongoClientOptions.Builder options = MongoConnection.getDefaultBuilder();
-        options.addServerMonitorListener(serverMonitorListener);
-        options.socketKeepAlive(socketKeepAlive);
+        MongoClientSettings.Builder options = MongoConnection.getDefaultBuilder();
+        options.applyToServerSettings((builder) -> {
```
rishabhdaim: No need for parentheses and curly braces for one line inside the lambda.
```suggestion
        options.applyToServerSettings(builder -> builder.addServerMonitorListener(serverMonitorListener));
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBConnection.java
```diff
@@ -61,16 +64,19 @@ final class MongoDBConnection {
...
+            builder.addServerMonitorListener(serverMonitorListener);
+        });
         if (socketTimeout > 0) {
-            options.socketTimeout(socketTimeout);
+            options.applyToSocketSettings((builder) -> {
```
rishabhdaim: Please see the above comment.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java
```diff
@@ -142,41 +153,53 @@ public void close() {
...
+                    builder.maxWaitTime(DEFAULT_MAX_WAIT_TIME, TimeUnit.MILLISECONDS);
```
rishabhdaim: please add a static import for TimeUnit.MILLISECONDS.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java
```diff
@@ -300,4 +324,61 @@ public static ReadConcernLevel readConcernLevel(ReadConcern readConcern) {
...
+     */
+    public static ServerAddress getPrimaryAddress(@NotNull MongoClient client) {
+        MongoClusterListener listener = getOakClusterListener(client);
+        if (listener != null) {
+            return listener.getPrimaryAddress();
+        }
+        return null;
+    }
+
+    private static MongoClusterListener getOakClusterListener(@NotNull MongoClient client) {
```
rishabhdaim: Make the return type of this API optional and make the client use fluent coding api's.

rishabhdaim:
```suggestion
    private static Optional<MongoClusterListener> getOakClusterListener(@NotNull MongoClient client) {
    return client.getClusterDescription().getClusterSettings().getClusterListeners().stream().filter(clusterListener -> clusterListener instanceof MongoClusterListener).map(clusterListener -> (MongoClusterListener) clusterListener).findFirst();
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java
```diff
@@ -300,4 +324,61 @@ public static ReadConcernLevel readConcernLevel(ReadConcern readConcern) {
...
+    }
+
+    /**
+     * Returns the {@ServerAddress} of the MongoDB cluster primary node.
+     *
+     * @param client the mongo client.
+     * @return {@ServerAddress} of the primary. {@null} if not connected or no listener was available.
+     */
+    public static ServerAddress getPrimaryAddress(@NotNull MongoClient client) {
+        MongoClusterListener listener = getOakClusterListener(client);
```
rishabhdaim: After making `getOakClusterListener()` to return optional this can be re-written as follows:

```suggestion
        MongoClusterListener listener = getOakClusterListener(client).map(l -> l.getPrimaryAddress()).orElse(null);
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java
```diff
@@ -300,4 +324,61 @@ public static ReadConcernLevel readConcernLevel(ReadConcern readConcern) {
...
+    }
+
+    /**
+     * Returns the {@ServerAddress} of the MongoDB cluster.
+     *
+     * @param client the mongo client.
+     * @return {@ServerAddress} of the cluster. {@null} if not connected or no listener was available.
+     */
+    public static ServerAddress getAddress(@NotNull MongoClient client) {
+        MongoClusterListener listener = getOakClusterListener(client);
```
rishabhdaim: See comment for `getPrimaryAddress()`

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoClusterListener.java
Joscorbe: The idea behind the Latch mechanism is to wait a few seconds until a new connection is fully established. This only happens when the connection is really fresh. After the first `clusterDescriptionChanged` is called, the variables are filled in and the latch is no longer used.

If the latch times out, the ServerAddress returned will be null, which shouldn't be a problem as it has the `Nullable` annotation and when it's used it is checked for null. For example: https://github.com/apache/jackrabbit-oak/blob/c6ddcc55bee3de915459af01e91edad32d538f3d/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoStatusTest.java#L207-L211

Joscorbe: Following your comment, I'm adding the `Nullable` annotation to the 3 methods in [MongoConnection.java](https://github.com/apache/jackrabbit-oak/blob/fada62153fe314c144ba6b8ccc9155438fe8f2a8/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java), so it clearly states the returned object can be null:
https://github.com/apache/jackrabbit-oak/blob/fada62153fe314c144ba6b8ccc9155438fe8f2a8/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java#L352

And the following ones.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoClusterListener.java
```diff
@@ -0,0 +1,82 @@
...
+                }
+                LATCH.countDown();
+            }
+        }
+    }
+
+    private void waitForLatch() {
+        try {
+            LATCH.await(LATCH_AWAIT_TIMEOUT, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {}
```
rishabhdaim, line 83: I would add a `log.error` msg here instead of swallowing the exception.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java
```diff
@@ -300,4 +324,61 @@ public static ReadConcernLevel readConcernLevel(ReadConcern readConcern) {
...
+     * Returns {@code true} if the MongoDB server is part of a Replica Set,
+     * {@code false} otherwise. The Replica Set Status is achieved by the
+     * ReplicaSetStatusListener, which is triggered whenever the cluster
+     * description changes.
+     *
+     * @param client the mongo client.
+     * @return {@code true} if part of Replica Set, {@code false} otherwise.
+     */
+    public static boolean isReplicaSet(@NotNull MongoClient client) {
+        MongoClusterListener listener = getOakClusterListener(client);
```
rishabhdaim: See the comment for `getPrimaryAddress()`

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java
```diff
@@ -188,9 +188,9 @@ static int checkDataStore(String... args) {
             NodeStore nodeStore = null;
             if (options.has(store)) {
                 String source = options.valueOf(store);
-                if (source.startsWith(MongoURI.MONGODB_PREFIX)) {
-                    MongoClientURI uri = new MongoClientURI(source);
-                    MongoClient client = new MongoClient(uri);
+                if (source.startsWith(MongoConnection.MONGODB_PREFIX)) {
```
rishabhdaim: Implementation of `options.valueOf(store)` clearly states that it can return an null value if it returns an `emptyList()`. I think we should handle the null check here as well.

```suggestion
                if (StringUtils.startsWith(source, MongoConnection.MONGODB_PREFIX)) {
```

where `StringUtils` is from `org.apache.commons.lang3.StringUtils`

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoClusterListener.java
Joscorbe: I have applied your recommendation. However, I don't think we should add a log message when no connected state is present, as this event can be executed in multiple situations (for example, when there is a disconnection), meaning we could flood the logs with messages that are not useful.

What I'm adding now is a single log message when `getPrimaryAddress()`, `getAddress()` or `isReplicaSet()` methods return a null value because the connection was not properly established. This way we ensure the log message will show up just once.

Joscorbe: I have changed the first part of the code to this one, I think it's equivalent and doesn't have to use the Google libraries for Optional and Iterables, but instead it uses Java API code:
`Optional<ServerDescription> ol = event.getNewDescription().getServerDescriptions()
            .stream().filter(s -> s.getState() == ServerConnectionState.CONNECTED).findFirst();`

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceTest.java
Joscorbe: Yes, I have re-introduced both tests using the property PROP_UPDATE_LIMIT

oak-store-document/pom.xml
Joscorbe: No, they aren't compatible. I have upgraded them in this file. The minimum MongoDB Driver version is now 4.7. I thought it could be a lower one like 4.0, but due to some breaking changes introduced in several classes they are not compatible.

Joscorbe:
Thanks @mreutegg and @rishabhdaim for your reviews. I think I have addressed all the suggestions in the code.
I would appreciate if you have more comments or give green light to merge it.

mreutegg (approved):
Looks good to me.

rishabhdaim (approved):

PR #297, opened 2021-06-04
OAK-9451 - Cold Standby SSL certificates should be configurable

ahanikel:
This patch enables one to

- configure server and client keys and certificate chains
- optionally enable client validation
- optionally enable validation of the server's and the client's SubjectDN against a regex pattern

In contrast to the server classes, the StandbyClient and StandbyClientSync did not have a builder yet. The addition of three new parameters to the constructor looked like too much, so I've added builders for these two as well. This made it necessary to change all the existing tests, that's why the PR touches so many files.

https://issues.apache.org/jira/browse/OAK-9451


ahanikel:
Patch has been merged, closing PR.

PR #299, opened 2021-06-24
OAK-9471 - Query for split documents by Revision GC times out

ankitaagar:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
```diff
@@ -136,7 +136,7 @@ protected SplitDocumentCleanUp createCleanUp(Set<SplitDocType> gcTypes,
             // makes any future similar problem more visible than long running
             // queries alone (15min is still long).
             Iterable<NodeDocument> iterable = filter(transform(getNodeCollection().find(query)
-                    .maxTime(15, TimeUnit.MINUTES),
+                    .maxTime(15, TimeUnit.MINUTES).hintString("_sdType_1__sdMaxRevTime_1"),
```
mreutegg: This name looks somewhat cryptic and generated by MongoDB. Is this something we can rely on? What happens when there is no such index? Will the query fail or MongoDB just silently ignore the hint? The reason why I'm asking is because there are cases where Oak does not automatically created that index. See https://github.com/apache/jackrabbit-oak/blob/72203f3928226a26ea22197aed9c8901605618f2/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java#L354

I'm also wondering whether it is possible to read the name of the index from MongoDB or use the other method variant `hint(Bson)`. I was hoping the latter does not depend on how MongoDB generates a name for the index.

ankitaagar: comments incorporated

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
```diff
@@ -127,6 +128,13 @@ protected SplitDocumentCleanUp createCleanUp(Set<SplitDocType> gcTypes,
         // With OAK-8351 this switched from 1 to 2 queries (see createQueries)
         // hence we iterate over the queries returned by createQueries
         List<Bson> queries = createQueries(gcTypes, sweepRevs, oldestRevTimeStamp);
+        BasicDBObject keys = null;
+        if(hasIndex(getNodeCollection(), SD_TYPE, SD_MAX_REV_TIME_IN_SECS)) {
+           keys = new BasicDBObject();
+           keys.put(SD_TYPE,1);
+           keys.put(SD_MAX_REV_TIME_IN_SECS, 1);
+        }
+
```
mreutegg: What do you think about moving this up to the constructor? That way, Revision GC will only check for the index once and now with every GC cycle.

mreutegg (changes requested):
Can you please add a test that validates Revision GC works fine on MongoDB when there is no compound index on `_sdType` and `_sdMaxRevTime`?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
stefan-egli: Marking this as resolved as the hint is now created in the constructor

stefan-egli (approved):
+1, lgtm.
One thing which is slightly unfortunate is that we do not have a (unit)test that reproduces the original problem. But it should not prevent us from rolling out a fix. If we later come up with a way to reproduce we can add it at that point as well.

PR #301, opened 2021-06-24
OAK-9473 - Cold Standby: configure decryption password for SSL key in OSGi

ahanikel:
Adding the option to configure a password in OSGi which can decrypt encrypted SSL keys.

Additionally I've renamed the configuration options `sslServerSubjectPattern` and `sslClientSubjectPattern` into a single `sslSubjectPattern` because we never need both in the same instance. I hope it's still ok to do that change given that we haven't had an Oak release since these were introduced.


dulceanu (approved):
LGTM

PR #304, opened 2021-06-28
OAK-9476: Clock difference check performed too late

mreutegg:


stefan-egli (approved):
+1, lgtm. 
(what's left being executed now before this check (for mongo at least) is MongoDocumentStore.ensureIndexes() which should be fine even if clocks are not in sync.)

mreutegg:
> MongoDocumentStore.ensureIndexes() which should be fine even if clocks are not in sync.

Right. That shouldn't be a problem because it does not deal with any operations that depend on `Revision` or other time dependent data structures.

PR #307, opened 2021-06-30
OAK-9479:  upgrade jackson-databind to 2.10.5.1

fabriziofortino:


thomasmueller (commented):
Wouldn't we want to change the version in oak-parent/pom.xml?

fabriziofortino:
@thomasmueller I wanted to limit the fix to oak-search-elastic since the vulnerability affects this module only. But you are right, I have modified the parent pom. Please review again.

nit0906 (approved):
+1

PR #309, opened 2021-07-01
OAK-9467 added utility object to wrap the segment file and last modif…

smiroslav:
…ication date

### DIFF
--- a/oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
+++ b/oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
@@ -23,6 +23,7 @@ import org.apache.jackrabbit.oak.commons.Buffer;
 import org.apache.jackrabbit.oak.segment.spi.monitor.IOMonitor;
 import org.apache.jackrabbit.oak.segment.spi.persistence.persistentcache.AbstractPersistentCache;
 import org.apache.jackrabbit.oak.segment.spi.persistence.persistentcache.SegmentCacheStats;
+import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -62,12 +63,6 @@ public class PersistentDiskCache extends AbstractPersistentCache {
     final AtomicBoolean cleanupInProgress = new AtomicBoolean(false);
 
     final AtomicLong evictionCount = new AtomicLong();
-
-    private static final Comparator<SegmentCacheEntry> sortedByAccessTime = (segmentCacheEntry1, segmentCacheEntry2) -> {
-        FileTime lastAccessFile1 = segmentCacheEntry1.getLastAccessTime();
-        FileTime lastAccessFile2 = segmentCacheEntry2.getLastAccessTime();
-        return lastAccessFile1.compareTo(lastAccessFile2);
-    };
 
     public PersistentDiskCache(File directory, int cacheMaxSizeMB, IOMonitor diskCacheIOMonitor) {
         this.directory = directory;
@@ -198,7 +193,7 @@ public class PersistentDiskCache extends AbstractPersistentCache {
                                 return new SegmentCacheEntry(path, FileTime.fromMillis(Long.MAX_VALUE));
                             }
                         })
-                        .sorted(sortedByAccessTime);
+                        .sorted(Comparator.naturalOrder());
 
                 StreamConsumer.forEach(segmentCacheEntryStream, (segmentCacheEntry, breaker) -> {
 
@@ -217,7 +212,7 @@ public class PersistentDiskCache extends AbstractPersistentCache {
         }
     }
 
-    private static class SegmentCacheEntry {
+    private static class SegmentCacheEntry implements Comparable<SegmentCacheEntry>{
         private Path path;
         private FileTime lastAccessTime;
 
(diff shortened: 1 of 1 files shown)


oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
```diff
@@ -193,16 +188,27 @@ public void cleanUp() {
...
+                            try {
+                                return new SegmentCacheEntry(path, Files.readAttributes(path, BasicFileAttributes.class).lastAccessTime());
+                            } catch (IOException e) {
+                                logger.error("Error while getting the last access time for {}", path.toFile().getName());
+                            }
+                            return new SegmentCacheEntry(null, FileTime.fromMillis(Long.MAX_VALUE));
+                        })
                         .sorted(sortedByAccessTime)
-                        .filter(filePath -> !filePath.toFile().isDirectory());
+                        .filter(segmentCacheEntry -> !segmentCacheEntry.getPath().toFile().isDirectory());
```
fabriziofortino: I think this filter can be moved before the first `map()`: there is no need to create `SegmentCacheEntry` for folders.

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
```diff
@@ -193,16 +188,27 @@ public void cleanUp() {
...
             try {
-                Stream<Path> segmentsPaths = Files.walk(directory.toPath())
+                Stream<SegmentCacheEntry> segmentCacheEntryStream = Files.walk(directory.toPath())
+                        .map(path -> {
+                            try {
+                                return new SegmentCacheEntry(path, Files.readAttributes(path, BasicFileAttributes.class).lastAccessTime());
+                            } catch (IOException e) {
+                                logger.error("Error while getting the last access time for {}", path.toFile().getName());
+                            }
+                            return new SegmentCacheEntry(null, FileTime.fromMillis(Long.MAX_VALUE));
```
fabriziofortino: I think it would be better to return the actual path here. If for some weird reason we cannot read any lastAccessTime we won't be able to clean up the cache at all.
```suggestion
                            return new SegmentCacheEntry(path, FileTime.fromMillis(Long.MAX_VALUE));
```

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
```diff
@@ -193,16 +188,27 @@ public void cleanUp() {
...
+                        .filter(segmentCacheEntry -> !segmentCacheEntry.getPath().toFile().isDirectory());
 
-                StreamConsumer.forEach(segmentsPaths, (path, breaker) -> {
+                StreamConsumer.forEach(segmentCacheEntryStream, (segmentCacheEntry, breaker) -> {
 
                     if (cacheSize.get() > maxCacheSizeBytes * 0.66) {
-                        cacheSize.addAndGet(-path.toFile().length());
-                        path.toFile().delete();
-                        evictionCount.incrementAndGet();
+                        if (segmentCacheEntry.getPath() != null) { // it can be null if error has occurred while processing the stream
```
fabriziofortino: this won't be needed if we always return the path

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
```diff
@@ -213,6 +219,23 @@ private void cleanUpInternal() {
         }
     }
 
+    private class SegmentCacheEntry {
```
fabriziofortino: cosmetics: since this class is internal, we could have the members as `public final` and remove the getters to be more concise. Moreover, the class should be `static`.

fabriziofortino (approved):
LGTM, just a few comments.

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
smiroslav: Agree, thanks.

jelmini: Minor: as a matter of style, I think it's better to move the last return inside the catch clause because it makes clear that there is an happy path result and an error path result.
```suggestion
                            } catch (IOException e) {
                                logger.error("Error while getting the last access time for {}", path.toFile().getName());
                                return new SegmentCacheEntry(null, FileTime.fromMillis(Long.MAX_VALUE));
                            }
```

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
```diff
@@ -63,15 +63,10 @@
...
-    private static final Comparator<Path> sortedByAccessTime = (path1, path2) -> {
-        try {
-            FileTime lastAccessFile1 = Files.readAttributes(path1, BasicFileAttributes.class).lastAccessTime();
-            FileTime lastAccessFile2 = Files.readAttributes(path2, BasicFileAttributes.class).lastAccessTime();
-            return lastAccessFile1.compareTo(lastAccessFile2);
-        } catch (IOException e) {
-            logger.error("A problem occurred while cleaning up the cache: ", e);
-        }
-        return 0;
+    private static final Comparator<SegmentCacheEntry> sortedByAccessTime = (segmentCacheEntry1, segmentCacheEntry2) -> {
```
fabriziofortino: One more thing here: since you now have `SegmentCacheEntry` you could make it implement `Comparable` and move this logic in there. Then you need to call `sorted` on the stream below without arguments.

smiroslav: Good suggestion, thanks!

dulceanu (approved):
LGTM

PR #310, opened 2021-07-02
OAK-9482: upgrade httpclient to 4.5.13

fabriziofortino:


averma21 (approved):

reschke (changes requested):
You should update httpcore as well.

fabriziofortino:
@reschke I did not because httpclient 4.5.13 is compiled/tested/released with httpcore 4.4.13 (https://github.com/apache/httpcomponents-client/blob/rel/v4.5.13/pom.xml#L69). The latest httpcore will be used in the next release of httpclient.

thomasmueller:
Test

reschke:
See https://www.cvedetails.com/cve/CVE-2020-13956/

thomasmueller (approved):

fabriziofortino:
@reschke correct, the update to v 4.5.13 is to avoid the vulnerability you mentioned (https://issues.apache.org/jira/browse/OAK-9482)

reschke:
Yep, I misread the CVE.

Anyway; I understand the desire to use this combination, but given the fact that there are bugfixes in httpcore I really do not understand why we wouldn't use it, There's a reason why that release was made.

reschke (approved):

PR #325, opened 2021-07-16
OAK-9498 enable parallel Maven build

kwin:


mreutegg:
I'm not sure this works correctly when multiple modules run tests concurrently. Some modules use a local MongoDB and may write to the same database. I think this needs some changes to existing tests first.

kwin:
This requires some more work as outlined in https://github.com/apache/jackrabbit-oak/pull/325#issuecomment-881444200

PR #326, opened 2021-07-16
OAK-9496 | Update zookeeper version to handle CVE-2018-8012

nit0906:


fabriziofortino (approved):

fabriziofortino (changes requested):
I have actually run `mvn dependency:tree`. The solr version we are using depends on zookeeper 3.5.7. We should use this version. In general, I don't understand the reason we need to explicitly have this dependency here.

nit0906:
@fabriziofortino  - I have removed the dependency as we discussed.

nit0906:
[ERROR] Tests run: 5, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 12.46 s <<< FAILURE! - in org.apache.jackrabbit.oak.osgi.OSGiIT

[ERROR] bundleStates(org.apache.jackrabbit.oak.osgi.OSGiIT)  Time elapsed: 0.14 s  <<< FAILURE!

java.lang.AssertionError: Bundle org.apache.jackrabbit.oak-solr-osgi [38] not active. have a look at the logs expected:<32> but was:<2>

	at org.junit.Assert.fail(Assert.java:88)



Seems like removing the dep, prevents oak-solr-osgi from starting up.

PR #327, opened 2021-07-17
OAK-9497 enable logging of Apache DS through Logback during tests and set groups without JNDI LDAP client

kwin:


mreutegg (approved):
Looks good to me, but I'm not too familiar with this part of the code.

Code changes are tests only and they pass.

PR #328, opened 2021-07-19
OAK-9501 don't trim stack traces in unit and integration tests

kwin:


mreutegg (approved):
I don't have a strong opinion about this, but it looks good to me.

PR #333, opened 2021-07-29
OAK-9141: fix assertEventually logic with async indexing

fabriziofortino:
The `assertEventually` logic wrongly mixes seconds and milliseconds. We have never had issues so far but this could result in flaky tests when async indexing is enabled.


mreutegg (approved):
All tests on Jenkins were successful. Looks good.

averma21 (approved):

PR #336, opened 2021-08-03
OAK-9522 Index cost estimation: prefer union query with path restriction

thomasmueller:


oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexPathRestrictionTest.java
```diff
@@ -104,6 +105,53 @@ public void pathTranformationWithNoPathRestriction() throws Exception {
...
+            testRootBuilder.child("n" + i).setProperty("foo", "bar");
+        }
+        commit();
+
+        FilterImpl f;
+
+        // //*[foo = 'bar']
+        f = createFilter(root, NT_BASE);
+        f.restrictProperty("foo", Operator.EQUAL, PropertyValues.newString("bar"));
+        // equality check: we assume 5 different values
```
averma21: I guess this is referring to {{FulltextIndexPlanner#DEFAULT_PROPERTY_WEIGHT}}. I think it would be nice to use that constant here - both in comment and assert statement below, so that it is easier to understand and we don't need to change it, if the default weight changes.

averma21 (approved):
LGTM. Added a small suggestion.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexPathRestrictionTest.java
thomasmueller: +1

nit0906 (approved):
lgtm

PR #338, opened 2021-08-05
OAK-9524: Retry failed data download (resuming from the point it stop…

averma21:
…ped) during indexing

### DIFF
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/CompositeException.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/CompositeException.java
@@ -16,17 +16,26 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+package org.apache.jackrabbit.oak.index.indexer.document;
 
-package org.apache.jackrabbit.oak.index.indexer.document.flatfile;
+import org.slf4j.Logger;
 
-import org.apache.jackrabbit.oak.index.indexer.document.CompositeException;
+public class CompositeException extends Exception {
 
-import java.io.File;
-import java.io.IOException;
+    final Iterable<Throwable> collectedThrowables;
 
-public interface SortStrategy {
+    public CompositeException(Iterable<Throwable> collectedThrowables) {
+        this.collectedThrowables = collectedThrowables;
+    }
 
-    File createSortedStoreFile() throws IOException, CompositeException;
+    public Iterable<Throwable> getCollectedThrowables() {
+        return collectedThrowables;
+    }
 
-    long getEntryCount();
+    public void logAllExceptions(String message, Logger log) {
+        for (Throwable throwable : getCollectedThrowables()) {
+            log.error(message, throwable);
+        }
+    }
+
 }
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/CompositeException.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/CompositeException.java
@@ -22,18 +22,8 @@ import org.slf4j.Logger;
 
 public class CompositeException extends Exception {
 
-    final Iterable<Throwable> collectedThrowables;
-
-    public CompositeException(Iterable<Throwable> collectedThrowables) {
-        this.collectedThrowables = collectedThrowables;
-    }
-
-    public Iterable<Throwable> getCollectedThrowables() {
-        return collectedThrowables;
-    }
-
     public void logAllExceptions(String message, Logger log) {
-        for (Throwable throwable : getCollectedThrowables()) {
+        for (Throwable throwable : getSuppressed()) {
             log.error(message, throwable);
         }
     }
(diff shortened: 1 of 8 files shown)


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/CompositeException.java
```diff
@@ -0,0 +1,41 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.jackrabbit.oak.index.indexer.document;
+
+import org.slf4j.Logger;
+
+public class CompositeException extends Exception {
```
thomasmueller, line 23: What would happen if we also implement "Throwable[] getSuppressed​()" to return the (array of) collectedThrowables? That way it would be more generally useful (I think). I'm not saying this class should be moved to oak-commons right now... also I would still implement the logAllExceptions method, as it looks useful.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -133,6 +135,52 @@ public NodeStateEntryTraverser create(LastModifiedRange lastModifiedRange) {
...
+        DocumentNodeState rootDocumentState = (DocumentNodeState) checkpointedState;
+        DocumentNodeStore nodeStore = (DocumentNodeStore) indexHelper.getNodeStore();
+
+        DocumentStoreSplitter splitter = new DocumentStoreSplitter(getMongoDocumentStore());
+        List<Long> lastModifiedBreakPoints = splitter.split(Collection.NODES, 0L ,10);
+        FlatFileNodeStoreBuilder builder = null;
+
+        while (flatFileStore == null && executionCount <= MAX_DOWNLOAD_RETRIES) {
+            try {
+                //TODO Use flatFileStore only if we have relative nodes to be indexed
```
thomasmueller: Not quite sure why this TODO was+is there... I would remove it

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -254,7 +258,9 @@ private void resumeFromPreviousState(List<LastModifiedRange> previousState,
             LastModifiedRange currentRange = previousState.get(i);
             LastModifiedRange nextRange = i < previousState.size() - 1 ? previousState.get(i+1) : null;
             if (nextRange != null && currentRange.checkOverlap(nextRange)) {
-                throw new IllegalStateException("Range overlap between " + currentRange + " and " + nextRange);
+                LastModifiedRange merged = currentRange.mergeWith(nextRange);
+                log.info("Range overlap between " + currentRange + " and " + nextRange + ". Using merged range " + merged);
+                currentRange = merged;
```
thomasmueller, line 264: Not sure if I understand the logic... If you merge the ranges, wouldn't you want to remove nextRange as well? Is that anyway done and so not necessary?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/CompositeException.java
averma21: That method is final, so we can't override it. Adding/getting suppressed exceptions is provided by Throwables class itself. So, keeping this class just for exception type identification and logging method.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
averma21: yes, you're right :) This didn't surface up during tests because having the second range as well was just duplicating the downloads but duplicates are finally removed while merging. 
Fixed the code to skip the next range if merge is done.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
averma21: done.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/CompositeException.java
thomasmueller: I see, so instead of overriding getSuppressed() we would have to call addSuppressed(Throwable exception) for each entry in the list. I think it would still make sense do to that, but I don't have a strong opinion.

thomasmueller (approved):

PR #339, opened 2021-08-12
OAK-9529 | Log message 'Another copy of the index update is already r…

nit0906:
…unning' should be info

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
@@ -501,9 +501,9 @@ public class AsyncIndexUpdate implements Runnable, Closeable {
             long currentTime = System.currentTimeMillis();
             if (leaseEndTime > currentTime) {
                 long leaseExpMsg = (leaseEndTime - currentTime) / 1000;
-                String err = String.format("Another copy of the index update is already running; skipping this update. " +
+                String msg = String.format("Another copy of the index update is already running; skipping this update. " +
                         "Time left for lease to expire %d s. Indexing can resume by %tT", leaseExpMsg, leaseEndTime);
-                log.info(err);
+                log.info(msg);
                 return;
             }
         }
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
@@ -107,6 +107,7 @@ public class AsyncIndexUpdate implements Runnable, Closeable {
      * Name of service property which determines the name of Async task
      */
     public static final String PROP_ASYNC_NAME = "oak.async";
+    private static final String CONCURRENT_EXCEPTIPN_MSG ="Another copy of the index update is already running; skipping this update. ";
     private static final Logger log = LoggerFactory
             .getLogger(AsyncIndexUpdate.class);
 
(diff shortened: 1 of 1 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -501,9 +501,9 @@ private void runWhenPermitted() {
             long currentTime = System.currentTimeMillis();
             if (leaseEndTime > currentTime) {
                 long leaseExpMsg = (leaseEndTime - currentTime) / 1000;
-                String err = String.format("Another copy of the index update is already running; skipping this update. " +
+                String msg = String.format("Another copy of the index update is already running; skipping this update. " +
                         "Time left for lease to expire %d s. Indexing can resume by %tT", leaseExpMsg, leaseEndTime);
-                indexStats.failed(new Exception(err, newConcurrentUpdateException()));
```
thomasmueller, line 506: Are you sure this doesn't have a side effect?

nit0906: @thomasmueller  - I don't think so, because here we actually mark the index lane as failing in case there is a concurrent update running (I suppose this was a choice), in case we just log an info and don't mark these stats as failing and simply return, it shouldn't have any side effects.

Basically choosing here if we want to mark the indexing cycle as failed or simply skip it till next run.

thomasmueller (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1050,6 +1051,7 @@ public void done(String now) {
         }
 
         public void failed(Exception e) {
+            boolean isConcurrentUpdateException = (e.getMessage() != null) && (e.getMessage().startsWith(CONCURRENT_EXCEPTIPN_MSG));
```
thomasmueller: what about adding "e instanceof ConcurrentUpdateException" ?

thomasmueller: I would make it multi-line, as the line is quite long.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1066,13 +1068,21 @@ public void failed(Exception e) {
...
+                if (isConcurrentUpdateException) {
+                    log.info("[{}]", name,  e.getMessage());
+                } else {
+                    log.warn("[{}] The index update failed", name, e);
+                }
             } else {
                 // subsequent occurrences
                 boolean warn = System.currentTimeMillis() - latestErrorWarn > ERROR_WARN_INTERVAL;
                 if (warn) {
-                    latestErrorWarn = System.currentTimeMillis();
```
thomasmueller, line 1074: I would keep latestErrorWarn where it is... same as you did above.

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
nit0906: @thomasmueller 
e instanceof ConcurrentUpdateException -> Actually it's not ConcurrentUpdateException but rather CommitFailedException - see https://github.com/apache/jackrabbit-oak/blob/trunk/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java#L1557#L1559 

And CommitFailedException is more generic so I didn't want to make use of instanceOf.

I will make it multi line though.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
nit0906: done

PR #347, opened 2021-08-20
OAK-9546 failing test for references of moved node

kwin:


kwin:
@mreutegg Any idea why the query in https://github.com/apache/jackrabbit-oak/blob/4dfed10836051aab6d12cd0dc1b022c0745ae8de/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/identifier/IdentifierManager.java#L222 does not find the references in that case?

mreutegg:
I think this is still relevant.

PR #351, opened 2021-08-23
[Do not merge]OAK-9552: Don't index except if it's oak:QueryIndexDefinition

tihom88:


tihom88:
In this change we are not logging invalid index-definitions in "sync lucene" index

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
```diff
@@ -72,6 +73,9 @@
 
     private static final Logger log = LoggerFactory.getLogger(IndexUpdate.class);
     private static final String TYPE_ELASTICSEARCH = "elasticsearch";
+    //This is used so that wrong index definitions are sparsely logged. After every 1000 indexing cycles with diff, wrong index definitions
```
thomasmueller, line 76: By default the indexing cycle is once every 5 seconds. That mean 5000 seconds, or 83 minutes. I think that's  a good value.

thomasmueller: And for sync indexes, once in 1000 updates is fine as well I think.

thomasmueller (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
```diff
@@ -280,10 +296,19 @@ private void collectIndexEditors(NodeBuilder definitions,
...
+                String primaryType = definition.getName(JcrConstants.JCR_PRIMARYTYPE);
                 if (type == null) {
                     // probably not an index def
                     continue;
                 }
+                /*
+                 Log a warning after every indexJcrTypeInvalidLogLimiter cycles of indexer where nodeState changed.
+                 */
+                if (!IndexConstants.INDEX_DEFINITIONS_NODE_TYPE.equals(primaryType)
+                        && totalExecutionCount % indexJcrTypeInvalidLogLimiter == 0) {
```
thomasmueller, line 308: This isn't relevant here, but just FYI:

"%" (modulo) is a relatively slow CPU instructions (same as division). Multiplication is faster, and bitwise and is even faster. 

Here, in theory, it would be possible to use bitwise and, but then indexJcrTypeInvalidLogLimiter would need to be a power of 2. 

To avoid modulo for e.g. a hash map, when calculating the bucket from the hash code, if the number of buckets isn't a power of 2, one can use multiplication & shift. See https://lemire.me/blog/2016/06/27/a-fast-alternative-to-the-modulo-reduction/

thomasmueller (approved):

tihom88:
Merged with https://github.com/apache/jackrabbit-oak/pull/357

PR #357, opened 2021-09-03
OAK-9552: Don't index except if it's oak:QueryIndexDefinition

tihom88:


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
```diff
@@ -72,6 +73,16 @@
 
     private static final Logger log = LoggerFactory.getLogger(IndexUpdate.class);
     private static final String TYPE_ELASTICSEARCH = "elasticsearch";
+    public static final String DEFAULT_indexJcrTypeInvalidLogLimiter = "1000";
```
thomasmueller: I would just inline the 1000 in the line below.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
```diff
@@ -72,6 +73,16 @@
...
+    //This is used so that wrong index definitions are sparsely logged. After every 1000 indexing cycles, index definitions
+    // with wrong nodetype will be logged.
+    private static final long indexJcrTypeInvalidLogLimiter = Long.valueOf(System.getProperty("oak.indexer.indexJcrTypeInvalidLogLimiter", DEFAULT_indexJcrTypeInvalidLogLimiter));
```
thomasmueller: This should be all caps. And then I think it can be public, so you can use it from the test.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
```diff
@@ -274,16 +284,35 @@ private static boolean hasAnyHiddenNodes(NodeBuilder builder){
         return false;
     }
 
+    private void sparseLoggger(String primaryType, String name) {
+        synchronized (countLock) {
```
thomasmueller: I don't really see a point in synchronizing it... it's just a warning message, why is it important that it's synchronized? Worst case (very unlikey), we see two warning messages, right? Or is there some deeper reason?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
```diff
@@ -274,16 +284,35 @@ private static boolean hasAnyHiddenNodes(NodeBuilder builder){
...
+            if ((cyclicExecutionCount >= indexJcrTypeInvalidLogLimiter)) {
```
thomasmueller: I liked the previous "executionCount % LIMIT == 0" better, because it's simpler.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
tihom88: I am not setting this in test using system property as it is "static final" --> this gets initialised first with default value whenever any test runs and then I am unable to change it in the test. I don't want to add a setter or not make it final just to pass the test. 

Will change it to all caps

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
tihom88: Well yes, it is not that important to log... will remove synchronize.
There can be double/ no-logs dependent on thread state

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
tihom88: well yes, it is complicated than executionCount % LIMIT but I think it is not that complicated. We are just resetting the count after LIMIT.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
fabriziofortino: You can replace `valueOf` with `parseLong` to avoid unneeded boxing.

```suggestion
    private static final long indexJcrTypeInvalidLogLimiter = Long.parseLong(System.getProperty("oak.indexer.indexJcrTypeInvalidLogLimiter", DEFAULT_indexJcrTypeInvalidLogLimiter));
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
```diff
@@ -274,16 +284,35 @@ private static boolean hasAnyHiddenNodes(NodeBuilder builder){
         return false;
     }
 
+    private void sparseLoggger(String primaryType, String name) {
```
fabriziofortino: typo: there are 3 gs in `sparseLoggger`

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/InvalidIndexTest.java
```diff
@@ -0,0 +1,197 @@
...
+        test1.setProperty("jcr:primaryType", invalidJcrPrimaryTypeForIndex, Type.NAME);
+        test1.setProperty("type", "lucene");
+        test1.setProperty("reindex", true);
+        test1.setProperty("async", "async");
+        root.commit();
+        runIndexerCyclesAfterEachNodeCommit(contentCount, true);
+        runEmptyAsyncIndexerCyclesWithoutNewContent(contentCount);
+
+        List<String> logs = customLogger.getLogs();
+        assertEquals(assertionLogPresent(logs, invalidJcrPrimaryTypeLog()), true);
```
fabriziofortino: You could use `assertTrue` here and in the tests below.

```suggestion
        assertTrue(assertionLogPresent(logs, invalidJcrPrimaryTypeLog()));
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
fabriziofortino: +1 on removing sync

fabriziofortino (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
thomasmueller: Making it public will keep it read-only (as it's final). I thought making it public so it can be read in the test.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
thomasmueller: having executionCount not going back to 0 would give us some advantage I think... we could log the value, and therefore see how often the method was called... but it's not directly needed.

fabriziofortino (approved):

PR #361, opened 2021-09-06
OAK-9565: NullPointerException on Session.logout()

mreutegg:
Test case with proposed fix.

### DIFF
--- a/oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/session/SessionLogoutTest.java
+++ b/oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/session/SessionLogoutTest.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


mreutegg:
`SessionStatsTest.testInitStackTraceEnabledAfterOpeningManySessions` fails because of a side effect of the new `SessionLogoutTest`. The new test provokes the issue described in OAK-9262. The initStackTrace feature and the test depends on a correct session counter, which is violated when `SessionLogoutTest` runs.

I'm going to change the fix to also include OAK-9262.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/session/SessionLogoutTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+import javax.jcr.Session;
+
+import org.apache.jackrabbit.test.AbstractJCRTest;
+
+public class SessionLogoutTest extends AbstractJCRTest {
+
+    public void testConcurrentSessionLogout() throws Exception {
+        Session s = getHelper().getReadWriteSession();
+        List<Exception> exceptions = new ArrayList<>();
```
stefan-egli: (Just a note to self: I was first wondering whether you'd need a concurrent version of List (eg ConcurrentLinkedQueue) - but Thread.join indeed, and somewhat obviously, honours the ['happens-before order' ](https://docs.oracle.com/javase/specs/jls/se7/html/jls-17.html#jls-17.4.5), so this works perfectly fine.)

stefan-egli (approved):

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/session/SessionLogoutTest.java
mreutegg: Hmm, right. But the List must still be thread-safe, which is currently not the case. Multiple threads may add exceptions to the list concurrently in the failure case. I'm going to change that.

PR #364, opened 2021-09-08
OAK-9568 : Support custom osgi LeaseFailureHandler

stefan-egli:


mreutegg:
Can you please add a test? E.g. to DocumentNodeStoreServiceTest.

stefan-egli:
> Can you please add a test? E.g. to DocumentNodeStoreServiceTest.

added test cases at 0480ce80d8bdf3785f9d5081763fb0efecc39206 : testing the default case is a bit simplistic, as all it does is to ensure an `UnsupportedOperationException` is thrown - which is due to the mocking framework in use doesn't support `bundle.stop()`. That's a bit weak - but then again, it would be quite some effort to boost up the mocking to just to ensure `bundle.stop()` does get called.

mreutegg:
That's fine, I was anyway more interested in the custom case and not the default one.

Hmm, but those two tests failed on Jenkins because they cannot connect to a MongoDB.

mreutegg:
One more thing. `LeaseFailureHandler` is currently not in an exported package. The package `org.apache.jackrabbit.oak.plugins.document.spi` is exported, but does not maintain a package-info.java with an explicit export version. What do you think about moving it to a new package `org.apache.jackrabbit.oak.plugins.document.spi.lease`?

stefan-egli:
+1, moved LeaseFailureHandler to new package.

mreutegg:
I think you also need to add the package to the list of exported packages in `oak-store-document/pom.xml`. This could be checked with new test in `oak-it-osgi`.

stefan-egli:
oups, right

mreutegg (approved):
Looks good to me now.

stefan-egli:
Thx for reviewing!

PR #367, opened 2021-09-09
OAK-9552: Don't index except if it's oak:QueryIndexDefinition

tihom88:


tihom88:
Rectified few failing tests.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
```diff
@@ -280,10 +288,25 @@ private void collectIndexEditors(NodeBuilder definitions,
...
                     continue;
                 }
+                /*
+                 Log a warning after every indexJcrTypeInvalidLogLimiter cycles of indexer where nodeState changed.
+                 and skip further execution for invalid nodetype of index definition.
+                 */
+                if (!IndexConstants.INDEX_DEFINITIONS_NODE_TYPE.equals(primaryType)) {
+                    // It is a cyclic counter which reset back to 0 after INDEX_JCR_TYPE_INVALID_LOG_LIMITER
+                    // This is to sparsely log this warning.
+                    if ((cyclicExecutionCount >= INDEX_JCR_TYPE_INVALID_LOG_LIMITER)) {
```
thomasmueller, line 303: With this logic, it will log after 1000 executions. That is, with a delay of 5 seconds, after 5000 seconds, or 83 minutes.

It would be better to log in the first call as well - you see I keep coming back to the original 

    executionCount % INDEX_JCR_TYPE_INVALID_LOG_LIMITER == 0

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
tihom88: Well the case is already taken care of as we are setting :
`private static volatile long cyclicExecutionCount = INDEX_JCR_TYPE_INVALID_LOG_LIMITER;`
So on first call itself the condition to log is met.

But I can change it to modulo, if it is about readability of code.

thomasmueller (approved):

fabriziofortino (approved):

PR #368, opened 2021-09-09
OAK-9572: index aggregations should ignore hidden properties

fabriziofortino:


tihom88:
@fabriziofortino shouldn't we also ignore hidden nodes?

thomasmueller (approved):

fabriziofortino:
@tihom88 hidden nodes seem to be already ignored. I have changed the unit test to include them as well. See https://github.com/fabriziofortino/jackrabbit-oak/commit/8028be77afcd2c842b50e151536faea530ed068c

PR #370, opened 2021-09-13
Oak-9566 | Improve Index Stats

nit0906:
* Show all indexes (not just the ones recently opened)
* Show Additional info such as hasPropertyIndex and hasHiddenOakLibsMount
* Show lastUpdatedTimestamp

### DIFF
--- a/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexInfoProvider.java
+++ b/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexInfoProvider.java
@@ -141,17 +141,13 @@ public class LuceneIndexInfoProvider implements IndexInfoProvider {
 
     private static void checkIfHiddenNodesExists(NodeState idxState, LuceneIndexInfo info) {
         // Check for hidden oak libs mount node that has indexed content for read only repo in composite store
+        // Also check for hidden property index node :property-index - present in case of hybrid indexes
         info.hasHiddenOakLibsMount = false;
+        info.hasPropertyIndexNode = false;
         for(String c : idxState.getChildNodeNames()) {
             if (c.startsWith(IndexDefinition.HIDDEN_OAK_MOUNT_PREFIX)) {
                 info.hasHiddenOakLibsMount = true;
-            }
-        }
-
-        // Now check for hidden property index node :property-index - present in case of hybrid indexes
-        info.hasPropertyIndexNode = false;
-        for(String c : idxState.getChildNodeNames()) {
-            if (c.equals(IndexDefinition.PROPERTY_INDEX)) {
+            } else if (c.equals(IndexDefinition.PROPERTY_INDEX)) {
                 info.hasPropertyIndexNode = true;
             }
         }
--- a/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexInfoProvider.java
+++ b/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexInfoProvider.java
@@ -149,6 +149,7 @@ public class LuceneIndexInfoProvider implements IndexInfoProvider {
         }
 
         // Now check for hidden property index node :property-index - present in case of hybrid indexes
+        info.hasPropertyIndexNode = false;
         for(String c : idxState.getChildNodeNames()) {
             if (c.equals(IndexDefinition.PROPERTY_INDEX)) {
                 info.hasPropertyIndexNode = true;
--- a/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexInfoProvider.java
+++ b/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexInfoProvider.java
@@ -33,8 +33,6 @@ import org.apache.jackrabbit.oak.plugins.index.IndexConstants;
 import org.apache.jackrabbit.oak.plugins.index.IndexInfo;
 import org.apache.jackrabbit.oak.plugins.index.IndexInfoProvider;
 import org.apache.jackrabbit.oak.plugins.index.IndexUtils;
-import org.apache.jackrabbit.oak.plugins.index.lucene.LuceneIndexConstants;
-import org.apache.jackrabbit.oak.plugins.index.lucene.LuceneIndexDefinition;
 import org.apache.jackrabbit.oak.plugins.index.lucene.directory.DirectoryUtils;
 import org.apache.jackrabbit.oak.plugins.index.lucene.directory.IndexConsistencyChecker;
 import org.apache.jackrabbit.oak.plugins.index.lucene.directory.OakDirectory;
@@ -87,6 +85,7 @@ public class LuceneIndexInfoProvider implements IndexInfoProvider {
         computeIndexDefinitionChange(idxState, info);
         computeLastUpdatedTime(idxState, info);
         computeAsyncIndexInfo(idxState, indexPath, info);
+        checkIfHiddenNodesExists(idxState, info);
         return info;
     }
 
@@ -140,6 +139,23 @@ public class LuceneIndexInfoProvider implements IndexInfoProvider {
         }
     }
 
+    private static void checkIfHiddenNodesExists(NodeState idxState, LuceneIndexInfo info) {
+        // Check for hidden oak libs mount node that has indexed content for read only repo in composite store
+        info.hasHiddenOakLibsMount = false;
+        for(String c : idxState.getChildNodeNames()) {
+            if (c.startsWith(IndexDefinition.HIDDEN_OAK_MOUNT_PREFIX)) {
+                info.hasHiddenOakLibsMount = true;
+            }
+        }
+
+        // Now check for hidden property index node :property-index - present in case of hybrid indexes
+        for(String c : idxState.getChildNodeNames()) {
+            if (c.equals(IndexDefinition.PROPERTY_INDEX)) {
+                info.hasPropertyIndexNode = true;
+            }
+        }
+    }
+
     private static void computeIndexDefinitionChange(NodeState idxState, LuceneIndexInfo info) {
         NodeState storedDefn = idxState.getChildNode(INDEX_DEFINITION_NODE);
         if (storedDefn.exists()) {
@@ -160,6 +176,8 @@ public class LuceneIndexInfoProvider implements IndexInfoProvider {
         long lastUpdatedTime;
         boolean indexDefinitionChanged;
         String indexDiff;
+        boolean hasHiddenOakLibsMount;
+        boolean hasPropertyIndexNode;
 
         public LuceneIndexInfo(String indexPath) {
             this.indexPath = indexPath;
@@ -208,6 +226,16 @@ public class LuceneIndexInfoProvider implements IndexInfoProvider {
         @Override
         public String getIndexDefinitionDiff() {
             return indexDiff;
+        }
+
+        @Override
+        public boolean hasHiddenOakLibsMount() {
+            return hasHiddenOakLibsMount;
+        }
+
+        @Override
+        public boolean hasPropertyIndexNode() {
+            return hasPropertyIndexNode;
         }
     }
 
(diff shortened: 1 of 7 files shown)


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexInfoProvider.java
```diff
@@ -140,6 +139,24 @@ private static void computeLastUpdatedTime(NodeState idxState, LuceneIndexInfo i
         }
     }
 
+    private static void checkIfHiddenNodesExists(NodeState idxState, LuceneIndexInfo info) {
+        // Check for hidden oak libs mount node that has indexed content for read only repo in composite store
+        info.hasHiddenOakLibsMount = false;
+        for(String c : idxState.getChildNodeNames()) {
+            if (c.startsWith(IndexDefinition.HIDDEN_OAK_MOUNT_PREFIX)) {
+                info.hasHiddenOakLibsMount = true;
```
fabriziofortino, line 149: `break`?

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexInfoProvider.java
```diff
@@ -140,6 +139,24 @@ private static void computeLastUpdatedTime(NodeState idxState, LuceneIndexInfo i
...
+            }
+        }
+
+        // Now check for hidden property index node :property-index - present in case of hybrid indexes
+        info.hasPropertyIndexNode = false;
+        for(String c : idxState.getChildNodeNames()) {
+            if (c.equals(IndexDefinition.PROPERTY_INDEX)) {
+                info.hasPropertyIndexNode = true;
```
fabriziofortino, line 151: `break` ?

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexInfoProvider.java
```diff
@@ -140,6 +139,24 @@ private static void computeLastUpdatedTime(NodeState idxState, LuceneIndexInfo i
...
+            }
+        }
+    }
```
fabriziofortino, line 154: Apart from terminating the conditions with a break, I think this logic can be simplified with a single for-loop.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexMBeanImpl.java
```diff
@@ -122,13 +133,13 @@ public TabularData getIndexStats() throws IOException {
             TabularType tt = new TabularType(LuceneIndexMBeanImpl.class.getName(),
                     "Lucene Index Stats", IndexStats.TYPE, new String[]{"path"});
             tds = new TabularDataSupport(tt);
-            Set<String> indexes = indexTracker.getIndexNodePaths();
-            for (String path : indexes) {
+            // Use indexPathService to get list of all the lucene indexes.
```
fabriziofortino, line 138: If I get it right, this is needed to show all the indexes while the `IndexTracker` returns the open ones only. Can we explain this in the comment?

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexMBeanImpl.java
```diff
@@ -269,6 +280,10 @@ public boolean isFailing() {
         return getFieldTermPrefixInfo(indexPath, field, Integer.MAX_VALUE, term);
     }
 
+    private void bindIndexInfoProviders(IndexInfoServiceImpl indexInfoService) {
```
fabriziofortino: this method is called only in the constructor. To improve the readability, I would move this logic in the constructor so the reader does not have to jump somewhere else to see what it does.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexMBeanImpl.java
```diff
@@ -675,7 +690,10 @@ private static Query newDepthQuery(int depth) {
                 "numDeletedDocs",
                 "nrtIndexSize",
                 "nrtIndexSizeStr",
-                "nrtNumDocs"
+                "nrtNumDocs",
+                "lastUpdatedTimeStamp",
```
fabriziofortino: Timestamp can be written with one or two words. Usually, in software, the single word version is used.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexInfoProvider.java
```diff
@@ -140,6 +139,24 @@ private static void computeLastUpdatedTime(NodeState idxState, LuceneIndexInfo i
         }
     }
 
+    private static void checkIfHiddenNodesExists(NodeState idxState, LuceneIndexInfo info) {
+        // Check for hidden oak libs mount node that has indexed content for read only repo in composite store
+        info.hasHiddenOakLibsMount = false;
+        for(String c : idxState.getChildNodeNames()) {
+            if (c.startsWith(IndexDefinition.HIDDEN_OAK_MOUNT_PREFIX)) {
```
fabriziofortino, line 148: I am not sure about this condition here. Instead of `startsWith` shouldn't we check for a node with name `:oak:mount-libs-index-data`? cc @thomasmueller

fabriziofortino (changes requested):

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexMBeanImpl.java
nit0906: done

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexMBeanImpl.java
nit0906: done

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexMBeanImpl.java
nit0906: done

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexInfoProvider.java
nit0906: @fabriziofortino  - I modified it to be in one for loop. I didn't add the break statement, because the we only have fixed 3-4 child nodes here. So break statement is not required as such.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexInfoProvider.java
nit0906: @fabriziofortino  - we have the same check currently in the isActiveIndex check in IndexName java - so this should work.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexInfoProvider.java
nit0906: Refer below comment on for loop

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexInfoProvider.java
nit0906: Refer below comment on for loop

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexMBeanImpl.java
fabriziofortino: @nit0906 I think this change is sound from a logical perspective. I am wondering though if this can cause some performance issues. The `indexPathService` returns all the indexes, even the ones that are not used (eg: an old version of an index). When we call `indexTracker.acquireIndexNode` (line 142) the `IndexSearcher` gets initialized. If the index is not present it gets downloaded. We can potentially download a lot of indexes only to serve the stats.

nit0906: @fabriziofortino  - Agreed with your point about the performance. 

We can ignore older versions of indexes here (using the same filter logic as is used in the FullTextIndex while creating the index plans) - but I am wondering if that would be correct ? Because for statistics, we would want to show the older indexes as well - if they are active (being indexed) and/or consuming disk space. 

Thoughts ? (cc : @thomasmueller  @tihom88 )

fabriziofortino: I think we could list all the indexes regardless if they are active or not (this would already be an improvement compared to what we have).
Only the active ones will have full stats info (size, number of docs, etc).
Opening a searcher, and potentially downloading an inactive huge index, only to compute the stats seems a waste of resources to me.

nit0906: @fabriziofortino  - how about, I send the isActive info in the IndexStats constructor and not call the getSearcher() for old/inactive indexes. 
This way, we can get size for them (which might be relevant). And we don't calculate the numDocs etc (for which we open the searcher).

fabriziofortino: It sounds good to me.

nit0906: @fabriziofortino  - it seems we init the searcher when we do acuireIndexNode from IndexTracker - so we wouldn't be able to avoid that. 

So it's either get stats for all indexes (disabled/old/active). Or get stats only for active ones and ignore disabled and old indexes completely.  We can't do the half and half approach as I thought earlier.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexInfoProvider.java
thomasmueller: `startsWith` makes sense in my view... The cases we have so far is all `:oak:mount-libs-index-data`, but someone might change the read-only mount names. I think the prefix will always stay like this.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexMBeanImpl.java
thomasmueller: We need to avoid downloading all indexes, that's for sure. If that means we have some unknown data (numDocs and so on), then that's fine. Should we set "unknown" to the value -1 or 0?

thomasmueller: I hope some of the info we can still list, even without potentially downloading the indexes (by calling acquireIndexNode) - for example lastUpdatedTimestamp, hasHiddenOakLibsMount, hasPropertyIndex, and size on disk. But I guess we can't list things like nrtNumDocs, numDocs, and so on. Right?

nit0906: @thomasmueller  - We can get lastUpdatedTimestamp, hasHiddenOakLibsMount, hasPropertyIndex, but not size on disk, as currently that is also calculated using IndexTracker's get Readers.

PR #372, opened 2021-09-20
OAK-9566 | Improve index stats

nit0906:
* Added support for Json format 
* Fixed issue with local index dir size - currently with composite node store - it was just reporting size of read only mount
* Added new stats for indexCreationTimestamp, ReindexCompletionTimestamp, suggestDirSize, If the index has hidden property index node and if it has a hidden read only libs mount.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
@@ -78,7 +78,7 @@ public class IndexPrinter implements InventoryPrinter {
             pw.print(JsopBuilder.prettyPrint(json.toString()));
         } else {
             //TODO Highlight if failing
-            printAsyncIndexInfo(pw, null,format);
+            printAsyncIndexInfo(pw, null, format);
             printIndexInfo(pw, null, format);
         }
     }
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
@@ -203,7 +203,7 @@ public class IndexPrinter implements InventoryPrinter {
             keyValue("    Estimated entry count    ", info.getEstimatedEntryCount(), pw, json, format);
         }
 
-        if (info.getType().equals("lucene")) {
+        if ("lucene".equals(info.getType())) {
             // Only valid for lucene type indexes, for others it will simply show false.
             keyValue("    Has hidden oak mount     ", info.hasHiddenOakLibsMount(), pw, json, format);
             keyValue("    Has property index       ", info.hasPropertyIndexNode(), pw, json, format);
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
@@ -20,8 +20,11 @@
 package org.apache.jackrabbit.oak.plugins.index.inventory;
 
 import java.io.PrintWriter;
+import java.text.SimpleDateFormat;
 import java.util.Calendar;
+import java.util.Date;
 import java.util.List;
+import java.util.TimeZone;
 
 import com.google.common.base.Strings;
 import com.google.common.collect.ArrayListMultimap;
@@ -183,10 +186,12 @@ public class IndexPrinter implements InventoryPrinter {
 
         if (info.getSizeInBytes() >= 0){
             keyValue("    Size                     ", IOUtils.humanReadableByteCount(info.getSizeInBytes()), pw, json, format);
+            keyValue("    Size (in Bytes)          ", info.getSizeInBytes(), pw, json, format);
         }
 
         if (info.getSuggestSizeInBytes() >= 0){
             keyValue("    Suggest size             ", IOUtils.humanReadableByteCount(info.getSuggestSizeInBytes()), pw, json, format);
+            keyValue("    Suggest size (in Bytes)  ", info.getSuggestSizeInBytes(), pw, json, format);
         }
 
         if (info.getEstimatedEntryCount() >= 0){
@@ -212,7 +217,10 @@ public class IndexPrinter implements InventoryPrinter {
     private static String formatTime(long time){
         Calendar cal = Calendar.getInstance();
         cal.setTimeInMillis(time);
-        return ISO8601.format(cal);
+        Date date = cal.getTime();
+        SimpleDateFormat outputFmt = new SimpleDateFormat("yyyy-mm-dd'T'HH:mm:ss.s z");
+        outputFmt.setTimeZone(TimeZone.getTimeZone("UTC")); //set timezone here
+        return outputFmt.format(date);
     }
 
     private static void keyValue(String key, Object value, PrintWriter pw, JsopBuilder json, Format format) {
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
@@ -31,6 +31,7 @@ import org.apache.felix.inventory.Format;
 import org.apache.felix.inventory.InventoryPrinter;
 import org.apache.jackrabbit.oak.api.jmx.IndexStatsMBean;
 import org.apache.jackrabbit.oak.commons.IOUtils;
+import org.apache.jackrabbit.oak.commons.json.JsopBuilder;
 import org.apache.jackrabbit.oak.plugins.index.AsyncIndexInfo;
 import org.apache.jackrabbit.oak.plugins.index.AsyncIndexInfoService;
 import org.apache.jackrabbit.oak.plugins.index.IndexInfo;
@@ -46,7 +47,8 @@ import static com.google.common.base.Preconditions.checkNotNull;
         property = {
             "felix.inventory.printer.name=oak-index-stats",
             "felix.inventory.printer.title=Oak Index Stats",
-            "felix.inventory.printer.format=TEXT"
+            "felix.inventory.printer.format=TEXT",
+                "felix.inventory.printer.format=JSON"
         })
 public class IndexPrinter implements InventoryPrinter {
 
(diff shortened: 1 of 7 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
```diff
@@ -46,7 +47,8 @@
         property = {
             "felix.inventory.printer.name=oak-index-stats",
             "felix.inventory.printer.title=Oak Index Stats",
-            "felix.inventory.printer.format=TEXT"
+            "felix.inventory.printer.format=TEXT",
+                "felix.inventory.printer.format=JSON"
```
fabriziofortino, line 54: minor: align formatting

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
```diff
@@ -66,97 +68,193 @@ public IndexPrinter(IndexInfoService indexInfoService, AsyncIndexInfoService asy
...
+            } else if (value instanceof Integer) {
+                json.value((Integer) value);
+            }
+        } else if (format == Format.TEXT) {
+            pw.printf(key+":%s%n",value);
+        }
+    }
+
+    // Wrappers around JsonBuilder that will result in NOOP if builder is null -
+    // These are just to avoid the mulitple if/else check while handling for both TEXT and JSON formats
```
fabriziofortino: typo
```suggestion
    // These are just to avoid the multiple if/else check while handling for both TEXT and JSON formats
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
```diff
@@ -66,97 +68,193 @@ public IndexPrinter(IndexInfoService indexInfoService, AsyncIndexInfoService asy
...
+        printWithNewLine(pw, "", format);
+
         for (String lane : asyncLanes) {
-            pw.println(lane);
+            printWithNewLine(pw, lane, format);
+            addJsonKey(json, lane);
             AsyncIndexInfo info = asyncIndexInfoService.getInfo(lane);
             if (info != null) {
-                        pw.printf("    Last indexed to      : %s%n", formatTime(info.getLastIndexedTo()));
+                startJsonObject(json);
```
fabriziofortino: `startJsonObject(json);` is called in both `if` and `else` blocks. You can remove from both an move it before the `if` block.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
```diff
@@ -66,97 +68,193 @@ public IndexPrinter(IndexInfoService indexInfoService, AsyncIndexInfoService asy
...
+            keyValue("    Estimated entry count    ", info.getEstimatedEntryCount(), pw, json, format);
         }
 
-        if (info.hasIndexDefinitionChangedWithoutReindexing()) {
-            pw.println("    Index definition changed without reindexing");
-            String diff = info.getIndexDefinitionDiff();
-            if (diff != null) {
-                pw.println("    "+diff);
-            }
+        if (info.getType() == "lucene") {
```
fabriziofortino: use `equals` instead of `==`

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinterTest.java
```diff
@@ -49,11 +53,30 @@ public void asyncIndexInfo() throws Exception {
...
+        when(asyncInfo.getInfo("foo-async"))
+                .thenReturn(new AsyncIndexInfo("foo-async", 0, 0, false, null));
+
+        String output = getPrintOutput(Format.JSON);
+
+        JsonObject json = JsonObject.fromJson(output, true);
+        Map<String, JsonObject> jsonMap = json.getChildren();
+        assertTrue(jsonMap.keySet().contains("Async Indexers State"));
+        int size  = Integer.parseInt(jsonMap.get("Async Indexers State").getProperties().get("Number of async indexer lanes"));
+        assertTrue(size == 2);
```
fabriziofortino:
```suggestion
        assertEquals(2, size);
```

fabriziofortino (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
nit0906: done

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
nit0906: done

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
nit0906: done

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinterTest.java
nit0906: done

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
nit0906: done

thomasmueller (commented):
Could you add an example output, in Json format, to https://issues.apache.org/jira/browse/OAK-9566 ?
Or (might be easier) add a test case with Json output... anyway having a test case would be good.

fabriziofortino (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
```diff
@@ -66,97 +68,190 @@ public IndexPrinter(IndexInfoService indexInfoService, AsyncIndexInfoService asy
...
+            keyValue("    Estimated entry count    ", info.getEstimatedEntryCount(), pw, json, format);
         }
 
-        if (info.hasIndexDefinitionChangedWithoutReindexing()) {
-            pw.println("    Index definition changed without reindexing");
-            String diff = info.getIndexDefinitionDiff();
-            if (diff != null) {
-                pw.println("    "+diff);
-            }
+        if (info.getType().equals("lucene")) {
```
fabriziofortino: just to avoid npe in case the type is not specified
```suggestion
        if ("lucene".equals(info.getType())) {
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
```diff
@@ -66,97 +68,190 @@ public IndexPrinter(IndexInfoService indexInfoService, AsyncIndexInfoService asy
...
+            json.key(key.trim());
+            if (value instanceof  String) {
+                json.value((String)value);
+            } else if (value instanceof  Long) {
+                json.value((Long)value);
+            } else if (value instanceof  Boolean) {
+                json.value((Boolean)value);
+            } else if (value instanceof Integer) {
+                json.value((Integer) value);
+            }
```
thomasmueller, line 245: I would probably handle "null" separately at the beginning (if value == null). Could throw an exception.
Also, "else" for all other cases is missing (could e.g. throw an exception)

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
```diff
@@ -66,97 +68,190 @@ public IndexPrinter(IndexInfoService indexInfoService, AsyncIndexInfoService asy
...
-            printTitle(pw, title);
-            pw.println();
+            printTitle(pw, title, format);
+            addJsonKey(json, type);
+            startJsonObject(json);
             for (IndexInfo info : typedInfo){
-                printIndexInfo(pw, info);
+                printIndexInfo(pw, json, info, format);
+                if (format == Format.TEXT) {
+                    if (info.hasIndexDefinitionChangedWithoutReindexing()) {
```
thomasmueller: The json output could also have this info I assume, right?

nit0906:
@thomasmueller -
>> Could you add an example output, in Json format, to https://issues.apache.org/jira/browse/OAK-9566 ?

Attached a sample o/p to the jira. Please note that the values in it are not correct and some are missing, since It's from an earlier revision and was not run on composite node store - but the format is same so it'll give an idea about how it's structured.

 >> Or (might be easier) add a test case with Json output... anyway having a test case would be good.

I added a couple of tests cases. See IndexPrinterTest#asyncIndexInfoJson and  IndexPrinterTest#indexInfoJson

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
nit0906: done

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
```diff
@@ -66,97 +68,190 @@ public IndexPrinter(IndexInfoService indexInfoService, AsyncIndexInfoService asy
...
+            keyValue("     Creation time           ", formatTime(info.getCreationTimestamp()), pw, json, format);
+        }
+
+        if (info.getReindexCompletionTimestamp() > 0){
+            keyValue("     Reindex completion time ", formatTime(info.getReindexCompletionTimestamp()), pw, json, format);
         }
 
         if (info.getSizeInBytes() >= 0){
-            pw.printf("    Size                    : %s%n", IOUtils.humanReadableByteCount(info.getSizeInBytes()));
+            keyValue("    Size                     ", IOUtils.humanReadableByteCount(info.getSizeInBytes()), pw, json, format);
```
thomasmueller, line 188: My personal opinion is that we should list all size data in the same way, e.g. all in GB. Examples:
* 1000 (= 1 TB)
* 1 (= 1 GB)
* 0.001 (= 1 MB)
* 0.000001 (= 1 B)

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
nit0906: @thomasmueller  - I disagree a bit here. Since this is human readable format (that would not be consumed by any code for calculations/alerting ), so I think, 1 B makes more sense than 0.000001 GB. 

If we want to use these for some calculations/alerting - then I agree with your point that they should be in the same unit (and in which case we would have to change the formatting even a bit more ), because here the o/p would be string - 10B, 10MB, 10GB and like that.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
nit0906: done

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
nit0906: yes - I was thinking that this diff was also in json, and might cause formatting issues - but it's not in json. I have modified for the json stats also to include this

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
thomasmueller: > I disagree a bit here.

That's fine, it's just my personal opinion. When I investigate a problem, in the Lucene index stats, it I usually search for "GB" and then look at only those that have this unit; or use grep "GB". I would miss entries with "TB" however. 

So I argue that even for a human, using the same unit would make sense, assuming there is a sensible default unit.

We could round to 3 digits after the comma (e.g. round up from 0.000001 GB => 0.001 GB).

thomasmueller: > If we want to use these for some calculations/alerting

Looking at the json output, I think it would make sense to use it for alerting... What about we list both the "human readable" as well as the "same unit" numbers?

thomasmueller:
I wonder if the timestamps would be better listed in UTC? Right now we have

    "Last indexed up to": "2021-09-17T12:52:02.619+05:30",

That would become

    "Last indexed up to": "2021-09-17T07:52:02.619Z",

Where the "Z" is from here: https://www.w3.org/TR/NOTE-datetime

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
nit0906: Makes sense. I will add both the String format (that appends  B, MB, GB  accordingly) and the numeric one (in bytes).

nit0906:
@thomasmueller 
>>> I wonder if the timestamps would be better listed in UTC?

Sure - I transformed them in UTC now. And also added another property to show the sized in bytes (numeric) that can be used for alerts etc. Added a new sample json here - https://issues.apache.org/jira/browse/OAK-9566?focusedCommentId=17417733&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17417733 .

thomasmueller:
> https://issues.apache.org/jira/browse/OAK-9566?focusedCommentId=17417733&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17417733

The example uses "2021-55-20T16:55:53.53 UTC", you have used seconds instead of month...

While "2021-09-20T16:55:53.53 UTC" is clear for the user, is not a standardized format... https://www.w3.org/TR/NOTE-datetime specifies "2021-09-17T07:52:02.619Z". 

If we want "human readable", then what about "2021-09-20 16:55:53 UTC" (no "T", and no fractional seconds).

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
```diff
@@ -66,97 +71,200 @@ public IndexPrinter(IndexInfoService indexInfoService, AsyncIndexInfoService asy
...
-        pw.println();
+        endJsonObject(json);
     }
 
     private static String formatTime(long time){
         Calendar cal = Calendar.getInstance();
         cal.setTimeInMillis(time);
-        return ISO8601.format(cal);
+        Date date = cal.getTime();
+        SimpleDateFormat outputFmt = new SimpleDateFormat("yyyy-mm-dd'T'HH:mm:ss.s z");
```
thomasmueller: mm is for minutes, but you need month, I think that would be MM. "ss.s" is incorrect, it shows seconds twice. I don't think we need milliseconds or fractional seconds, but if you want them then this needs to be fixed.

For human readable, what about

    SimpleDateFormat outputFmt = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss 'UTC'");

For ISO 8601, I think this would be correct:

    SimpleDateFormat outputFmt = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'");

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
```diff
@@ -66,97 +71,200 @@ public IndexPrinter(IndexInfoService indexInfoService, AsyncIndexInfoService asy
...
+        outputFmt.setTimeZone(TimeZone.getTimeZone("UTC"));
+        return outputFmt.format(date);
+    }
+
+    private static void keyValue(String key, Object value, PrintWriter pw, JsopBuilder json, Format format) {
+        // In case the key is null or an emppty String or value is null,
+        // throw an IllegalArgumentException.
+        if (value == null || key == null || key.equals("")) {
```
thomasmueller: I would support null values.

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
nit0906: @thomasmueller 
I actually added the check post your comment earlier on - 

> I would probably handle "null" separately at the beginning (if value == null). Could throw an exception.

I can remove it .

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
nit0906: Changed the format to look like this now 
`2021-09-21T07:16:53 UTC`

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
nit0906: done

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexPrinter.java
nit0906: `2021-09-21T07:16:53Z` Set format to 
`SimpleDateFormat outputFmt = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'");`

PR #376, opened 2021-09-27
OAK-9584 add failing test when accessing rep:authorizableId via expanded

kwin:
name

@anchela @mreutegg Any idea why this is failing for the namespace prefix "rep"?

### DIFF
--- a/oak-jcr/pom.xml
+++ b/oak-jcr/pom.xml
@@ -52,6 +52,7 @@
       org.apache.jackrabbit.test.api.WorkspaceCopyTest#testCopyNodesLocked                             <!-- OAK-118 -->
       org.apache.jackrabbit.test.api.WorkspaceMoveSameNameSibsTest#testMoveNodesOrderingSupportedByParent <!-- OAK-118 -->
       org.apache.jackrabbit.test.api.WorkspaceMoveTest#testMoveNodesLocked                             <!-- OAK-118 -->
+      org.apache.jackrabbit.oak.jcr.ValidNamesTest#testRepNamespaceUri                                 <!-- OAK-74 -->
 
       <!-- Locking : not fully implemented -->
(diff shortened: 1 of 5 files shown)


mreutegg:
Meanwhile discussion continued in OAK-9584. Issue is caused by invalid namespace URI as mentioned in OAK-74.

@kwin do you want to keep this PR open? You could add the test to the list of known issues in https://github.com/apache/jackrabbit-oak/blob/403801af2ccf727880019c18efefe95b83fdc3ae/oak-jcr/pom.xml#L38 and then merge.

kwin:
@mreutegg Where is the Maven property `known.issues` (passed as system property to maven-surefire-plugin and maven-failsafe-plugin) evaluated? Despite adding my test method to this property it is still evaluated: https://ci-builds.apache.org/blue/organizations/jenkins/Jackrabbit%2Foak-trunk-pr/detail/PR-376/4/pipeline

kwin:
It seems to work only for test cases extending [AbstractJCRTest, JUnit3](https://github.com/apache/jackrabbit/blob/ed3124e5fe223dada33ce6ddf53bc666063c3f2f/jackrabbit-jcr-tests/src/main/java/org/apache/jackrabbit/test/AbstractJCRTest.java#L476) (and therefore leveraging https://github.com/apache/jackrabbit/blob/ed3124e5fe223dada33ce6ddf53bc666063c3f2f/jackrabbit-jcr-tests/src/main/java/org/apache/jackrabbit/test/JCRTestResult.java#L103) but not for ones extending [AbstractRepositoryTest, JUnit4](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/AbstractRepositoryTest.java) as the latter does not implement this filtering logic. This would require a custom [TestRule](https://junit.org/junit4/javadoc/4.12/org/junit/rules/TestRule.html) (similar to http://www.javabyexamples.com/conditional-ignore-with-junit-test-rule)

PR #378, opened 2021-09-29
OAK-9587 - Add an attribute to enforce a strict index tag check

etugarev:
Jira: https://issues.apache.org/jira/browse/OAK-9587

Changes:
- for _lucene_ and _property_ indexes implemented a 'tag' policy for strict tag checking against the query, was hesitating whether boolean is enough, but decided to leave more room if we need to add more statuses in the future. _Solr_ indexes does not use tags so they were out of scope.
- removed unused imports here and there (courtesy of intelliJ, not mine)
- ensured that adding of **selectionPolicy** does not cause reindex.

These changes should be backward compatible as they do not change the existing behaviour.


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexConstants.java
```diff
@@ -114,6 +114,11 @@
      */
     String INDEX_TAGS = "tags";
 
+    /**
+     * The property to indicate tags matching policy. Currently, only 'TagsMatchingPolicy.STRICT' has effect
+     */
+    String TAGS_MATCHING_POLICY = "tagsMatchingPolicy";
```
tihom88: I think INDEX_TAGS_MATCHING_POLICY should be used.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/IndexDefinition.java
```diff
@@ -288,6 +288,8 @@
     @Nullable
     private final String[] indexTags;
 
+    private final String tagsMatchingPolicy;
```
tihom88: indexTagsMatchingPolicy

etugarev: fixed

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexConstants.java
etugarev: fixed

tihom88 (approved):

thomasmueller (changes requested):
This seems like a big change, and we should not rush it. It needs more time to review.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndex.java
```diff
@@ -183,6 +184,10 @@ private static boolean wrongIndex(ChildNodeEntry entry, Filter filter, NodeState
             }
             // no tag matches
             return true;
+        } else if (tags != null && IndexSelectionPolicy.TAG.equals(
```
thomasmueller: I would say "tags != null && " is not needed here: let's assume someone doesn't set any tag in the index definition, but uses selectionPolicy="tag". I would expect that the index is never used in this case. But in your logic, it might be used. We would also need a test case for this.

Moving String[] tags = getOptionalStrings(definition, IndexConstants.INDEX_TAGS); is then also not needed.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndexPlanner.java
```diff
@@ -426,6 +427,9 @@ private boolean wrongIndex() {
             }
             // no tag matches
             return true;
+        } else if (tags != null && IndexSelectionPolicy.TAG.equals(definition.getIndexSelectionPolicy())) {
```
thomasmueller: Same as in the other case.

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndex.java
etugarev: done

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndexPlanner.java
etugarev: done

thomasmueller (approved):

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/util/IndexDefinitionBuilder.java
```diff
@@ -624,6 +622,7 @@ private static Tree getOrCreateChild(Tree tree, String name){
                 IndexConstants.USE_IF_EXISTS,
                 IndexConstants.QUERY_PATHS,
                 IndexConstants.INDEX_TAGS,
+                IndexConstants.INDEX_SELECTION_POLICY,
```
averma21, line 625: This class is deprecated. You probably need to change org.apache.jackrabbit.oak.plugins.index.search.util.IndexDefinitionBuilder
That class is also extended by ElasticIndexDefinitionBuilder.

averma21: @tihom88 I think probably you did this refactoring. Any inputs?

etugarev: @averma21 this deprecated class is still used, so i changed it. I also changed `org.apache.jackrabbit.oak.plugins.index.search.util.IndexDefinitionBuilder` and added tests there.

thomasmueller: Hm, we need to move it to org.apache.jackrabbit.oak.plugins.index.search.util.IndexDefinitionBuilder... The Lucene class might still be there for compatibility, but we need to use the org.apache.jackrabbit.oak.plugins.index.search.util.IndexDefinitionBuilder. The clients that need selectionPolicy need to be migrated. We might want to deprecate the Lucene one.

etugarev: @thomasmueller @averma21 I agree, there are things to improve (like code reorganisation), but this was not in scope of this PR. Currently both builders support it, so this change should be transparent. Also, switching the clients to a `org.apache.jackrabbit.oak.plugins.index.search.util.IndexDefinitionBuilder` would require some additional work.

thomasmueller: what if we add the method to both classes? (the lucene one and the new one).

etugarev: We can, but I don't think we should invest more than needed into deprecated class. It will be eventually replaced with non deprecated version which has `selctionPolicy()`. I will leave it as is.

PR #381, opened 2021-10-04
OAK-9590 Reliable Event Processing

thomasmueller:
I think this should cover the gap we have in our documentation.


fabriziofortino (approved):

stefan-egli (approved):
lgtm. I think it's good to have an example of how it can be done reliably (there are of course other ways, but having this one documented is a good thing).

mreutegg:
This may not cover all use cases for reliable event processing. You may have a requirement to get events for all changes that happened on a node. With the documented approach it is possible to miss a change and only get an event for the most recent change on the node after a crash.

PR #383, opened 2021-10-05
OAK-9576: Multithreaded download synchronization issues

averma21:


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/DefaultMemoryManager.java
```diff
@@ -109,7 +110,8 @@ private void configureMemoryListener() {
         log.info("Setting up a listener to monitor pool '{}' and trigger batch save " +
                 "if memory drop below {} GB (max {})", pool.getName(), minMemoryBytes/ONE_GB, humanReadableByteCount(maxMemory));
         pool.setCollectionUsageThreshold(minMemoryBytes);
-        checkMemory(usage);
+        // todo - should we check and block in the beginning? This creates problem in case of download resume.
```
thomasmueller: Why does it create problems?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/DefaultMemoryManager.java
```diff
@@ -209,11 +211,15 @@ public void handleNotification(Notification notification,
                     .getType()
                     .equals(MemoryNotificationInfo.MEMORY_COLLECTION_THRESHOLD_EXCEEDED)) {
                 if (sufficientMemory.get()) {
-                    CompositeData cd = (CompositeData) notification
-                            .getUserData();
-                    MemoryNotificationInfo info = MemoryNotificationInfo
-                            .from(cd);
-                    checkMemory(info.getUsage());
+                    synchronized (sufficientMemory) {
+                        if (sufficientMemory.get()) {
```
thomasmueller: sufficientMemory is AtomicBoolean. I don't see why you would want to synchronize on it. Is the problem that you want to protect against concurrent calls on checkMemory? If yes, then checkMemory should be synchronized instead.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/StoreAndSortStrategy.java
```diff
@@ -62,8 +63,12 @@ public StoreAndSortStrategy(Iterable<NodeStateEntry> nodeStates, PathElementComp
 
     @Override
     public File createSortedStoreFile() throws IOException {
-        File storeFile = writeToStore(storeDir, getStoreFileName());
-        return sortStoreFile(storeFile);
+        try {
+            File storeFile = writeToStore(storeDir, getStoreFileName());
+            return sortStoreFile(storeFile);
+        } finally {
+            nodeStates.close();
```
thomasmueller: It sounds weird that createSortedStoreFile has a side effect of closing the nodeStates. Why?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTask.java
```diff
@@ -172,14 +173,24 @@ private boolean registerWithMemoryManager() {
...
             log.error(taskID + " could not complete download ", e);
         } finally {
             phaser.arriveAndDeregister();
+            log.info("{} entered finally block.", taskID);
+            if (dataDumpNotifyingPhaser != null) {
+                log.info("{} Data dump phaser not null after task completion. Notifying memory listener.", taskID);
+                dataDumpNotifyingPhaser.arriveAndDeregister();
+            }
+            if (MemoryManager.Type.JMX_BASED.equals(memoryManager.getType())) {
+                memoryManager.deregisterClient(registrationID);
```
thomasmueller, line 179: I wonder why do we even need a callback method from the memory manager... What about if each client (of the memory manager) just calls the memory manager when needed? That would reduce the need of synchronization.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTask.java
```diff
@@ -172,14 +173,24 @@ private boolean registerWithMemoryManager() {
...
+            }
+            try {
+                nodeStates.close();
```
thomasmueller, line 182: Here again we close the nodeStates.... why here? What about having a separate close() method in this class?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTask.java
```diff
@@ -239,7 +246,9 @@ private void addEntry(NodeStateEntry e) throws IOException {
         //Here logic differs from NodeStateEntrySorter in sense that
         //Holder line consist only of json and not 'path|json'
         NodeStateHolder h = new StateInBytesHolder(e.getPath(), jsonText);
-        entryBatch.add(h);
+        synchronized (this) {
```
thomasmueller: We now synchronise all access to entryBatch in the caller of entryBatch. What about moving all the synchronization to entryBatch itself? That would be easier to understand in my view. Or hopefully synchronization isn't needed at all (that would even be easier).

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTask.java
```diff
@@ -263,14 +274,18 @@ private synchronized void reset() {
...
+            entryBatch.sort(comparator);
         }
-        entryBatch.sort(comparator);
         Stopwatch w = Stopwatch.createStarted();
         File newtmpfile = File.createTempFile("sortInBatch", "flatfile", sortWorkDir);
         long textSize = 0;
         try (BufferedWriter writer = FlatFileStoreUtils.createWriter(newtmpfile, compressionEnabled)) {
+            // no concurrency issue with this traversal because addition to this list is only done in #addEntry which, for
+            // a given TraverseAndSortTask object will only be called from same thread
             for (NodeStateHolder h : entryBatch) {
```
thomasmueller: Ah, no, here we don't synchronize on entryBatch! That's inconsistent synchronization, and can cause big problems I think. (The way around it would be to clone entryBatch... but again, I would prefer if we don't need any synchronization).

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseWithSortStrategy.java
```diff
@@ -88,7 +89,7 @@
     private ArrayList<NodeStateHolder> entryBatch = new ArrayList<>();
```
thomasmueller, line 91: Java has a synchronized list...

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseWithSortStrategy.java
```diff
@@ -99,11 +100,15 @@
...
-        writeToSortedFiles();
-        return sortStoreFile();
+        try {
+            logFlags();
+            configureMemoryListener();
+            sortWorkDir = createdSortWorkDir(storeDir);
+            writeToSortedFiles();
+            return sortStoreFile();
+        } finally {
+            nodeStates.close();
```
thomasmueller: Here again, closing the nodeStates as a side effect of another method... I find it weird.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreTest.java
```diff
@@ -294,7 +295,7 @@ public NodeStateEntryTraverser create(LastModifiedRange range) {
                     null, range) {
                 @Override
                 public @NotNull Iterator<NodeStateEntry> iterator() {
-                    Map<String, Long> times = new HashMap<>();
+                    Map<String, Long> times = new LinkedHashMap<>(); // should be sorted in increasing order of value i.e. lastModificationTime
```
thomasmueller, line 297: LinkedHashMap doesn't sort in increasing order of the key. To sort, you would need to use a TreeMap.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreTest.java
```diff
@@ -348,11 +349,12 @@ int getTotalProvidedDocCount() {
...
-     * @return a map with keys denoting timestamp and values denoting paths which were created at those timestamps.
+     * @return a map with keys denoting timestamp and values denoting paths which were created at those timestamps. An
+     * iterator over the map entries would be in the increasing order of timestamps.
      */
-    private Map<Long, List<String>> createPathsWithTimestamps() {
-        Map<Long, List<String>> map = new HashMap<>();
-        for( int i = 1; i <= 10; i++) {
+    private LinkedHashMap<Long, List<String>> createPathsWithTimestamps() {
+        LinkedHashMap<Long, List<String>> map = new LinkedHashMap<>();
+        for( int i = 1; i <= 15; i++) {
```
thomasmueller, line 356: Why 15?

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/StoreAndSortStrategy.java
averma21: yes, this looks weird since this class is not the owner of nodeStates. This has been done to prevent OOM for multi threaded download. We need to keep on closing the nodeStates as the download tasks keep on finishing. This is mainly for TraverseAndSortTask (tasks created for parallel download) but due to code flow same pattern had to be followed here.
We could improve this but some more refactoring would be needed.

averma21: Oh I see you mean that particular method - createSortedStoreFile. 
I will see if we can add close method as you suggested.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTask.java
averma21: see https://github.com/apache/jackrabbit-oak/pull/383#discussion_r727798177

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreTest.java
averma21: here sorting is needed in increasing order of value.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/DefaultMemoryManager.java
averma21: As per the current code, we don't want to call checkMemory again, if sufficientMemory is already false, hence this approach.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTask.java
averma21: As discussed, this is needed because we want the clients to tell memory manager that they have finished dumping their data so that it could set the memory low flag to false, without depending on actual garbage collection run. For implementing this synchronization, we need to list of clients which would be iterated upon to inform about memory low condition and for providing the clients the phaser/latch on which synchronization would be done.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreTest.java
averma21: This is test class. I just increased the data in the test a bit.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/DefaultMemoryManager.java
averma21: I don't see a purpose of checking the memory during the beginning. There is nothing to do in this case. The only thing we could do is not to proceed further, but such kind of check could be added even before this stage. 

This one creates a problem during resume, in case memory is low due to previous run. It won't have any registered clients who still need to dump data (since previous run task objects would be unreachable) and checkMemory waits for clients to dump data. So the process blocks.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTask.java
averma21: The synchronization is not just for entryBatch, but it couples a few things along. For example, see `memoryLow` method which is also synchronized on `this` object and works with phaser. So using a concurrent data structure for entryBatch won't be enough.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -71,7 +71,7 @@
     protected List<NodeStateIndexerProvider> indexerProviders;
     protected final IndexerSupport indexerSupport;
     private final Set<String> indexerPaths = new HashSet<>();
-    private static final int MAX_DOWNLOAD_RETRIES = Integer.parseInt(System.getProperty("oak.indexer.maxDownloadRetries", "3"));
+    private static final int MAX_DOWNLOAD_ATTEMPTS = Integer.parseInt(System.getProperty("oak.indexer.maxDownloadRetries", "5")) + 1;
 
     public DocumentStoreIndexerBase(IndexHelper indexHelper, IndexerSupport indexerSupport) throws IOException {
```
fabriziofortino: this should not throw IOE

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -167,10 +162,20 @@ private FlatFileStore buildFlatFileStore(NodeState checkpointedState, CompositeI
...
                 previousDownloadDirs.add(builder.getFlatFileStoreDir());
             }
+            if (executionCount < MAX_DOWNLOAD_ATTEMPTS) {
+                try {
+                    log.info("Waiting for {} millis before retrying", backOffTimeInMillis);
+                    Thread.sleep(backOffTimeInMillis);
+                    backOffTimeInMillis *= 2;
+                } catch (InterruptedException e) {
+                    log.error("Interrupted while waiting before retrying download ", e);
+                    e.printStackTrace();
```
fabriziofortino: this should not be needed since the instruction above should print the stack trace

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -54,6 +54,17 @@
      */
```
fabriziofortino, line 53: remove import of `LastModifiedRange`

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -52,6 +52,8 @@
 import java.util.stream.Stream;
 
 import static com.google.common.base.Charsets.UTF_8;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileNodeStoreBuilder.DEFAULT_NUMBER_OF_DATA_DUMP_THREADS;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileNodeStoreBuilder.PROP_THREAD_POOL_SIZE;
 import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileStoreUtils.createWriter;
 import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileStoreUtils.getSortedStoreFileName;
```
fabriziofortino, line 59: The link in the javadoc below should be `FlatFileNodeStoreBuilder#PROP_THREAD_POOL_SIZE`

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTask.java
```diff
@@ -71,10 +73,9 @@
     private long memoryUsed;
     private final File sortWorkDir;
     private final List<File> sortedFiles = new ArrayList<>();
-    private final ArrayList<NodeStateHolder> entryBatch = new ArrayList<>();
+    private final LinkedList<NodeStateHolder> entryBatch = new LinkedList<>();
     private NodeStateEntry lastSavedNodeStateEntry;
     private final String taskID;
-    private final AtomicBoolean dumpData = new AtomicBoolean(false);
```
fabriziofortino, line 77: remove import of `AtomicBoolean`

fabriziofortino (changes requested):

fabriziofortino (approved):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/DefaultMemoryManager.java
thomasmueller: Ok, in this case could you remove the TODO and write e.g. "We don't check memory here, as there is no good way here to free up memory in case it is low."?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/DefaultMemoryManager.java
thomasmueller: Hm this not clear... I don't currently see an answer to the question "why you would want to synchronize on it"

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTask.java
```diff
@@ -172,22 +171,28 @@ private boolean registerWithMemoryManager() {
...
+                log.info("{} Data dump phaser not null after task completion. Notifying memory listener.", taskID);
+                dataDumpNotifyingPhaser.arriveAndDeregister();
+            }
+            if (MemoryManager.Type.JMX_BASED.equals(memoryManager.getType())) {
+                memoryManager.deregisterClient(registrationID);
+            }
+            try {
+                nodeStates.close();
+            } catch (IOException e) {
+                log.error("{} could not close NodeStateEntryTraverser", taskID);
```
thomasmueller: Could you log the stack trace as well? Just add ", e" to the log.error call.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTask.java
thomasmueller: I don't understand the exact use for synchronization... which data structure needs to be protected from concurrent operations?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTask.java
```diff
@@ -263,15 +269,21 @@ private synchronized void reset() {
...
+            if (entryBatch.isEmpty()) {
+                return;
+            }
+            entryBatch.sort(comparator);
         }
-        entryBatch.sort(comparator);
         Stopwatch w = Stopwatch.createStarted();
         File newtmpfile = File.createTempFile("sortInBatch", "flatfile", sortWorkDir);
         long textSize = 0;
+        long size = entryBatch.size();
```
thomasmueller, line 272: Here we access entryBatch without synchronized - again, I feel it's better to use "save" data structures (e.g. in java.util.concurrent), instead of synchronizing ourselves. If we need to synchronize ourselves, then we need to do it consistently, and write sufficient test cases that verify all possible concurrency problems (which is very hard).

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTask.java
```diff
@@ -263,15 +269,21 @@ private synchronized void reset() {
...
         try (BufferedWriter writer = FlatFileStoreUtils.createWriter(newtmpfile, compressionEnabled)) {
-            for (NodeStateHolder h : entryBatch) {
+            // no concurrency issue with this traversal because addition to this list is only done in #addEntry which, for
+            // a given TraverseAndSortTask object will only be called from same thread
+            while (!entryBatch.isEmpty()) {
+                NodeStateHolder h = entryBatch.removeFirst();
```
thomasmueller, line 277: This is possibly another case of unsynchronized access to entryBatch

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTask.java
averma21: Removed synchronization.

thomasmueller (approved):

PR #385, opened 2021-10-05
OAK-9592 - Log exceptions when retrying find

ankitaagar:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -550,6 +550,8 @@ public NodeDocument call() throws Exception {
                 return findUncached(collection, key, docReadPref);
             } catch (MongoException e) {
                 ex = e;
+                LOG.error("Read Fails with an exception" + e);
+                e.printStackTrace();
```
```suggestion
                LOG.error("findUncachedWithRetry : read fails with an exception" + e, e);
```

stefan-egli: would adding the exception stacktrace to the log make sense or is that too verbose? (e.printStackTrace() otherwise just prints to standard out)

reschke:
~~~
LOG.error("findUncachedWithRetry: read fails with an exception: " + e, e);
~~~

stefan-egli (approved):

PR #390, opened 2021-10-18
Script to check for inconsistent version nodes and generate a report

Joscorbe:


stefan-egli (changes requested):
As discussed I agree that this might not necessarily have to be in oak-run as it was likely caused by a non-open source tool. But then again, it might also be useful for others as blueprints/reuse of similar checks etc. So from that point of view I'm fine with having it in oak-run.

Few points to consider addressing:
* is there a corresponding oak jira ticket? if not, might make sense to have it for tracking purpose.
* the class does not have a javadoc with some high-level explanation as to what the tool does
* the ProgressThread has an empty catch block which swallows an InterruptedException. Nothing dramatic, as it's only the status background thread plus it would not normally happen - but might still be useful to have at least a printStackTrace there.
* Was wondering if the admin username should be hardcoded or not made an argument as well - in theory I guess another username might also have admin privileges and execute this traversal.

Additional points more as remarks:
* the tool does not integrate into the oak-run Command framework as other tools herein do. But since this is not exactly meant to be a generic, frequently used tool but rather something very specific, temporary, it might make sense to keep it outside of that framework to not set wrong expectations.
* was wondering about the use of static volatile variables and alternatives. But then again it works fine and is appropriate for this case I think

Joscorbe:
Link to the OAK issue I have created: https://issues.apache.org/jira/browse/OAK-9602

stefan-egli (approved):
+1, lgtm. Should I go ahead and merge or do you have some changes pending?

Joscorbe:
@stefan-egli Yes, it should be ready, sorry I didn't notice your message until now.
Thanks for reviewing!

PR #393, opened 2021-10-19
OAK-9598: reduce server load on reindex operations

fabriziofortino:
This PR changes the ES indexes lifecycle. At reindexing time, a new index gets created and populated while the old one is still used for queries. The new index is optimized for bulk loads (to reduce server load). Once the reindex process ends, the new index gets enabled and the old one is deleted.


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
```diff
@@ -223,19 +223,18 @@ private boolean shouldReindex(NodeBuilder definition, NodeState before,
...
 
+        //Do not attempt reindex of disabled indexes
+        if (type != null && TYPE_DISABLED.equals(type.getValue(Type.STRING))) {
+            return false;
+        }
+
         //Async indexes are not considered for reindexing for sync indexing
         // Skip this check for elastic index
         // TODO : See if the check to skip elastic can be handled in a better way - maybe move isMatchingIndexNode to IndexDefinition ?
         if (!TYPE_ELASTICSEARCH.equals(type.getValue(Type.STRING)) && !isMatchingIndexMode(definition)){
```
thomasmueller: Could we get a NPE here if type is null? (I understand this code was there before... but now it's more obvious... Also, missing space before "{"

nit0906 (commented):
I was wondering if it might be possible to add a test that verifies the changes in this PR, the current test verifies if reindexing works fine, but we are not testing which remote index is actually returning the results. Would it be possible to check that the old index is still used for queries during reindex process, and then once reindex is completed, the new index is used and the old one has been deleted.

fabriziofortino:
@nit0906 I wanted to add a test but unfortunately reindex in tests seems to work only with synchronous indexes. A test like that would require async indexing. Even with that, there could be race conditions where reindexing finishes faster than expected. To work this around we can introduce a [toxyproxy](https://www.testcontainers.org/modules/toxiproxy/) between the test and the ES container to add latency while reindexing.

I have created a separate issue to track this task.

nit0906 (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
```diff
@@ -242,29 +240,27 @@ private boolean shouldReindex(NodeBuilder definition, NodeState before,
...
-        // in case of OutOfBand indexing during content import, there is no hidden node created for elastic (not even :status)
-        // So, we log a warn and return false to avoid unnecessary reindexing. The warning is only if the someone added the new index node and forgot to add
+        // In case of elasticsearch, indexed data is stored remotely and not under hidden nodes, so in case of OutOfBand
+        // indexing during content import, there is no hidden node created for elastic (not even :status)
+        // So, we log a warning and return false to avoid unnecessary reindexing. The warning is  displayed only if
+        // someone added the new index node and forgot to add
         // the reindex flag, in case OutOfBand Indexing has been performed, warning can be ignored.
         // Also, in case the new elastic node has been added with reindex = true , this method would have already returned true
         if (result && TYPE_ELASTICSEARCH.equals((type.getValue(Type.STRING)))) {
             log.warn("Found a new elastic index node [{}]. Please set the reindex flag = true to initiate reindexing." +
```
thomasmueller, line 259: Nit: missing space after the dot. This was there before.

thomasmueller (commented):
Great!

PR #399, opened 2021-10-25
OAK-9603: fix index name management in oak-search-elastic

fabriziofortino:


oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexNameHelper.java
```diff
@@ -98,8 +75,8 @@ public static String getRemoteIndexName(ElasticIndexDefinition indexDefinition)
      * <p>
      * The resulting file name would be truncated to MAX_NAME_LENGTH
      */
-    private static String getElasticSafeIndexName(String indexPath) {
-        String name = StreamSupport
+    public static String getElasticSafeIndexName(String indexPrefix, String indexPath) {
+        String name = indexPrefix + "." + StreamSupport
```
averma21, line 87: Since this method is supposed to return a safe name, i think we should check indexPrefix for safety here and thrown an exception if it's not safe as done in ElasticConnection class.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexWriter.java
```diff
@@ -102,11 +101,12 @@ public boolean close(long timestamp) throws IOException {
         return bulkProcessorHandler.close();
     }
 
-    protected void provisionIndex(long seed) throws IOException {
+    protected void provisionIndex() throws IOException {
         final IndicesClient indicesClient = elasticConnection.getClient().indices();
         // check if index already exists
-        final String indexName = ElasticIndexNameHelper.getRemoteIndexName(indexDefinition, seed);
-        boolean exists = indicesClient.exists(new GetIndexRequest(indexName), RequestOptions.DEFAULT);
+        boolean exists = elasticConnection.getClient().indices().exists(
```
averma21, line 135: Will we now be creating index using alias? (the class level indexName has been obtained from getIndexAlias())

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexNameHelper.java
fabriziofortino: The check on the `indexPrefix` is performed in the `ElasticConnection` init phase (called once when the `ElasticProviderService` gets initialized). I did not add the check here because this method is called frequently.
I could (1) add the check anyway (2) add a comment to explain why the prefix does not need to be checked. I prefer option 2. What do you think?

averma21: In that case, option 2 is fine and I think instead of comment, adding a javadoc which clarifies the expectations would be better.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexWriter.java
fabriziofortino: The `indexName` depends on the `reindex` flag: it's the full name on reindex operations or the alias during incremental indexing. I can see this is not clear. The problem lays in the classes' design: the index provisioning logic is spread across 3 different classes (`ElasticIndexWriter`, `ElasticIndexWriterFactory`, `ElasticIndexEditorContext`) making it quite ununderstandable. I have taken the opportunity to refactor it with the intent of making it more understandable: all these parts (create seed, provision index, enable index) have been moved to `ElasticIndexWriter`. Please let me know if it makes more sense now.

averma21: yes it was complicated and this is much better now :)

averma21 (approved):

PR #405, opened 2021-11-01
OAK-9609: Override persistentCacheIncludes with framework property

mreutegg:


stefan-egli (approved):
+1
PS: would it make sense to add a note to the Configuration.persistentCacheIncludes description about the "::" syntax?

mreutegg:
Yes, good point about documenting this.

PR #414, opened 2021-11-11
OAK-9621: add tests for mandatory child nodes (both residual and non-residual)

kwin:


kwin:
I would like to add the tests (as I don't think this is covered in the TCK). The behaviour is fine though and fully compliant with JCR 2.0 spec.

reschke:
I'm ok with adding it here, but it might be better to have a TCK test for this, so that Jackrabbit classic is covered as well. (Or alternatively add this to the jackrabbit-core tests).

reschke (changes requested):
It's good to add test coverage. I'm however sceptical that Oak's behavior is actually correct, and would recommend that we have a discussion about this on the mailing list (will follow up later).

PR #419, opened 2021-11-17
OAK-9624 add query caller logging

joerghoh:
Add fucntionality which is able to log the class which is issuing a JCR query. The detection is based on an analysis of the callstack with the ability to ignore a defined set of packages/names.

* the configuration of an ignoreList is done via the OSGI configuration (QueryEngineSettingsService)
* if the ignoreList is empty (non-configured), nothing is done, and there is no overhead.
* If configured it displays the caller based on the callstack, ignoring stackframes which start with the classnames on the ignoreList.
* If a more verbose loglevel is used, also full stacktraces are logged (helps to refine the configuration)
* Dedicated loggers are created for internal queries (which either are completely matched by elements of the ignoreList or which have the ``/* oak-internal */`` as annotation) or external queries (everything else).
* For external queries a single log line is logged on INFO (stacktrace on DEBUG), for internal queries the single log line is logged on DEBUG (stacktrace on TRACE).

There might be a small overhead in execution time (getting the stack isn't that cheap), but for the frequent and time-critical queries (the ones marked with ``/* oak-internal */`` this isn't done.

When these elements are put onto the ignoreList:
```
org.apache.jackrabbit.oak
java.lang
org.apache.sling.jcr.resource.internal.helper.JcrResourceUtil.query
org.apache.sling.jcr.resource.internal.helper.jcr.BasicQueryLanguageProvider
org.apache.sling.resourceresolver.impl.providers.stateful.AuthenticatedResourceProvider.findResources
org.apache.sling.resourceresolver.impl.helper.ResourceResolverControl.findResources
org.apache.sling.resourceresolver.impl.ResourceResolverImpl.findResources
```

The correct calling classes of JCR Queries and Sling's ``ResourceResolver.findResource`` are logged.

### DIFF
--- a/oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
+++ b/oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
@@ -139,5 +139,14 @@ public interface QueryEngineSettingsMBean {
 
     @Description("Get the query validator data as a JSON string.")
     String getQueryValidatorJson();
-
+    
+    
+//    @Description("Set or remove Java Package Name to ignore in Call Trace analysis")
+    void setIgnoredPackageNamesinCallTrace(
+            @Description("package names")
+            @Name("package names")
+            String[] packageNames);
+    
+//    @Description("Get the Java package Names to ignore in Call trace analysis")
+    String[] getIgnoredPackageNamesinCallTrace();
 }
--- a/oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
+++ b/oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
@@ -141,16 +141,14 @@ public interface QueryEngineSettingsMBean {
     @Description("Get the query validator data as a JSON string.")
     String getQueryValidatorJson();
     
-    
     /**
      * Set or remove java package/class names which are ignored from finding the 
      * invoking class for queries.
      * 
-     * it can be either Java package names or fully-qualified class names (package + class name).
+     * It can be either Java package names or fully-qualified class names (package + class name).
      * 
      * @param classNames the class names to be ignored.
      */
-    
     @Description("Set or remove Java Package Name to ignore in Call Trace analysis")
     void setIgnoredClassNamesinCallTrace(
             @Description("package or fully qualified class names")
--- a/oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
+++ b/oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
@@ -16,6 +16,7 @@
  */
 package org.apache.jackrabbit.oak.api.jmx;
 
+import org.jetbrains.annotations.NotNull;
 import org.osgi.annotation.versioning.ProviderType;
 
 @ProviderType
@@ -141,12 +142,21 @@ public interface QueryEngineSettingsMBean {
     String getQueryValidatorJson();
     
     
-//    @Description("Set or remove Java Package Name to ignore in Call Trace analysis")
-    void setIgnoredClassNamesinCallTrace(
-            @Description("package names")
-            @Name("package names")
-            String[] packageNames);
+    /**
+     * Set or remove java package/class names which are ignored from finding the 
+     * invoking class for queries.
+     * 
+     * it can be either Java package names or fully-qualified class names (package + class name).
+     * 
+     * @param classNames the class names to be ignored.
+     */
     
-//    @Description("Get the Java package Names to ignore in Call trace analysis")
-    String[] getIgnoredClassNamesinCallTrace();
+    @Description("Set or remove Java Package Name to ignore in Call Trace analysis")
+    void setIgnoredClassNamesinCallTrace(
+            @Description("package or fully qualified class names")
+            @Name("class names")
+            @NotNull String[] classNames);
+    
+    @Description("Get the Java package / fully qualified class names to ignore when finding the caller of query")
+    @NotNull String[] getIgnoredClassNamesinCallTrace();
 }
--- a/oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
+++ b/oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
@@ -142,11 +142,11 @@ public interface QueryEngineSettingsMBean {
     
     
 //    @Description("Set or remove Java Package Name to ignore in Call Trace analysis")
-    void setIgnoredPackageNamesinCallTrace(
+    void setIgnoredClassNamesinCallTrace(
             @Description("package names")
             @Name("package names")
             String[] packageNames);
     
 //    @Description("Get the Java package Names to ignore in Call trace analysis")
-    String[] getIgnoredPackageNamesinCallTrace();
+    String[] getIgnoredClassNamesinCallTrace();
 }
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
@@ -60,6 +60,8 @@ import org.apache.jackrabbit.oak.api.ContentRepository;
 import org.apache.jackrabbit.oak.api.ContentSession;
 import org.apache.jackrabbit.oak.api.Descriptors;
 import org.apache.jackrabbit.oak.api.Root;
+import org.apache.jackrabbit.oak.api.jmx.Description;
+import org.apache.jackrabbit.oak.api.jmx.Name;
 import org.apache.jackrabbit.oak.api.jmx.QueryEngineSettingsMBean;
 import org.apache.jackrabbit.oak.api.jmx.RepositoryManagementMBean;
 import org.apache.jackrabbit.oak.commons.IOUtils;
@@ -974,6 +976,15 @@ public class Oak {
         void setFullTextComparisonWithoutIndex(boolean fullTextComparisonWithoutIndex) {
             this.settings.setFullTextComparisonWithoutIndex(fullTextComparisonWithoutIndex);
         }
+        
+        public void setIgnoredPackageNamesinCallTrace(String[] packageNames) {
+            settings.setIgnoredPackageNamesinCallTrace(packageNames);
+        }
+        
+        public String[] getIgnoredPackageNamesinCallTrace() {
+            return settings.getIgnoredPackageNamesinCallTrace();
+        }
+        
     }
 
     public static class OakDefaultComponents {
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
@@ -60,8 +60,6 @@ import org.apache.jackrabbit.oak.api.ContentRepository;
 import org.apache.jackrabbit.oak.api.ContentSession;
 import org.apache.jackrabbit.oak.api.Descriptors;
 import org.apache.jackrabbit.oak.api.Root;
-import org.apache.jackrabbit.oak.api.jmx.Description;
-import org.apache.jackrabbit.oak.api.jmx.Name;
 import org.apache.jackrabbit.oak.api.jmx.QueryEngineSettingsMBean;
 import org.apache.jackrabbit.oak.api.jmx.RepositoryManagementMBean;
 import org.apache.jackrabbit.oak.commons.IOUtils;
@@ -977,11 +975,11 @@ public class Oak {
             this.settings.setFullTextComparisonWithoutIndex(fullTextComparisonWithoutIndex);
         }
         
-        public void setIgnoredClassNamesinCallTrace(String[] packageNames) {
+        public void setIgnoredClassNamesinCallTrace(@NotNull String[] packageNames) {
             settings.setIgnoredClassNamesinCallTrace(packageNames);
         }
         
-        public String[] getIgnoredClassNamesinCallTrace() {
+        public @NotNull String[] getIgnoredClassNamesinCallTrace() {
             return settings.getIgnoredClassNamesinCallTrace();
         }
         
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
@@ -977,12 +977,12 @@ public class Oak {
             this.settings.setFullTextComparisonWithoutIndex(fullTextComparisonWithoutIndex);
         }
         
-        public void setIgnoredPackageNamesinCallTrace(String[] packageNames) {
-            settings.setIgnoredPackageNamesinCallTrace(packageNames);
+        public void setIgnoredClassNamesinCallTrace(String[] packageNames) {
+            settings.setIgnoredClassNamesinCallTrace(packageNames);
         }
         
-        public String[] getIgnoredPackageNamesinCallTrace() {
-            return settings.getIgnoredPackageNamesinCallTrace();
+        public String[] getIgnoredClassNamesinCallTrace() {
+            return settings.getIgnoredClassNamesinCallTrace();
         }
         
     }
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
@@ -27,6 +27,7 @@ import org.apache.jackrabbit.oak.query.stats.QueryStatsMBeanImpl;
 import org.apache.jackrabbit.oak.query.stats.QueryStatsReporter;
 import org.apache.jackrabbit.oak.spi.query.QueryLimits;
 import org.apache.jackrabbit.oak.stats.StatisticsProvider;
+import org.jetbrains.annotations.NotNull;
 
 /**
  * Settings of the query engine.
@@ -187,11 +188,11 @@ public class QueryEngineSettings implements QueryEngineSettingsMBean, QueryLimit
         return queryValidator;
     }
     
-    public void setIgnoredClassNamesinCallTrace(String[] packageNames) {
+    public void setIgnoredClassNamesinCallTrace(@NotNull String[] packageNames) {
         classNamesIgnoredInCallTrace = packageNames;
     }
     
-    public String[] getIgnoredClassNamesinCallTrace() {
+    public @NotNull String[] getIgnoredClassNamesinCallTrace() {
         return classNamesIgnoredInCallTrace;
     }
 
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
@@ -18,6 +18,8 @@
  */
 package org.apache.jackrabbit.oak.query;
 
+import java.util.Arrays;
+
 import org.apache.jackrabbit.oak.api.StrictPathRestriction;
 import org.apache.jackrabbit.oak.api.jmx.QueryEngineSettingsMBean;
 import org.apache.jackrabbit.oak.query.stats.QueryStatsMBean;
@@ -81,7 +83,7 @@ public class QueryEngineSettings implements QueryEngineSettingsMBean, QueryLimit
 
     private final QueryStatsMBeanImpl queryStats = new QueryStatsMBeanImpl(this);
     
-    private String[] javaPackagesIgnoredInCallTrace = new String[0];
+    private String[] classNamesIgnoredInCallTrace = new String[0];
 
     /**
      * StatisticsProvider used to record query side metrics.
@@ -185,12 +187,12 @@ public class QueryEngineSettings implements QueryEngineSettingsMBean, QueryLimit
         return queryValidator;
     }
     
-    public void setIgnoredPackageNamesinCallTrace(String[] packageNames) {
-        javaPackagesIgnoredInCallTrace = packageNames;
+    public void setIgnoredClassNamesinCallTrace(String[] packageNames) {
+        classNamesIgnoredInCallTrace = packageNames;
     }
     
-    public String[] getIgnoredPackageNamesinCallTrace() {
-        return javaPackagesIgnoredInCallTrace;
+    public String[] getIgnoredClassNamesinCallTrace() {
+        return classNamesIgnoredInCallTrace;
     }
 
     @Override
@@ -202,6 +204,7 @@ public class QueryEngineSettings implements QueryEngineSettingsMBean, QueryLimit
                 ", fullTextComparisonWithoutIndex=" + fullTextComparisonWithoutIndex +
                 ", sql2Optimisation=" + sql2Optimisation +
                 ", fastQuerySize=" + fastQuerySize +
+                ", classNamesIgnoredInCallTrace=" + Arrays.toString(classNamesIgnoredInCallTrace) +
                 '}';
     }
     
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
@@ -80,6 +80,8 @@ public class QueryEngineSettings implements QueryEngineSettingsMBean, QueryLimit
     private StrictPathRestriction strictPathRestriction = StrictPathRestriction.DISABLE;
 
     private final QueryStatsMBeanImpl queryStats = new QueryStatsMBeanImpl(this);
+    
+    private String[] javaPackagesIgnoredInCallTrace = new String[0];
 
     /**
      * StatisticsProvider used to record query side metrics.
@@ -181,6 +183,14 @@ public class QueryEngineSettings implements QueryEngineSettingsMBean, QueryLimit
 
     public QueryValidator getQueryValidator() {
         return queryValidator;
+    }
+    
+    public void setIgnoredPackageNamesinCallTrace(String[] packageNames) {
+        javaPackagesIgnoredInCallTrace = packageNames;
+    }
+    
+    public String[] getIgnoredPackageNamesinCallTrace() {
+        return javaPackagesIgnoredInCallTrace;
     }
 
     @Override
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettingsService.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettingsService.java
@@ -84,7 +84,7 @@ public class QueryEngineSettingsService {
                         + "which executed this query. This java package is the first package in the call trace"
                         + "which does not not start with the any of the provided fully qualfied class names (packagename + classname)"
                 )
-        String[] ignoredClassNamesinCallTrace();
+        String[] ignoredClassNamesinCallTrace() default {};
 
     }
 
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettingsService.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettingsService.java
@@ -79,12 +79,12 @@ public class QueryEngineSettingsService {
         String getStrictPathRestrictionsForIndexes() default DISABLED_STRICT_PATH_RESTRICTION;
         
         @AttributeDefinition(
-                name="Java package names to ignore when finding caller",
+                name="Fully qualified class names to ignore when finding caller",
                 description="If non-empty the query engine logs the query statement plus the java package "
                         + "which executed this query. This java package is the first package in the call trace"
-                        + "which does not not start with the any of the provided package names"
+                        + "which does not not start with the any of the provided fully qualfied class names (packagename + classname)"
                 )
-        String[] ignoredPackageNamesinCallTrace();
+        String[] ignoredClassNamesinCallTrace();
 
     }
 
@@ -130,7 +130,7 @@ public class QueryEngineSettingsService {
             logMsg(QUERY_FAIL_TRAVERSAL, QueryEngineSettings.OAK_QUERY_FAIL_TRAVERSAL);
         }
 
-        queryEngineSettings.setIgnoredPackageNamesinCallTrace(config.ignoredPackageNamesinCallTrace());
+        queryEngineSettings.setIgnoredClassNamesinCallTrace(config.ignoredClassNamesinCallTrace());
 
         boolean fastQuerySizeSysProp = QueryEngineSettings.DEFAULT_FAST_QUERY_SIZE;
         boolean fastQuerySizeFromConfig = config.fastQuerySize();
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettingsService.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettingsService.java
@@ -77,6 +77,14 @@ public class QueryEngineSettingsService {
                         "the queryPaths of the index is taken into account."
         )
         String getStrictPathRestrictionsForIndexes() default DISABLED_STRICT_PATH_RESTRICTION;
+        
+        @AttributeDefinition(
+                name="Java package names to ignore when finding caller",
+                description="If non-empty the query engine logs the query statement plus the java package "
+                        + "which executed this query. This java package is the first package in the call trace"
+                        + "which does not not start with the any of the provided package names"
+                )
+        String[] ignoredPackageNamesinCallTrace();
 
     }
 
@@ -121,6 +129,8 @@ public class QueryEngineSettingsService {
         } else {
             logMsg(QUERY_FAIL_TRAVERSAL, QueryEngineSettings.OAK_QUERY_FAIL_TRAVERSAL);
         }
+
+        queryEngineSettings.setIgnoredPackageNamesinCallTrace(config.ignoredPackageNamesinCallTrace());
 
         boolean fastQuerySizeSysProp = QueryEngineSettings.DEFAULT_FAST_QUERY_SIZE;
         boolean fastQuerySizeFromConfig = config.fastQuerySize();
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
@@ -37,7 +37,6 @@ public class QueryInformationLogger {
      * @param ignoredClassNames entries to be ignored. If empty, the method is a no-op.
      */
     public static void logCaller(@NotNull String statement, @NotNull String[] ignoredClassNames) {
-        
         if (ignoredClassNames.length == 0) {
             return;
         }
@@ -80,8 +79,8 @@ public class QueryInformationLogger {
      * @return the calling class; if all classes of the call trace are in packages which are ignored,
      *   an empty string is returned.
      */
-    private static String getInvokingClass(String[] ignoredJavaPackages) {
-        
+    @NotNull
+    private static  String getInvokingClass(@NotNull String[] ignoredJavaPackages) {
         // With java9 we would use https://docs.oracle.com/javase/9/docs/api/java/lang/StackWalker.html
         final StackTraceElement[] callStack = Thread.currentThread().getStackTrace();
         
@@ -101,7 +100,6 @@ public class QueryInformationLogger {
         return ""; 
     }
     
-    
     /**
      * Check if statement is explicitly marked to be internal
      * @param statement the JCR query statement to check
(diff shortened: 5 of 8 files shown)


oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
```diff
@@ -139,5 +139,14 @@ void setQueryValidatorPattern(
 
     @Description("Get the query validator data as a JSON string.")
     String getQueryValidatorJson();
-
+    
+    
+//    @Description("Set or remove Java Package Name to ignore in Call Trace analysis")
```
anchela: why do you comment this?
if i look at other methods, the '@Description' tag is above the method with some human readable information.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
```diff
@@ -139,5 +139,14 @@ void setQueryValidatorPattern(
...
+    void setIgnoredClassNamesinCallTrace(
+            @Description("package names")
```
anchela: see above.... maybe you have to remove it here....?

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
```diff
@@ -139,5 +139,14 @@ void setQueryValidatorPattern(
...
+            @Name("package names")
+            String[] packageNames);
+    
+//    @Description("Get the Java package Names to ignore in Call trace analysis")
```
anchela: same as on line 144

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
```diff
@@ -60,6 +60,8 @@
 import org.apache.jackrabbit.oak.api.ContentSession;
 import org.apache.jackrabbit.oak.api.Descriptors;
 import org.apache.jackrabbit.oak.api.Root;
+import org.apache.jackrabbit.oak.api.jmx.Description;
+import org.apache.jackrabbit.oak.api.jmx.Name;
```
anchela: why are these 2 imports needed? i didn't look at it in an IDE.... so i might be mistaken..... but i don't see these 2 imports used in the 2 methods added below.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
```diff
@@ -139,5 +139,14 @@ void setQueryValidatorPattern(
...
     @Description("Get the query validator data as a JSON string.")
     String getQueryValidatorJson();
-
+    
+    
+//    @Description("Set or remove Java Package Name to ignore in Call Trace analysis")
+    void setIgnoredClassNamesinCallTrace(
+            @Description("package names")
+            @Name("package names")
+            String[] packageNames);
```
anchela: i would add @notnull annotation

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
```diff
@@ -139,5 +139,14 @@ void setQueryValidatorPattern(
...
+    
+//    @Description("Get the Java package Names to ignore in Call trace analysis")
+    String[] getIgnoredClassNamesinCallTrace();
```
anchela: i would add @Notnull annotation for the return value. that helps simplifying the code and makes the contract clear. consumers with the right IDE settings will be notified if they add redundant checks for null.

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
```diff
@@ -974,6 +976,15 @@ public String toString() {
         void setFullTextComparisonWithoutIndex(boolean fullTextComparisonWithoutIndex) {
             this.settings.setFullTextComparisonWithoutIndex(fullTextComparisonWithoutIndex);
         }
+        
+        public void setIgnoredClassNamesinCallTrace(String[] packageNames) {
```
anchela: also here: @notnull annotation to the param

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
```diff
@@ -974,6 +976,15 @@ public String toString() {
...
+            settings.setIgnoredClassNamesinCallTrace(packageNames);
+        }
+        
+        public String[] getIgnoredClassNamesinCallTrace() {
```
anchela: add @notnull annotation for the return value

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
```diff
@@ -182,6 +186,14 @@ public String getQueryValidatorJson() {
     public QueryValidator getQueryValidator() {
         return queryValidator;
     }
+    
+    public void setIgnoredClassNamesinCallTrace(String[] packageNames) {
```
anchela: add @notnull annotations to the param

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
```diff
@@ -80,6 +82,8 @@
     private StrictPathRestriction strictPathRestriction = StrictPathRestriction.DISABLE;
 
     private final QueryStatsMBeanImpl queryStats = new QueryStatsMBeanImpl(this);
+    
+    private String[] classNamesIgnoredInCallTrace = new String[0];
```
anchela, line 87: sounds like a reasonable default..... and when reading it i noticed about the missing @nonull annotations.... because the default might get replaced by 'null' upon set if you don't specify it to be not-null.....

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
```diff
@@ -182,6 +186,14 @@ public String getQueryValidatorJson() {
     public QueryValidator getQueryValidator() {
         return queryValidator;
     }
+    
+    public void setIgnoredClassNamesinCallTrace(String[] packageNames) {
+        classNamesIgnoredInCallTrace = packageNames;
```
anchela, line 192: with the annotation you can avoid having to check for null param here.... which would overwrite the default empty-string value..... i like the code in this form and i would avoid redundant checks for null by adding the annotation.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
```diff
@@ -182,6 +186,14 @@ public String getQueryValidatorJson() {
...
+    }
+    
+    public String[] getIgnoredClassNamesinCallTrace() {
```
anchela: @notnull return value annotation

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettingsService.java
```diff
@@ -77,6 +77,14 @@
...
         )
         String getStrictPathRestrictionsForIndexes() default DISABLED_STRICT_PATH_RESTRICTION;
+        
+        @AttributeDefinition(
+                name="Fully qualified class names to ignore when finding caller",
+                description="If non-empty the query engine logs the query statement plus the java package "
+                        + "which executed this query. This java package is the first package in the call trace"
+                        + "which does not not start with the any of the provided fully qualfied class names (packagename + classname)"
+                )
+        String[] ignoredClassNamesinCallTrace();
```
anchela: @notnull annotation for the return value :)

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettingsService.java
```diff
@@ -77,6 +77,14 @@
                         "the queryPaths of the index is taken into account."
         )
         String getStrictPathRestrictionsForIndexes() default DISABLED_STRICT_PATH_RESTRICTION;
+        
+        @AttributeDefinition(
+                name="Fully qualified class names to ignore when finding caller",
+                description="If non-empty the query engine logs the query statement plus the java package "
```
anchela, line 83: if you prefer not to add the notnull annotation you should explain what a null return value means..... but i would suggest to avoid return null....

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
```diff
@@ -0,0 +1,110 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class QueryInformationLogger {
+    
+    private static final Logger INTERNAL_QUERIES_LOG = LoggerFactory.getLogger(QueryInformationLogger.class.getName() + ".internalQuery");
+    private static final Logger EXTERNAL_QUERIES_LOG = LoggerFactory.getLogger(QueryInformationLogger.class.getName() + ".externalQuery");
+    
+    protected static final String OAK_INTERNAL_MARKER = "/* oak-internal */";
```
anchela: that /* oak-internal */ is no specific for the QueryInformationLogger, right?
so, i would rather have it present with some generic IndexConstants object...... e.g. org.apache.jackrabbit.oak.plugins.index.IndexConstants. would that work?

also i don't like too much that it's protected.... can't explain just a gut feeling :)

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
```diff
@@ -0,0 +1,110 @@
...
+    
+    /**
+     * Logs the caller of the query based on the callstack. To refine the result, all stack frames
+     * are ignored, for which the entry (consisting of package name, classname and method name) starts
+     * with one of the entries provided by ignoredClassNames.
+     * 
+     * @param statement the query statement
+     * @param ignoredClassNames entries to be ignored. If empty, the method is a no-op.
+     */
+    public static void logCaller (String statement, String[] ignoredClassNames) {
```
anchela: i would again add @notnull annotations for the 2 params... otherwise you have to explain what a null value means and how it is handled. so far you only explain what an empty string means....

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
```diff
@@ -0,0 +1,110 @@
...
+ * KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+package org.apache.jackrabbit.oak.query;
+
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class QueryInformationLogger {
```
anchela, line 22: as far as i can see the QueryInformationLogger only has static methods and static fields.
so, it's a utility right? IMHO it should in this case come with a private constructor that avoids having instances created.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
```diff
@@ -0,0 +1,110 @@
...
+     * Logs the caller of the query based on the callstack. To refine the result, all stack frames
+     * are ignored, for which the entry (consisting of package name, classname and method name) starts
+     * with one of the entries provided by ignoredClassNames.
+     * 
+     * @param statement the query statement
+     * @param ignoredClassNames entries to be ignored. If empty, the method is a no-op.
+     */
+    public static void logCaller (String statement, String[] ignoredClassNames) {
+        
+        if (ignoredClassNames.length == 0) {
```
anchela, line 40: :) that one would result in NPE if the 'ignoredClassNames' array were to be null..... that's why i like the notnull annotations.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+        
+        if (isOakInternalQuery(statement)) {
+            INTERNAL_QUERIES_LOG.debug("Oak-internal query, query=[{}]", statement);
+            return;
+        }
+        
+        String callingClass = getInvokingClass (ignoredClassNames);
+        if (callingClass.isEmpty()) {
+            INTERNAL_QUERIES_LOG.debug("Oak-internal query, query=[{}]", statement);
```
anchela, line 53: why is it an internal query if the callingClass is empty? 
for me as an outsider that feels a bit odd.... maybe add a comment explaining it?

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
```diff
@@ -0,0 +1,110 @@
...
+        if (ignoredClassNames.length == 0) {
+            return;
+        }
+        
+        if (isOakInternalQuery(statement)) {
+            INTERNAL_QUERIES_LOG.debug("Oak-internal query, query=[{}]", statement);
+            return;
+        }
+        
+        String callingClass = getInvokingClass (ignoredClassNames);
```
anchela: formatting: extra white space between getInvokingClass and (

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Retrieves the calling class and method from the call stack; this is determined by unwinding
+     * the stack until it finds a combination of full qualified classname + method (separated by ".") which
+     * do not start with any of the values provided by the ignoredJavaPackages parameters.
+     * 
+     * @param ignoredJavaPackages
+     * @return the calling class; if all classes of the call trace are in packages which are ignored,
+     *   an empty string is returned.
+     */
+    private static String getInvokingClass (String[] ignoredJavaPackages) {
```
anchela: - add @notnull annotations for param and return value
- formatting: extra whitespace between method name and (

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
```diff
@@ -0,0 +1,110 @@
...
+            }
+        }
+    }
+    
+    /**
+     * Retrieves the calling class and method from the call stack; this is determined by unwinding
+     * the stack until it finds a combination of full qualified classname + method (separated by ".") which
+     * do not start with any of the values provided by the ignoredJavaPackages parameters.
+     * 
+     * @param ignoredJavaPackages
```
anchela, line 78: javadoc description is missing..... but ok.... it's a private class and it's obvious from the context.... 
so really minor nitpicking :)

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
```diff
@@ -0,0 +1,110 @@
...
+            if (!foundMatch) {
+                return stackFrame.getClassName() + "." + stackFrame.getMethodName();
+            }
+        }
+        return ""; 
+    }
+    
+    
+    
+    private static boolean isOakInternalQuery (String statement) {
```
anchela: add @notnull annotation

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
```diff
@@ -0,0 +1,110 @@
...
+        return statement.endsWith(OAK_INTERNAL_MARKER);
```
anchela: that one would raise a NPE if the statement were to be null.... and without the annotation the default would be that it actually can be null..... but the check for null can be omitted if you make the contract clear... you never want a null query statement.

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryInformationLoggerTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+import org.apache.jackrabbit.oak.commons.junit.LogCustomizer;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.event.Level;
+
+public class QueryInformationLoggerTest {
+    
+    private static final Logger LOG = LoggerFactory.getLogger(QueryInformationLoggerTest.class);
```
anchela: i might be mistaken.... but i haven't seen this LOG field used.... if it's not -> drop.

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryInformationLoggerTest.java
```diff
@@ -0,0 +1,143 @@
...
+        LogCustomizer internal = LogCustomizer.forLogger(INTERNAL_LOG).enable(Level.TRACE).create();
+        try {
+            external.starting();
+            internal.starting();
+            String query = "SELECT * FROM [cq:Page]";
+            QueryInformationLogger.logCaller(query, ignoredClasses);
+            assertEquals("Exactly 1 entry must be written",1,external.getLogs().size());
+            String logEntry = external.getLogs().get(0);
+            assertTrue(logEntry.contains("org.junit.runners.model")); // org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
+            assertEquals(0,internal.getLogs().size());
```
anchela: formatting: i would add a space after the ,

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryInformationLoggerTest.java
```diff
@@ -0,0 +1,143 @@
...
+    public void regularQueryTest() {
+        
+        LogCustomizer external = LogCustomizer.forLogger(EXTERNAL_LOG).enable(Level.INFO).create();
+        LogCustomizer internal = LogCustomizer.forLogger(INTERNAL_LOG).enable(Level.TRACE).create();
+        try {
+            external.starting();
+            internal.starting();
+            String query = "SELECT * FROM [cq:Page]";
+            QueryInformationLogger.logCaller(query, ignoredClasses);
+            assertEquals("Exactly 1 entry must be written",1,external.getLogs().size());
```
anchela: formatting: i would add spaces after each , for the sake of readability

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryInformationLoggerTest.java
```diff
@@ -0,0 +1,143 @@
...
+    public void withStacktrace() {
+        
+        LogCustomizer external = LogCustomizer.forLogger(EXTERNAL_LOG).enable(Level.DEBUG).create();
+        LogCustomizer internal = LogCustomizer.forLogger(INTERNAL_LOG).enable(Level.TRACE).create();
+        try {
+            external.starting();
+            internal.starting();
+            String query = "SELECT * FROM [cq:Page]";
+            QueryInformationLogger.logCaller(query, ignoredClasses);
+            assertEquals("Exactly 2 entries must be written",2,external.getLogs().size()); //
```
anchela: formatting: add a space after the ,

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryInformationLoggerTest.java
```diff
@@ -0,0 +1,143 @@
...
+        
+        LogCustomizer external = LogCustomizer.forLogger(EXTERNAL_LOG).enable(Level.DEBUG).create();
+        LogCustomizer internal = LogCustomizer.forLogger(INTERNAL_LOG).enable(Level.TRACE).create();
+        try {
+            external.starting();
+            internal.starting();
+            String query = "SELECT * FROM [cq:Page]";
+            QueryInformationLogger.logCaller(query, ignoredClasses);
+            assertEquals("Exactly 2 entries must be written",2,external.getLogs().size()); //     
+            assertEquals(0,internal.getLogs().size());
```
anchela: formatting: add a space after the ,

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryInformationLoggerTest.java
```diff
@@ -0,0 +1,143 @@
...
+        LogCustomizer external = LogCustomizer.forLogger(EXTERNAL_LOG).enable(Level.DEBUG).create();
+        LogCustomizer internal = LogCustomizer.forLogger(INTERNAL_LOG).enable(Level.INFO).create();
+        try {
+            external.starting();
+            internal.starting();
+            String query = "SELECT * FROM [cq:Page] " + QueryInformationLogger.OAK_INTERNAL_MARKER;
+            QueryInformationLogger.logCaller(query, ignoredClasses);
+            
+            // On INFO no logs are written
+            assertEquals(0,external.getLogs().size()); //
```
anchela: - formatting: add a space after ,
- i think the // at the end of the line are not needed, right? the comment is already above

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryInformationLoggerTest.java
```diff
@@ -0,0 +1,143 @@
...
+        LogCustomizer internal = LogCustomizer.forLogger(INTERNAL_LOG).enable(Level.INFO).create();
+        try {
+            external.starting();
+            internal.starting();
+            String query = "SELECT * FROM [cq:Page] " + QueryInformationLogger.OAK_INTERNAL_MARKER;
+            QueryInformationLogger.logCaller(query, ignoredClasses);
+            
+            // On INFO no logs are written
+            assertEquals(0,external.getLogs().size()); //     
+            assertEquals(0,internal.getLogs().size());
```
anchela: formatting: add a space after ,

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryInformationLoggerTest.java
```diff
@@ -0,0 +1,143 @@
...
+            internal.starting();
+            String query = "SELECT * FROM [cq:Page]";
+            QueryInformationLogger.logCaller(query, allClassesIgnored);
+            try {
+                throw new AssertionError();
+            } catch (AssertionError e) {
+                e.printStackTrace();
+            }
+            
+            assertEquals(0,external.getLogs().size());
```
anchela: formatting again :)
you could simplify as `assertTrue(external.getLogs().isEmpty()`.. but that's probably a matter of taste.

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryInformationLoggerTest.java
```diff
@@ -0,0 +1,143 @@
...
+            String query = "SELECT * FROM [cq:Page]";
+            QueryInformationLogger.logCaller(query, allClassesIgnored);
+            try {
+                throw new AssertionError();
+            } catch (AssertionError e) {
+                e.printStackTrace();
+            }
+            
+            assertEquals(0,external.getLogs().size());    
+            assertEquals(0,internal.getLogs().size()); // would require DEBUG
```
anchela: - formatting again :)
- you could simplify as assertTrue(external.getLogs().isEmpty().. but that's probably a matter of taste.

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryInformationLoggerTest.java
```diff
@@ -0,0 +1,143 @@
...
+    public void testWithAllStackEntriesIgnored_DEBUG() {
+        LogCustomizer external = LogCustomizer.forLogger(EXTERNAL_LOG).enable(Level.DEBUG).create();
+        LogCustomizer internal = LogCustomizer.forLogger(INTERNAL_LOG).enable(Level.DEBUG).create();
+        try {
+            external.starting();
+            internal.starting();
+            String query = "SELECT * FROM [cq:Page]";
+            QueryInformationLogger.logCaller(query, allClassesIgnored);
+            
+            assertEquals(0,external.getLogs().size());
```
anchela: see above. formatting and potential simplification.

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryInformationLoggerTest.java
```diff
@@ -0,0 +1,143 @@
...
+        LogCustomizer external = LogCustomizer.forLogger(EXTERNAL_LOG).enable(Level.DEBUG).create();
+        LogCustomizer internal = LogCustomizer.forLogger(INTERNAL_LOG).enable(Level.DEBUG).create();
+        try {
+            external.starting();
+            internal.starting();
+            String query = "SELECT * FROM [cq:Page]";
+            QueryInformationLogger.logCaller(query, allClassesIgnored);
+            
+            assertEquals(0,external.getLogs().size());    
+            assertEquals(1,internal.getLogs().size()); // query statement
```
anchela: see above.... formatting

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryInformationLoggerTest.java
```diff
@@ -0,0 +1,143 @@
...
+        LogCustomizer external = LogCustomizer.forLogger(EXTERNAL_LOG).enable(Level.DEBUG).create();
+        LogCustomizer internal = LogCustomizer.forLogger(INTERNAL_LOG).enable(Level.TRACE).create();
+        try {
+            external.starting();
+            internal.starting();
+            String query = "SELECT * FROM [cq:Page]";
+            QueryInformationLogger.logCaller(query, allClassesIgnored);
+            
+            // On INFO no logs are written
+            assertEquals(0,external.getLogs().size());
```
anchela: - formatting: missing space after ,
- potential simplification see above

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryInformationLoggerTest.java
```diff
@@ -0,0 +1,143 @@
...
+        LogCustomizer internal = LogCustomizer.forLogger(INTERNAL_LOG).enable(Level.TRACE).create();
+        try {
+            external.starting();
+            internal.starting();
+            String query = "SELECT * FROM [cq:Page]";
+            QueryInformationLogger.logCaller(query, allClassesIgnored);
+            
+            // On INFO no logs are written
+            assertEquals(0,external.getLogs().size());    
+            assertEquals(2,internal.getLogs().size()); // query statement + Stacktrace
```
anchela: formattign again :)

anchela (changes requested):
@joerghoh, i am not too familiar with the query code and the PR looks really good and nice to me. so, it's only minor stuff i would like you to check again.

i would suggest though to consistently add notnull annotations to make the API contract clear (and avoid having to check for null everywhere).
and then there were a few formatting issues....

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettingsService.java
joerghoh: I have never seen this annotation being used in a class annotated with ```@interface```, as this is normally not instantiated directly, but just created by some OSGI magic. Will add an explicit default value though.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
joerghoh: Good point, ```QueryEngine.INTERNAL_SQL2_QUERY``` is it.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
joerghoh: Fixed

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
joerghoh: Fixed

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
joerghoh: removed

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
joerghoh: added a private constructor

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
```diff
@@ -139,5 +140,23 @@ void setQueryValidatorPattern(
...
+     */
+    
+    @Description("Set or remove Java Package Name to ignore in Call Trace analysis")
+    void setIgnoredClassNamesinCallTrace(
+            @Description("package or fully qualified class names")
+            @Name("class names")
+            @NotNull String[] classNames);
+    
+    @Description("Get the Java package / fully qualified class names to ignore when finding the caller of query")
+    @NotNull String[] getIgnoredClassNamesinCallTrace();
```
anchela, line 159: hi @joerghoh i looked at your patch again in my IDE not just the diff on github. and my IDE complains about a typo in this method name. it should be `getIgnoredClassNamesInCallTrace`

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
```diff
@@ -139,5 +140,23 @@ void setQueryValidatorPattern(
...
+     * Set or remove java package/class names which are ignored from finding the 
+     * invoking class for queries.
+     * 
+     * it can be either Java package names or fully-qualified class names (package + class name).
+     * 
+     * @param classNames the class names to be ignored.
+     */
+    
+    @Description("Set or remove Java Package Name to ignore in Call Trace analysis")
+    void setIgnoredClassNamesinCallTrace(
```
anchela, line 153: my IDE complains about a typo in the method name:  should be `setIgnoredClassNamesInCallTrace`

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
```diff
@@ -974,6 +974,15 @@ public String toString() {
         void setFullTextComparisonWithoutIndex(boolean fullTextComparisonWithoutIndex) {
             this.settings.setFullTextComparisonWithoutIndex(fullTextComparisonWithoutIndex);
         }
+        
+        public void setIgnoredClassNamesinCallTrace(@NotNull String[] packageNames) {
```
anchela, line 978: same here.... `setIgnoredClassNamesInCallTrace`

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
```diff
@@ -974,6 +974,15 @@ public String toString() {
...
+            settings.setIgnoredClassNamesinCallTrace(packageNames);
+        }
+        
+        public @NotNull String[] getIgnoredClassNamesinCallTrace() {
```
anchela, line 982: and here `getIgnoredClassNamesInCallTrace`

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
```diff
@@ -0,0 +1,120 @@
...
+            return;
+        }
+        
+        String callingClass = getInvokingClass(ignoredClassNames);
+        if (callingClass.isEmpty()) {
+            // If all stackframes are covered by the ignoredClassNames, it's something which is
+            // explicitly wanted not to be logged along other queries for which this is not true.
+            INTERNAL_QUERIES_LOG.debug("Oak-internal query, query=[{}]", statement);
+            if (INTERNAL_QUERIES_LOG.isTraceEnabled()) {
+                try {
```
anchela, line 55: this try-catch looks a bit odd.... i wonder if it could be replaced by just writing:
`INTERNAL_QUERIES_LOG.trace("providing requested stacktrace", new RuntimeException("requested stacktrace");`
but i am not 100% sure....

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
```diff
@@ -0,0 +1,120 @@
...
+                    throw new RuntimeException("requested stacktrace");
+                } catch (RuntimeException e) {
+                    INTERNAL_QUERIES_LOG.trace("providing requested stacktrace", e);
+                }
+            }
+        } else {
+            EXTERNAL_QUERIES_LOG.info("query=[{}], caller=[{}]",statement, callingClass);
+            if (EXTERNAL_QUERIES_LOG.isDebugEnabled()) {
+                try {
```
anchela, line 64: same here

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
```diff
@@ -0,0 +1,120 @@
...
+    
+    /**
+     * Check if statement is explicitly marked to be internal
+     * @param statement the JCR query statement to check
+     * @return true the query is marked to be an internal query
+     */
+    private static boolean isOakInternalQuery(@NotNull String statement) {
+        return statement.endsWith(QueryEngine.INTERNAL_SQL2_QUERY);
+    }
+
```
anchela: extra lines until end of class -> drop

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryInformationLoggerTest.java
```diff
@@ -0,0 +1,141 @@
+package org.apache.jackrabbit.oak.query;
+
```
anchela, line 14: one line too much :)

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryInformationLoggerTest.java
```diff
@@ -0,0 +1,141 @@
...
+            
+            assertEquals(0, external.getLogs().size());    
+            assertEquals(0, internal.getLogs().size()); // would require DEBUG
+        } finally {
+            external.finished();
+            internal.finished();
+        }     
+    }  
+    
+
```
anchela: remove 1 empty line

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryInformationLoggerTest.java
```diff
@@ -0,0 +1,141 @@
+package org.apache.jackrabbit.oak.query;
```
anchela: missing license header.....

anchela (commented):
sent you a patch with the whitespace/header findings.
regarding the try-catch.... i am not totally sure.... but maybe you want to give it a try.
i didn't fix the typo in the method names though.... wdyt about this one?

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
```diff
@@ -0,0 +1,111 @@
...
+                final String classMethod = stackFrame.getClassName() + "." + stackFrame.getMethodName();
+                if (classMethod.startsWith(packageName)) {
+                    foundMatch = true;
+                    continue;
+                }
+            }
+            if (!foundMatch) {
+                return stackFrame.getClassName() + "." + stackFrame.getMethodName();
+            }
+        }
```
jsedding, line 99: The logic in this loop seems a bit hard to read and could be more efficient (it calls `stackFrame.getClassName()` only once per `stackFrame` and `pkg`).

```suggestion
        for (StackTraceElement stackFrame : callStack) {
            final String className = stackFrame.getClassName();
            if (Stream.of(ignoredJavaPackages).noneMatch(pkg -> className.startsWith(pkg))) {
                return className + "." + stackFrame.getMethodName();
            }
        }
```

jsedding (commented):
Thanks for getting this logging added, I've struggled before to find out where my bad query was coming from!

Probably I don't have a full understanding of the issue you are trying to solve. But don't we already detect when a query is slow, iterates too many nodes etc? Would it not be possible to get a stack-trace only when one of these conditions is met and we want to actually log it?

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryInformationLogger.java
joerghoh: Thanks for the suggestion, I will slightly adapt it and add the ability to ignore specific method calls as well.

Regarding the implementation: We want to explicitly be able to find the creators of _all_ queries, not just the slow ones.

joerghoh:
I redefined the logging strategy with @thomasmueller  and this approach does not seem to be too useful. I will close this PR and come up with a new one.
Thanks to @anchela  and @jsedding for your help, I will incorporate your feedback into the next iteration of this feature.

PR #421, opened 2021-11-24
backporting of upgrading netty dependency to 4.1.68.final jira OAK-9611

ArunOnCloud:


reschke (changes requested):
Please update the PR so that it only contains necessary changes; right now the whole pom.xml is a a huge diff.

PR #431, opened 2021-12-02
introduce reliability tests on elastic

fabriziofortino:


oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnectionRule.java
```diff
@@ -52,28 +53,19 @@
...
+    private final String elasticConnectionString;
+
     public ElasticConnectionRule(String elasticConnectionString) {
         this.elasticConnectionString = elasticConnectionString;
     }
 
     public ElasticsearchContainer elastic;
 
-    /*
-    Executed once in the test class' execution lifecycle, after the execution of apply()
```
nit0906, line 68: I don't recall exactly, but I think the purpose of getting the docker connection in the before method here was to make sure that docker container is run only once during the whole test class' execution - otherwise it would probably run and get destroyed after every test - leading to increased test execution time.

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnectionRule.java
```diff
@@ -126,61 +124,56 @@ private void downloadSimilaritySearchPluginIfNotExists(String localPluginPath, S
...
+        return ElasticConnection.newBuilder()
+                .withIndexPrefix(INDEX_PREFIX + System.currentTimeMillis())
+                .withConnectionParameters(ElasticConnection.DEFAULT_SCHEME,
+                        containerIpAddress, port)
+                .withApiKeys(null, null)
+                .build();
     }
 
-    public void closeElasticConnection() throws IOException {
+    public void closeElasticConnection(ElasticConnection elasticConnection) throws IOException {
```
nit0906: I couldn't find where this is getting called from now ?

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnectionRule.java
fabriziofortino: I thought the same. But this was actually not needed. I confirm all tests in a class run against the same docker container.

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnectionRule.java
fabriziofortino: good catch! This is not needed anymore since the container gets removed when the tests are complete.

nit0906: @fabriziofortino 
I think not having a cleanup method that closes ElasticConnection and deletes test indexes might create problems if tests are run on a non docker bases elastic server - I am not sure if we would really need to do that. 
But maybe we can add TODO statement there in case it's needed in future.

Rest all looks good to me.

nit0906 (approved):

thomasmueller (approved):

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnectionRule.java
fabriziofortino: You are right. I forgot it was possible to run the tests without testcontainers. I added back the delete index but I have moved it in `ElasticAbstractQueryTest`. I think it makes more sense to have it there.

PR #442, opened 2021-12-16
OAK-9644: Flaky test due to HashSet

mreutegg:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/ClusterView.java
```diff
@@ -187,22 +187,13 @@ private String asJson(final long viewSeqNum, final boolean viewFinal, final Stri
         builder.key("id").value(clusterId);
         builder.key("me").value(localId);
         builder.key("active").array();
-        for (Iterator<Integer> it = activeIds.iterator(); it.hasNext();) {
```
jelmini, line 190: A potential alternative:
`activeIds.stream().sorted().forEachOrdered(builder::value);`
The `sorted()` operator achieves the same than the `TreeSet`. 
Usually, `forEach()` and `forEachOrdered()` should have the same behaviour with a sequential stream, but in theory `forEach()` behaviour  is undefined, because in case of a parallel stream it could apply items in any order.

mreutegg: I like this suggestion and updated the PR accordingly. Thanks!

mreutegg:
@stefan-egli WDYT?

stefan-egli:
+1, lgtm.
I'll go ahead and merge (hope that's fine with you)

PR #443, opened 2021-12-16
OAK-9553: Add config support for multiple mounts in CompositeNodeStore

jelmini:
Add MountInfoConfig configuration factory, which can be used to
configure multiple read-only mounts.
MountInfoProviderService can now be configured with the list of expected
mounts, so it can wait for all of them before registering a
MountInfoProvider.
For backward-compatibility, MountInfoProviderService can still be used
to configure a single read-only mount, but all such properties are now
marked as deprecated.

### DIFF
--- a/oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/MountInfoConfig.java
+++ b/oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/MountInfoConfig.java
@@ -20,57 +20,53 @@ package org.apache.jackrabbit.oak.composite;
 
 import java.util.Arrays;
 import java.util.List;
-import java.util.Map;
-import org.apache.felix.scr.annotations.Activate;
-import org.apache.felix.scr.annotations.Component;
-import org.apache.felix.scr.annotations.Property;
-import org.apache.felix.scr.annotations.PropertyUnbounded;
-import org.apache.felix.scr.annotations.Service;
-import org.apache.jackrabbit.oak.commons.PropertiesUtil;
 import org.osgi.framework.BundleContext;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.metatype.annotations.AttributeDefinition;
+import org.osgi.service.metatype.annotations.Designate;
+import org.osgi.service.metatype.annotations.ObjectClassDefinition;
 
 import static java.util.stream.Collectors.toList;
 
-@Component(label = "Apache Jackrabbit Oak MountInfo Config", configurationFactory = true, metatype = true)
-@Service(MountInfoConfig.class)
+@Component(service = MountInfoConfig.class)
+@Designate(ocd = MountInfoConfig.Props.class, factory = true)
 public class MountInfoConfig {
 
-    private static final String[] PROP_MOUNTED_PATHS_DEFAULT = {};
+    @ObjectClassDefinition(name = "MountInfoConfig Properties")
+    public @interface Props {
 
-    @Property(label = "Mounted paths",
-        unbounded = PropertyUnbounded.ARRAY,
-        description = "Paths which are part of private mount"
-    )
-    private static final String PROP_MOUNT_PATHS = "mountedPaths";
+        String DEFAULT_MOUNT_NAME = "private";
+
+        @AttributeDefinition(
+            name = "Mounted paths",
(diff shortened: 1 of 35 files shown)


oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/MountInfoConfig.java
```diff
@@ -0,0 +1,115 @@
...
+import org.apache.felix.scr.annotations.Component;
+import org.apache.felix.scr.annotations.Property;
+import org.apache.felix.scr.annotations.PropertyUnbounded;
+import org.apache.felix.scr.annotations.Service;
+import org.apache.jackrabbit.oak.commons.PropertiesUtil;
+import org.osgi.framework.BundleContext;
+
+import static java.util.stream.Collectors.toList;
+
+@Component(label = "Apache Jackrabbit Oak MountInfo Config", configurationFactory = true, metatype = true)
```
kwin: For new components rather use the official OSGi component annotations: https://github.com/apache/jackrabbit-oak/blob/9a2cf021772ac577d161f1409ada6edd9642a441/oak-parent/pom.xml#L548

oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/MountInfoConfig.java
```diff
@@ -0,0 +1,115 @@
...
+
+    MountInfoConfig(String mountName, List<String> paths, List<String> pathsSupportingFragments, boolean readOnly) {
+        this.mountName = mountName;
+        this.paths = trimAll(paths);
+        this.pathsSupportingFragments = trimAll(pathsSupportingFragments);
+        this.readOnly = readOnly;
+    }
+
+    @Activate
+    private void activate(BundleContext bundleContext, Map<String, ?> config) {
```
kwin: You should use component property types to ease dealing with defaults and generating meta type information at the same time.

oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/MountInfoConfig.java
kwin: Compare also with the ongoing effort tracked in https://issues.apache.org/jira/browse/OAK-6741. Mixing both approaches in the same module shouldn't be a problem.

oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/package-info.java
```diff
@@ -55,7 +55,7 @@
  *  This is obviously correct but may be slow.
  *  {@link org.apache.jackrabbit.oak.composite.CompositionContext#getContributingStores(java.lang.String, java.util.function.Function)}
  */
-@Version("0.3.0")
+@Version("0.4.0")
```
smiroslav, line 58: According to guideline below, this should be 1.0.0
https://semver.org/#how-do-i-know-when-to-release-100

smiroslav (approved):

oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/MountInfoConfig.java
jelmini: Thanks for your suggestion, @kwin. I have migrated the classes touched by this PR to the official OSGi component annotations.

oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/MountInfoConfig.java
jelmini: In the context of moving to the official OSGi component annotations, the properties are now defined on an `@interface` annotated with `@ObjectClassDefinition` and `@AttributeDefinition`.

oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/package-info.java
jelmini: As discussed, bumping now to 1.0.0 could break user code. Best done in a separate PR.

jelmini:
Added integration tests using Pax Exam in oak-store-composite, validating the new configuration of multiple read-only mounts, as well as backward compatibility with previous configuration.

smiroslav (approved):

PR #444, opened 2021-12-16
OAK-9647: oak-search-elastic: nullCheckEnabled and notNullCheckEnabled queries not working as expected

fabriziofortino:


oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/TermQueryBuilderFactory.java
```diff
@@ -128,20 +115,17 @@ public static TermQueryBuilder newNullPropQuery(String propName) {
         } else if (pr.first != null && pr.last != null) {
             return newRangeQuery(propertyName, first, last,
                     pr.firstIncluding, pr.lastIncluding);
-        } else if (pr.first != null && pr.last == null) {
+        } else if (pr.first != null) {
```
thomasmueller, line 118: This change seems to be unrelated to the oak issue. I see that technically, it's not needed.

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/PropertyIndexCommonTest.java
```diff
@@ -171,6 +180,56 @@ public void propertyExistenceQuery() throws Exception {
...
+
+        Tree test = root.getTree("/").addChild("test");
+        createNodeWithType(test, "a", "oak:TestNode").setProperty("propa", "a");
+        createNodeWithType(test, "b", "oak:TestNode").setProperty("propa", "c");
+        createNodeWithType(test, "c", "oak:TestNode").setProperty("propb", "e");
+        root.commit();
+
+        String query = "select [jcr:path] from [oak:TestNode] where [propa] is null";
+        String explanation = explain(query);
+        assertThat(explanation, containsString(indexOptions.getIndexType() + ":test2(/oak:index/test2) "));
```
thomasmueller: I think it is insufficient. The old test verified the elastic query, however the new test doesn't. The test wouldn't fail even if the query string isn't correct, as the Oak query engine would filter out the data.

thomasmueller (changes requested):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/TermQueryBuilderFactory.java
fabriziofortino: Correct. This code was inherited from the initial Elastic implementation. This check is redundant and can be safely removed to tidy up the code.

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/PropertyIndexCommonTest.java
fabriziofortino: Right. I have re-worked the test to make sure the Lucene test is verified.

thomasmueller (approved):
This looks correct to me... The Lucene test case verifies that Lucene uses the ":notNullProps:" / ":nullPros:", while for Elastic we assume it is done correctly as we don't have the Elastic query in the plan. I think it would be nice if for Elastic, the explain plan would also contain the Elastic query, but I guess this can cause a big output.

fabriziofortino:
@thomasmueller the elastic query is a json string. We could verify that is correct but I think it's not needed because we check that the index is used and the results are correct.

PR #446, opened 2021-12-23
OAK-9651: Protection against very large queries

tihom88:


oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineImpl.java
```diff
@@ -151,12 +151,19 @@
         } else {
             LOG.debug("Parsing {} statement: {}", language, statement);
         }
+        QueryEngineSettings settings = context.getSettings();
+        if (statement.length() > (settings.getQueryLengthErrorLimit())){
+            LOG.error("Too large query: " + statement);
+            throw new RuntimeException("Query length "+ statement.length() + " is larger than max supported query length: " + settings.getQueryLengthErrorLimit());
```
thomasmueller: Maybe use ParseException instead of RuntimeException? We could use another exception as well, e.g. IllegalArgumentException, but we use ParseException below as well; I think it's better than to introduce a new exception type.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
```diff
@@ -93,6 +93,22 @@
 
     private String[] classNamesIgnoredInCallTrace = new String[] {};
 
+
+    private static final String OAK_QUERY_LENGTH_WARN_LIMIT = "oak.query.length.warn.limit";
+    private static final String OAK_QUERY_LENGTH_ERROR_LIMIT = "oak.query.length.error.limit";
+
+    private final long queryLengthWarnLimit = Long.getLong(OAK_QUERY_LENGTH_WARN_LIMIT, 102400); // 100 KB
```
thomasmueller: So let's use "1024 * 1024" here (1 MiB)

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
```diff
@@ -93,6 +93,22 @@
...
+    private final long queryLengthErrorLimit = Long.getLong(OAK_QUERY_LENGTH_ERROR_LIMIT, 1048576); //1MB (1024KB)
```
thomasmueller: We currently want to see if there are warnings, and not throw errors, so 100 * 1024 * 1024 would be better I think (100 MiB). BTW did you know that newer Java versions supports "_" in numbers, so 1_000_000 would be 1 million. But I think using 100 * 1024 is OK (serves the same purpose).

thomasmueller (changes requested):

thomasmueller (approved):

PR #448, opened 2022-01-04
OAK-9653 Add unit test

schulm:
Jira: https://issues.apache.org/jira/browse/OAK-9653

I've just copied the ``descendantTest()`` and added the index tag to the query which leads to a failing test.

### DIFF
--- a/oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/IndexQueryCommonTest.java
+++ b/oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/IndexQueryCommonTest.java
@@ -55,6 +55,7 @@ public abstract class IndexQueryCommonTest extends AbstractQueryTest {
         indexDefn = createTestIndexNode(index, indexOptions.getIndexType());
         TestUtil.useV2(indexDefn);
         indexDefn.setProperty(FulltextIndexConstants.EVALUATE_PATH_RESTRICTION, true);
+        indexDefn.setProperty("tags", "x");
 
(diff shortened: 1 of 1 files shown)


oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/IndexQueryCommonTest.java
```diff
@@ -142,6 +143,24 @@ public void descendantTest() throws Exception {
...
+        test.addChild("a");
+        test.addChild("b");
+        root.commit();
+
+        assertEventually(() -> {
+            Iterator<String> result = executeQuery(
+                    "select [jcr:path] from [nt:base] where isdescendantnode('/test') option (index tag x)",
+                    "JCR-SQL2").iterator();
+            assertTrue(result.hasNext());
+            assertEquals("/test/a", result.next());
```
thomasmueller, line 158: I believe the test would pass even if no index is used, or the wrong index. I believe it would be better if the test verifies the right index is used, by executing "explain select" and then checking the plan.

I wonder which Jira issue this test refers to?

thomasmueller (changes requested):

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/IndexQueryCommonTest.java
schulm: Hi  @thomasmueller I've just created https://issues.apache.org/jira/browse/OAK-9653

PR #449, opened 2022-01-04
OAK-9654 : more logging for a RuntimeException

stefan-egli:
Plus an almost forgotten [minor cleanup](https://github.com/apache/jackrabbit-oak/commit/8c5bb742c056e0a150caff8f5ce910bb5c92391e) on /Jenkinsfile


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CachingCommitValueResolver.java
```diff
@@ -148,7 +149,13 @@ public String resolve(@NotNull Revision changeRevision,
             // only cache committed states
             // e.g. branch commits may be merged later and
             // the commit value will change
-            commitValueCache.put(changeRevision, value);
+            try {
+                commitValueCache.put(changeRevision, value);
+            } catch (RuntimeException re) {
+                LoggerFactory.getLogger(getClass()).error("resolve: RuntimeException with " +
+                        changeRevision + " - " + value, re);
+                throw re;
```
mreutegg: How about turning this into a warn message and *not* re-throw the exception? I understand the issue is probably with Guava because both key and value are non-null. Guavas LocalCache.put() even checks this.

stefan-egli: Good point, it probably doesn't matter if that `put` fails - so a rethrow isnt strictly necessary.
Changed that now [here](https://github.com/apache/jackrabbit-oak/pull/449/commits/5a7b331af20345d2e75a068fc183cd1b0c1886c8)

mreutegg (approved):
Looks good now. Can you please update the JIRA issue summary to better reflect the change?

stefan-egli:
thx. yes, jira ticket is updated now.

PR #452, opened 2022-01-06
OAK-9653 Adding the index tag option interferes with regex properties

thomasmueller:


schulm (approved):
I've tested the change locally and it fixes the issue it caused. Thanks, but I'm not sure if some other options like index name will cause troubles together with the regex property field.

nit0906 (approved):
lgtm

PR #453, opened 2022-01-06
OAK-9656: Recovery runs mistakenly when system clock jumps ahead

mreutegg:
Test to reproduce problem.

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
@@ -146,22 +146,21 @@ public class LastRevRecoveryAgent {
             // Check our own lease before running recovery for another
             // clusterId (OAK-9656)
             long now = revisionContext.getClock().getTime();
-            ClusterNodeInfoDocument me = null;
             if (clusterId != revisionContext.getClusterId()) {
                 // Get leaseEnd from our own cluster node info, unless
                 // we are doing recovery on startup for the clusterId
                 // we want to acquire. Then it's fine to go ahead with
                 // an expired lease.
-                me = missingLastRevUtil.getClusterNodeInfo(revisionContext.getClusterId());
-            }
-            if (me != null && me.isRecoveryNeeded(now)) {
-                log.warn("Own clusterId {} has a leaseEnd {} ({}) older than current time {} ({}). " +
-                                "Refusing to run recovery on clusterId {}.",
-                        revisionContext.getClusterId(),
-                        me.getLeaseEndTime(), asISO8601(me.getLeaseEndTime()),
-                        now, asISO8601(now),
-                        clusterId);
-                return 0;
+                ClusterNodeInfoDocument me = missingLastRevUtil.getClusterNodeInfo(revisionContext.getClusterId());
+                if (me != null && me.isRecoveryNeeded(now)) {
+                    log.warn("Own clusterId {} has a leaseEnd {} ({}) older than current time {} ({}). " +
+                                    "Refusing to run recovery on clusterId {}.",
+                            revisionContext.getClusterId(),
+                            me.getLeaseEndTime(), asISO8601(me.getLeaseEndTime()),
+                            now, asISO8601(now),
+                            clusterId);
+                    return 0;
+                }
             }
             // Check if _lastRev recovery needed for this cluster node
             // state is Active && current time past leaseEnd
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
@@ -153,13 +153,13 @@ public class LastRevRecoveryAgent {
                 // an expired lease.
                 ClusterNodeInfoDocument me = missingLastRevUtil.getClusterNodeInfo(revisionContext.getClusterId());
                 if (me != null && me.isRecoveryNeeded(now)) {
-                    log.warn("Own clusterId {} has a leaseEnd {} ({}) older than current time {} ({}). " +
-                                    "Refusing to run recovery on clusterId {}.",
-                            revisionContext.getClusterId(),
-                            me.getLeaseEndTime(), asISO8601(me.getLeaseEndTime()),
-                            now, asISO8601(now),
+                    String msg = String.format(
+                            "Own clusterId %s has a leaseEnd %s (%s) older than current time %s (%s). " +
+                                    "Refusing to run recovery on clusterId %s.",
+                            revisionContext.getClusterId(), me.getLeaseEndTime(),
+                            asISO8601(me.getLeaseEndTime()), now, asISO8601(now),
                             clusterId);
-                    return 0;
+                    throw new DocumentStoreException(msg);
                 }
             }
             // Check if _lastRev recovery needed for this cluster node
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
@@ -141,11 +141,20 @@ public class LastRevRecoveryAgent {
     public int recover(int clusterId, long waitUntil)
             throws DocumentStoreException {
         ClusterNodeInfoDocument nodeInfo = missingLastRevUtil.getClusterNodeInfo(clusterId);
-        ClusterNodeInfoDocument me = missingLastRevUtil.getClusterNodeInfo(revisionContext.getClusterId());
 
         if (nodeInfo != null) {
+            // Check our own lease before running recovery for another
+            // clusterId (OAK-9656)
             long now = revisionContext.getClock().getTime();
-            if (me != null && me.isActive() && me.getLeaseEndTime() < now) {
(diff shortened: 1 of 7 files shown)


mreutegg:
@Joscorbe can you please review?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/RecoveryWithIncorrectClockTest.java
```diff
@@ -0,0 +1,97 @@
...
+        LastRevRecoveryAgent agent = ns2.getLastRevRecoveryAgent();
+        try {
+            assertTrue(agent.isRecoveryNeeded());
+            agent.performRecoveryIfNeeded();
+        } catch (DocumentStoreException e) {
+            // this is expected. any of the above LastRevRecoveryAgent calls
+            // can and should fail because the lease for ns2 expired
+        }
+
+        ClusterNodeInfoDocument infoDoc = store.find(Collection.CLUSTER_NODES, "" + 1);
```
Joscorbe: Can't we just write `"1"` instead of `"" + 1`?

Joscorbe (approved):

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/RecoveryWithIncorrectClockTest.java
mreutegg: Yes, you are right. Don't know what I was thinking here. Probably copied the line and adjusted.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
```diff
@@ -142,9 +143,29 @@ public int recover(int clusterId, long waitUntil)
...
+            long now = revisionContext.getClock().getTime();
+            ClusterNodeInfoDocument me = null;
+            if (clusterId != revisionContext.getClusterId()) {
+                // Get leaseEnd from our own cluster node info, unless
+                // we are doing recovery on startup for the clusterId
+                // we want to acquire. Then it's fine to go ahead with
+                // an expired lease.
+                me = missingLastRevUtil.getClusterNodeInfo(revisionContext.getClusterId());
+            }
+            if (me != null && me.isRecoveryNeeded(now)) {
```
stefan-egli: Since `me` is only used within the above 
```if (clusterId != revisionContext.getClusterId()) {```
block, this block here could be moved 1 line up into that block. That could make it slightly easier to read - since the scope it then smaller. wdyt?

File: oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/RecoveryWithIncorrectClockTest.java [hunk 3]
```diff
@@ -0,0 +1,98 @@
...
+        clock.waitUntil(now + TimeUnit.HOURS.toMillis(1));
+        long ahead = clock.getTime();
+        // ns2 now sees an expired lease for clusterId 1
+        LastRevRecoveryAgent agent = ns2.getLastRevRecoveryAgent();
+        try {
+            assertTrue(agent.isRecoveryNeeded());
+            agent.performRecoveryIfNeeded();
+        } catch (DocumentStoreException e) {
+            // this is expected. any of the above LastRevRecoveryAgent calls
+            // can and should fail because the lease for ns2 expired
```
stefan-egli, line 82: does that mean both lines in the `try/catch` block can fail? in that case, wouldn't it make sense to put each one in an individual `try/catch` block (as otherwise you're not testing that both lines fail)

stefan-egli (commented):
Two comments and a question:
in the ticket I don't fully understand what this means (more to understand the problem than anything else) :
> The current implementation attempts to run recovery and fails because it eventually notices its own lease expired. It appears other clusterIds are modified anyway and then fail the next lease update because the leaseEnd does not match the expected value.

What I don't understand is whether it fails ("eventually notices") in the same run or the next. Also since the 2nd sentence states that it fails on the "next lease update" - but unclear whether that's triggered by the time-jumped node or by any other one.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
mreutegg: Good suggestion. Will do.

mreutegg:
> What I don't understand is whether it fails ("eventually notices") in the same run or the next.

At some point while running recovery the process may notice that its lease expired and fail the recovery with an exception. This may happen when the recovery reads or updates documents and the `LeaseCheckDocumentStoreWrapper` refuses to perform the operation.

> Also since the 2nd sentence states that it fails on the "next lease update" - but unclear whether that's triggered by the time-jumped node or by any other one.

That's indeed confusing. The second sentence is about the impact on the node that has the correct time. The node with the incorrect time wants to run recovery on the clusterId of the node with the correct time. At this point the node with the correct time will observe its entry in the clusterNodes collection updated by someone else and conclude that its lease must have expired.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/RecoveryWithIncorrectClockTest.java
mreutegg: > does that mean both lines in the try/catch block can fail?

At least one of the calls should fail. Without the fix `agent.performRecoveryIfNeeded()` fails with an exception. I didn't want the test to prescribe where the implementation should fail. But with the current fix no exception is actually thrown and the test does not detect it.

I'll change both the test to fail when no exception is thrown and update LastRevRecoveryAgent.recover() to throw an exception when it detects its own lease timed out. Let me know what you think after that change.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
```diff
@@ -142,9 +143,28 @@ public int recover(int clusterId, long waitUntil)
...
+                // an expired lease.
+                ClusterNodeInfoDocument me = missingLastRevUtil.getClusterNodeInfo(revisionContext.getClusterId());
+                if (me != null && me.isRecoveryNeeded(now)) {
+                    String msg = String.format(
+                            "Own clusterId %s has a leaseEnd %s (%s) older than current time %s (%s). " +
+                                    "Refusing to run recovery on clusterId %s.",
+                            revisionContext.getClusterId(), me.getLeaseEndTime(),
+                            asISO8601(me.getLeaseEndTime()), now, asISO8601(now),
+                            clusterId);
+                    throw new DocumentStoreException(msg);
```
mreutegg, line 162: @stefan-egli now the method throws an exception instead of returning zero. WDYT?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/RecoveryWithIncorrectClockTest.java
```diff
@@ -0,0 +1,100 @@
...
+        long now = clock.getTime();
+        // simulate a sudden jump of time to the system clock
+        clock.waitUntil(now + TimeUnit.HOURS.toMillis(1));
+        long ahead = clock.getTime();
+        // ns2 now sees an expired lease for clusterId 1
+        LastRevRecoveryAgent agent = ns2.getLastRevRecoveryAgent();
+        try {
+            assertTrue(agent.isRecoveryNeeded());
+            agent.performRecoveryIfNeeded();
+            fail("recovery must fail with expired lease");
```
mreutegg, line 79: And the test will fail if the two calls above do not throw an exception.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
stefan-egli: ok, I'm assuming callers will be able to handle this exception? in that case this indeed seems better (as it should be a rather unfrequent case, so an exception seems appropriate)

stefan-egli:
Thx for the explanations, makes sense now

stefan-egli (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/LastRevRecoveryAgent.java
mreutegg: The code path hitting this potential exception is via LastRevRecoveryJob.run().

https://github.com/apache/jackrabbit-oak/blob/954d62c5543b97ac9da51ac670e7b54cc09ab5ca/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java#L1011

It does not handle the exception, but I think that's fine. The job is scheduled at a fixed rate here:

https://github.com/apache/jackrabbit-oak/blob/954d62c5543b97ac9da51ac670e7b54cc09ab5ca/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java#L822

It would just run again when it's time, but more likely the lease mechanism shuts down the service before this happens as a reaction to the expired lease.

stefan-egli: ack, thx!

PR #359, opened 2021-09-03
OAK-9564: Make the renewLease mechanism a bit more tolerant against unexpected updates

Joscorbe:
- Added a new RuntimeID field to ensure the update is coming from the same ClusterNode.
- Added a new condition to ensure not older updates are executed.

If an update comes from the same RuntimeID and it's newer, we will still accept it even if it was not expected at that moment.

PR #464, opened 2022-01-16
OAK-9564 (Bugfixed): Make the renewLease mechanism a bit more tolerant against unexpected updates

Joscorbe:
This PR #455, which is a revert of #359 because of a discovered bug after it was merged.
The new commit resolves the bug, improves the logging of the newly added field "Runtime_id", and adds new tests to reproduce and cover the detected bug to allow to acquire recovered and disposed clusters properly.


Joscorbe:
@mreutegg I think now for real, can you review? Thanks!

mreutegg (approved):
Looks good to me.

PR #465, opened 2022-01-17
OAK-9662 | Support not equal condition in query from Index

nit0906:


oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/PropertyValueImpl.java
```diff
@@ -125,12 +125,7 @@ public void restrict(FilterImpl f, Operator operator, PropertyValue v) {
                     f.restrictPath(v.getValue(Type.STRING), PathRestriction.EXACT);
                 }
             } else {
-                if (operator == Operator.NOT_EQUAL && v != null) {
-                    // "x <> 1" also means "x is not null"
-                    f.restrictProperty(pn, Operator.NOT_EQUAL, null, propertyType);
```
thomasmueller, line 130: If we keep this restriction (in addition to the f.restrictProperty(pn, operator, v, propertyType)), then property indexes can continue to use it.

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/PropertyValueImpl.java
nit0906: @thomasmueller  - propertyIndex could still use it - because in restrictProperty method in FilterImpl, I added new varaibles to handle inequality, which would effectively leave the old behaviour for PropertyIndex as it is. 

Changes in Filter.java and FilterImpl.java should handle this. wdyt ?

thomasmueller: No, it might not work correctly... I think the property index (incorrectly) relies on value to be "null" for the is-not-null condition, here:

ContentMirrorStoreStrategy

    long count(Filter filter, NodeState root, NodeState indexMeta, final String indexStorageNodeName,
            Set<String> values, int max) {
        NodeState index = indexMeta.getChildNode(indexStorageNodeName);
        long count = -1;
        if (values == null) {
            // property is not null

Yes, it seems to be a bug in this index implementation. But it would no longer work as it did work, I think. And to me it seems there are not enough test cases in the property index, maybe.

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/PropertyValueImpl.java
nit0906: Oh..I see.
>And to me it seems there are not enough test cases in the property index, maybe.
Yes, because no tests are failing with current changes.

Also, I tried to add both the property restrictions i.e
f.restrictProperty(pn, Operator.NOT_EQUAL, null, propertyType);
and 
f.restrictProperty(pn, operator, v, propertyType));

But it seems that the FilterImpl only returns the first one when we try and get all the PropertyRestrictions for a given property, so only the first of the 2 restrictions are honored.

nit0906: @thomasmueller  - thanks. I understand now.

>And to me it seems there are not enough test cases in the property index, maybe.

Yes, all the tests were passing and I was unaware of this usage in ContentMirrorStoreStrategy. 

I have made some changes in the latest commit - 
https://github.com/apache/jackrabbit-oak/pull/465/commits/f9d45d8f8f422cdf9f27f1aca794ae73e9b5e0bc. Now adding 2 propertyRestrictions - so that the old behaviour with PropertyIndex is not broken.

Can you please re review ?

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
```diff
@@ -1237,13 +1238,18 @@ private static Query createQuery(String propertyName, PropertyRestriction pr,
                 } else if (pr.isNotNullRestriction()) {
                     // not null. For date lower bound of zero can be used
                     return NumericRangeQuery.newLongRange(propertyName, 0L, Long.MAX_VALUE, true, true);
+                } else if (pr.isNot && pr.not != null) {
+                    // -[property]=[value]
+                    BooleanQuery bool = new BooleanQuery();
+                    // This will exclude entries with [property]=[value]
+                    bool.add(NumericRangeQuery.newLongRange(propertyName, pr.not.getValue(LONG), pr.not.getValue(LONG), true, true), MUST_NOT);
```
averma21, line 1245: Except for the default case, the boolean queries are not returned. I think return statement is missing and the tests only test for strings, so these cases didn't get detected.

nit0906: good catch. Thanks. Fixed it now - https://github.com/apache/jackrabbit-oak/pull/465/commits/b45147efafc2fe77c2b576e1ebe1672426256838

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexQueryCommonTest.java
```diff
@@ -57,4 +57,26 @@ public void descendantTestWithIndexTagExplain() throws Exception {
...
+    public String getContainsValueFortestInequalityQueryWithoutAncestorFilter_native() {
+        return "\"filter\":[{\"exists\":{\"field\":\"propa\",\"boost\":1.0}},{\"bool\":" +
+                "{\"must_not\":[{\"term\":{\"propa.keyword\":{\"value\":\"bar\",\"boost\":1.0}}}]";
+    }
+
+    @Override
+    public String getContainsValueFortestEqualityInequalityCombined_native() {
+        return "\"filter\":[{\"term\":{\":ancestors\":{\"value\":\"/test\",\"boost\":1.0}}},{\"term\":{\"propb.keyword\":{\"value\":\"world\",\"boost\":1.0}}}," +
+                "{\"exists\":{\"field\":\"propa\",\"boost\":1.0}},{\"bool\":{\"must_not\":[{\"term\":{\"propa.keyword\":{\"value\":\"bar\",\"boost\":1.0}}}]";
+    }
```
averma21, line 81: can we add one more test for the case `value is not null`, or does it exist already?

nit0906: @averma21  - yes there are multiple tests. 
testEqualityQuery_native - > tests simple equality condition
testEqualityInequalityCombined_native -> equality and inequality
testInequalityQuery_native -> tests inequality with nodes created with this property and also without this property.

So effectively the condition that x<>1 should return nodes where x is set and not equal to 1 is being tested. I hope this is what you were asking for - right ?

averma21: I was thinking of query on the lines of -
```
"select * from [nt:unstructured] where [x] is not null"
```
The execution flow of this query would also go through the changes done in PropertyValueImpl.java, right?

nit0906: Added here https://github.com/apache/jackrabbit-oak/pull/465/commits/988b46ea53eba0e523c34a319b1c914cf7422d57

averma21 (approved):

thomasmueller (approved):

PR #467, opened 2022-01-18
OAK-9670 : log at WARN level for fulltext query without index 

blackfor:
Fulltext queries cannot be handled by traversal, so we need to highlight prominently when a fulltext query has been issues but no appropriate index can be found.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
@@ -1201,9 +1201,9 @@ public class QueryImpl implements Query {
         }
 
         if (potentiallySlowTraversalQuery || bestIndex == null) {
-            // Log error for fulltext queries without index, since these cannot return results
+            // Log warning for fulltext queries without index, since these cannot return results
             if(!filter.getFulltextConditions().isEmpty()) { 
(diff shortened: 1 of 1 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -1201,7 +1201,13 @@ private SelectorExecutionPlan getBestSelectorExecutionPlan(
         }
 
         if (potentiallySlowTraversalQuery || bestIndex == null) {
-            LOG.debug("no proper index was found for filter {}", filter);
+            // Log error for fulltext queries without index, since these cannot return results
+            if(!filter.getFulltextConditions().isEmpty()) { 
+                LOG.error("Fulltext query without index for filter {}; no results will be returned", filter);
```
thomasmueller: Nit: I think it should be "WARN" to be in-line with other messages (e.g. "The query read more than...") -- I would use errors for data corruption and so on only.

thomasmueller (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
blackfor: I was unsure on this too; my thinking was that it should be an ERROR because the query will _never_ return results - so it's not that it's not performant, it's actually effectively failing.

PR #469, opened 2022-01-18
OAK-9671 Increase dynamicBoost and dynamicBoostLite full-text test co…

FrancoisZhang:
…verage

### DIFF
--- a/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
+++ b/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
@@ -25,6 +25,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Deque;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
@@ -1551,7 +1552,7 @@ public class LucenePropertyIndex extends FulltextIndex {
         if (FieldNames.FULLTEXT.equals(fieldName)) {
             Query subQuery = new BooleanQuery();
             if (pr.indexDefinition.isDynamicBoostLiteEnabled()) {
-                subQuery = new TermQuery(new Term(FieldNames.SIMILARITY_TAGS, text.toLowerCase()));
+                subQuery = getSimTagsQueryTerm(text);
             } else if (augmentor != null) {
                 subQuery = augmentor.getQueryTerm(text, analyzer, pr.indexDefinition.getDefinitionNodeState());
             }
@@ -1567,6 +1568,29 @@ public class LucenePropertyIndex extends FulltextIndex {
         }
 
         return ret;
+    }
+
+    private static Query getSimTagsQueryTerm(String text) {
+        if (text == null || text.equals("")) {
+            return null;
+        }
+        LOG.debug("getSimTagsQueryTerm Text: {}", text);
+
+        String[] termArr = text.replaceAll("\\\\", "").split("\\s+");
+        Set<String> terms = new HashSet<>(Arrays.asList(termArr));
+        LOG.debug("getSimTagsQueryTerm terms: {}", terms);
+        if (terms.size() > 10) {
+            LOG.debug("Not adding query terms as number of terms {} exceeds maximum permissible value of {}", terms.size(), 10);
+            return null;
+        }
+
+        BooleanQuery query = new BooleanQuery();
+        for (String fragment : terms) {
+            Term term = new Term(FieldNames.SIMILARITY_TAGS, fragment.toLowerCase());
+            query.add(new TermQuery(term), BooleanClause.Occur.SHOULD);
+            LOG.debug("Added simTags query term: {}", fragment.toLowerCase());
+        }
+        return query;
     }
 
     static Query tokenToQuery(String text, String fieldName, Analyzer analyzer) {
--- a/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
+++ b/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
@@ -1551,7 +1551,7 @@ public class LucenePropertyIndex extends FulltextIndex {
         if (FieldNames.FULLTEXT.equals(fieldName)) {
             Query subQuery = new BooleanQuery();
             if (pr.indexDefinition.isDynamicBoostLiteEnabled()) {
-                subQuery = new TermQuery(new Term(FieldNames.SIMILARITY_TAGS, text));
+                subQuery = new TermQuery(new Term(FieldNames.SIMILARITY_TAGS, text.toLowerCase()));
             } else if (augmentor != null) {
                 subQuery = augmentor.getQueryTerm(text, analyzer, pr.indexDefinition.getDefinitionNodeState());
             }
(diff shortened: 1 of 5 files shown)


oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/dynamicBoost/DynamicBoostTest.java
```diff
@@ -77,101 +77,280 @@ protected void createTestIndexNode() throws Exception {
...
+    // dynamic boost: query should be case insensitive
+    @Test
+    public void testQueryDynamicBoostCaseInsensitive() throws Exception {
+        createAssetsIndexAndProperties(false, false);
+        prepareTestAssets();
+
+        assertQuery("//element(*, dam:Asset)[jcr:contains(., 'FLOWER')]", XPATH, Arrays.asList("/test/asset1", "/test/asset2"));
+    }
+
+    @Ignore //todo: bug? wildcard doesn't work at all for dynamic boost term
```
fabriziofortino: this cannot work with standard dynamic boost since the terms are the field names. I would remove this test

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/dynamicBoost/DynamicBoostTest.java
```diff
@@ -77,101 +77,280 @@ protected void createTestIndexNode() throws Exception {
...
+        assertQuery("select [jcr:path] from [dam:Asset] where contains(*, 'blue flower')", SQL2,
+                Arrays.asList("/test/asset1", "/test/asset2", "/test/asset3"));
+    }
+
+    @Test
+    public void testQueryDynamicBoostOr() throws Exception {
+        createAssetsIndexAndProperties(false, false);
+        prepareTestAssets();
+
+        assertQuery("select [jcr:path] from [dam:Asset] where contains(*, 'blue OR coffee')", SQL2,
```
fabriziofortino: I guess an explicit `AND` would not work here too. This is the expected behavior though.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/dynamicBoost/DynamicBoostTest.java
```diff
@@ -77,101 +77,280 @@ protected void createTestIndexNode() throws Exception {
...
+                Arrays.asList("/test/asset2", "/test/asset3"));
+    }
+
+    @Ignore // todo: failed, minus doesn't work
```
fabriziofortino: I would remove this test too. Lucene syntax cannot be supported with dynamic boost.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/dynamicBoost/DynamicBoostTest.java
```diff
@@ -77,101 +77,280 @@ protected void createTestIndexNode() throws Exception {
...
+    }
+
+    @Test
+    public void testQueryMixOr() throws Exception {
+        createAssetsIndexAndProperties(false, false);
+        prepareTestAssets();
+        // explicit OR works as expected
+        assertQuery("//element(*, dam:Asset)[jcr:contains(., 'blue OR long')]", XPATH,
+                Arrays.asList("/test/asset1", "/test/asset2", "/test/asset3"));
+    }
```
fabriziofortino: I actually don't know how the latter works. We should investigate this.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/dynamicBoost/DynamicBoostTest.java
```diff
@@ -77,101 +77,280 @@ protected void createTestIndexNode() throws Exception {
...
+        assertEquals(
+                "[dam:Asset] as [a] /* lucene:test-index(/oak:index/test-index) :fulltext:plant simtags:plant ft:(\"plant\")\n  where contains([a].[*], 'plant') */",
+                explain("//element(*, dam:Asset)[jcr:contains(., 'plant')]", XPATH));
+
+        assertQuery("//element(*, dam:Asset)[jcr:contains(., 'plant')]", XPATH,
+                Arrays.asList("/test/asset1", "/test/asset2", "/test/asset3"));
+        assertQuery("//element(*, dam:Asset)[jcr:contains(., 'flower')]", XPATH, Arrays.asList("/test/asset1", "/test/asset2"));
+    }
+
+    @Ignore // dynamic boost lite: dynamic boost lite don't work as case insensitive like dynamic boost, todo: log a bug?
```
fabriziofortino: yes, this is a bug

fabriziofortino (approved):
Good work. I would remove the unit tests for the dynamicboost scenario we cannot support.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/dynamicBoost/DynamicBoostTest.java
FrancoisZhang: make sense, removed it

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/dynamicBoost/DynamicBoostTest.java
FrancoisZhang: make sense, removed it

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
```diff
@@ -1551,7 +1551,7 @@ private static Query tokenToQuery(String text, String fieldName, PlanResult pr,
         if (FieldNames.FULLTEXT.equals(fieldName)) {
             Query subQuery = new BooleanQuery();
             if (pr.indexDefinition.isDynamicBoostLiteEnabled()) {
-                subQuery = new TermQuery(new Term(FieldNames.SIMILARITY_TAGS, text));
+                subQuery = new TermQuery(new Term(FieldNames.SIMILARITY_TAGS, text.toLowerCase()));
```
FrancoisZhang: @fabriziofortino here is place I'd like to change for fixing case insensitive query over dynamicBoostLite

fabriziofortino: @FrancoisZhang that's a possibility but we are going to apply other possible transformations that the analyzer does.

Could you instead try to modify this code with:

```java
subQuery = tokenToQuery(text, FieldNames.SIMILARITY_TAGS, analyzer);
```

In this way, we use the lucene analyzer that applies a number of transformations. I think this was the main issue with dynamic boost lite (my bad). Could you try this and verify the other tests work? Thanks

FrancoisZhang: @fabriziofortino quick test seems doesn't work. can you check latest code? I use a new created method to make it work the same as dynamicBoost query. which is actually implemented in AEM DAM code base. new method I create here is try to mimic with that and make them work as the same way.

FrancoisZhang: @fabriziofortino  I verified it again, tokenToQuery does great to make it fully supporting the lucene fulltext query syntax. I made another commit to use that instead and added some tests to verify it's query ability. It should make more sense here. Only concern is it will make it doesn't match the behavior of DynamicBoost in Lucene and Elastic. More like a product decision here though.

fabriziofortino (approved):

fabriziofortino:
Thanks @FrancoisZhang 🍟 🍺

PR #470, opened 2022-01-19
OAK-9672 - Adding checks to support setting null values in oak-commons JSON

klcodanr:
Currently performing the following will result in a NullPointerException as null values for the properties or child objects are not checked:

```
  JsonObject a = new JsonObject(true);
  a.getChildren().put("test", (JsonObject) null);
```

This PR introduces checks to ensure null children and null properties can be set on a JsonObject without the toString method failing.

### DIFF
--- a/oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsonObject.java
+++ b/oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsonObject.java
@@ -146,15 +146,19 @@ public class JsonObject {
     }
 
     private static void toJson(JsopBuilder buf, JsonObject obj) {
-        buf.object();
-        for (String name : obj.props.keySet()) {
-            buf.key(name).encodedValue(obj.props.get(name));
+        if(obj == null){
+            buf.value(null);
+        } else {
+            buf.object();
+            for (String name : obj.props.keySet()) {
+                buf.key(name).encodedValue(obj.props.get(name));
+            }
+            for (String name : obj.children.keySet()) {
+                buf.key(name);
+                toJson(buf, obj.children.get(name));
+            }
+            buf.endObject();
         }
-        for (String name : obj.children.keySet()) {
-            buf.key(name);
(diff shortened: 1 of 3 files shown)


oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsonObject.java
```diff
@@ -146,15 +146,19 @@ public String toString() {
     }
 
     private static void toJson(JsopBuilder buf, JsonObject obj) {
-        buf.object();
-        for (String name : obj.props.keySet()) {
-            buf.key(name).encodedValue(obj.props.get(name));
+        if(obj == null){
```
reschke: broken whitespace

reschke (changes requested):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsonObject.java
reschke:
```suggestion
        if (obj == null) {
```

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsopBuilder.java
```diff
@@ -208,12 +210,20 @@ public JsopBuilder value(String value) {
...
         buff.append(value);
         needComma = true;
         return this;
     }
 
+    private int strLength(@Nullable String value) {
+        if (value != null) {
+            return value.length();
+        } else {
+            return 4;// null = 4 chars
```
thomasmueller: Suggested change:

            return "null".length();

thomasmueller (changes requested):

klcodanr:
@reschke  / @thomasmueller  I believe I have addressed the concerns raised, could you please re:review?

reschke (approved):

thomasmueller (approved):

PR #471, opened 2022-01-24
OAK-9675 allow authorizable properties defined by a mixin

enapps-enorman:
Adds a configuration option for allowed authorizable properties mixin types

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/AuthorizablePropertiesImpl.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/AuthorizablePropertiesImpl.java
@@ -20,10 +20,12 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
+import java.util.stream.Stream;
 
 import javax.jcr.RepositoryException;
 import javax.jcr.Value;
 import javax.jcr.nodetype.ConstraintViolationException;
+import javax.jcr.nodetype.NodeType;
 import javax.jcr.nodetype.PropertyDefinition;
 
 import org.apache.jackrabbit.JcrConstants;
@@ -234,9 +236,34 @@ class AuthorizablePropertiesImpl implements AuthorizableProperties {
         }
         ReadOnlyNodeTypeManager nodeTypeManager = authorizable.getUserManager().getNodeTypeManager();
         PropertyDefinition def = nodeTypeManager.getDefinition(parent, property, true);
-        if (def.isProtected() || (authorizablePath.equals(parent.getPath())
-                && !def.getDeclaringNodeType().isNodeType(UserConstants.NT_REP_AUTHORIZABLE))) {
+        if (def.isProtected()) {
+            // exclude all protected properties
             return null;
+        } else if (authorizablePath.equals(parent.getPath())) {
+            // non-protected properties on the root must be defined by the expected
+            //  primary type or one of the configured mixin types
+            Boolean allowed = null;
+            NodeType declaringNodeType = def.getDeclaringNodeType();
+            if (declaringNodeType.isNodeType(UserConstants.NT_REP_AUTHORIZABLE)) {
+                // defined by the expected primary type so allowed
+                allowed = Boolean.TRUE;
+            } else {
+                // OAK-9675 - not declared by the primary type, so let's check for a mixin
+                if (declaringNodeType.isMixin()) {
+                    // check if the property is defined by any of the configured mixin type
+                    String[] allowedMixins = authorizable.getUserManager().getConfig().getConfigValue(UserConstants.PARAM_AUTHORIZABLE_PROPERTIES_MIXINS, null, String[].class);
+                    if (allowedMixins != null) {
+                        // check if the actual mixin name is one of the configured items
+                        String actualMixinName = declaringNodeType.getName();
+                        allowed = Stream.of(allowedMixins).anyMatch(actualMixinName::equals);
+                    }
+                }
+            }
+
+            if (!Boolean.TRUE.equals(allowed)) {
+                // not defined by an allowed node type
+                return null;
+            }
         } // else: non-protected property somewhere in the subtree of the user tree.
 
         return property;
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/AuthorizablePropertiesImpl.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/AuthorizablePropertiesImpl.java
@@ -242,11 +242,11 @@ class AuthorizablePropertiesImpl implements AuthorizableProperties {
         } else if (authorizablePath.equals(parent.getPath())) {
             // non-protected properties on the root must be defined by the expected
             //  primary type or one of the configured mixin types
-            Boolean allowed = null;
+            boolean allowed = false;
             NodeType declaringNodeType = def.getDeclaringNodeType();
             if (declaringNodeType.isNodeType(UserConstants.NT_REP_AUTHORIZABLE)) {
                 // defined by the expected primary type so allowed
-                allowed = Boolean.TRUE;
+                allowed = true;
             } else {
                 // OAK-9675 - not declared by the primary type, so let's check for a mixin
                 if (declaringNodeType.isMixin()) {
@@ -260,7 +260,7 @@ class AuthorizablePropertiesImpl implements AuthorizableProperties {
                 }
             }
 
-            if (!Boolean.TRUE.equals(allowed)) {
+            if (!allowed) {
                 // not defined by an allowed node type
                 return null;
             }
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserConfigurationImpl.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserConfigurationImpl.java
@@ -176,6 +176,13 @@ public class UserConfigurationImpl extends ConfigurationBase implements UserConf
                 name = "RFC7613 Username Comparison Profile",
                 description = "Enable the UsercaseMappedProfile defined in RFC7613 for username comparison.")
         boolean enableRFC7613UsercaseMappedProfile() default false;
+
+        // for OAK-9675
+        @AttributeDefinition(
+                name = "Authorizable Properties Mixin Types",
+                description = "Optional configuration defining the names of additional " +
+                        " mixin types that are allowed to define authorizable properties.")
+        String[] authorizablePropertiesMixins();
     }
 
     private static final UserAuthenticationFactory DEFAULT_AUTH_FACTORY = new UserAuthenticationFactoryImpl();
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserConfigurationImpl.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserConfigurationImpl.java
@@ -180,7 +180,7 @@ public class UserConfigurationImpl extends ConfigurationBase implements UserConf
         // for OAK-9675
         @AttributeDefinition(
                 name = "Authorizable Properties Mixin Types",
-                description = "Optional configuration defining the names of additional " +
+                description = "Optional configuration defining the names of any " +
                         " mixin types that are allowed to define authorizable properties.")
         String[] authorizablePropertiesMixins();
     }
--- a/oak-core/src/test/java/org/apache/jackrabbit/oak/security/user/AuthorizablePropertiesDefinedByConfiguredMixinTest.java
+++ b/oak-core/src/test/java/org/apache/jackrabbit/oak/security/user/AuthorizablePropertiesDefinedByConfiguredMixinTest.java
@@ -1,0 +1,129 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.jackrabbit.oak.security.user;
+
+import static org.apache.jackrabbit.oak.spi.nodetype.NodeTypeConstants.NODE_TYPES_PATH;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+import java.util.Iterator;
+
+import javax.jcr.RepositoryException;
+import javax.jcr.Value;
+import javax.jcr.ValueFactory;
+
+import org.apache.jackrabbit.JcrConstants;
+import org.apache.jackrabbit.api.security.user.User;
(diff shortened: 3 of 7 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/AuthorizablePropertiesImpl.java
```diff
@@ -234,9 +236,34 @@ private PropertyState getAuthorizableProperty(@NotNull Tree authorizableTree, @N
...
         PropertyDefinition def = nodeTypeManager.getDefinition(parent, property, true);
-        if (def.isProtected() || (authorizablePath.equals(parent.getPath())
-                && !def.getDeclaringNodeType().isNodeType(UserConstants.NT_REP_AUTHORIZABLE))) {
+        if (def.isProtected()) {
+            // exclude all protected properties
             return null;
+        } else if (authorizablePath.equals(parent.getPath())) {
+            // non-protected properties on the root must be defined by the expected
+            //  primary type or one of the configured mixin types
+            Boolean allowed = null;
```
anchela: i would change that to 
`boolean allowed = false;`
and then set it to `allowed = true` as i don't see any benefit of a 'null' value.
what was the reason for this?

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/AuthorizablePropertiesImpl.java
```diff
@@ -234,9 +236,34 @@ private PropertyState getAuthorizableProperty(@NotNull Tree authorizableTree, @N
...
+            NodeType declaringNodeType = def.getDeclaringNodeType();
+            if (declaringNodeType.isNodeType(UserConstants.NT_REP_AUTHORIZABLE)) {
+                // defined by the expected primary type so allowed
+                allowed = Boolean.TRUE;
```
anchela: my understanding of the original improvement request is that allowing all types of properties is too permissive the the intention is to limit what properties can be stored below a given user/group node.

see the following sentence from the original report:
> This is in support of a use case where we want a stricter constraint on what is allowed to be stored in an authorizable property. 

given that `rep:Authorizables `allows for any residiual property of any type, single and multi-valued, this seems to defeat the purpose. my understanding of the request is that _only_ properties defined by the configured mixin types should be allow instead of allow any residual property as it would still be the case here.

as far as i can see this requirement isn't covered by the tests either.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/AuthorizablePropertiesImpl.java
```diff
@@ -234,9 +236,34 @@ private PropertyState getAuthorizableProperty(@NotNull Tree authorizableTree, @N
...
+                    String[] allowedMixins = authorizable.getUserManager().getConfig().getConfigValue(UserConstants.PARAM_AUTHORIZABLE_PROPERTIES_MIXINS, null, String[].class);
+                    if (allowedMixins != null) {
+                        // check if the actual mixin name is one of the configured items
+                        String actualMixinName = declaringNodeType.getName();
+                        allowed = Stream.of(allowedMixins).anyMatch(actualMixinName::equals);
+                    }
+                }
+            }
+
+            if (!Boolean.TRUE.equals(allowed)) {
```
anchela: this should just be 
`if (!allowed)`

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserConfigurationImpl.java
```diff
@@ -176,6 +176,13 @@
                 name = "RFC7613 Username Comparison Profile",
                 description = "Enable the UsercaseMappedProfile defined in RFC7613 for username comparison.")
         boolean enableRFC7613UsercaseMappedProfile() default false;
+
+        // for OAK-9675
+        @AttributeDefinition(
+                name = "Authorizable Properties Mixin Types",
+                description = "Optional configuration defining the names of additional " +
```
anchela: the 'additional' strikes me here..... my reading of the improvement request (as pointed out before) would be to limit authorizable properties to those defined by the configured mixin types and not allow them 'in addition'.

as by this comment:
`The proposed improvement here is to add an optional configuration property that would define the names of mixin types that are allowed to define authorizable properties.  Any property definition defined by a mixin type in this set would be included, and anything else would be excluded as before`.

anchela (changes requested):
the patch IMHO doesn't cover the required use case which was that authorizable properties are too permissive today and that the goal was to limit it to dedicated mixin types.

also, the patch only changes the get of authorizable properties and not the set, imho the same limitations should be applied for the set. in addition we should discuss if/how the configured mixin types should be automatically added to the user node or the properties node in the subtree in order to make sure it is present without having to explicitly obtain the Node beforehand. this in particular applies to sequences of 'setProperty' with relative path and subsequent 'getProperty'.

finally, there should be additional test for set-get sequences, setting/getting without mixin set (-> should not work), setting/getting with relative paths that store properties somewhere in the subtree and not directly below the user/group node.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/AuthorizablePropertiesImpl.java
enapps-enorman: Sure, I'll make that cleanup change as the current impl only needs true or false.  This code was leftover from a failed experiment where a third state would have been relevant.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/AuthorizablePropertiesImpl.java
enapps-enorman: ok

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserConfigurationImpl.java
enapps-enorman: Fair enough, I'll change "additional" to "any" to provide better clarity.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/AuthorizablePropertiesImpl.java
enapps-enorman: Yes, perhaps I didn't explain that well enough.  This proposal is mostly about reading authorizable properties.  I was able to deal with the setting of those authorizable properties using ACLs and already existing (and custom) code.

Basically my intention is to utilize ACLs to limit the setting of properties.  Essentially, I have re-configured the DefaultAuthorizableActionProvider (as below) so each user is only allowed "rep:alterProperties" rights on their home folder.  So they would only be able to update properties that already exist.  Then I have a custom AuthorizableActionProvider component that applies additional actions that will add the appropriate mixin type to the home folders and populates initial values for the properties that the end user should have access to alter.

For example, via OSGi configuration like this:
`        "org.apache.jackrabbit.oak.spi.security.user.action.DefaultAuthorizableActionProvider":{
            "userPrivilegeNames":[
                "jcr:read", 
                "rep:alterProperties"
            ],
            "groupPrivilegeNames":[
                "jcr:read"
            ],
            "enabledActions":[
                "org.apache.jackrabbit.oak.spi.security.user.action.AccessControlAction"
            ]
        }
`

Do you see any flaw with that approach?

enapps-enorman:
> also, the patch only changes the get of authorizable properties and not the set, imho the same limitations should be applied for the set. in addition we should discuss if/how the configured mixin types should be automatically added to the user node or the properties node in the subtree in order to make sure it is present without having to explicitly obtain the Node beforehand. this in particular applies to sequences of 'setProperty' with relative path and subsequent 'getProperty'.

Please see my related comment at https://github.com/apache/jackrabbit-oak/pull/471#discussion_r794754078

I'm not sure it would be practical to try to automatically add a mixin type during the setProperty calls, as there could be multiple choices on which mixin to use.  This is why I prefer to add the mixin to the home node in a custom ActionProvider when the user/group is created.

Also this proposal doesn't change any behavior for authorizable properties defined for relative subnodes under the home node.  Access to mixin defined properties on the subnodes wasn't restricted like was for properties on the root home folder.

> 
> finally, there should be additional test for set-get sequences, setting/getting without mixin set (-> should not work), setting/getting with relative paths that store properties somewhere in the subtree and not directly below the user/group node.

FYI: getting a mixin defined property without this fix was already covered by [AuthorizablePropertiesImplTest#testGetPropertyDeclaredByDifferentType](https://github.com/apache/jackrabbit-oak/blob/c424e1c179092594c365034c7b3f6eceb6cf2b42/oak-core/src/test/java/org/apache/jackrabbit/oak/security/user/AuthorizablePropertiesImplTest.java#L192)

Also, I didn't intend to suggest that the properties defined by the mixin were the only properties that could be set, only that those properties should not be excluded from the get calls.

enapps-enorman:
@anchela I've committed in some cleanup code changes based on your feedback.  I'll wait for your responses to my other replies regarding if any additional changes are required.

anchela:
@enapps-enorman , thanks for your comments. the problem is see is that authorizable properties are not limited to the user/group node. the API allows to set and later read them from any subtree defined below a given user/group node but the structure of those nodes is an implementation detail. so, i was looking for a test that would reflect what happens when properties are set with a relative path and how the mixin types come to play in that scenario (without the application having to read those subtree nodes and apply the mixins).

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/AuthorizablePropertiesImpl.java
anchela: @enapps-enorman , let me try to rephrase to make sure i get your use case right:
- you want to make sure that a user can only modify and read the properties stored in in the subtree of it's home node
- upon user/group creation you create a set of properties
- but in addition you enforce particular types/values of those properties by adding mixins to the user/group node
- however, properties defined by those mixins will not be returned by Authorizable.getProperty.

so, your use-case is not about limiting the set of properties that can be written by adding mixins but rather making sure those defined by mixins are also returned as authorizable properties.
is that correct?

enapps-enorman:
@anchela The mixin handling proposed here should not have any impact at all on any state stored under a subtree of the user/group node.  If you look at AuthorizablePropertiesImpl line 242 it is checking if the path equals the user/group home path before doing any of the new mixin checking and handing.  So all of the behavior for those subtree scenarios is the same as it was before.  I believe that the subtree handling has always included all the properties that exist there regardless of the mixin types or primary type of the subtree nodes.  If I am not mistaken, the properties stored on the main home folder was always the only one that was checking where the property definition came from.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/AuthorizablePropertiesImpl.java
enapps-enorman: @anchela Yes, that is basically correct.  My proposal is not really about limiting the set of properties that can be written, I simply wish to have a more specific value constraint defined for certain property definitions (as defined by a mixin type that I have assigned).  And I don't want the existence of that more specific property definition (with the value constraint) to make the property invisible from the Authorizable.getProperty call.

If you are curious, the more specific value constraint idea is motivated by security concerns about ensuring that self-registered end users of the platform can not abuse their self-managed user profile properties to store invalid or unreasonably long values to the repository.  For an extreme example, we don't want the user to do something like encoding a large binary file as text and storing that value as the "nickname" property in their user profile.  We want some reasonable max length (or other) value constraint on the property definition so extremely invalid values will never be stored even if the end user somehow manages to get around the form post validation checking.

anchela: @enapps-enorman , thanks for the insight.... and yes.... i am always curious to understand the reason behind a feature request or a patch. so thanks..... and it helps.

i am still wondering if it would make sense to automatically add the configured mixin types to the node that holds the authorizable-properties. that would avoid that developers have to make some assumptions about the way user properties are being stored in the repository.... in particular if the properties are stored with an relative path.... because i guess you would not want such huge binaries to be uploaded to <user-node>/some/subtree either, would you?

anchela:
@enapps-enorman , i have to check the code again regarding the handling in the subtree because i thought it was different. my point is that properties somewhere in the subtree should have the same mixin-type checks.
so, large-binary-constraint should also be applied in the subtree and the application writing the properties should not have to care about the internal structure used.... that's why i would suggest to automatically add the mixin(s) when creating a authorizable-properties node.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/AuthorizablePropertiesImpl.java
enapps-enorman: Yes, it is true that we don't want invalid values in the subtree either.  However, I believe that scenario could already be handled with the previously existing code via ACLs (by denying the rights to create children) and/or via pre-creating the possible subtree nodes with specific primary types that defines what properties are allowed.  Only the properties of the root home folder could not be handled that way which is why I proposed this enhancement.

In fact the above is exactly what we are trying to do with a custom AuthorizableActionProvider component that pre-creates a "private" subfolder with a custom primary type (and an appropriate ACL) whenever a new user is created.

enapps-enorman:
> i have to check the code again regarding the handling in the subtree because i thought it was different. my point is that properties somewhere in the subtree should have the same mixin-type checks.

@anchela Well, I don't think we can assume that every subtree node would need the same mixin type value.  Each subtree node would likely have a unique set of possible properties and a different set of value constraints on those properties.  Wouldn't handing this at the oak level become complex if you have to map specific mixintype (or primarytype) to subtree paths?  

We were using a custom AuthorizableActionProvider component to pre-create the subtree nodes with the right types and appropriate ACLs when the user/group are created to avoid this complexity.

FYI: I am working on a set of unit tests to demonstrate what we have been doing.  I'll try to finish that up tomorrow and add it to this PR.  Perhaps seeing it in practice will help to evaluate the proposal so we can reach a resolution.

anchela:
@enapps-enorman , one more comment regarding

> I'm not sure it would be practical to try to automatically add a mixin type during the setProperty calls, as there could be multiple choices on which mixin to use. This is why I prefer to add the mixin to the home node in a custom ActionProvider when the user/group is created.

it's perfectly fine if you add the mixin(s) as part of an authorizable-action for your use-case, which apparently only stores properties directly below the user node.
however, this improvement should not be limited to your use-case but should work on an general level. instead of trying to add the 'right' mixin I would therefore suggest to add all configured mixin types.
we may also consider limiting the configuration option by setting a cardinality (2 or 5) that would prevent the number of mixins added to be unnecessary big. after all one can always leverage supertypes if there is the need for more.

enapps-enorman:
> @enapps-enorman , i have to check the code again regarding the handling in the subtree because i thought it was different. 

Please do check the code, because I am fairly certain that it works as I described.

> that's why i would suggest to automatically add the mixin(s) when creating a authorizable-properties node.

To me this seems like you would just be guessing about what the intent is, which I don't think is a good solution.  For example, you may want a different mixin for users vs groups.   Or you may want a different mixin for users who belong to a certain organization.

enapps-enorman:
> which apparently only stores properties directly below the user node.

This is not true, we do use properties stored under subnodes as well.  We use a custom primary type and acls to control the read/write access to the subnodes.  None of the subnode handling code required the changes proposed here as that part already worked ok.

Anyways, I have committed the additional tests to demonstrate what we are doing at [4b18b72](https://github.com/apache/jackrabbit-oak/pull/471/commits/4b18b728f1470bc403484c8b9c4965e94120a05b) for you to review and consider.

Can we please evaluate this PR on the narrow goal of making the (non subnode) authorizable properties that are defined by a configured mixin readable as authorizable properties?  All the other topics you keep bringing up seem to be out of scope and I would hope that those could be discussed and considered separately.

Trying to get this PR accepted has been frustrating and I've already spent more time on this than I wanted to, so I would appreciate an answer as to whether the PR will be merged or not.

enapps-enorman:
Since there is no forward progress in getting this approved, I'm withdrawing this contribution from consideration.  Please don't use any of the code from this PR in your product.

PR #476, opened 2022-01-29
OAK-9651: Protection against very large queries (#446)

andrewmkhoury:
Backport to 1.22 branch - OAK-9651: Protection against very large queries


nit0906 (approved):
lgtm

PR #478, opened 2022-01-31
fix links to javadoc

kwin:


mreutegg (approved):
Looks good to me. 

Though, I still like https://jackrabbit.apache.org/jcr/jcr-api.html as a starting point and overview for the JCR and Jackrabbit API.

kwin:
@mreutegg Should we use  https://jackrabbit.apache.org/jcr/jcr-api.html as target for the JCR 2.0 menu entry instead?

reschke (changes requested):
I have to say that I find the changes that were already applied to "https://jackrabbit.apache.org/jcr/jcr-api.html" very confusing.

They confuse "Jackrabbit API" (as the API exposed by all Jackrabbit modules) with the "Jackrabbit JCR Extension API".  Please back these out and just add a note that that specific module has moved.

kwin:
> They confuse "Jackrabbit API" (as the API exposed by all Jackrabbit modules) with the "Jackrabbit JCR Extension API".

Please clarify or point me to a documentation where I can read that up?
Actually explicitly mentioning Jackrabbit API is beneficial on that page as that is the only part which is also implemented by Oak IIUC.

reschke:
I'm not sure what you're asking for.

This web page is and always was about the API exposed by Jackrabbit (not Oak). The module "jackrabbit-api" is only part of that. It's good to point out that it has moved somewhere else, but that's it. Putting a link to Oak's API docs asif it described the "Jackrabbit API" is completely confusing.

Ok, I'll fix that myself.

oak-doc/src/site/markdown/oak_api/overview.md
```diff
@@ -15,12 +15,13 @@
    limitations under the License.
   -->
 
-Oak API
+Oak and Jackrabbit API
```
reschke, line 18: I would leave the title alone.

oak-doc/src/site/site.xml
```diff
@@ -37,8 +37,7 @@ under the License.
     </menu>
     <menu name="Main APIs">
       <item href="https://s.apache.org/jcr-2.0-spec/index.html" name="JCR API" />
-      <item href="https://jackrabbit.apache.org/jcr/jcr-api.html" name="Jackrabbit API" />
-      <item href="oak_api/overview.html" name="Oak API" />
+      <item href="oak_api/overview.html" name="Oak/Jackrabbit API" />
     </menu>
```
reschke, line 41: Same here; dropping the reference to the Jackrabbit API page is ok, but claiming that the Oak page defines the "Jackrabbit API" is not.

reschke (changes requested):

oak-doc/src/site/site.xml
kwin: >  but claiming that the Oak page defines the "Jackrabbit API" is not.

I don't think I do that. It is just that the Jackrabbit API is now being maintained at Oak and if you look for the most recent documentation you find it there and no longer in "JR2". But feel free to propose something else.

reschke: "Jackrabbit API" is IMHO Jackrabbit's API. Just as "Oak API" is Oak's API. Both are different from the "Jackrabbit JCR API extensions" (which moved from Jackrabbit to Oak).

It probably would be best to have a totally separate doc page about that API.

kwin: `jackrabbit-(oak)-api` is the artifact id and its name is "Jackrabbit API" (https://github.com/apache/jackrabbit-oak/blob/85c7f002219f0c8435e4bdacbb8fe56dcf5efe83/oak-jackrabbit-api/pom.xml#L30). I never heard of "Jackrabbit JCR API extensions" but I do agree that Jackrabbit exposes API apart from Jackrabbit API which makes it a bit confusing. BTW Oak also has a dedicated API bundle which is only part of the API exposed by Oak (https://github.com/apache/jackrabbit-oak/tree/trunk/oak-api).
I would strongly suggest to keep the original meaning of "Jackrabbit API" (i.e. the part which moved to Oak) and to use another name for the full API exposed by Jackrabbit.

PR #481, opened 2022-02-02
OAK-9683 Bump elasticsearch version to 7.16.3

AngelaFabregues:
The following tests are not yet successfully executed:

ElasticIndexProviderServieTest.defaultSetup
ElasticSimilarQuesryTest.vectorSimilarityElastiknnIndexConfiguration  <--- expected angular but was cosine.


oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
```diff
@@ -36,7 +36,7 @@
 import org.elasticsearch.client.RequestOptions;
```
fabriziofortino, line 36: at line 148, `BulkProcessor.builder` is now deprecated. You just need to use the overloaded method that takes the name as the last parameter. You could use `this.indexName + "-bulk-processor"`

fabriziofortino (changes requested):
Good work overall!

There are a few changes needed:

- `oak-run-elastic` does not compile because of a size enforcement in the pom. It seems the new Elastic version it's a bit bigger. You need to increase the enforcer from 98MB to 102MB https://github.com/apache/jackrabbit-oak/blob/trunk/oak-run-elastic/pom.xml#L36-L40
- the elastiknn test fails because there was a name change from `angular` to `cosine` in a [recent version](https://github.com/alexklibisz/elastiknn/releases/tag/7.13.3.1). Just change the expected value to `cosine`.
- the `ElasticIndexProviderServiceTest` fails because something has changed in the Elastic client exception handling: a RuntimeException is returned instead of an IOException. Just change the type to `Exception` here https://github.com/apache/jackrabbit-oak/blob/trunk/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnection.java#L129

fabriziofortino (approved):
LGTM 👍

PR #483, opened 2022-02-07
OAK-9681 oak-doc-railroad-macro build broken by h2db update

thomasmueller:
A quick fix... I didn't verify yet whether it doesn't break the documentation; I will do that now.


thomasmueller:
Ah sorry it contains unrelated changes (not sure why)

PR #484, opened 2022-02-07
OAK-9681 oak-doc-railroad-macro build broken by h2db update

thomasmueller:
I now tested "mvn site" and it seems to work properly


reschke (approved):
I'll trust you :-)

PR #485, opened 2022-02-08
OAK-9612 Convert ReadOnlyBuilder's exceptions into RepositoryExcept…

joerghoh:
…ions


joerghoh:
(when merging please adjust the jira ID in the commit message ... of course it's OAK-9612!)

oak-store-spi/src/main/java/org/apache/jackrabbit/oak/spi/state/ReadyOnlyBuilderUnsupportedOperationException.java
```diff
@@ -0,0 +1,12 @@
+package org.apache.jackrabbit.oak.spi.state;
+
+/**
+ * Indicates that a modification operation was tried to execute on a read-only builder.
+ * It should be used instead of throwing plain UnsupportedOperationExceptions in that situation.
+ */
+public class ReadyOnlyBuilderUnsupportedOperationException extends UnsupportedOperationException {
```
anchela: Maybe `ReadyOnlyBuilderException` would be sufficiently clear and a bit shorter than `ReadyOnlyBuilderUnsupportedOperationException`

anchela (approved):
that patch looks good to me.
i would consider renaming the exception to something slightly shorter though.... but not a requirement.

oak-store-spi/src/main/java/org/apache/jackrabbit/oak/spi/state/ReadyOnlyBuilderUnsupportedOperationException.java
joerghoh: done

oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/IndexUtils.java
```diff
@@ -142,7 +142,7 @@ public static void checkLibsIsReadOnly(Persistence p) throws RepositoryException
         try {
             libsNode.addNode("illegal");
             Assert.fail();
-        } catch (UnsupportedOperationException e) {
+        } catch (RepositoryException e) {
```
mreutegg, line 145: I think it would be better if the test also check if the cause of the RepositoryException is a ReadOnlyBuilderException.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/SessionDelegate.java
```diff
@@ -212,6 +213,8 @@ public void refreshAtNextAccess() {
             } finally {
                 postPerform(sessionOperation, t0);
             }
+        } catch (ReadyOnlyBuilderException e) {
+            throw new RepositoryException(e);
```
mreutegg: What is the rationale behind throwing a RepositoryException vs ConstraintViolationException?

I'm not saying it's wrong. Just want to understand the reason why one is used instead of the other.

mreutegg (changes requested):
There should be some tests in oak-jcr that exercise the new catch blocks.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/SessionDelegate.java
joerghoh: We actually discussed it in the OAK-9612 to throw a ```ConstraintViolationException```. Will change it.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/IndexUtils.java
joerghoh: That is an implementation detail of oak-jcr which I don't want to leak into into oak-lucene.

joerghoh:
@mreutegg I added additional test coverage and changed the exception as indicated.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/IndexUtils.java
mreutegg: Fair enough. Though, the test did specifically check for just this implementation detail that an UnsupportedOperationException is thrown when an attempt is made to add a node on a read-only mount.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/delegate/SessionDelegateTest.java
```diff
@@ -68,4 +74,38 @@ public void testRefreshUnaware() {
         delegate.refresh(false);
         verify(pp, times(6)).refresh();
     }
+    
+    @Test(expected = RepositoryException.class)
```
mreutegg: Shouldn't these tests now expect a `ConstraintViolationException`?

oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/IndexUtils.java
mreutegg, line 145: Catch `ConstraintViolationException` instead?

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/delegate/SessionDelegateTest.java
joerghoh: The contract states just a ```RepositoryException```, so it makes sense just to test for that.
Although one could argue, that a change in that code, which also throws a ```RepositoryException``` (or an exception inheriting from it) would be caught as well.

As this is the test for this implementation,I will change it to the ```ConstraintViolationException```.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/IndexUtils.java
joerghoh: Won't change it here. This test is not responsible to test the implementation of the JCR, so it should just rely on the API contract.

mreutegg: > This test is not responsible to test the implementation of the JCR, so it should just rely on the API contract.

Well, currently the test does check an implementation detail. It catches a UnsupportedOperationException. But you are right the method does not `UnsupportedOperationException` nor `ConstraintViolationException`.

mreutegg (approved):

joerghoh:
@mreutegg / @anchela:  I don't have enough karma/permissions to merge this PR, can you do it for me?

PR #486, opened 2022-02-08
OAK-9680 Add support for Azure SAS URIs to oak-segment-azure

jelmini:
Add support for SAS URIs to utility code.


smiroslav:
Looks good to me. 

As a minor comment, not sure that introducing Environment class brings obvious benefits right now.

smiroslav (approved):

jelmini:
@smiroslav I had to introduce the `Environment` class to be able to test the code. Mockito cannot mock `System.getenv()` because it's a system class. Alternatively, it's possible to hack a workaround using reflection to obtain the underlying mutable Map from `System.getenv()`. This works with Java up until 15 with a big warning, but stops working from Java 16 forward (because "illegal reflective access" really becomes illegal). 
Is there another approach for testing this code you could suggest?

smiroslav:
Thanks @jelmini for clarification. There are some test libraries that can inject values for system properties, but I like your approach better.

PR #489, opened 2022-02-10
OAK-9691 Improve fulltext query syntax support for ElasticSearch

FrancoisZhang:


oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/dynamicBoost/DynamicBoostTest.java
```diff
@@ -147,6 +146,14 @@ public void testQueryFullTextWithTitle() throws Exception {
         // wildcard works
         assertQuery("//element(*, dam:Asset)[jcr:contains(., 'title*')]", XPATH,
                 Arrays.asList("/test/asset1", "/test/asset2", "/test/asset3"));
+
```
fabriziofortino: not sure if the whole `testQueryFullTextWithTitle` should stay in this class since it does not test dynamic boost.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/dynamicBoost/DynamicBoostTest.java
```diff
@@ -224,7 +231,12 @@ public void testQueryDynamicBoostLiteWildcard() throws Exception {
         createAssetsIndexAndProperties(true, true);
         prepareTestAssets();
 
-        assertQuery("select [jcr:path] from [dam:Asset] where contains(*, 'blu*')", SQL2, Arrays.asList("/test/asset3"));
+        assertQuery("select [jcr:path] from [dam:Asset] where contains(*, 'blu?')", SQL2, Arrays.asList("/test/asset3"));
+        assertQuery("select [jcr:path] from [dam:Asset] where contains(*, 'bl?e')", SQL2, Arrays.asList("/test/asset3"));
+        assertQuery("select [jcr:path] from [dam:Asset] where contains(*, '?lue')", SQL2, Arrays.asList("/test/asset3"));
+        assertQuery("select [jcr:path] from [dam:Asset] where contains(*, 'coff*')", SQL2, Arrays.asList("/test/asset2"));
+        assertQuery("select [jcr:path] from [dam:Asset] where contains(*, 'co*ee')", SQL2, Arrays.asList("/test/asset2"));
+        assertQuery("select [jcr:path] from [dam:Asset] where contains(*, '*ffee')", SQL2, Arrays.asList("/test/asset2"));
```
fabriziofortino, line 210: nice test set 👍

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticFullTextAsyncTest.java
```diff
@@ -264,16 +259,104 @@ public void fulltextWithModifiedNodeScopeIndex() throws Exception {
...
+            assertQuery("//*[jcr:contains(., 'hello world')] ", XPATH, Arrays.asList("/test/nodeb", "/test/nodec"));
+
+            // exclude
+            assertQuery("//*[jcr:contains(., 'hello -world')] ", XPATH, Arrays.asList("/test/nodea"));
+
+            // explicit OR
+            assertQuery("//*[jcr:contains(., 'ocean OR world')] ", XPATH, Arrays.asList("/test/nodea", "/test/nodeb", "/test/nodec"));
+
+            // analyzed field
+            assertQuery("//*[jcr:contains(., 'soccer')] ", XPATH, Arrays.asList("/test/nodeb"));
```
fabriziofortino: can we add a few queries using the lucene syntax on a single analyzed query? eg `//*[jcr:contains(a, 'Hell*')]`

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
```diff
@@ -779,41 +779,13 @@ private static QueryBuilder referenceConstraint(String uuid) {
...
-            }
-
-            if (allowLeadingWildCard) {
-                return queryStringQuery(text).field(fieldName).defaultOperator(Operator.AND);
-            } else {
-                return simpleQueryStringQuery(text).analyzeWildcard(true).field(fieldName).defaultOperator(Operator.AND);
-            }
+        LOG.debug("fullTextQuery for text: '{}', fieldName: '{}'", text, fieldName);
+        QueryStringQueryBuilder queryStringQueryBuilder = queryStringQuery(text).defaultOperator(Operator.AND)
+                .type(MultiMatchQueryBuilder.Type.CROSS_FIELDS);
```
fabriziofortino: I think I have discovered the reason some tests in Elastic were not successful. In Lucene, we rewrite the query escaping some reserved chars (including `/`, `|`, `&`). Without this Lucene queries would be exactly the same as Elastic query_string queries.

If we use the same rewrite function (the modifier needs to be widened from `protected` to `public`) we get the same results. Changes in `oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/IndexQueryCommonTest.java ` can be reverted.

```suggestion
        QueryStringQueryBuilder queryStringQueryBuilder = queryStringQuery(FulltextIndex.rewriteQueryText(text))
                .defaultOperator(Operator.AND)
                .type(MultiMatchQueryBuilder.Type.CROSS_FIELDS);
```

fabriziofortino (changes requested):

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticFullTextAsyncTest.java
```diff
@@ -264,16 +259,104 @@ public void fulltextWithModifiedNodeScopeIndex() throws Exception {
...
+        assertEventually(() -> assertQuery("//*[jcr:contains(@analyzed_field, 'SUN.JPG')] ", XPATH, Collections.singletonList("/test/a")));
 
         // add nodeScopeIndex at a later stage
-        index.getChild("indexRules").getChild("nt:base").getChild("properties")
-                .getChild("analyzed_field").setProperty(FulltextIndexConstants.PROP_NODE_SCOPE_INDEX, true);
+        index.getChild("indexRules")
+                .getChild("nt:base")
+                .getChild("properties")
+                .getChild("analyzed_field")
+                .setProperty(FulltextIndexConstants.PROP_NODE_SCOPE_INDEX, true);
```
nit0906: Wouldn't we need to reindex if we are changing the index definition later on in the test?

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/IndexQueryCommonTest.java
```diff
@@ -279,7 +279,7 @@ public void containsPathNum() throws Exception {
         a.setProperty("name", "/segment1/segment2/segment3");
         root.commit();
 
-        String stmt = "//*[jcr:contains(., '/segment1/segment2')]";
+        String stmt = "//*[jcr:contains(., '\\\\/segment1\\\\/segment2')]";
```
nit0906: Would the user need to escape characters like this while querying on paths ?

nit0906 (changes requested):
Added a couple of comments - Also, since we are looking at bringing parity among lucene and elastic here - maybe it would make sense to have a common Test for both elastic and lucene fulltext that runs with both configurations - something like IndexQueryCommonTest which defines common tests and is then extended by ElasticIndexQueryCommonTest and LuceneIndexQueryCommonTest

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/IndexQueryCommonTest.java
nit0906: I suppose Fabrizio's comment above regarding FulltextIndex.rewriteQueryText(text) would handle this and user might not need to escape in queries - right ?

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticFullTextAsyncTest.java
FrancoisZhang: @nit0906 not sure, actually no change here, the diff is just from IDE code format, I should not do the format here actually. let me see if can easily revert the format change here.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/dynamicBoost/DynamicBoostTest.java
FrancoisZhang: @fabriziofortino agreed, it should be in a separate class actually, let me see if there is exists one to put.

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/IndexQueryCommonTest.java
FrancoisZhang: @nit0906 agreed, this should be reverted after using FulltextIndex.rewriteQueryText(text), I will fix it.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
FrancoisZhang: @fabriziofortino makes sense, I just fix that way and revert the test change

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticFullTextAsyncTest.java
FrancoisZhang: @fabriziofortino thanks, but I also found there is a legacy test defaultAnalyzer in same class which is covered for analyzed field. hence may no need to create dup tests here.

FrancoisZhang:
@nit0906 @fabriziofortino Thanks a lot for review, I made changes based on your review. And also move the fulltext test over the IndexQueryCommonTest, so that it will test over both elastic and lucene. it ended up a clean commit.
Can you help to review it again? thanks

fabriziofortino (approved):
LGTM

nit0906 (approved):
LGTM

PR #490, opened 2022-02-10
OAK-9686: replace Elastic RHLC with Elastic Java Client

AngelaFabregues:
This is the elasticserach client migration approach that I'm following. Two clients share the same transport layer.


AngelaFabregues:
I've increased the max jar file from run-elastic to try to pass the checks. Now to 120MB.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnection.java
```diff
@@ -108,11 +121,25 @@ public RestHighLevelClient getClient() {
                         Header[] headers = new Header[]{new BasicHeader("Authorization", "ApiKey " + apiKeyAuth)};
                         builder.setDefaultHeaders(headers);
                     }
-                    client = new RestHighLevelClient(builder);
+                    hlClient = new RestHighLevelClient(builder);
```
thomasmueller: Let's say there is an exception or error (RuntimeException, OutOfMemoryError,...) after setting the hlClient but before setting the esClient. In this case, we end up having hlClient != null and esClient == null, which is a weird state... 

To protect against that, the only 100% save way I would know is encapsulate esClient + hlClient in a separate object (a "ClientPair" class) and assign that... but it would be overly complicated I think. An alternative is to do the following (untested):

    RestHighLevelClient rc = new RestHighLevelClient(builder);
    ElasticsearchClient ec = new ElasticsearchClient(new RestClientTransport(
            	        rc.getLowLevelClient(), new JacksonJsonpMapper()));
    // "atomically" assign to the fields
    hlClient = rc;
    esClient = ec;

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnection.java
```diff
@@ -108,11 +121,25 @@ public RestHighLevelClient getClient() {
...
+        return hlClient;
+    }
+
+    /**
+     * Gets the NEW Elasticsearch Client
+     * @return the new Elasticsearch client
+     */
+    public ElasticsearchClient getElasticsearchClient() {
+    	getClient();
+	    return esClient;
```
thomasmueller: It looks like a tab character here... In our code (well, my code at least), we use spaces instead of tabs (4 spaces). Could you configure your IDE to use spaces instead of tabs? I know it's not the Eclipse default...

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnection.java
```diff
@@ -90,15 +95,23 @@ private ElasticConnection(@NotNull String indexPrefix, @NotNull String scheme, @
...
+     * @return the old Elasticsearch client
+     */
     public RestHighLevelClient getClient() {
         if (isClosed.get()) {
             throw new IllegalStateException("Already closed");
         }
 
-        // double-checked locking to get good performance and avoid double initialization
-        if (client == null) {
+        if (hlClient == null) {
```
thomasmueller: If we do use double-checked locking, then we need to use esClient == null here, otherwise there is a possibility that hlClient != null but esClient isn't assigned yet (race condition).

And, at some point in the future, I assume we want to get rid of the hlClient field. So, it's better to not use it where possible.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnection.java
```diff
@@ -90,15 +95,23 @@ private ElasticConnection(@NotNull String indexPrefix, @NotNull String scheme, @
...
+     * It double-checks locking to get good performance and avoid double initialization
+     * @deprecated
+     * @return the old Elasticsearch client
+     */
     public RestHighLevelClient getClient() {
         if (isClosed.get()) {
             throw new IllegalStateException("Already closed");
         }
 
-        // double-checked locking to get good performance and avoid double initialization
```
thomasmueller, line 98: This comment still applies, so I would keep it.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnection.java
```diff
@@ -135,8 +162,8 @@ public boolean isAvailable() {
 
     @Override
     public synchronized void close() throws IOException {
-        if (client != null) {
-            client.close();
+        if (hlClient != null) {
```
thomasmueller: It is slightly weird to check hlClient != null but then use esClient... I think this would be easier to read:

    if (esClient != null) {
        esClient._transport().close();
    }

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexCleaner.java
```diff
@@ -138,13 +138,11 @@ public void run() {
...
             }
-            DeleteIndexRequest deleteIndexRequest = new DeleteIndexRequest(indicesToDelete.toArray(new String[]{}));
-            if (deleteIndexRequest.indices() != null && deleteIndexRequest.indices().length > 0) {
-                String indexString = Arrays.toString(deleteIndexRequest.indices());
-                AcknowledgedResponse acknowledgedResponse = elasticConnection.getClient().indices().delete(deleteIndexRequest, RequestOptions.DEFAULT);
-                LOG.info("Deleting remote indices {}", indexString);
-                if (!acknowledgedResponse.isAcknowledged()) {
-                    LOG.error("Could not delete remote indices " + indexString);
+            if(!indicesToDelete.isEmpty()) {
+            	DeleteIndexResponse response = elasticConnection.getElasticsearchClient().indices().delete(i->i.index(indicesToDelete));
```
thomasmueller: Please add spaces around ->

thomasmueller (changes requested):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnection.java
AngelaFabregues: Forget about the following, it can be easily done checking the esClient==null as well as you point out in another comment. I will do that. Thanks.
> I don't see this case as possible. I trust you and I assume it is a protective strategy to implement the code. I will try to understand and follow.
> What I would like to assess beforehand is whether we are even contemplating the option of merging the code before it the client migration (recoding) has been completed. If so, yes we should be protective and I should put more attention to it.

AngelaFabregues:
@fabriziofortino , the new ElasticConnection implementation makes access to the LowClient depend on the deprecated RestHighLevelClient.
I couldn't find how to add the filter_path to the new java client without using the low-level client. The ElasticQueryScanner depends on it.
How should we proceed?
1. Ignoring the filter_path and making the ElasticQueryScanner use the new java client (instead of the low level one).
2. Modifying the ElasticConnection (a bit back) to remove the dependency from the low level to the deprecated RestHighLevelClient
3. Ignoring the issue for now and dealing with it whenever the index creation and ingestion are also migrated.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnection.java
fabriziofortino: the clients are now atomically returned in a Tuple.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnection.java
```diff
@@ -108,11 +123,41 @@ public RestHighLevelClient getClient() {
...
-        return client;
+        return Tuple.tuple(Tuple.tuple(esClient, esAsyncClient), hlClient);
+    }
+
+    /**
+     * Gets the Elasticsearch Client
+     * @return the Elasticsearch client
+     */
+    public ElasticsearchClient getClient() {
+        return getClients().v1().v1();
```
nit0906: I know this is temporary and we will eventually remove the hlc later on - but I was wondering is we can still maybe return a map from getClients() and then do something like getClients().get("ES_JAVA_CLIENT") and so on..

The v1().v1() is pretty confusing (especially when in context of clients) - v1() immediately makes me think of client/api versioning and not of getting the first value from tupple.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnection.java
```diff
@@ -135,8 +180,11 @@ public boolean isAvailable() {
 
     @Override
     public synchronized void close() throws IOException {
-        if (client != null) {
-            client.close();
+        if (esClient != null) {
+        	esClient._transport().close();
```
nit0906: is esAsyncClient also closed as part of this call ? Or should we close that separately ? Since it's created independently - I would assume it would need to be closed separately too (but I might be wrong here.)

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
```diff
@@ -161,7 +160,7 @@ private BulkProcessor initBulkProcessor() {
     }
 
     protected BiConsumer<BulkRequest, ActionListener<BulkResponse>> requestConsumer() {
-        return (request, bulkListener) -> elasticConnection.getClient().bulkAsync(request, RequestOptions.DEFAULT, bulkListener);
+        return (request, bulkListener) -> elasticConnection.getOldClient().bulkAsync(request, RequestOptions.DEFAULT, bulkListener);
```
nit0906, line 164: Can we add a TODO here and anyplace where we are using the oldClient, mentioning that it would have to be removed in future with a link to issue that's preventing us to remove it now ?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnection.java
```diff
@@ -108,11 +123,41 @@ public RestHighLevelClient getClient() {
...
                         builder.setDefaultHeaders(headers);
                     }
-                    client = new RestHighLevelClient(builder);
+                    RestClient httpClient = builder.build();
+                    hlClient = new RestHighLevelClientBuilder(httpClient).setApiCompatibilityMode(true).build();
+
+                    ElasticsearchTransport transport = new RestClientTransport(
+                            httpClient, new JacksonJsonpMapper());
+                    esClient = new ElasticsearchClient(transport);
+                    esAsyncClient = new ElasticsearchAsyncClient(transport);
```
nit0906: Could there be a case where esClient could be null but not esAsyncClient ?

nit0906:
```
[INFO] --- maven-bundle-plugin:5.1.1:baseline (baseline) @ oak-search-elastic ---
[INFO] artifact org.apache.jackrabbit:oak-search-elastic: checking for updates from Nexus
[INFO] artifact org.apache.jackrabbit:oak-search-elastic: checking for updates from central
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  30:33 min
[INFO] Finished at: 2022-05-12T14:28:57
[INFO] ------------------------------------------------------------------------
[ERROR] GC overhead limit exceeded -> [Help 1]
```


The jenkins build on the PR for oak-search-elastic seems to be failing with GC overhead limit exceeded - we should probably look into this.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnection.java
fabriziofortino: Makes sense. I have created an internal class (`Clients`) to wrap up all the clients. I preferred it to a Map since it's strongly typed (no need to cast).

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnection.java
fabriziofortino: Both clients are based on the same transport layer. We just need to close one. Added a comment to clarify that.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
fabriziofortino: Done

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnection.java
fabriziofortino: no, this cannot happen

fabriziofortino:
> ```
> [ERROR] GC overhead limit exceeded -> [Help 1]
> ```
> 
> The jenkins build on the PR for oak-search-elastic seems to be failing with GC overhead limit exceeded - we should probably look into this.

@nit0906 I have seen this but it's not clear what triggers it. I would expect an OOM when the tests are executed. The OOM happens later though. Can we enable full debug on jenkins to see if that helps in understanding where this comes from?

nit0906:
@fabriziofortino 

> Can we enable full debug on jenkins to see if that helps in understanding where this comes from?

This seems to be some apache ci jenkins - I am not entirely sure how to change the debug logging, but it seems it runs from this jenkins file - https://github.com/apache/jackrabbit-oak/blob/trunk/Jenkinsfile 

I think if we add the -X option , in the mvn clean verify command, it should log at debug level. But I think we might need to change that in the trunk branch (changing in the feature branch wouldn't work I guess since the job https://ci-builds.apache.org/job/Jackrabbit/job/oak-trunk-pr/configure here is picking from the trunk of apache repo.)

fabriziofortino:
@nit0906 the problem was with the bundle. I have fine-tuned the dependencies (we can get rid of all lucene dependencies when the migration to ES Java Client is complete) and the build is now successful.

PR #493, opened 2022-02-14
OAK-9695 correctly evaluate property type from definition while check if

kwin:
protected

### DIFF
--- a/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/NodeDelegate.java
+++ b/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/NodeDelegate.java
@@ -212,7 +212,7 @@ public class NodeDelegate extends ItemDelegate {
         // property definitions. Iterate through them to check whether
         // there's a matching, protected one.
         if (protectedResidual) {
-            Tree definition = findMatchingResidualPropertyDefinition(null, types, propertyType, true);
+            Tree definition = findMatchingResidualPropertyDefinition(types, propertyType);
             if (definition != null && TreeUtil.getBoolean(definition, JCR_PROTECTED)) {
                 return true;
             }
@@ -615,13 +615,13 @@ public class NodeDelegate extends ItemDelegate {
         return findMatchingResidualPropertyDefinition(fuzzyMatch, types, propertyType.isArray(), definedType, undefinedType, exactTypeMatch);
     }
 
-    private Tree findMatchingResidualPropertyDefinition(Tree fuzzyMatch, List<Tree> types, Type<?> propertyType, boolean exactTypeMatch) {
+    private Tree findMatchingResidualPropertyDefinition(List<Tree> types, Type<?> propertyType) {
         String definedType = propertyType.toString();
         String undefinedType = UNDEFINED.toString();
         if (propertyType.isArray()) {
             undefinedType = UNDEFINEDS.toString();
         }
-        return findMatchingResidualPropertyDefinition(fuzzyMatch, types, propertyType.isArray(), definedType, undefinedType, exactTypeMatch);
+        return findMatchingResidualPropertyDefinition(null, types, propertyType.isArray(), definedType, undefinedType, true);
     }
 
     private Tree findMatchingResidualPropertyDefinition(Tree fuzzyMatch, List<Tree> types, boolean isMultiValue, String definedType, String undefinedType, boolean exactTypeMatch) {
--- a/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/NodeDelegate.java
+++ b/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/NodeDelegate.java
@@ -193,14 +193,14 @@ public class NodeDelegate extends ItemDelegate {
         return false;
     }
 
-    boolean isProtected(String property) throws InvalidItemStateException {
+    boolean isProtected(String propertyName, Type<?> propertyType) throws InvalidItemStateException {
         Tree tree = getTree();
         Tree typeRoot = sessionDelegate.getRoot().getTree(NODE_TYPES_PATH);
         List<Tree> types = TreeUtil.getEffectiveType(tree, typeRoot);
 
         boolean protectedResidual = false;
         for (Tree type : types) {
-            if (contains(TreeUtil.getNames(type, REP_PROTECTED_PROPERTIES), property)) {
+            if (contains(TreeUtil.getNames(type, REP_PROTECTED_PROPERTIES), propertyName)) {
                 return true;
             } else if (!protectedResidual) {
                 protectedResidual = TreeUtil.getBoolean(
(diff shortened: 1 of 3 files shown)


oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/PropertyDelegate.java
```diff
@@ -92,7 +92,7 @@ public Status getStatus() {
 
     @Override
     public boolean isProtected() throws InvalidItemStateException {
-        return getParent().isProtected(name);
+        return getParent().isProtected(name, state.getType());
```
mreutegg: `state` may be `null` when the property is stale. E.g. another session removed the property.

```suggestion
        return getParent().isProtected(name, getPropertyState().getType());
```
The implementation actually has another similar problem. `getParent()` may also return null :-/ I'll create a separate issue for that.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/NodeDelegate.java
```diff
@@ -617,6 +612,19 @@ private Tree findMatchingPropertyDefinition(
         }
 
         // Then look through any residual property definitions
+        return findMatchingResidualPropertyDefinition(fuzzyMatch, types, propertyType.isArray(), definedType, undefinedType, exactTypeMatch);
+    }
+
+    private Tree findMatchingResidualPropertyDefinition(Tree fuzzyMatch, List<Tree> types, Type<?> propertyType, boolean exactTypeMatch) {
```
mreutegg: The only call of this method is from here: https://github.com/apache/jackrabbit-oak/pull/493/files#diff-75abd61c57624d67f0c501c17199bf4824b5d56a03c22329c1e717c88044aa1cR215 with `fuzzyMatch` set to `null` and `exactTypeMatch` set to `true`. Those parameters are therefore unnecessary.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/PropertyTest.java
```diff
@@ -274,4 +285,23 @@ public void testPropertyIteratorSize() throws Exception{
         assertEquals(3, pitr.getSize());
         assertEquals(3, Iterators.size(pitr));
     }
+    
+    public void testSetAndRemoveUnprotectedProperty() throws ValueFormatException, VersionException, LockException, ConstraintViolationException, RepositoryException {
```
mreutegg: There is no need to declare `ValueFormatException`, `VersionException`, `LockException`, `ConstraintViolationException`. They are all `RepositoryException`.
```suggestion
    public void testSetAndRemoveUnprotectedProperty() throws RepositoryException {
```

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/PropertyTest.java
```diff
@@ -274,4 +285,23 @@ public void testPropertyIteratorSize() throws Exception{
...
+    	Property property = node.setProperty(BOOLEAN_PROP_NAME, true);
+    	assertNotNull(property);
+    	property.setValue(false);
+    	superuser.save();
+    	property.remove();
+    	superuser.save();
+    }
+    
+    public void testSetProtectedResidualProperty() throws ValueFormatException, VersionException, LockException, ConstraintViolationException, RepositoryException {
```
mreutegg: Same as above.
```suggestion
    public void testSetProtectedResidualProperty() throws RepositoryException {
```

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/PropertyTest.java
```diff
@@ -274,4 +285,23 @@ public void testPropertyIteratorSize() throws Exception{
...
+    	Value uriValue = ValueFactoryImpl.getInstance().createValue("http://example.com", PropertyType.URI);
+    	try {
+    		node.setProperty("test", uriValue);
+    		fail("Setting protected property (according to residual propery type definition) must throw ConstraintViolationException");
```
mreutegg:
```suggestion
    		fail("Setting protected property (according to residual property type definition) must throw ConstraintViolationException");
```

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/PropertyTest.java
```diff
@@ -274,4 +285,23 @@ public void testPropertyIteratorSize() throws Exception{
...
         assertEquals(3, Iterators.size(pitr));
     }
+    
+    public void testSetAndRemoveUnprotectedProperty() throws ValueFormatException, VersionException, LockException, ConstraintViolationException, RepositoryException {
+    	Property property = node.setProperty(BOOLEAN_PROP_NAME, true);
+    	assertNotNull(property);
+    	property.setValue(false);
+    	superuser.save();
+    	property.remove();
+    	superuser.save();
```
mreutegg: Please replace tabs with spaces.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/PropertyTest.java
```diff
@@ -274,4 +285,23 @@ public void testPropertyIteratorSize() throws Exception{
...
+    }
+    
+    public void testSetProtectedResidualProperty() throws ValueFormatException, VersionException, LockException, ConstraintViolationException, RepositoryException {
+    	Value uriValue = ValueFactoryImpl.getInstance().createValue("http://example.com", PropertyType.URI);
+    	try {
+    		node.setProperty("test", uriValue);
+    		fail("Setting protected property (according to residual propery type definition) must throw ConstraintViolationException");
+    	} catch (ConstraintViolationException e) {
```
mreutegg: Please replace tabs with spaces.

mreutegg (changes requested):
In general I agree with the approach and the fix, but see my suggestions for minor changes.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/NodeDelegate.java
kwin: It is also called from https://github.com/apache/jackrabbit-oak/pull/493/files#diff-75abd61c57624d67f0c501c17199bf4824b5d56a03c22329c1e717c88044aa1cR615 with both arguments having non-predictable values.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/PropertyTest.java
kwin: Done in https://github.com/apache/jackrabbit-oak/pull/493/commits/43443990ba235b9468b1f551e0400318d8561849

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/PropertyTest.java
kwin: Done in https://github.com/apache/jackrabbit-oak/pull/493/commits/43443990ba235b9468b1f551e0400318d8561849

PR #494, opened 2022-02-15
OAK-9696 Improve query syntax support for dynamicBoost in ElasticSearch

FrancoisZhang:


thomasmueller (approved):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexHelper.java
```diff
@@ -276,6 +276,11 @@ private static void mapIndexRules(ElasticIndexDefinition indexDefinition, XConte
             mappingBuilder.endObject();
         }
 
+        mappingBuilder.startObject(ElasticIndexDefinition.DYNAMIC_BOOST_TAGS)
+                .field("type", "text")
+                .field("analyzer", "oak_analyzer")
+                .endObject();
+
```
fabriziofortino: I would change this. We currently support one or more dynamicBoost properties. Setting this to a fixed field won't work in case more than one field is configured. The same mapping can be moved in the dynamicBoost for-loop after line 276. I would also rename the field name using the `:fulltext` suffix:

```java
 mappingBuilder.startObject(pd.nodeName + FieldNames.FULLTEXT)
    .field("type", "text")
    .field("analyzer", "oak_analyzer")
    .endObject();
```

This would require changes in `ElasticDocument` and `ElasticRequestHandler`

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
```diff
@@ -568,7 +568,7 @@ private boolean visitTerm(String propertyName, String text, String boost, boolea
                 if (boost != null) {
                     fullTextQuery.boost(Float.parseFloat(boost));
                 }
-                BoolQueryBuilder shouldBoolQueryWrapper = boolQuery().should(fullTextQuery);
+                BoolQueryBuilder shouldBoolQueryWrapper = boolQuery().must(fullTextQuery);
```
fabriziofortino: We could simplify this by removing a level of nesting: `fullTextQuery` can be passed as `must` clause at line 575, the rest (dynamic score queries) should go in the `should` clause.

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticDynamicBoostQueryTest.java
```diff
@@ -152,15 +152,53 @@ public void testQueryDynamicBoostOrder() throws Exception {
...
+            assertQuery("select [jcr:path] from [dam:Asset] where contains(@title, 'bl?e')", SQL2, Arrays.asList("/test/asset3"));
+            assertQuery("select [jcr:path] from [dam:Asset] where contains(@title, '?lue')", SQL2, Arrays.asList("/test/asset3"));
+            assertQuery("select [jcr:path] from [dam:Asset] where contains(@title, 'coff*')", SQL2, Arrays.asList("/test/asset2"));
+            assertQuery("select [jcr:path] from [dam:Asset] where contains(@title, 'co*ee')", SQL2, Arrays.asList("/test/asset2"));
+            assertQuery("select [jcr:path] from [dam:Asset] where contains(@title, '*ffee')", SQL2, Arrays.asList("/test/asset2"));
+        });
+    }
+
+    @Test
+    public void testQueryDynamicBoosSpace() throws Exception {
```
fabriziofortino: typo 

```suggestion
    public void testQueryDynamicBoostSpace() throws Exception {
```

fabriziofortino (changes requested):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
FrancoisZhang: @fabriziofortino thanks, that makes sense! let me change that way

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexHelper.java
FrancoisZhang: @fabriziofortino Thanks, I just refined the PR with another commit over the review. Can you review again to see if all meet what you suggested here?

fabriziofortino: @FrancoisZhang I don't think this works. The mapping field in the for-loop overwrites itself since the name is always the same. The name has to be composed (see my suggestion above). I would create it with `pd.nodeName + FieldNames.FULLTEXT` (you can get rid of DYNAMIC_BOOST_TAGS). This would require some changes in `ElasticRequestHandler` since the fields to query need to be re-computed.

FrancoisZhang: @fabriziofortino thanks, that makes sense. I will change that way.
I actually think the origin way I put here with fixed field should also works, the reason the query is cross fields and don't really care if they're stored in single field or multiple fields. I could also put multiple fields, both should works IMO and fixed will be easier.

FrancoisZhang: @fabriziofortino just pushed another commit which change it back to a fixed field for storing all terms for multiple boost fields. as the query will do cross fields query. so either single field or multiple fields should works here. thanks

FrancoisZhang: @fabriziofortino OK, just pushed more tests to cover multiple dynamic boost cases. It's ready for merging if you also approve it, thanks!

FrancoisZhang: @fabriziofortino per discussion, I refined more to store the tags into fulltext field. it ended up a more clean commit now. thanks!

fabriziofortino (approved):

PR #495, opened 2022-02-16
Backport OAK-9587 - Add an attribute to enforce a strict index tag check

etugarev:
This PR includes:

- Backport of of https://github.com/apache/jackrabbit-oak/pull/378
- Updated documentation

PR #496, opened 2022-02-16
Backport OAK-9653 Adding the index tag option interferes with regex properties

etugarev:
This PR rely on #495 and includes a partial backport of #452

Due to major structure difference the following were not backported

- changes related to elastic search
- tests


thomasmueller (commented):
There are some missing tests... https://github.com/apache/jackrabbit-oak/pull/452/files contains descendantTestWithIndexTagExplain() for example, which is missing here

thomasmueller:
Ah you mentioned the missing tests... yes it makes sense. I will try to merge this PR.

PR #499, opened 2022-02-22
OAK-9700: RevisionGC may fail with NPE

mreutegg:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -805,7 +805,10 @@ private long getNumPreviousDocuments() {
 
         private void addPreviousDocuments(Iterator<String> ids) throws IOException {
             while (ids.hasNext()) {
-                prevDocIdsToDelete.add(ids.next());
+                String id = ids.next();
+                if (id != null) {
+                    prevDocIdsToDelete.add(id);
+                }
```
stefan-egli, line 814: Was thinking to add a log.debug in case it is null. In which case you'd also need something meaningful to log - so maybe the `doc` or just the `id` (eg as `deletedId`) could be passed to this method and that could be logged. (and just at debug shouldn't hurt)
```suggestion
                } else if (log.isDebugEnabled()) {
                    log.debug("addPreviousDocuments: null id found via deletedId={}", deletedId);
                }
```

stefan-egli (commented):
+1 for catcing the null. Maybe we could also add a log.debug in case of null?

stefan-egli (approved):
but I wasn't meant to block the PR, so here's the +1 :)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
mreutegg: I like this suggestion and will update the PR.

PR #500, opened 2022-02-22
OAK-9702 AzureBlobStoreBackend: no container auto-creation with SAS URI

jelmini:


oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
```diff
@@ -165,7 +165,8 @@ public void init() throws DataStoreException {
             try {
                 Utils.setProxyIfNeeded(properties);
                 containerName = (String) properties.get(AzureConstants.AZURE_BLOB_CONTAINER_NAME);
-                createBlobContainer = PropertiesUtil.toBoolean(properties.getProperty(AzureConstants.AZURE_CREATE_CONTAINER), true);
+                createBlobContainer = PropertiesUtil.toBoolean(properties.getProperty(AzureConstants.AZURE_CREATE_CONTAINER), true) 
+                    && PropertiesUtil.toString(properties.getProperty(AzureConstants.AZURE_SAS), "").isEmpty();
```
smiroslav, line 169: I think it should rather be controlled with property `azureCreateContainer`. Code here assumes that SAS token is generated for container, but it can be account level token as well.

jelmini: Good point. I will reject this PR. I will on the other hand recreate a PR with only the test code, which is still useful.

PR #503, opened 2022-02-23
OAK-9705 Explain Query tool doesn't show the correct ES query when su…

FrancoisZhang:
…ggest queries are made

### DIFF
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
@@ -206,6 +206,9 @@ public class ElasticRequestHandler {
                         );
                     }
                 }
+            }  else if (propertyRestrictionQuery.startsWith(SUGGEST_PREFIX)) {
+                String suggestQueryString = propertyRestrictionQuery.replace(SUGGEST_PREFIX, "");
+                boolQuery.must(suggestionMatchQuery(suggestQueryString));
             } else {
                 boolQuery.must(queryStringQuery(propertyRestrictionQuery));
             }
--- a/oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticSuggestQueryTest.java
+++ b/oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticSuggestQueryTest.java
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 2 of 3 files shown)


oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticSuggestQueryTest.java
```diff
@@ -0,0 +1,122 @@
+/*
```
nit0906, line 1: There's a test class for suggest use cases - IndexSuggestionCommonTest which is extended by ElasticIndexSuggestionCommonTest and LuceneIndexSuggestionCommonTest. Maybe we can add new tests there only - this way it's easier to maintain - also we do similar tests for both lucene and elastic.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
```diff
@@ -206,6 +206,9 @@ public BoolQueryBuilder baseQuery() {
                         );
                     }
                 }
+            }  else if (propertyRestrictionQuery.startsWith(SUGGEST_PREFIX)) {
+                String suggestQueryString = propertyRestrictionQuery.replace(SUGGEST_PREFIX, "");
```
nit0906, line 210: Maybe a comment here on why we did this ?

nit0906 (changes requested):
Requested a couple of changes.

thomasmueller (approved):

fabriziofortino (changes requested):
I think the logic can be improved. I looked more into the existing code and the `explain query` does not work well in all cases, not just for suggestions.

The problem is that we always return the `baseQuery` but the different iterators could enrich the query. As an example, the async iterator (responsible for most of the queries) composes the query in the following way:

https://github.com/apache/jackrabbit-oak/blob/trunk/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java#L195-L233 

So, for normal queries, we are potentially missing sorting and aggregations clauses.

To implement the explain properly we should instantiate the specific iterator and get the query without actually executing it. This is already okay for spellcheck and suggest iterator. The `ElasticResultRowAsyncIterator` needs to be changed since it runs the query at instantiation time.

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticSuggestQueryTest.java
FrancoisZhang: @nit0906 I tried there first, but realized it's lack of explain query etc. facilities in ElasticAbstractQueryTest, made it needs lots of additional code for making it possible.

FrancoisZhang:
@fabriziofortino  Thanks! let me use a new ticket and PR for a more generic solution for all explain issue. Let me close this PR since it's only cover suggestion query issue.

PR #508, opened 2022-03-02
OAK-9738 Parallel merge

Ewocker:


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -65,6 +65,24 @@
      */
     static final String PROP_THREAD_POOL_SIZE = "oak.indexer.dataDumpThreadPoolSize";
 
+    /**
+     * Default value for {@link #PROP_MERGE_THREAD_POOL_SIZE}
+     */
+    static final String DEFAULT_NUMBER_OF_MERGE_TASK_THREADS = "4";
```
thomasmueller: I would make this an "int".

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -65,6 +65,24 @@
...
+    /**
+     * System property for specifying number of threads for parallel merge when using {@link MultithreadedTraverseWithSortStrategy}
+     */
+    static final String PROP_MERGE_THREAD_POOL_SIZE = "oak.indexer.mergeTaskThreadPoolSize";
+
+    /**
+     * Default value for {@link #PROP_MERGE_TASK_BATCH_SIZE}
+     */
+    static final String DEFAULT_NUMBER_OF_FILES_PER_MERGE_TASK = "64";
```
thomasmueller: int

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -29,17 +29,12 @@
...
+import java.io.*;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Comparator;
-import java.util.List;
+import java.util.*;
```
thomasmueller: I would not use star imports.  I don't see an advantage. See also https://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -48,12 +43,13 @@
...
 import java.util.concurrent.Phaser;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.stream.Stream;
 
 import static com.google.common.base.Charsets.UTF_8;
-import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileNodeStoreBuilder.DEFAULT_NUMBER_OF_DATA_DUMP_THREADS;
-import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileNodeStoreBuilder.PROP_THREAD_POOL_SIZE;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileNodeStoreBuilder.*;
```
thomasmueller, line 50: The problem with star imports here is: if you want to find where DEFAULT_NUMBER_OF_DATA_DUMP_THREADS comes from, and you have multiple star imports, then you can't easily find out.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -264,6 +269,9 @@ void createInitialTasks(NodeStateEntryTraverserFactory nodeStateEntryTraverserFa
                     if (!existingSortWorkDir.isDirectory()) {
                         log.info("Not a directory {}. Skipping it.", existingSortWorkDir.getAbsolutePath());
                         continue;
+                    } else if (existingSortWorkDir.getName().equals("merge")) {
+                        log.info("Intermediate Merge Directory. Skipping it.", existingSortWorkDir.getAbsolutePath());
```
thomasmueller: The {} is missing in the message.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -408,6 +402,172 @@ public void run() {
...
+     *    (if final merge) merge all intermediate merge files and create sorted file
+     *      <li>add all merged files to merged list</li>
+     * </ol>
+     */
+    private class MergeRunner implements Runnable {
+        private final ArrayList<File> mergedFiles = new ArrayList<File>();
+        private final File sortedFile;
+        private int nextMergedLength = 0;
+        private final ExecutorService executorService;
+        private final int threadPoolSize = Integer.parseInt(System.getProperty(PROP_MERGE_THREAD_POOL_SIZE, DEFAULT_NUMBER_OF_MERGE_TASK_THREADS));
```
thomasmueller: Here, we can use Integer.getInteger:
https://docs.oracle.com/javase/9/docs/api/java/lang/Integer.html#getInteger-java.lang.String-int-

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -408,6 +402,172 @@ public void run() {
...
+        private final int batchMergeSize = Integer.parseInt(System.getProperty(PROP_MERGE_TASK_BATCH_SIZE, DEFAULT_NUMBER_OF_FILES_PER_MERGE_TASK));
```
thomasmueller: same

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -408,6 +402,172 @@ public void run() {
...
+            this.executorService = Executors.newFixedThreadPool(threadPoolSize);
+        }
+
+        private ArrayList<File> getUnmergedFiles(int size) {
+            ArrayList<File> unmergedFiles = new ArrayList<File>();
+            for (File f : sortedFiles) {
+                if (!mergedFiles.contains(f) && f != MERGE_POISON_PILL) {
+                    unmergedFiles.add(f);
+                }
+                if (unmergedFiles.size() == size) {
```
thomasmueller: I would use >=

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -408,6 +402,172 @@ public void run() {
...
+                boolean exceptionsCaught = false;
+                for (Future<File> result : results) {
+                    try {
+                        mergeTarget.add(result.get());
+                    } catch (Throwable e) {
+                        throwables.add(e);
+                        exceptionsCaught = true;
+                    }
+                }
+                log.debug("Completed merge result collection {}. Arriving at phaser now.", !exceptionsCaught ? "fully" : "partially");
```
thomasmueller: What about 
    exceptionsCaught ? "partially" : "fully"

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -408,6 +402,172 @@ public void run() {
...
+            while (true) {
+                while (!sortedFiles.contains(MERGE_POISON_PILL) && sortedFiles.size() <= nextMergedLength) {
+                    // waiting for n files to be merged in a batch
+                }
+                if (sortedFiles.contains(MERGE_POISON_PILL)) {
+                    break;
+                }
+
+                ArrayList<File> mergeTarget = getUnmergedFiles(batchMergeSize);
+                System.out.println(mergeTarget);
```
thomasmueller: Use "log." instead

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -408,6 +402,172 @@ public void run() {
...
+
+        public MergeRunner(File sortedFile) {
+            this.sortedFile = sortedFile;
+            this.executorService = Executors.newFixedThreadPool(threadPoolSize);
+        }
+
+        private ArrayList<File> getUnmergedFiles(int size) {
+            ArrayList<File> unmergedFiles = new ArrayList<File>();
+            for (File f : sortedFiles) {
+                if (!mergedFiles.contains(f) && f != MERGE_POISON_PILL) {
```
thomasmueller: It looks like we merge the first n files. In my view, it is important to merge the _smallest_ n files.

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
Ewocker: I followed the property for `DEFAULT_NUMBER_OF_DATA_DUMP_THREADS` which is also a string
Should I change it all to `int` instead?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
Ewocker: I followed the property for DEFAULT_NUMBER_OF_DATA_DUMP_THREADS which is also a string
Should I change it all to int instead?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
Ewocker: make sense

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
Ewocker: make sense

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
Ewocker: make sense

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
Ewocker: added

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
Ewocker: make sense

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
Ewocker: make sense

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
Ewocker: updated for all

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
Ewocker: updated for all

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
Ewocker: removed, this is debug log message I forget to remove

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
Ewocker: make sense, was copied from original code somewhere.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
Ewocker: added change

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -264,6 +278,9 @@ void createInitialTasks(NodeStateEntryTraverserFactory nodeStateEntryTraverserFa
                     if (!existingSortWorkDir.isDirectory()) {
                         log.info("Not a directory {}. Skipping it.", existingSortWorkDir.getAbsolutePath());
                         continue;
+                    } else if (existingSortWorkDir.getName().equals("merge")) {
```
averma21: Since this file name "merge" is used at multiple places, it would be good to make it a constant and reference it.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -408,6 +411,170 @@ public void run() {
...
+    }
+
+    /**
+     * Class responsible for -
+     * <ol>
+     *     <li>Watching {@link #taskQueue} for new tasks</li>
+     *     <li>Submitting those tasks to an {@link ExecutorService}</li>
+     *     <li>Collecting the results (sorted files) created by those tasks into one place</li>
+     * </ol>
+     */
```
averma21: This javadoc needs to be changed .

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -408,6 +411,170 @@ public void run() {
...
+            return unmergedFiles.subList(0, (size - 1) > unmergedFiles.size() ? unmergedFiles.size() : (size - 1));
+        }
+
+        @Override
+        public void run() {
+            nextMergedLength += batchMergeSize;
+            Phaser mergeTaskPhaser = new Phaser(1);
+            List<Future<File>> results = new ArrayList<>();
+
+            while (true) {
```
averma21: I think we can avoid this busy wait by converting `sortedFiles` to a [BlockingQueue](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/BlockingQueue.html) . The TraverseAndSort tasks keep on adding files to that queue and this thread consumes from that blocking queue and keeps on adding files to a local list.
If the list size grows beyond the batch size, it is sorted, merged and cleared. WDYT?
We already use a blocking queue which can be taken as an example - `taskQueue`

averma21 (commented):
Would be good to improve the class level java doc of `MultithreadedTraverseWithSortStratgey` to reflect the latest changes.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
Ewocker: done

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
Ewocker: done

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
Ewocker: changed

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -408,6 +454,174 @@ public void run() {
...
+        public void run() {
+            Phaser mergeTaskPhaser = new Phaser(1);
+            List<Future<File>> results = new ArrayList<>();
+            List<File> mergeTarget = new ArrayList<>();
+            int count = 0;
+
+            while (true) {
+                try {
+                    unmergedFiles.add(sortedFiles.take());
+                    if (sortedFiles.contains(MERGE_POISON_PILL) || unmergedFiles.contains(MERGE_POISON_PILL)) {
```
averma21: Since this is a blocking queue now and we pick element by element, I don't see a benefit of doing `contains` check. You could just check the element removed from the queue and break. That would simply this code and you would also not need to drain sortedFiles to unmergedFiles later.

Ewocker: The reason I added that wide range check is so that it proceed directly to the final merge and we don't create anymore tasks if poison pill shows up.
Otherwise if there are 2000 files left to be merged, it would create more tasks and wait for those to finish before we can go for a final merge step which MIGHT(?) introduce more wait time.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -304,14 +364,18 @@ void createInitialTasks(NodeStateEntryTraverserFactory nodeStateEntryTraverserFa
...
     }
 
     @Override
     public File createSortedStoreFile() throws IOException, CompositeException {
         String watcherThreadName = "watcher";
+        String mergerThreadName = "merger";
         Thread watcher = new Thread(new TaskRunner(), watcherThreadName);
         watcher.start();
+        File sortedFile = new File(storeDir, getSortedStoreFileName(compressionEnabled));
+        Thread merger = new Thread(new MergeRunner(sortedFile), mergerThreadName);
```
thomasmueller: Could you set setDaemon(true) please? (For watcher as well.) So that the JVM will stop if this is the only living thread. I see you have already set the thread name, that's good.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -408,6 +454,174 @@ public void run() {
...
+                        count++;
+                        mergeTarget.clear();
+                        mergeTarget = getSmallestUnmergedFiles(batchMergeSize);
+                        Callable<File> mergeTask = new MergeTask(mergeTarget, mergeTaskPhaser,
+                                new File(mergeDir, String.format("intermediate-%s", count)));
+                        markAsMerged(mergeTarget);
+                        results.add(executorService.submit(mergeTask));
+                    }
+                } catch (InterruptedException e) {
+                    log.error("Failed while draining from sortedFiles {}", e);
```
thomasmueller: Actually, here the right code would be:

    log.error("Failed while draining from sortedFiles", e);

... this will log the stack trace.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -59,12 +59,30 @@
...
+    static final int DEFAULT_NUMBER_OF_DATA_DUMP_THREADS = 4;
     /**
      * System property for specifying number of threads for parallel download when using {@link MultithreadedTraverseWithSortStrategy}
      */
     static final String PROP_THREAD_POOL_SIZE = "oak.indexer.dataDumpThreadPoolSize";
 
+    /**
+     * Default value for {@link #PROP_MERGE_THREAD_POOL_SIZE}
+     */
+    static final int DEFAULT_NUMBER_OF_MERGE_TASK_THREADS = 4;
```
thomasmueller: Sorry... I thought we only want _one_ thread that concurrently merges. Now we have 4... that would mean we can have 4 times 65 = 260 files open at the same time... I think 1 merge thread is more efficient.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTaskTest.java
```diff
@@ -34,8 +34,7 @@
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Queue;
-import java.util.concurrent.Callable;
-import java.util.concurrent.Phaser;
+import java.util.concurrent.*;
```
thomasmueller: Please don't use star imports.

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
Ewocker: make sense

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
Ewocker: make sense

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
Ewocker: sure.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTaskTest.java
Ewocker: yep....
need to figure out why intellij always do that for me...

Ewocker: found https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports

thomasmueller (changes requested):
This looks good now, but what I'm missing is test cases! Specially a test case that tests this in isolation: given a (synthetic, large) list of files, run the MergeRunner and see if it really picks he 64 smallest files and merges them, and then in the next run does the same, and so on.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
thomasmueller: > Otherwise if there are 2000 files left to be merged, it would create more tasks and wait for those to finish

I think, actually it would be better to merge at most 64 files at a time (always the smallest 64). I think it's better than to merge 2000 files at the same time: that would result in a lower cache hit rate.

thomasmueller (changes requested):
A good test case is missing (I think I mentioned this before).
And we should never merge more than 64 files at a time.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunner.java
```diff
@@ -0,0 +1,295 @@
...
+     * Constructor.
+     * @param sortedFiles thread safe list containing files to be merged.
+     * @param comparator comparator used to help with sorting of node state entries.
+     * @param mergeDir directory where sorted files will be created.
+     * @param compressionEnabled if true, the created files would be compressed
+     */
+    MergeRunner(File sortedFile, BlockingQueue<File> sortedFiles, File mergeDir, Comparator comparator,
+                Phaser phaser, boolean compressionEnabled) throws IOException {
+        this.mergeDir = mergeDir;
+        FileUtils.forceMkdir(mergeDir);
```
thomasmueller: Could you defer creating the directory to the merge part please? I think it's better to not do too much stuff (do any IO) in the constructor itself.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunner.java
```diff
@@ -0,0 +1,295 @@
...
+        this.compressionEnabled = compressionEnabled;
+        this.sortedFiles = sortedFiles;
+        this.sortedFile = sortedFile;
+        this.throwables = new ConcurrentLinkedQueue<>();
+        this.comparator = comparator;
+        this.phaser = phaser;
+        this.executorService = Executors.newFixedThreadPool(threadPoolSize);
```
thomasmueller: Could you defer creating the executorService to the merge part as well? Two reasons:

(A) If only the constructor is called, but there is some exception, then we have a dangling executorService.

(B) That would allow having setters for threadPoolSize and batchMergeSize, which can then simplify testing (no need to change the system property).

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunnerTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+    @After
+    public void after() {
+        lc.finished();
+    }
+
+    @Test
+    public void test() throws IOException {
+        lc.starting();
+        System.setProperty(FlatFileNodeStoreBuilder.PROP_MERGE_TASK_BATCH_SIZE, "3");
```
thomasmueller: If you are setting the system properties in the test, then you would need to use 

    try {
        old = System.getProperty...
        System.setProperty...
    } finally {
        System.setProperty old...
    }

But! It's better to use a setter, and not change the system properties (see above).

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunnerTest.java
```diff
@@ -0,0 +1,176 @@
...
+        PathElementComparator pathComparator = new PathElementComparator();
+        Comparator<NodeStateHolder> comparator = (e1, e2) -> pathComparator.compare(e1.getPathElements(), e2.getPathElements());
+        BlockingQueue<File> sortedFiles = new LinkedBlockingQueue<>();
+        Phaser mergePhaser = new Phaser(1);
+        Runnable mergeRunner = new MergeRunner(sortedFile, sortedFiles, mergeDir, comparator, mergePhaser, false);
+        Thread merger = new Thread(mergeRunner, "test-merger");
+        merger.setDaemon(true);
+
+        // Adding test files in predefined order
+        ArrayList<Integer> filenameOrder = new ArrayList<>(Arrays.asList(7, 1, 11, 9, 8, 6, 12, 4, 5, 13, 3, 2, 10));
```
thomasmueller: I'm afraid I don't understand this sequence... Is it random numbers? I think it would make sense to use a formula, and not just constants in the test case... If you really want random numbers, then you could use Random.nextInt(n)

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunnerTest.java
```diff
@@ -0,0 +1,176 @@
...
+        for (int filename: filenameOrder.subList(0,4)) {
+            sortedFiles.add(testFiles.get(filename-1));
```
thomasmueller: Why minus one?

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunnerTest.java
```diff
@@ -0,0 +1,176 @@
...
+        List<String> resultList = Lists.newArrayList();
+        for (int i = 0; i <= 90; i++) {
+            String line = String.format("/%05d|{}", i);
+            lineBuffer.add(line);
+            resultList.add(line);
+            if (lineBuffer.size() == nextFileSize) {
+                String text = String.join(newline, lineBuffer);
+                FileUtils.writeStringToFile(testFile, text, UTF_8);
+                testFiles.add(testFile);
+                nextFileSize += 1;
```
thomasmueller: nextFileSize++?

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunnerTest.java
```diff
@@ -0,0 +1,176 @@
...
+    private void generateTestFiles(File tmpDir) throws IOException {
+        int nextFileSize = 1;
+        File testFile = new File(tmpDir, Integer.toString(nextFileSize));
+        List<String> lineBuffer = Lists.newArrayList();
+        List<String> resultList = Lists.newArrayList();
+        for (int i = 0; i <= 90; i++) {
+            String line = String.format("/%05d|{}", i);
+            lineBuffer.add(line);
+            resultList.add(line);
+            if (lineBuffer.size() == nextFileSize) {
```
thomasmueller: You loop 90 times, and if the buffer size matches the next file size, you switch to the new file... but it's not easy (for me) to understand how many files are generated that way... Wouldn't it be easier to understand to loop over the list of files?

What about making the files smaller and smaller? That way, it would be easy to verify they are merged in the right order (just an idea).

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunnerTest.java
```diff
@@ -0,0 +1,176 @@
...
+                add(new File(mergeDir, "intermediate-2"));
+                add(new File(mergeDir, "final-1"));
+                add(new File(mergeDir, "final-2"));
+            }
+        }));
+        expectedLogOutput.add("Total batch sorted files length is 18");
+        assertEquals("final merge log output should be expected", String.join(newline, expectedLogOutput), String.join(newline, actualLogOutput.subList(7, 12)));
+    }
+
+    private void generateTestFiles(File tmpDir) throws IOException {
```
thomasmueller: What about passing a parameter "fileCount"?

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunnerTest.java
```diff
@@ -0,0 +1,176 @@
...
+                nextFileSize += 1;
+                testFile = new File(tmpDir, Integer.toString(nextFileSize));
+                lineBuffer.clear();
+            }
+        }
+        expectedSortedFileStr = String.join(newline, resultList) + newline;
+    }
+
+//    private static List<File> getResourceFolderFiles (String folder) {
+//        ClassLoader loader = Thread.currentThread().getContextClassLoader();
```
thomasmueller: Please remove unused code

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunner.java
Ewocker: make sense

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunner.java
Ewocker: mmake sense

Ewocker: I actually make it a parameter. don't think we would ever change the value, at least threadpool size cannot be changed.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunnerTest.java
Ewocker: make it part of the param

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunnerTest.java
Ewocker: I come up with this sequence randomly, but I don't really want this being randomly.
If it is randomly, logs will be very hard to predict.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunnerTest.java
Ewocker: filename is 1, 2, 3, 4, 5, 6... to 13.  filename equals to the number of line in the file.
but index starts with 0.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunnerTest.java
Ewocker: added

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunnerTest.java
Ewocker: removed

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunnerTest.java
Ewocker: logic changed

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunnerTest.java
Ewocker: did something similar but not exactly the same.
change to use nodeState for generating the lines.
so if fileCount is 3 => we get file with name `1`, `2`, `3`
and each file `1` contains 1 line, file `2` contains 2 lines and so on.
line contents are randomized but sorted.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunnerTest.java
thomasmueller: You can use a random number generator with a seed. In this case, the sequence is random, but reproducible.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunnerTest.java
thomasmueller: Yes... but it seems complicated to me to add one to each entry in filenameOrder, and then subtract one. It feels unnecessary.

thomasmueller (changes requested):

thomasmueller (approved):
I think the test is now much easier to understand and maintain! Thanks!

PR #510, opened 2022-03-03
OAK-9713 sas uri support in oak-upgrade

ArunOnCloud:
sas uri support in oak-upgrade

### DIFF
--- a/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
+++ b/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
@@ -100,20 +100,31 @@ public final class AzureUtilities {
 
     public static CloudBlobDirectory cloudBlobDirectoryFrom(StorageCredentials credentials,
             String uri, String dir) throws URISyntaxException, StorageException {
+        return cloudBlobDirectoryFrom(credentials, uri, dir, false);
+    }
+
+    public static CloudBlobDirectory cloudBlobDirectoryFrom(StorageCredentials credentials,
+            String uri, String dir, boolean skipCreateContainer) throws URISyntaxException, StorageException {
         StorageUri storageUri = new StorageUri(new URI(uri));
         CloudBlobContainer container = new CloudBlobContainer(storageUri, credentials);
-
-        container.createIfNotExists();
-
+        if (!skipCreateContainer) {
+            container.createIfNotExists();
+        }
         return container.getDirectoryReference(dir);
     }
 
     public static CloudBlobDirectory cloudBlobDirectoryFrom(String connection, String containerName,
             String dir) throws InvalidKeyException, URISyntaxException, StorageException {
+        return cloudBlobDirectoryFrom(connection, containerName, dir, false);
+    }
+
+    public static CloudBlobDirectory cloudBlobDirectoryFrom(String connection, String containerName,
+            String dir, boolean skipCreateContainer) throws InvalidKeyException, URISyntaxException, StorageException {
         CloudStorageAccount cloud = CloudStorageAccount.parse(connection);
         CloudBlobContainer container = cloud.createCloudBlobClient().getContainerReference(containerName);
-        container.createIfNotExists();
-
+        if (!skipCreateContainer) {
+            container.createIfNotExists();
+        }
         return container.getDirectoryReference(dir);
     }
 
--- a/oak-upgrade/src/main/java/org/apache/jackrabbit/oak/upgrade/cli/node/SegmentAzureFactory.java
+++ b/oak-upgrade/src/main/java/org/apache/jackrabbit/oak/upgrade/cli/node/SegmentAzureFactory.java
@@ -24,6 +24,8 @@ import java.io.IOException;
 import java.net.URISyntaxException;
 import java.security.InvalidKeyException;
 
+import com.microsoft.azure.storage.StorageCredentialsSharedAccessSignature;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.jackrabbit.oak.segment.SegmentNodeStoreBuilders;
 import org.apache.jackrabbit.oak.segment.azure.AzurePersistence;
 import org.apache.jackrabbit.oak.segment.azure.AzureUtilities;
@@ -48,6 +50,8 @@ public class SegmentAzureFactory implements NodeStoreFactory {
     private final String connectionString;
     private final String containerName;
     private final String dir;
+    private final String sasToken;
+    private final boolean skipCreateContainer;
 
     private int segmentCacheSize;
     private final boolean readOnly;
@@ -61,6 +65,8 @@ public class SegmentAzureFactory implements NodeStoreFactory {
         private String uri;
         private String connectionString;
         private String containerName;
+        private String sasToken;
+        private boolean skipCreateContainer;
 
         public Builder(String dir, int segmentCacheSize, boolean readOnly) {
             this.dir = dir;
@@ -87,6 +93,14 @@ public class SegmentAzureFactory implements NodeStoreFactory {
             this.containerName = containerName;
             return this;
         }
+        public Builder sasToken(String sasToken) {
+            this.sasToken = sasToken;
+            return this;
+        }
+        public Builder skipCreateContainer(boolean skipCreateContainer) {
+            this.skipCreateContainer = skipCreateContainer;
+            return this;
+        }
 
         public SegmentAzureFactory build() {
             return new SegmentAzureFactory(this);
@@ -101,6 +115,8 @@ public class SegmentAzureFactory implements NodeStoreFactory {
         this.dir = builder.dir;
         this.segmentCacheSize = builder.segmentCacheSize;
         this.readOnly = builder.readOnly;
+        this.sasToken = builder.sasToken;
+        this.skipCreateContainer = builder.skipCreateContainer;
     }
 
     @Override
@@ -143,11 +159,16 @@ public class SegmentAzureFactory implements NodeStoreFactory {
         CloudBlobDirectory cloudBlobDirectory = null;
 
         if (accountName != null && uri != null) {
-            String key = System.getenv("AZURE_SECRET_KEY");
-            StorageCredentials credentials = new StorageCredentialsAccountAndKey(accountName, key);
-            cloudBlobDirectory = AzureUtilities.cloudBlobDirectoryFrom(credentials, uri, dir);
+            StorageCredentials credentials;
+            if (!StringUtils.isBlank(sasToken)) {
+                credentials = new StorageCredentialsSharedAccessSignature(sasToken);
+            } else {
+                String key = System.getenv("AZURE_SECRET_KEY");
+                credentials = new StorageCredentialsAccountAndKey(accountName, key);
+            }
+            cloudBlobDirectory = AzureUtilities.cloudBlobDirectoryFrom(credentials, uri, dir, skipCreateContainer);
         } else if (connectionString != null && containerName != null) {
-            cloudBlobDirectory = AzureUtilities.cloudBlobDirectoryFrom(connectionString, containerName, dir);
+            cloudBlobDirectory = AzureUtilities.cloudBlobDirectoryFrom(connectionString, containerName, dir, skipCreateContainer);
         }
 
         if (cloudBlobDirectory == null) {
--- a/oak-upgrade/src/main/java/org/apache/jackrabbit/oak/upgrade/cli/parser/MigrationOptions.java
+++ b/oak-upgrade/src/main/java/org/apache/jackrabbit/oak/upgrade/cli/parser/MigrationOptions.java
@@ -104,6 +104,8 @@ public class MigrationOptions {
 
     private final Boolean addSecondaryMetadata;
 
+    private final Boolean skipCreateAzureContainer;
+
     public MigrationOptions(MigrationCliArguments args) throws CliArgumentException {
         this.disableMmap = args.hasOption(OptionParserFactory.DISABLE_MMAP);
         this.copyBinaries = args.hasOption(OptionParserFactory.COPY_BINARIES);
@@ -168,6 +170,7 @@ public class MigrationOptions {
         } else {
             this.addSecondaryMetadata = Boolean.getBoolean(ADD_SECONDARY_METADATA_PROP);
         }
+        this.skipCreateAzureContainer = args.hasOption(OptionParserFactory.SKIP_CREATE_AZURE_CONTAINER);
     }
 
     public boolean isCopyBinaries() {
@@ -347,6 +350,10 @@ public class MigrationOptions {
         return isDstFbs() || isDstFds() || isDstS3() || isDstAzure();
     }
 
+    public boolean isSkipCreateAzureContainer() {
+        return skipCreateAzureContainer;
+    }
+
     public void logOptions() {
         if (disableMmap) {
             log.info("Disabling memory mapped file access for Segment Store");
@@ -406,6 +413,10 @@ public class MigrationOptions {
 
         if (addSecondaryMetadata) {
             log.info("Secondary metadata will be added");
+        }
+
+        if (skipCreateAzureContainer) {
+            log.info("Skipping creation of the Azure container, if it does not yet exists");
         }
 
         log.info("Cache size: {} MB", cacheSizeInMB);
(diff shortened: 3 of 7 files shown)


oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
```diff
@@ -116,6 +116,12 @@ public static CloudBlobDirectory cloudBlobDirectoryFrom(String connection, Strin
 
         return container.getDirectoryReference(dir);
     }
+    public static CloudBlobDirectory cloudBlobDirectoryFromSasUri(String connection, String containerName,
```
smiroslav: Not sure this method is needed, since the only difference in this method compared to ```cloudBlobDirectoryFrom``` is that container creation is skipped. Otherwise in the method implementation I do not see anything specific to SAS URI.

oak-upgrade/src/main/java/org/apache/jackrabbit/oak/upgrade/cli/node/SegmentAzureFactory.java
```diff
@@ -87,6 +90,10 @@ public Builder containerName(String containerName) {
             this.containerName = containerName;
             return this;
         }
+        public Builder sasToken(String sasToken) {
```
smiroslav, line 96: Is it necessary to explicitly set ```sasToken```? I have impression that it can be the part of ```connectionString```.
https://docs.microsoft.com/en-us/azure/storage/common/storage-configure-connection-string#create-a-connection-string-using-a-shared-access-signature

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
mattvryan: I'm concerned about this.  How accessible is this method?  In particular, is it exported so that it can be called outside the bundle?  I worry about security implications - it's giving access to cloud storage based on any valid connection string passed in to the method.

jelmini: @mattvryan This class seem to be exported and is accessed from the `oak-upgrade` module. However, I don't think it's creating a security issue. This class just wraps the Azure SDK. If someone has valid credentials, then it could just as well use the Azure SDK directly to successfully connect.

oak-upgrade/src/main/java/org/apache/jackrabbit/oak/upgrade/cli/parser/StoreType.java
```diff
@@ -166,6 +168,7 @@ public StoreFactory createFactory(String[] paths, MigrationDirection direction,
                                 direction == MigrationDirection.SRC)
                         .connectionString(config.get(KEY_CONNECTION_STRING))
                         .containerName(config.get(KEY_CONTAINER_NAME))
+                        .sasToken(config.get(KEY_SHARED_ACCESS_SIGNATURE))
```
jelmini: Setting the SAS token here is not strictly necessary because it's by definition already part of the connection string (given that it's extracted from the connection string by the call to `parseAzureConfigurationFromCustomConnection(path)` just above).
However, in the `else` branch below, if a SAS token is part of the URL parsed by `parseAzureConfigurationFromUri(path)` (where URL has form `https://path/container?sasToken`) then it should be set in the builder, otherwise the `AZURE_SECRET_KEY` env var will be used in that case.

oak-upgrade/src/main/java/org/apache/jackrabbit/oak/upgrade/cli/node/SegmentAzureFactory.java
jelmini: I think it can be useful in the case where an URL of form `https://path/container?sasToken` is used instead of a connection string.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
jelmini: I agree with @smiroslav that this new method is misleading. What do we need is the ability to skip creating the container. I suggest to add a boolean parameter `skipCreateContainer` to the existing `cloudBlobDirectoryFrom()` methods. The value could then come from `MigrationOptions`, so that when running the upgrade tool, the flag can be set at the command line.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
```diff
@@ -116,6 +116,15 @@ public static CloudBlobDirectory cloudBlobDirectoryFrom(String connection, Strin
 
         return container.getDirectoryReference(dir);
     }
+    public static CloudBlobDirectory cloudBlobDirectoryFrom(String connection, String containerName,
```
smiroslav, line 121: I believe this method is not necessary. Rather previous signature could be changed, and add **skipCreateContainer** parameter. Seems it is only used in org.apache.jackrabbit.oak.run.DataStoreCommandTest.StoreFixture.AzureSegmentStoreFixture#init

oak-upgrade/src/main/java/org/apache/jackrabbit/oak/upgrade/cli/node/SegmentAzureFactory.java
smiroslav: Ok, I see that this is set after parsing the connection in org.apache.jackrabbit.oak.upgrade.cli.parser.StoreType.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
jelmini: @smiroslav  It's me that suggested to add an overload of `cloudBlobDirectoryFrom()`, in order to keep backward compatibility in case 3rd-party code is using it. Do you think it's not necessary?

If we keep them, however, one overload should call the second, instead of re-implementing it:
```java
public static CloudBlobDirectory cloudBlobDirectoryFrom(String connection, String containerName,
            String dir) throws InvalidKeyException, URISyntaxException, StorageException {
  return cloudBlobDirectoryFrom(connection, containerName, dir, false);
}
```

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
```diff
@@ -116,6 +116,15 @@ public static CloudBlobDirectory cloudBlobDirectoryFrom(String connection, Strin
...
+                                                            String dir,boolean skipCreateContainer) throws InvalidKeyException, URISyntaxException, StorageException {
```
jelmini: Code formatting: for consistency with the rest of the code, please 
* add spaces after commas (int this line) and  before `{` (in the `if` below) 
* don't align parameter lists with the opening method `(`, a simple indent is sufficient:
```java
public static CloudBlobDirectory cloudBlobDirectoryFrom(String connection, String containerName,
        String dir, boolean skipCreateContainer) throws InvalidKeyException, URISyntaxException, StorageException {
...
}
```

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
smiroslav: @jelmini you are right. Util classes are also part of the public API and we should maintain compatibility. 
Your suggested approach can be used to prevent code duplication.

oak-upgrade/src/main/java/org/apache/jackrabbit/oak/upgrade/cli/node/SegmentAzureFactory.java
```diff
@@ -143,11 +159,16 @@ private AzurePersistence createAzurePersistence() throws StorageException, URISy
...
+            if (!StringUtils.isBlank(sasToken)) {
+                credentials = new StorageCredentialsSharedAccessSignature(sasToken);
+            } else {
+                String key = System.getenv("AZURE_SECRET_KEY");
+                credentials = new StorageCredentialsAccountAndKey(accountName, key);
+            }
             cloudBlobDirectory = AzureUtilities.cloudBlobDirectoryFrom(credentials, uri, dir);
         } else if (connectionString != null && containerName != null) {
-            cloudBlobDirectory = AzureUtilities.cloudBlobDirectoryFrom(connectionString, containerName, dir);
+            cloudBlobDirectory = AzureUtilities.cloudBlobDirectoryFrom(connectionString, containerName, dir,skipCreateContainer);
```
jelmini: Formatting: space after comma.

oak-upgrade/src/main/java/org/apache/jackrabbit/oak/upgrade/cli/parser/OptionParserFactory.java
```diff
@@ -174,5 +176,6 @@ private static void addMiscOptions(OptionParser op) {
         op.accepts(SKIP_CHECKPOINTS, "Don't copy checkpoints on the full segment->segment migration");
         op.accepts(FORCE_CHECKPOINTS, "Copy checkpoints even if the --include,exclude,merge-paths option is specified");
         op.accepts(ADD_SECONDARY_METADATA, "Adds the metadata required by secondary store");
+        op.accepts(SKIP_CREATE_AZURE_CONTAINER, "Create azure container if not exist");
```
jelmini: Text should be the other way around: "Skip creating the Azure container if it does not exist".

oak-upgrade/src/main/java/org/apache/jackrabbit/oak/upgrade/cli/parser/MigrationOptions.java
```diff
@@ -408,6 +415,10 @@ public void logOptions() {
             log.info("Secondary metadata will be added");
         }
 
+        if (!skipCreateAzureContainer) {
```
jelmini: Log message and `skipCreateAzureContainer` semantics are mismatched. I suggest:
```java
if (skipCreateAzureContainer) {
    log.info("Skipping creation of the Azure container, if it does not yet exists");
}
```

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
jelmini, line 121: Formatting: Please add an empty line between methods.

oak-upgrade/src/main/java/org/apache/jackrabbit/oak/upgrade/cli/parser/StoreType.java
```diff
@@ -176,6 +179,7 @@ public StoreFactory createFactory(String[] paths, MigrationDirection direction,
                         migrationOptions.getCacheSizeInMB(), direction == MigrationDirection.SRC)
                         .accountName(config.get(KEY_ACCOUNT_NAME))
                         .uri(config.get(KEY_STORAGE_URI))
+                        .sasToken(config.get(KEY_SHARED_ACCESS_SIGNATURE))
```
jelmini, line 182: Shouldn't `skipCreateContainer` be set here as well?

oak-upgrade/src/test/java/org/apache/jackrabbit/oak/upgrade/cli/SegmentTarToSegmentAzureWithSASUriTest.java
```diff
@@ -0,0 +1,140 @@
...
+import org.slf4j.LoggerFactory;
+
+import static com.microsoft.azure.storage.blob.SharedAccessBlobPermissions.READ;
+import static com.microsoft.azure.storage.blob.SharedAccessBlobPermissions.WRITE;
+import static com.microsoft.azure.storage.blob.SharedAccessBlobPermissions.ADD;
+import static com.microsoft.azure.storage.blob.SharedAccessBlobPermissions.CREATE;
+import static com.microsoft.azure.storage.blob.SharedAccessBlobPermissions.LIST;
+
+
+public class SegmentTarToSegmentAzureWithSASUriTest extends AbstractOak2OakTest {
```
jelmini, line 42: Please check formatting in this file. In several lines, spaces are missing. Applying the default IDE reformatter should be enough.

oak-upgrade/src/test/java/org/apache/jackrabbit/oak/upgrade/cli/SegmentTarToSegmentAzureWithSASUriTest.java
```diff
@@ -0,0 +1,140 @@
...
+        return source;
+    }
+
+    @Override
+    protected NodeStoreContainer getDestinationContainer() {
+        return destination;
+    }
+
+    @Before
+    public void prepare() throws Exception {
```
jelmini, line 78: This method overrides its superclass method and is duplicating a lot of code from there. I suggest to call the superclass method to avoid duplication:

```java
    @Override
    public void prepare() throws Exception {
        CloudBlobContainer container = azurite.getContainer(DEFAULT_CONTAINER_NAME);
        this.sasToken = container.generateSharedAccessSignature(policy(READ_WRITE), null);
        
        super.prepare();
    }
```
Adding the `@Override` annotation is also beneficial to make it clear that this method is actually an override.

oak-upgrade/src/test/java/org/apache/jackrabbit/oak/upgrade/cli/SegmentTarToSegmentAzureWithSASUriTest.java
```diff
@@ -0,0 +1,140 @@
...
+        String[] args = getArgs();
+        log.info("oak2oak {}", Joiner.on(' ').join(args));
+
+        OakUpgrade.main(args);
+        createSession();
+    }
+
+    @Override
+    protected String[] getArgs() {
+        String destinationString = destination.getDescription() + ";" + KEY_SHARED_ACCESS_SIGNATURE + "=" + sasToken;
```
jelmini: This code could be simplified and made more clear. I suggest:

```java
StringBuilder builder = new StringBuilder();
String[] properties = destination.getDescription().split(";");
for (String str : properties) {
    if (!str.startsWith(ACCOUNT_KEY)) {
        builder.append(str).append(";");
    }
}
String destinationString = builder.toString() + ";" + KEY_SHARED_ACCESS_SIGNATURE + "=" + sasToken;
return new String[] { source.getDescription(), destinationString, "--" + OptionParserFactory.SKIP_CREATE_AZURE_CONTAINER };
```

jelmini:
Please document the possibility to use SAS URIs in `upgrade_usage.txt`. The new `skip-create-azure-container` option should be documented there as well.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
smiroslav: @ArunOnCloud I think this is still not resolved. to prevent code duplication, the suggestion from @jelmini should be applied.

ArunOnCloud: @smiroslav Yes I missed re-implementation part.  I will resolve it in next commit, if everything else is ok.

smiroslav (approved):

oak-upgrade/src/main/java/org/apache/jackrabbit/oak/upgrade/cli/node/SegmentAzureFactory.java
```diff
@@ -143,11 +159,16 @@ private AzurePersistence createAzurePersistence() throws StorageException, URISy
...
-            String key = System.getenv("AZURE_SECRET_KEY");
-            StorageCredentials credentials = new StorageCredentialsAccountAndKey(accountName, key);
+            StorageCredentials credentials;
+            if (!StringUtils.isBlank(sasToken)) {
+                credentials = new StorageCredentialsSharedAccessSignature(sasToken);
+            } else {
+                String key = System.getenv("AZURE_SECRET_KEY");
+                credentials = new StorageCredentialsAccountAndKey(accountName, key);
+            }
             cloudBlobDirectory = AzureUtilities.cloudBlobDirectoryFrom(credentials, uri, dir);
```
jelmini: Shouldn't we call `AzureUtilities.cloudBlobDirectoryFrom(credentials, uri, dir, skipCreateContainer)` here as well? If credentials of type `StorageCredentialsSharedAccessSignature` are used, then the container creation may need to be skipped. 
We should add a new overload: `AzureUtilities#cloudBlobDirectoryFrom(StorageCredentials credentials,
            String uri, String dir, boolean skipCreateContainer)`

jelmini (approved):

kwin:
@smiroslav What should happen with this PR now? It was approved more than three months ago but never merged. Also I see no comment in here or the connected JIRA explaining why it is still lingering.

PR #513, opened 2022-03-09
OAK-9717 - Prevent Merge from Removing nodes

klcodanr:
Changes:

 - Ensuring custom index definitions do not remove nodes from product indexes in the IndexDefMergerUtils
 - Clean up / refactoring tests to reduce the number of test files


oak-run/src/main/java/org/apache/jackrabbit/oak/index/merge/IndexDefMergerUtils.java
```diff
@@ -163,7 +165,9 @@ private static JsonObject mergeChild(String path, String child, int level, JsonO
         if (level == 0 && USE_PRODUCT_CHILD_LEVEL_0.contains(child)) {
             return p;
         }
-        if (isSameJson(a, p) || isSameJson(c, p)) {
+        if(c == null && p != null){
+            return p;
+        }else if (isSameJson(a, p) || isSameJson(c, p)) {
```
thomasmueller: space before "else" is missing

oak-run/src/main/java/org/apache/jackrabbit/oak/index/merge/IndexDefMergerUtils.java
```diff
@@ -163,7 +165,9 @@ private static JsonObject mergeChild(String path, String child, int level, JsonO
         if (level == 0 && USE_PRODUCT_CHILD_LEVEL_0.contains(child)) {
             return p;
         }
-        if (isSameJson(a, p) || isSameJson(c, p)) {
+        if(c == null && p != null){
```
thomasmueller: Spaces around "()" are missing

oak-run/src/test/resources/org/apache/jackrabbit/oak/index/merge/merge.txt
```diff
@@ -13,12 +13,6 @@
   "product": {"o": {"a": 1, "p": 20}},
   "expected": {"merges": ["/oak:index/test-2", "/oak:index/test-1-custom-1"], "o": {"a": 2, "c": 10, "p": 20 }}
 },
-{
-  "ancestor": {"o": {"a": 1}},
-  "custom": {},
-  "product": {"o": {"a": 2}},
-  "expected":"Conflicts detected: [Could not merge value; path=/o property=a; ancestor=1; custom=null; product=2]"
-},
```
thomasmueller, line 21: Instead of removing the test case, what about just changing the "expected"?

oak-run/src/main/java/org/apache/jackrabbit/oak/index/merge/IndexDefMergerUtils.java
```diff
@@ -148,6 +148,8 @@ private static String mergeProperty(String path, String property, JsonObject anc
             return cp;
         } else if (Objects.equals(ap, cp)) {
             return pp;
+        } else if (ap == null) {
+            return cp;
```
thomasmueller: It would be good to add a comment, e.g.

    // if it doesn't exist in the ancestor, prefer customization over the product value

However, I'm not quite sure why this behavior is needed... shouldn't we prefer pp (product)? Or maybe cp == pp ?

oak-run/src/main/java/org/apache/jackrabbit/oak/index/merge/IndexDefMergerUtils.java
```diff
@@ -163,7 +165,9 @@ private static JsonObject mergeChild(String path, String child, int level, JsonO
         if (level == 0 && USE_PRODUCT_CHILD_LEVEL_0.contains(child)) {
             return p;
         }
-        if (isSameJson(a, p) || isSameJson(c, p)) {
+        if(c == null && p != null){
+            return p;
```
thomasmueller: I would add a comment here as well, e.g.

    // if customer removed it, then we put it back

thomasmueller (changes requested):

oak-run/src/main/java/org/apache/jackrabbit/oak/index/merge/IndexDefMergerUtils.java
klcodanr: My thought is that if the custom index defined the _same_ property name with a different value as is now in the product index, but that property was not present before, that we should preserve that custom value as the developer of the custom index would expect their customizations to be preserved not undone.

oak-run/src/main/java/org/apache/jackrabbit/oak/index/merge/IndexDefMergerUtils.java
klcodanr: Fixed

oak-run/src/main/java/org/apache/jackrabbit/oak/index/merge/IndexDefMergerUtils.java
klcodanr: Fixed

oak-run/src/test/resources/org/apache/jackrabbit/oak/index/merge/merge.txt
klcodanr: Done

oak-run/src/main/java/org/apache/jackrabbit/oak/index/merge/IndexDefMergerUtils.java
klcodanr: As per offline conversation, this _really_ shouldn't be the case and it would be inconsistent between missing and just different properties so both cases should be considered conflicts. Update to add tests as well.

thomasmueller (approved):

PR #516, opened 2022-03-11
OAK-9723 Allow comparing index definitions against a file

thomasmueller:


oak-run/src/main/java/org/apache/jackrabbit/oak/index/merge/IndexDiff.java
```diff
@@ -192,7 +192,17 @@ private static JsonObject addParent(String key, JsonObject obj) {
     static JsonObject compareIndexes(String directory, String index1, String index2) {
         Path indexPath = Paths.get(directory);
         JsonObject target = new JsonObject(true);
-        compareIndexesIndexesInDirectory(indexPath, index1, index2, target);
+        compareIndexesInDirectory(indexPath, index1, index2, target);
+        return target;
+    }
+
+    public static JsonObject compareIndexesAgainstBase(String directory, String indexBaseFile) {
```
klcodanr, line 199: Would it make sense to allow setting an output stream for the diff? That way you could either pipe it to standard out or write the diff to a file for further analysis.

thomasmueller: Sure it's possible. Right now we write it to stdout here:

    System.out.println("Comparing indexes in " + indexBaseFile + " against all indexes in " +
                        " directory \"" +
                        compareDirectory + "\"");
    System.out.println(IndexDiff.compareIndexesAgainstBase(compareDirectory, indexBaseFile));

Instead of stdout we could write it as a file to targetDirectory.

... but on the other hand, I don't need to further analyze it: for my use case, it is sufficient. So I argue: we can still change it quite easily once we need it...

PR #518, opened 2022-03-14
OAK-9724 Handle duplicate functions gracefully in indexing

thomasmueller:


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMaker.java
```diff
@@ -283,8 +283,10 @@ protected boolean indexTypeOrderedFields(Document doc, String pname, int tag, Pr
             }
 
             if (f != null && includePropertyValue(property, 0, pd)) {
-                doc.add(f);
-                fieldAdded = true;
+                if (doc.getField(f.name()) == null) {
+                    doc.add(f);
+                    fieldAdded = true;
+                }
```
nit0906, line 301: Should we add a log in the else part here , which might prompt the developer to remove the duplicity form the index def ?

nit0906, line 289: Should we add a log in the else part here , which might prompt the developer to remove the duplicity form the index def ?

nit0906 (changes requested):
LGTM , just a minor change requested around additional logging.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMaker.java
thomasmueller: Hm, we could, but then we need to limit the logging to e.g. one message per minute or so (otherwise there might be millions of log messages)

nit0906 (approved):

PR #520, opened 2022-03-16
OAK-9726 Improve index purge old version commands logs

FrancoisZhang:


reschke:
Please refer to *OAK* Jira tickets in commit messages and PRs.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -38,43 +43,59 @@
...
+     * @param purgeThresholdMillis  the threshold of time length since last time index time to determine, will purge if exceed that
+     * @param indexPaths            the index path or parent path
+     *
+     * @throws IOException
+     * @throws CommitFailedException
+     */
+    public void execute(NodeStore nodeStore, boolean isReadWriteRepository, long purgeThresholdMillis, List<String> indexPaths) throws IOException, CommitFailedException {
         if (!isReadWriteRepository) {
-            LOG.info("Repository connected in read-only mode. Use '--read-write' for write operations");
+            LOG.info("Repository is opened in read-only mode");
```
thomasmueller: I prefer the message above (with --read-write)

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -38,43 +43,59 @@
...
+            return;
```
thomasmueller: The old code allowed to just list the indexes that can be removed. Now this feature is gone... Why? I would prefer to keep it (it looks like it's quite easy to keep it).

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
```diff
@@ -38,43 +43,59 @@
...
+            purgeOldIndexVersion(nodeStore, purgeIndexList);
+            LOG.info("Index purging done, took '{}' ms", (System.currentTimeMillis() - purgeStart));
+        } else {
+            LOG.info("No indexes are found to be purged");
+        }
+    }
+
+    public List<IndexVersionOperation> getPurgeIndexes(NodeStore nodeStore, long purgeThresholdMillis, List<String> indexPaths) throws IOException, CommitFailedException {
+        List<IndexVersionOperation> purgeIndexList = new ArrayList<>();
+        LOG.info("Start execute purge index over node type '{}' over index paths '{}'", nodeStore.getClass().getSimpleName(), indexPaths);
```
thomasmueller: I would log "Getting indexes to purge ..." as in this method we don't purge.

thomasmueller (changes requested):

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
FrancoisZhang: @thomasmueller thanks, the reason I adjust it is the message fit for the case from oak run command line, but doesn't apply for the case if call it from tools image. Let me add it back, but bit earlier over PurgeOldIndexVersionCommand class

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
FrancoisZhang: @thomasmueller thanks, makes sense, will change it.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersion.java
FrancoisZhang: @thomasmueller yes, getting the list index is done in getPurgeIndexes and it will work when call from tools image. yes, I think makes sense to also list the indexes in oak run command line. will fix it.

FrancoisZhang:
@reschke sorry for that, just created an OAK ticket and fix the message in PR

@thomasmueller I have refined the commit over your code review, thanks!

thomasmueller (approved):

oak-run/src/main/java/org/apache/jackrabbit/oak/run/PurgeOldIndexVersionCommand.java
```diff
@@ -17,17 +17,24 @@
...
+import joptsimple.OptionParser;
+import joptsimple.OptionSet;
+import joptsimple.OptionSpec;
 
 public class PurgeOldIndexVersionCommand implements Command {
+    private static final Logger LOG = LoggerFactory.getLogger(PurgeOldIndexVersionCommand.class);
+
     private long threshold;
     private List<String> indexPaths;
     private long DEFAULT_PURGE_THRESHOLD = TimeUnit.DAYS.toMillis(5); // 5 days in millis
```
fabriziofortino, line 40: this should be `final`

FrancoisZhang: @fabriziofortino true, i's a legacy I didn't notice that. I will refine it together in next PR. seems need more logging here

PR #521, opened 2022-03-16
Update queary-engine.md changed code error.

stjepanB:
Variable row was not defined, so I changed it with already existing RowIterator it.


thomasmueller:
Great! Thanks a lot!

PR #524, opened 2022-03-21
OAK-9727 - ElasticSearch Index Function Support

klcodanr:
Most of the changes are to correct issues with the tests. 

The primary changes are to introduce a delay between the test changes and test execution to ensure that the Elastic index is updated by the time the assertions are run in the [FunctionIndexCommonTest.java](https://github.com/apache/jackrabbit-oak/compare/trunk...klcodanr:elastic-function-support?expand=1#diff-2d0a75bdc45709baff8c41f725c939f9f37e9414752b37ad03e335c2d40b33c7). This was implemented as an overridden method so that the Lucene indexing does not have any delays.  (see postCommitHook) 

The other change is to ensure that the functional restrictions are translated into Document properties when creating the [ElasticIndexDefinition.java](https://github.com/apache/jackrabbit-oak/compare/trunk...klcodanr:elastic-function-support?expand=1#diff-c047e881bb3fbf15e48f90204a88d5ac9b8f10c3ac6310098e40151a111f836d) and including them in the [ElasticIndexPlanner.java](https://github.com/apache/jackrabbit-oak/compare/trunk...klcodanr:elastic-function-support?expand=1#diff-747aaeec4a6e472c51175546720cc83729ce63ee0e44af12911a36b6e35b8d4c).

I did also reduce the re-indexing delay in the [ElasticTestRepositoryBuilder.java](https://github.com/apache/jackrabbit-oak/compare/trunk...klcodanr:elastic-function-support?expand=1#diff-102e564f471507afd5abb0fb9715a26201b9a46f2f3a3d9588a1fd1e9428e4fa) to reduce the impact of waiting for the changes to be indexed.

### DIFF
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
@@ -139,7 +139,7 @@ public class ElasticIndexDefinition extends IndexDefinition {
                         rule.getFunctionRestrictions().stream()))
                 .filter(pd -> pd.index) // keep only properties that can be indexed
                 .collect(Collectors.groupingBy(pd -> {
-                    if(pd.function != null){
+                    if (pd.function != null) {
                         return pd.function;
                     } else {
                         return pd.name;
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
@@ -18,21 +18,23 @@
  */
 package org.apache.jackrabbit.oak.plugins.index.elastic;
 
-import org.apache.jackrabbit.oak.api.Type;
-import org.apache.jackrabbit.oak.plugins.index.search.IndexDefinition;
-import org.apache.jackrabbit.oak.plugins.index.search.PropertyDefinition;
-import org.apache.jackrabbit.oak.spi.state.NodeState;
-import org.jetbrains.annotations.NotNull;
+import static org.apache.jackrabbit.oak.plugins.index.search.util.ConfigUtil.getOptionalValue;
+import static org.apache.jackrabbit.oak.plugins.index.search.util.ConfigUtil.getOptionalValues;
 
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
-import static org.apache.jackrabbit.oak.plugins.index.search.util.ConfigUtil.getOptionalValue;
-import static org.apache.jackrabbit.oak.plugins.index.search.util.ConfigUtil.getOptionalValues;
+import org.apache.jackrabbit.oak.api.Type;
+import org.apache.jackrabbit.oak.plugins.index.search.FieldNames;
+import org.apache.jackrabbit.oak.plugins.index.search.IndexDefinition;
+import org.apache.jackrabbit.oak.plugins.index.search.PropertyDefinition;
+import org.apache.jackrabbit.oak.spi.state.NodeState;
+import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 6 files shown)


oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
```diff
@@ -133,9 +135,16 @@ public ElasticIndexDefinition(NodeState root, NodeState defn, String indexPath,
 
         this.propertiesByName = getDefinedRules()
                 .stream()
-                .flatMap(rule -> StreamSupport.stream(rule.getProperties().spliterator(), false))
+                .flatMap(rule -> Stream.concat(StreamSupport.stream(rule.getProperties().spliterator(), false),
+                        rule.getFunctionRestrictions().stream()))
                 .filter(pd -> pd.index) // keep only properties that can be indexed
-                .collect(Collectors.groupingBy(pd -> pd.name));
+                .collect(Collectors.groupingBy(pd -> {
+                    if(pd.function != null){
```
thomasmueller: Nit: Please add spaces around ().

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticFunctionIndexCommonTest.java
```diff
@@ -16,42 +16,61 @@
...
 
+    @Override
+    protected String getIndexProvider() {
+        return "elasticsearch:";
+    }
+
+    @Override
+    protected void postCommitHook() {
+        try {
+            TimeUnit.SECONDS.sleep(2);
```
thomasmueller, line 56: Of course that's not that great... I assume we can't use synchronous indexes, and can't commit changes to Elasticsearch explicitly...

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
```diff
@@ -18,21 +18,23 @@
...
 import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
-import static org.apache.jackrabbit.oak.plugins.index.search.util.ConfigUtil.getOptionalValue;
-import static org.apache.jackrabbit.oak.plugins.index.search.util.ConfigUtil.getOptionalValues;
+import org.apache.jackrabbit.oak.api.Type;
+import org.apache.jackrabbit.oak.plugins.index.search.FieldNames;
```
fabriziofortino: unused import

thomasmueller (approved):

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticFunctionIndexCommonTest.java
```diff
@@ -16,42 +16,61 @@
...
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        }
+    }
+
```
fabriziofortino: We should not use this approach. We were using something similar in the past but tests were flaky.
As an alternative, I would remove the `postCommitHook` method and use `TestUtils.assertEventually` in the tests (see `IndexSuggestionCommonTest` for an example).

fabriziofortino (approved):
LGTM. I have just left a couple of comments.

This PR implements the Function feature as it is implemented in Lucene. That's okay for the time being. This is not optimized though.

The screenshot below shows that this feature leads to very long property names in the underneath index. We should avoid properties not defined in the mapping (using a strict mapping) to have more control over the index. This would require a lot more changes. I would create a separate PR for this optimization.

Screenshot 2022-03-21 at 17 16 22

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticFunctionIndexCommonTest.java
klcodanr: From my investigation while working on this, I didn't see a good way to either force the index to update or to just perform indexing asynchronously.  I opened another ticket to track improving the performance of this test:

https://issues.apache.org/jira/browse/OAK-9733

PR #525, opened 2022-03-21
OAK-9734 Index purge should prevent fully delete index definition whi…

FrancoisZhang:
…ch is is read-only repo

please ignore tests part above just yet, I’d really like you review the code change itself carefully to see if the change I made there is proper enough. some changes includes:

- add check over hidden oak mount. For inactive indexes, they will be disabled when hidden oak mount exists, and will be fully deleted when oak mount doesn’t exists anymore

- for disabled indexes, it was skipped for operation. But I added additional logic to check hidden oak mount as well, if oak mount not there anymore, it will do further fully deletion as well.

- since the index job completion time isn’t set proper per bug GRANITE-38843, we cannot assume some index isn’t active if no such field. Hence I changed the logic from original code. but not very confident enough for this per some legacy tests and assumption there. Please help to comment on that

- I got difficulty to create new tests for setup the hidden oak mount in current tests since the lack of support of composite node store in current fixture of oak-run code. Can you help to give some suggestion for how to resolve that?

### DIFF
--- a/oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
+++ b/oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
@@ -18,6 +18,11 @@
  */
 package org.apache.jackrabbit.oak.indexversion;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+
 import org.apache.jackrabbit.oak.api.Type;
 import org.apache.jackrabbit.oak.commons.PathUtils;
 import org.apache.jackrabbit.oak.plugins.index.search.IndexDefinition;
@@ -25,11 +30,6 @@ import org.apache.jackrabbit.oak.plugins.index.search.spi.query.IndexName;
 import org.apache.jackrabbit.oak.spi.state.NodeState;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.LinkedList;
-import java.util.List;
 
 /*
     Main operation of this class is to mark IndexName's with operations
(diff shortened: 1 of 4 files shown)


thomasmueller (commented):
It looks good in general... I would like to check if code+branch coverage for the changed code is nearly 100%, because it's easy to miss some edge case in the tests. Also, I need to verify if the asserts in the tests are correct

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -70,58 +70,93 @@ public String toString() {
...
+                long reindexCompletionTimeInMillis = PurgeOldVersionUtils.getMillisFromString(reindexCompletionTime);
+                long currentTimeInMillis = System.currentTimeMillis();
+                if (currentTimeInMillis - reindexCompletionTimeInMillis > purgeThresholdMillis) {
+                    LOG.info("Found active index {} is longer enough", indexName);
+                    return true;
+                } else {
+                    LOG.info("The last index time '{}' isn't longer enough for: {}", reindexCompletionTime, indexName);
+                }
+            } else {
+                LOG.warn("{} property is not set for index {}", IndexDefinition.REINDEX_COMPLETION_TIMESTAMP, indexName);
```
thomasmueller, line 144: I think the logic is correct... but how can we set the value if it is missing? Currently I'm not sure where we should set the property... Maybe we could set it right here (I think we have write access), to the current date.

oak-run/src/test/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersionTest.java
```diff
@@ -80,39 +85,63 @@ public void deleteOldIndexCompletely() throws Exception {
...
-        Assert.assertTrue("Index:" + "fooIndex-4-custom-2" + " deleted", fixture.getNodeStore().getRoot().getChildNode("oak:index").getChildNode("fooIndex-4-custom-2").exists());
+        Assert.assertFalse("Index:" + "fooIndex-2" + " deleted", indexRootNode.getChildNode("fooIndex-2").exists());
+        Assert.assertFalse("Index:" + "fooIndex-2-custom-1" + " deleted", indexRootNode.getChildNode("fooIndex-2-custom-1").exists());
+        Assert.assertFalse("Index:" + "fooIndex-3" + " deleted", indexRootNode.getChildNode("fooIndex-3").exists());
+        Assert.assertFalse("Index:" + "fooIndex-3-custom-1" + " deleted", indexRootNode.getChildNode("fooIndex-3-custom-1").exists());
+        Assert.assertFalse("Index:" + "fooIndex-3-custom-2" + " deleted", indexRootNode.getChildNode("fooIndex-3-custom-2").exists());
+        Assert.assertFalse("Index:" + "fooIndex" + " deleted", indexRootNode.getChildNode("fooIndex").exists());
+        Assert.assertEquals(indexRootNode.getChildNode("fooIndex-4").getProperty("type").getValue(Type.STRING), "disabled");
```
thomasmueller: Nit: I would swap the parameters (the literal "disabled" should come first)

thomasmueller (approved):
It looks OK to me.

oak-run/src/test/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersionTest.java
FrancoisZhang: @thomasmueller makes sense, missed this legacy tests, just made another commit for refining it

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
FrancoisZhang: @thomasmueller actually not, since the job will run twice, first run it in readonly mode for getting purging list. which is especially prevent do heavy segment clone and uploading for golden publish from beginning. hence this isn't a good place to do writing. seems still be better do this during AEM init.

PR #530, opened 2022-03-28
OAK-9744 Add check for StackOverflowError

Ewocker:
This PR prevent `StackOverflowError` for recursive call between function `isIndexActive` and `isIndexActiveMerged`.
It returns `true` by default if error is catched.

### DIFF
--- a/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/IndexName.java
+++ b/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/IndexName.java
@@ -245,9 +245,9 @@ public class IndexName implements Comparable<IndexName> {
             }
             return isIndexActiveMerged(indexNode, rootState);
         } catch (StackOverflowError e) {
-            LOG.error("Fail to check index activeness for {} with error {}", indexPath, e);
+            LOG.error("Fail to check index activeness for {} due to StackOverflowError", indexPath, e);
+            return true;
         }
-        return true;
     }
 
     private static boolean isIndexActiveMerged(NodeState indexNode, NodeState rootState) {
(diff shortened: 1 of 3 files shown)


oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/IndexName.java
```diff
@@ -233,16 +233,21 @@ public String nextCustomizedName() {
...
+                if (c.startsWith(":oak:mount-")) {
+                    return true;
+                }
+            }
+            return isIndexActiveMerged(indexNode, rootState);
+        } catch (StackOverflowError e) {
+            LOG.error("Fail to check index activeness for {} with error {}", indexPath, e);
         }
-        return isIndexActiveMerged(indexNode, rootState);
+        return true;
```
thomasmueller: OK, so if we get a StackOverflowError, then the index is marked active... This is OK in my view. I think it would be clearer if the "return" is moved into the "catch" part.

thomasmueller (changes requested):
We are missing a test case for this.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/IndexName.java
Ewocker: done

Ewocker:
> We are missing a test case for this.

added test case

oak-search/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,38 @@
+<!--
```
nit0906, line 1: Why do we need this separate logback config for ?

Ewocker: I believe I need this to use the LogCustomizer for comparing my log output

nit0906: ok

thomasmueller (approved):

PR #534, opened 2022-03-31
OAK-9743 oak-run indexing doesn't support facets

thomasmueller:


oak-run/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/LuceneIndexer.java
```diff
@@ -111,11 +114,22 @@ private void writeToIndex(Document doc, String path) throws IOException {
...
                 definition,
                 indexingRule,
                 path
         );
     }
+
+    @Override
+    public FacetsConfig getFacetsConfig() {
+        if (facetsConfig == null){
+            facetsConfig = FacetHelper.getFacetsConfig(definitionBuilder);
```
klcodanr, line 130: Would this be a good thing to log so the user can diagnose?

klcodanr (approved):

oak-run/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/LuceneIndexer.java
thomasmueller: I don't know if it's really needed... the old code just seems completely broken by design, and the TODO even admits that... it's just, the TODO was never fixed, and there is a complete lack of tests in this area. I would rather fix that instead of adding a way to analyze... just my opinion.

oak-run/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/LuceneIndexerTest.java
```diff
@@ -54,6 +56,29 @@ public void nodeIndexed_WithIncludedPaths() throws Exception{
...
+        NodeState defn = idxb.build();
+        IndexDefinition idxDefn = new IndexDefinition(root, defn, "/oak:index/testIndex");
+
+        NodeBuilder builder = root.builder();
+        LuceneIndexer indexer = new LuceneIndexer(idxDefn, mock(LuceneIndexWriter.class), builder,
+                mock(FulltextBinaryTextExtractor.class), mock(IndexingProgressReporter.class));
+
+        FacetsConfig config1 = indexer.getFacetsConfig();
+        FacetsConfig config2 = indexer.getFacetsConfig();
+        assertTrue(config1 == config2);
```
thomasmueller, line 78: Yes, I know it's kind of cheating... whitebox testing...

nit0906 (changes requested):
@thomasmueller  - Changes look good to me.

Regarding the test,  I think we can possibly add a test here https://github.com/apache/jackrabbit-oak/blob/trunk/oak-run/src/test/java/org/apache/jackrabbit/oak/index/DocumentStoreIndexerIT.java , where we should be able to reindex using oak-run and then test a facet query in assertion for correct results. I am ok if we want to take this as a separate task too, since this change is important to go in. 

But if it's easier to write the test case - then would appreciate if we can do it as part of this PR only.

nit0906 (approved):

thomasmueller:
> Regarding the test, I think we can possibly add a test here https://github.com/apache/jackrabbit-oak/blob/trunk/oak-run/src/test/java/org/apache/jackrabbit/oak/index/DocumentStoreIndexerIT.java , where we should be able to reindex using oak-run and then test a facet query in assertion for correct results. I am ok if we want to take this as a separate task too, since this change is important to go in.
> 
> But if it's easier to write the test case - then would appreciate if we can do it as part of this PR only.

I fully agree! I'll create an issue for that.

PR #535, opened 2022-03-31
OAK-9740 - Offset / Limit Option

klcodanr:
Adds support for setting the offset / limit via options within the query body in the form:

    SELECT * FROM [nt:base] OPTION(LIMIT 10, OFFSET 10)
   
Note this also supports XPath:

    /jcr:root/test/*/nt:resource[@jcr:encoding] option(offset 2, limit 75)

Any settings on the query will override the values parsed out of the query options.

### DIFF
--- a/oak-api/src/main/java/org/apache/jackrabbit/oak/api/QueryEngine.java
+++ b/oak-api/src/main/java/org/apache/jackrabbit/oak/api/QueryEngine.java
@@ -21,6 +21,7 @@ import static java.util.Collections.emptyMap;
 import java.text.ParseException;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
 import org.osgi.annotation.versioning.ProviderType;
@@ -89,6 +90,24 @@ public interface QueryEngine {
      */
     Result executeQuery(
             String statement, String language, long limit, long offset,
+            Map<String, ? extends PropertyValue> bindings,
+            Map<String, String> mappings) throws ParseException;
+
+    /**
+     * Execute a query and get the result.
+     *
+     * @param statement the query statement
+     * @param language the language
+     * @param limit the maximum result set size (may not be negative but may be empty)
+     * @param offset the number of rows to skip (may not be negative but may be empty)
+     * @param bindings the bind variable value bindings
+     * @param mappings namespace prefix mappings
+     * @return the result
+     * @throws ParseException if the statement could not be parsed
+     * @throws IllegalArgumentException if there was an error executing the query
+     */
+    Result executeQuery(
+            String statement, String language, Optional<Long> limit, Optional<Long> offset,
             Map<String, ? extends PropertyValue> bindings,
             Map<String, String> mappings) throws ParseException;
     
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineImpl.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineImpl.java
@@ -264,7 +264,7 @@ public abstract class QueryEngineImpl implements QueryEngine {
         List<Query> queries = parseQuery(statement, language, context, mappings);
 
         long actualLimit = getValue(queries, limit, Query::getLimit, Long.MAX_VALUE);
-        long actualOffset = getValue(queries, limit, Query::getOffset, 0L);
+        long actualOffset = getValue(queries, offset, Query::getOffset, 0L);
 
         for (Query q : queries) {
             q.setExecutionContext(context);
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineImpl.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineImpl.java
@@ -24,14 +24,15 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.jackrabbit.oak.api.PropertyValue;
 import org.apache.jackrabbit.oak.api.QueryEngine;
 import org.apache.jackrabbit.oak.api.Result;
-import org.apache.jackrabbit.oak.namepath.impl.LocalNameMapper;
 import org.apache.jackrabbit.oak.namepath.NamePathMapper;
+import org.apache.jackrabbit.oak.namepath.impl.LocalNameMapper;
 import org.apache.jackrabbit.oak.namepath.impl.NamePathMapperImpl;
 import org.apache.jackrabbit.oak.query.ast.NodeTypeInfoProvider;
 import org.apache.jackrabbit.oak.query.stats.QueryStatsData.QueryExecutionStats;
@@ -243,18 +244,12 @@ public abstract class QueryEngineImpl implements QueryEngine {
             Map<String, String> mappings) throws ParseException {
         return executeQuery(statement, language, Long.MAX_VALUE, 0, bindings, mappings);
     }
-    
+
     @Override
     public Result executeQuery(
-            String statement, String language, long limit, long offset,
+            String statement, String language, Optional<Long> limit, Optional<Long> offset,
             Map<String, ? extends PropertyValue> bindings,
             Map<String, String> mappings) throws ParseException {
-        if (limit < 0) {
-            throw new IllegalArgumentException("Limit may not be negative, is: " + limit);
-        }
-        if (offset < 0) {
-            throw new IllegalArgumentException("Offset may not be negative, is: " + offset);
-        }
 
         // avoid having to deal with null arguments
         if (bindings == null) {
@@ -266,23 +261,26 @@ public abstract class QueryEngineImpl implements QueryEngine {
 
         ExecutionContext context = getExecutionContext();
         List<Query> queries = parseQuery(statement, language, context, mappings);
-        
+
+        long actualLimit = getLimit(queries, limit);
+        long actualOffset = getOffset(queries, offset);
+
         for (Query q : queries) {
             q.setExecutionContext(context);
-            q.setLimit(limit);
-            q.setOffset(offset);
+            q.setLimit(actualLimit);
+            q.setOffset(actualOffset);
             if (bindings != null) {
                 for (Entry<String, ? extends PropertyValue> e : bindings.entrySet()) {
                     q.bindValue(e.getKey(), e.getValue());
                 }
             }
-            q.setTraversalEnabled(traversalEnabled);            
+            q.setTraversalEnabled(traversalEnabled);
         }
 
         boolean mdc = false;
         try {
             long start = System.nanoTime();
-            Query query = prepareAndSelect(queries); 
+            Query query = prepareAndSelect(queries);
             query.getQueryExecutionStats().execute(System.nanoTime() - start);
             mdc = setupMDC(query);
             return query.executeQuery();
@@ -292,7 +290,38 @@ public abstract class QueryEngineImpl implements QueryEngine {
             }
         }
     }
-    
+
+    @Override
+    public Result executeQuery(
+            String statement, String language, long limit, long offset,
+            Map<String, ? extends PropertyValue> bindings,
+            Map<String, String> mappings) throws ParseException {
+        return executeQuery(statement, language, Optional.of(limit), Optional.of(offset), bindings, mappings);
+
+    }
+
+    private long getOffset(List<Query> queries, Optional<Long> passedOffset) {
+        if (!passedOffset.isPresent()) {
+            return queries.stream().filter(q -> q.getOffset() >= 0).findFirst().map(Query::getOffset)
+                    .orElse(0L);
+        }
+        if (passedOffset.get() < 0) {
+            throw new IllegalArgumentException("Offset may not be negative, is: " + passedOffset.get());
+        }
+        return passedOffset.get();
+    }
+
+    private long getLimit(List<Query> queries, Optional<Long> passedLimit) {
+        if (!passedLimit.isPresent()) {
+            return queries.stream().filter(q -> q.getLimit() > 0).findFirst().map(Query::getLimit)
+                    .orElse(Long.MAX_VALUE);
+        }
+        if (passedLimit.get() < 0) {
+            throw new IllegalArgumentException("Limit may not be negative, is: " + passedLimit.get());
+        }
+        return passedLimit.get();
+    }
+
     /**
      * Prepare all the available queries and by based on the {@link QuerySelectionMode} flag return
      * the appropriate.
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineImpl.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineImpl.java
@@ -302,7 +302,7 @@ public abstract class QueryEngineImpl implements QueryEngine {
 
     private long getOffset(List<Query> queries, Optional<Long> passedOffset) {
         if (!passedOffset.isPresent()) {
-            return queries.stream().filter(q -> q.getOffset() >= 0).findFirst().map(Query::getOffset)
+            return queries.stream().map(Query::getOffset).filter(Optional::isPresent).map(Optional::get).findFirst()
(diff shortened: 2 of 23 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/query/xpath/Statement.java
```diff
@@ -376,6 +376,21 @@ private static void appendQueryOptions(StringBuilder buff, QueryOptions queryOpt
...
+            buff.append("offset ");
+            buff.append(queryOptions.offset);
+            optionCount++;
+        }
+        if (queryOptions.limit != -1) {
+            if (optionCount > 0) {
+                buff.append(", ");
+            }
+            buff.append("limit ");
+            buff.append(queryOptions.limit);
```
thomasmueller: missing optionCount++... I know, it's not needed here really, but it's easy to forget to add it later...
Actually, it's probably better to add the options first into an ArrayList<String>, and then use String.join(", ", list)...

oak-core/src/main/java/org/apache/jackrabbit/oak/query/SQL2Parser.java
```diff
@@ -290,6 +294,14 @@ private SelectorImpl parseSelector() throws ParseException {
 
         return factory.selector(nodeTypeInfo, selectorName);
     }
+
+    private long readBase10WholeNumber() throws ParseException {
+        String label = readName();
+        if (!label.matches("\\d*") || label.isEmpty() || label.length() > 19 /* Length of MAX_LONG */) {
```
thomasmueller: I don't think the "if" condition is needed; instead I would use catch (NumberFormatException nfe) (same as below).

oak-core/src/main/java/org/apache/jackrabbit/oak/query/SQL2Parser.java
```diff
@@ -290,6 +294,14 @@ private SelectorImpl parseSelector() throws ParseException {
 
         return factory.selector(nodeTypeInfo, selectorName);
     }
+
+    private long readBase10WholeNumber() throws ParseException {
```
thomasmueller: I would rename to readNumber()

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineImpl.java
```diff
@@ -292,7 +290,38 @@ public Result executeQuery(
...
+            return queries.stream().map(Query::getOffset).filter(Optional::isPresent).map(Optional::get).findFirst()
+                    .orElse(0L);
+        }
+        if (passedOffset.get() < 0) {
+            throw new IllegalArgumentException("Offset may not be negative, is: " + passedOffset.get());
+        }
+        return passedOffset.get();
+    }
+
+    private long getLimit(List<Query> queries, Optional<Long> passedLimit) {
```
thomasmueller: static?

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineImpl.java
```diff
@@ -292,7 +290,38 @@ public Result executeQuery(
...
+    @Override
+    public Result executeQuery(
+            String statement, String language, long limit, long offset,
+            Map<String, ? extends PropertyValue> bindings,
+            Map<String, String> mappings) throws ParseException {
+        return executeQuery(statement, language, Optional.of(limit), Optional.of(offset), bindings, mappings);
+
+    }
+
+    private long getOffset(List<Query> queries, Optional<Long> passedOffset) {
```
thomasmueller: (I didn't compile this sorry) but could it be static?

thomasmueller (changes requested):
It is a partial review

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineImpl.java
```diff
@@ -292,7 +290,38 @@ public Result executeQuery(
...
+        }
+        if (passedOffset.get() < 0) {
+            throw new IllegalArgumentException("Offset may not be negative, is: " + passedOffset.get());
+        }
+        return passedOffset.get();
+    }
+
+    private static long getLimit(List<Query> queries, Optional<Long> passedLimit) {
+        if (!passedLimit.isPresent()) {
+            return queries.stream().map(Query::getLimit).filter(Optional::isPresent).map(Optional::get).findFirst()
```
thomasmueller: I would check for IllegalArgumentException as well here

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineImpl.java
```diff
@@ -292,7 +290,38 @@ public Result executeQuery(
...
+                    .orElse(Long.MAX_VALUE);
+        }
+        if (passedLimit.get() < 0) {
+            throw new IllegalArgumentException("Limit may not be negative, is: " + passedLimit.get());
+        }
+        return passedLimit.get();
```
thomasmueller: I would call the getter only once... Maybe:

```
long result = passedLimit.orElse(...)
if (result < 0) {
    throw new IllegalArgumentException("Limit may not be negative, is: " + result);
}
return result;
```

thomasmueller (commented):
Sorry, again I didn't have enough time to do a full review...

oak-core/src/main/java/org/apache/jackrabbit/oak/query/xpath/Statement.java
klcodanr: Makes sense, I updated to add to a list and then join the list.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineImpl.java
klcodanr: Both methods made static.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineImpl.java
klcodanr: I don't believe an IllegalArgumentException can be thrown here as getLimit is only returning the previously parsed limit optional

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineImpl.java
```diff
@@ -292,7 +290,39 @@ public Result executeQuery(
...
+            return queries.stream().map(Query::getLimit).filter(Optional::isPresent).map(Optional::get).findFirst()
+                    .orElse(Long.MAX_VALUE);
+        }
+        long limit = passedLimit.get();
+        if (limit < 0) {
+            throw new IllegalArgumentException("Limit may not be negative, is: " + passedLimit.get());
+        }
+        return limit;
+    }
+
```
fabriziofortino: these two functions do pretty much the same. They can be replaced by:

```java
private static long getValue(List<Query> queries, Optional<Long> value, Function<Query, Optional<Long>> getter, Long defaultValue) {
        if (!value.isPresent()) {
            return queries.stream().map(getter::apply).filter(Optional::isPresent).map(Optional::get).findFirst()
                    .orElse(defaultValue);
        }
        if (value.get() < 0) {
            throw new IllegalArgumentException("Value may not be negative, is: " + value.get());
        }
        return value.get();
    }
```

oak-core/src/main/java/org/apache/jackrabbit/oak/query/xpath/Statement.java
```diff
@@ -353,29 +355,23 @@ private static void appendQueryOptions(StringBuilder buff, QueryOptions queryOpt
...
-            optionCount++;
+            optionValues.add("index tag [" + queryOptions.indexTag + "]");
+        }
+        if (queryOptions.offset != -1) {
+            optionValues.add("offset " + queryOptions.offset);
+        }
+        if (queryOptions.limit != -1) {
+            optionValues.add("limit " + queryOptions.limit);
         }
+        buff.append(optionValues.stream().collect(Collectors.joining(", ")));
```
fabriziofortino: this can be simplified with 
```suggestion
        buff.append(String.join(", ", optionValues));
```

oak-doc/src/site/markdown/query/query-engine.md
```diff
@@ -153,6 +154,26 @@ This is supported for both XPath and SQL-2, as follows:
     order by name()
     option(traversal ok)
 
+
+#### Query Option Offset / Limit
+
+`@since Oak 1.44.0 (OAK-9740)`
+
+By setting the offset / limit of a query you can set the limits and offsets set on the query object.
+Note, this setting will be overriden by any settings made via the `Query#setOffset` or `Query#setLimit` methods.
```
fabriziofortino: typo
```suggestion
Note, this setting will be overridden by any settings made via the `Query#setOffset` or `Query#setLimit` methods.
```

oak-api/src/main/java/org/apache/jackrabbit/oak/api/QueryEngine.java
```diff
@@ -91,6 +92,24 @@ Result executeQuery(
...
+     * @param limit the maximum result set size (may not be negative but may be empty)
+     * @param offset the number of rows to skip (may not be negative but may be empty)
+     * @param bindings the bind variable value bindings
+     * @param mappings namespace prefix mappings
+     * @return the result
+     * @throws ParseException if the statement could not be parsed
+     * @throws IllegalArgumentException if there was an error executing the query
+     */
+    Result executeQuery(
+            String statement, String language, Optional<Long> limit, Optional<Long> offset,
```
fabriziofortino, line 110: Some people suggest NOT using `Optional` as type parameters but only as return types. There are a lot of threads around this topic (see https://medium.com/@yassinhajaj/optionals-are-bad-practices-still-bad-practices-if-everyone-practices-them-6ec5a66c30aa).

IMHO it depends on the case. In this case, though, I would be in favor of removing it. The main reason is that `Optional` and `Long` are two wrappers. This double wrapping is hard to read and suboptimal for the compiler.

@thomasmueller I would like to hear your opinion on this

fabriziofortino (commented):
LGTM.

I have left a few comments on possible improvements.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/QueryEngine.java
thomasmueller: @fabriziofortino 

I don't have a strong opinion about this.

As for the performance reason: for a method that is called a lot (like: put, log,...) , I would avoid Optional for performance reasons. But executeQuery isn't called that often, so that's not a good reason to avoid it.

I see some advantages of Optional over magic values (-1 for infinity: make sure you always treat the magic value correctly), and some disadvantages (a bit harder to read).

I would leave it up to the author, I think...

oak-core/src/main/java/org/apache/jackrabbit/oak/query/SQL2Parser.java
```diff
@@ -182,6 +182,10 @@ public Query parse(final String query, final boolean initialise) throws ParseExc
                     } else if (readIf("TAG")) {
                         options.indexTag = readLabel();
                     }
+                } else if (readIf("OFFSET")) {
+                    q.setOffset(readNumber());
```
thomasmueller, line 186: So we don't support bind variable here... I think it perfectly fine: I anyway wouldn't want people to use offset for pagination.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryOptions.java
```diff
@@ -21,6 +21,8 @@ public class QueryOptions {
     public Traversal traversal = Traversal.DEFAULT;
     public String indexName;
     public String indexTag;
+    public long limit = -1;
```
thomasmueller: Here we use magic values... Shouldn't we use Optional as well here?

oak-api/src/main/java/org/apache/jackrabbit/oak/api/QueryEngine.java
thomasmueller: I think Optional has some advantage for the case where we have multiple "layers", like we have here: one could set the limit using the JCR API Query.setLimit (I think), or via query statement "... option(limit <n>)". If we use magic values, then it's harder to read the code (for humans). I think. Depending on which one has higher precedence:

    long limit = queryLimit.orElse(statementLimit.orElse(Long.MAX_VALUE));

or

    long limit = statementLimit.orElse(queryLimit.orElse(Long.MAX_VALUE));

If you want to do the same with magic values (say, -1), it would be:

    long limit = queryLimit >= 0 ? queryLimit : statementLimit >= 0 ? statementLimit : Long.MAX_VALUE;

or

    long limit = statementLimit >= 0 ? statementLimit : queryLimit >= 0 ? queryLimit : Long.MAX_VALUE;

... which is harder to read.

But then we should use Optional consistently (avoid magic values where possible).

klcodanr: @thomasmueller I agree. While it is double wrapping, it makes it clearer what the intent is. 

It also _seems_ like if you try to use a null Long Java will auto box the long -> Long resulting in the wrong method being called and failures executing queries. At least I think that's the issue. I tried doing this and running the tests and I _think_ that this is causing the tests to fail because it's calling the wrong method. As you said this can be handled through magic values but that makes it harder to evaluate if the user really intended (or messed up w/) Long.MAX_VALUE/-1 or just didn't set a value.

adamcin: Just a drive-by comment here, but instead of `Optional<Long>`, could you define a `Limit` type to serve as the wrapper for the limit long, and an `Offset` type to serve as the wrapper for the offset long? Then you would also eliminate the potential for callers accidentally reversing the order of the arguments.

Give both types factory methods like `.of(long)` and `.none()` and an instance method `Optional<Long> toOptional()`, or `boolean isPresent()` and `long getValue()`.

thomasmueller: I agree with you @adamcin , but I think this case here doesn't warrant this yet. Another solution would be to have a class "LimitOffset" with a builder (methods "limit" and "offset"), but I think the number of callers of this method, and the number of parameters so far, is so low that it's not needed yet.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -1102,13 +1113,16 @@ private SelectorExecutionPlan getBestSelectorExecutionPlan(
...
                 AdvancedQueryIndex advIndex = (AdvancedQueryIndex) index;
-                long maxEntryCount = limit;
-                if (offset > 0) {
-                    if (offset + limit < 0) {
+
+                long localLimit = limit.orElse(Long.MAX_VALUE);
+                long localOffset = offset.orElse(0L);
+                long maxEntryCount = localLimit;
+                if (localOffset > 0) {
+                    if (localOffset + localLimit < 0) {
```
thomasmueller: This is not really related to your patch...

Yes it works, Long.MAX_VALUE + Long.MAX_VALUE == -2.
We could also use QueryImpl.saturatedAdd, which is slower, as it's using BigInteger. 
I wonder, should we move the the logic outside of the loop?

klcodanr: It's a good point though, updated

klcodanr:
@thomasmueller  I believe I've addressed your feedback, can you re:review?

oak-core/src/main/java/org/apache/jackrabbit/oak/query/xpath/XPathToSQL2Converter.java
```diff
@@ -397,7 +398,11 @@ private Statement convertToStatement(String query) throws ParseException {
                     } else if (readIf("tag")) {
                         options.indexTag = readIdentifier();
                     }
-                } else {
+                } else if (readIf("offset")) {
+                    options.offset = Optional.of(readNumber());
+                } else if (readIf("limit")) {
+                    options.limit = Optional.of(readNumber());
+                }else {
```
thomasmueller: missing space

thomasmueller (approved):

PR #536, opened 2022-04-02
OAK-9747 Download resume needs to save progress and handle hidden node on exception

Ewocker:
## Problem
```
V1, V2, H1, H2, H3, H4, V3, V4, V5, V6, V7
```
Suppose TraverseAndSortTask has downloaded till H3 and last dump is `V1`. As per current code, TraverseAndSortTask has last-saved reference to `V2`, and last dump is till `V1` (due to memory low). 
Now mongo connection break up. With current implementation The progress will be saved as last-saved reference to `V2`, and last dumped `V1`.
This will result in duplication of download.

## Fix
Change will make the state to be as last-saved reference to `H3`__(handle hidden node)__, and last dumped `V2`__(dump all progress on exception)__.

### DIFF
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
@@ -19,16 +19,6 @@
 
 package org.apache.jackrabbit.oak.index.indexer.document;
 
-import java.io.Closeable;
-import java.io.File;
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.Predicate;
-
 import com.codahale.metrics.MetricRegistry;
 import com.google.common.base.Stopwatch;
 import com.google.common.io.Closer;
@@ -65,6 +55,16 @@ import org.apache.jackrabbit.oak.spi.state.NodeStore;
 import org.apache.jackrabbit.oak.stats.StatisticsProvider;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.io.Closeable;
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Predicate;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileNodeStoreBuilder.OAK_INDEXER_SORTED_FILE_PATH;
@@ -115,7 +115,8 @@ public abstract class DocumentStoreIndexerBase implements Closeable{
         }
 
         private MongoNodeStateEntryTraverserFactory(RevisionVector rootRevision, DocumentNodeStore documentNodeStore,
-                                                    MongoDocumentStore documentStore, Logger traversalLogger, CompositeIndexer indexer, Predicate<String> pathPredicate) {
+                                                    MongoDocumentStore documentStore, Logger traversalLogger, CompositeIndexer indexer,
+                                                    Predicate<String> pathPredicate) {
             this.rootRevision = rootRevision;
             this.documentNodeStore = documentNodeStore;
             this.documentStore = documentStore;
@@ -140,8 +141,7 @@ public abstract class DocumentStoreIndexerBase implements Closeable{
                                     throw new RuntimeException(e);
                                 }
                                 traversalLogger.trace(id);
-                            })
-                            .withPathPredicate((pathPredicate != null) ? pathPredicate : indexer::shouldInclude);
+                            });
         }
     }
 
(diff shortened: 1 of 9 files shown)


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/NodeStateEntryTraverser.java
```diff
@@ -176,12 +175,6 @@ private boolean includeId(String id) {
         }
 
         String path = Utils.getPathFromId(id);
-
-        //Exclude hidden nodes from index data
-        if (NodeStateUtils.isHiddenPath(path)){
-            return false;
-        }
-
         return pathPredicate.test(path);
```
thomasmueller: I think we should also move pathPredicate.test, not only NodeStateUtils.isHiddenPath... right? I mean, assuming there is a path predicate like "/content/project" then we only store the intermediate progress for nodes within "/content/project"... and that could in theory be: no node at all. So on each scale up / scale down, we download from scratch.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -172,6 +172,7 @@ private FlatFileStore buildFlatFileStore(NodeState checkpointedState, CompositeI
                         .withBlobStore(indexHelper.getGCBlobStore())
                         .withPreferredPathElements((preferredPathElements != null) ? preferredPathElements : indexer.getRelativeIndexedNodeNames())
                         .addExistingDataDumpDir(indexerSupport.getExistingDataDumpDir())
+                        .withPathPredicate((pathPredicate != null) ? pathPredicate : indexer::shouldInclude)
```
thomasmueller: We have removed the check for hidden nodes (NodeStateUtils.isHiddenPath), but I don't see where we add the check in the higher level (that is: here), in case there is a pathPredicate... Does pathPredicate do a check for hidden nodes? If yes, why did we check for both pathPredicate.test(path) AND hidden nodes before? (Maybe it's correct, but possibly not).

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/NodeStateEntryTraverser.java
Ewocker: removed and refactored

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
Ewocker: removed and refactored

check is in TraverseAndSortTask

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunner.java
```diff
@@ -297,14 +332,20 @@ private class Task implements Callable<File> {
...
-        public File call() {
+        public File call() throws Exception {
             try {
-                if (merge(mergeTarget, mergedFile)) {
-                    log.info("merge complete for {}", mergedFile.getName());
-                    return mergedFile;
+                String mergedFileName = mergedFile.getName();
+                if (mergeCancelled.get()) {
+                    log.debug("merge cancelled, skipping merge task");
+                    throw new Exception("merge skipped for " + mergedFileName);
```
thomasmueller: It would be good to use a different exception here, for example java.io.EOFException

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTask.java
```diff
@@ -176,7 +181,17 @@ public List<File> call() {
 
             return sortedFiles;
         } catch (IOException e) {
-            log.error(taskID + " could not complete download ", e);
+            log.error(taskID + " could not complete download with ", e);
+        } catch (Exception e) {
```
thomasmueller: Not sure if we need to catch all exceptions... Often it is better, yes. But what is the exception you expect? Sometimes it's better to explicitly mention the exception you expect (e.g. EOFException), and have separate code for the exception types you don't expect (java.lang.Exception)

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunner.java
Ewocker: changed

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseAndSortTask.java
Ewocker: This will not be `EOFException` because we do want to skip those, and will catch the __merge failure__ case which originally was throwing just `Exception` and now I change it to `RuntimeException` (correct me if there's a better one... not sure what is correct to use here). So changing it to now catch `RuntimeException` instead.

thomasmueller (approved):

thomasmueller:
Merged at 2022-04-20

PR #537, opened 2022-04-06
OAK-9746 : DefaultSyncHandler.syncProperties should sync Supplier type propertiesv

daniancu:
Add support for synchronising user profile properties that implement Supplier interface.

### DIFF
--- a/oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/TestIdentityProvider.java
+++ b/oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/TestIdentityProvider.java
@@ -16,12 +16,14 @@
  */
 package org.apache.jackrabbit.oak.spi.security.authentication.external;
 
+import java.math.BigDecimal;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Supplier;
 
 import javax.jcr.Credentials;
 import javax.jcr.SimpleCredentials;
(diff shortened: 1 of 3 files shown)


anchela:
hi @daniancu , i will look into this tomorrow. one trivial thing just before i get started: could you please create JIRA ticket in https://issues.apache.org/jira/browse/OAK-9745?jql=project%20%3D%20%22Jackrabbit%20Oak%22%20?
we need a dedicated jira for any kind of modifications to the code base in order to be able to make sure we have a comprehensive list for releases (and also for awareness of the community). thanks a lot.

oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/TestIdentityProvider.java
```diff
@@ -283,4 +285,20 @@ public ForeignExternalGroup() {
             return ImmutableList.of();
         }
     }
+    public static class UserWithSupplierProperties extends TestUser {
```
anchela: trivial nitpicking: there should be an extra line separating the UserWithSupplierProperties from the previous class.

oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/basic/DefaultSyncContextTest.java
```diff
@@ -1201,6 +1201,36 @@ public void testSyncPropertiesRemapped() throws Exception {
         }
     }
 
+    @Test
```
anchela, line 1207: minor wish: i don't do super consistently either but having a javadoc above the test inciting for which JIRA ticket this test was added is usually helpful if one needs to investigate later on.
since a dedicated jira ticket is anyway needed for the commit message you may want to add it as a comment here at the same time.

anchela (changes requested):
the patch looks very good. thanks a lot!
i just have 2 trivial findings that you may want to address.

but what we need is a dedicated OAK JIRA ticket and the commit message pointing to that ticket. except for trivial javadoc typos we don't make changes to the code base without a ticket reference. so that is the only thing that is really required before i can merge the PR.

oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/basic/DefaultSyncContextTest.java
anchela: thanks!

anchela (approved):

PR #538, opened 2022-04-06
OAK-9748 Parallel Merge does not handle download retry correctly

Ewocker:
## Problem
The current way of dealing with download retry is to build a new FlatFileStore folder each time a retry happens. Merge daemon thread creates a merge folder under that FlatFileStore folder and dump intermediate merge progress there.
When there's a download retry, the retried process load all the previously created FlatFileStore into its progress, however it does not take in the already merged intermediate files.

The unit test does not run into this issue because only a small number of intermediate batch merged files are generated and batch merge size by default is 64 files. So intermediate merge does not happen.

## Fix Implementation
Adding ability to forcefully stop merge process, and in retry load the previous merged files into sorted files list to be merged later.

### DIFF
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunner.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunner.java
@@ -39,6 +39,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.Phaser;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 
 import static com.google.common.base.Charsets.UTF_8;
@@ -100,6 +101,20 @@ import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFile
  *         </ol>
  *     </li>
  * </ol>
+ *
+ * <h3>Force Stop Explanation -</h3>
+ * <ol>
+ *     <li>On receiving MERGE_FORCE_STOP_POISON_PILL in SORTED_FILE_QUEUE from parent thread, all un-started tasks will be skipped.</li>
+ *     <li>No more tasks will be created on MERGE_FORCE_STOP_POISON_PILL message is received.</li>
+ *     <li>Running merge will continue to finish.</li>
+ *     <li>Final merge will not be performed.</li>
+ *     <li>This will not result in lose of data under conditions that -</li>
+ *          <ol>
+ *              <li>Files will not be removed until merge task is completed. Parent thread should add unmerged files to the SORTED_FILE_QUEUE on retry.</li>
+ *              <li>Merged files will still be under the merge folder. Parent thread should add those files to the SORTED_FILE_QUEUE on retry.</li>
+ *          </ol>
+ *     </li>
+ * </ol>
  */
 public class MergeRunner implements Runnable {
     private static final Logger log = LoggerFactory.getLogger(MergeRunner.class);
@@ -136,9 +151,17 @@ public class MergeRunner implements Runnable {
     private final Phaser phaser;
 
     /**
-     * This poison pill is added to {@link #sortedFiles} to indicate that download phase has completed.
+     * This poison pill is added to {@link #sortedFiles} to indicate that download phase has completed successfully and
+     * merge should advance to the final merge stage.
      */
     public static final File MERGE_POISON_PILL = new File("");
+
+    /**
+     * This poison pill is added to {@link #sortedFiles} to indicate that download phase has failed and
+     * merge should shut down immediately and not advance to the final merge stage.
+     */
+    public static final File MERGE_FORCE_STOP_POISON_PILL = new File("merge-force-stop-poison-pill");
+    private final AtomicBoolean mergeCancelled;
 
     /**
      * Constructor.
@@ -158,6 +181,7 @@ public class MergeRunner implements Runnable {
         this.phaser = phaser;
         this.batchMergeSize = batchMergeSize;
         this.threadPoolSize = threadPoolSize;
+        this.mergeCancelled = new AtomicBoolean(false);
     }
 
     private boolean merge(List<File> files, File outputFile) {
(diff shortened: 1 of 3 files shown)


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunner.java
```diff
@@ -254,6 +274,14 @@ public void run() {
             }
         }
         log.info("Waiting for batch sorting tasks completion");
+
+        // Parent thread signals to stop immediately
+        if (unmergedFiles.contains(MERGE_FORCE_STOP_POISON_PILL)) {
+            log.info("Merger receives force stop signal, shutting down all merge tasks");
+            executorService.shutdownNow();
```
thomasmueller: This cancels tasks via Thread.interrupt(), which is very problematic: is closes all open files and sockets, and this can cause issues. See also https://stackoverflow.com/questions/2020992/is-thread-interrupt-evil#:~:text=The%20main%20problem%20with%20Thread,interrupt%20the%20thread%20right%20away. - and we had issues in Oak and AEM as well. 

In my view, we must not use shutdownNow here.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreTest.java
```diff
@@ -49,8 +49,8 @@
 import static java.util.Arrays.asList;
 import static java.util.Collections.singleton;
 import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileNodeStoreBuilder.OAK_INDEXER_SORT_STRATEGY_TYPE;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileNodeStoreBuilder.PROP_MERGE_TASK_BATCH_SIZE;
+import static org.junit.Assert.*;
```
thomasmueller: Please don't use star imports.

thomasmueller (changes requested):
This is a partial review.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunner.java
```diff
@@ -297,14 +332,18 @@ private class Task implements Callable<File> {
             mergeTaskPhaser.register();
         }
 
+
         @Override
         public File call() {
             try {
-                if (merge(mergeTarget, mergedFile)) {
+                if (mergeCancelled.get()) {
+                    log.debug("merge cancelled, skipping merge task");
```
thomasmueller, line 341: So this is logged... but I don't see where it's actually skipped: it will still return mergedFile...

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunner.java
```diff
@@ -297,14 +332,18 @@ private class Task implements Callable<File> {
...
+                } else if (merge(mergeTarget, mergedFile)) {
                     log.info("merge complete for {}", mergedFile.getName());
                     return mergedFile;
+                } else {
+                    log.error("merge failed for {}", mergedFile.getName());
```
thomasmueller: Here as well: the code flow will return mergedFile (after the finally).

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunner.java
Ewocker: add throw

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunner.java
Ewocker: add throw

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreTest.java
Ewocker: removed

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunner.java
```diff
@@ -297,14 +332,20 @@ private class Task implements Callable<File> {
             mergeTaskPhaser.register();
         }
 
+
         @Override
-        public File call() {
+        public File call() throws Exception {
```
thomasmueller, line 337: Instead of throwing "Exception", a more specific exception should be thrown. There are two options here:
(a) use you own exception class, or
(b) use an existing one. 

I think appropriate would be java.io.EOFException (end-of-file): https://docs.oracle.com/javase/9/docs/api/java/io/EOFException.html

thomasmueller: I wonder, do we have a test case that checks an exception is thrown at the right time?

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeRunner.java
Ewocker: I rebase https://github.com/apache/jackrabbit-oak/pull/536 on top of this PR, since the unit tests of these two kind of go together and cannot be tested without another one (both are bugs, so want to have unit test that is checking the right state). So the same comment is addressed there as well. Feel free to close this one and review the other.

PR #549, opened 2022-04-21
OAK-9750: Oak-run-elastic: Add right tika dependency

tihom88:


oak-run-elastic/pom.xml
```diff
@@ -218,6 +218,11 @@
             <version>${project.version}</version>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>org.apache.tika</groupId>
+            <artifactId>tika-core</artifactId>
+            <version>1.24.1</version>
```
nit0906: we should pick this up from the oak-parent pom properties. https://github.com/apache/jackrabbit-oak/blob/trunk/oak-parent/pom.xml#L62 

<version>${tika.version}</version>

thomasmueller: +1, we should specify the version only once

reschke: FWIW; Tika 2.x has been around since July. Minimally, we should update Tika to the latest version from the maintenance branch.

oak-run-elastic/pom.xml
```diff
@@ -221,7 +221,7 @@
         <dependency>
             <groupId>org.apache.tika</groupId>
             <artifactId>tika-core</artifactId>
-            <version>1.24.1</version>
+            <version>${tika.version}</version>
         </dependency>
```
reschke, line 225: I believe it would be better to just remove the version tag here (I think than then decl in the parent pom applies)

oak-run-elastic/pom.xml
tihom88: Created Jira [OAK-9752](https://issues.apache.org/jira/browse/OAK-9752) to upgrade tika version.

tihom88:
Updated tika version in oak-search-elastic.

PR #550, opened 2022-04-25
OAK-9751 : handled cases where path changes in lucene exceeds max bui…

rishabhdaim:
…lder size

### DIFF
--- a/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/hybrid/LuceneJournalPropertyBuilder.java
+++ b/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/hybrid/LuceneJournalPropertyBuilder.java
@@ -91,11 +91,9 @@ class LuceneJournalPropertyBuilder implements JournalPropertyBuilder<LuceneDocum
                     if (!first) {
                         reader.read(',');
                     }
+                    String s = reader.readString();
                     if (sizeWithinLimits()) {
-                        indexedNodes.put(path, reader.readString());
-                    } else {
-                        // return if max limit reached for builder to avoid overflow exception
-                        return;
+                        indexedNodes.put(path, s);
                     }
                 }
             } while (reader.matches(','));
--- a/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/hybrid/LuceneJournalPropertyBuilder.java
+++ b/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/hybrid/LuceneJournalPropertyBuilder.java
@@ -93,6 +93,9 @@ class LuceneJournalPropertyBuilder implements JournalPropertyBuilder<LuceneDocum
                     }
                     if (sizeWithinLimits()) {
                         indexedNodes.put(path, reader.readString());
+                    } else {
+                        // return if max limit reached for builder to avoid overflow exception
+                        return;
                     }
                 }
             } while (reader.matches(','));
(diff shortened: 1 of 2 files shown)


oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/hybrid/LuceneJournalPropertyBuilderTest.java
```diff
@@ -53,7 +53,25 @@ public void nullOrEmptyJson() throws Exception{
...
         assertTrue(Iterables.isEmpty(((IndexedPaths)builder2.build())));
     }
+    @Test
+    public void addJsonLessThanMaxBuilderSize() throws Exception {
+        String a = null;
+        for (int i = 0; i < 499; i++) {
+            a = "{\"/var/eventing/jobs/foo/2022/4/19/14/27/af96fcfa9e32_8589" + i + "\" :[\"/oak:index/foo\",\"/oak:index/bar\"]}";
+            builder.addSerializedProperty(a);
+        }
+        assertEquals(createdIndexPathMap((IndexedPaths)builder.build()).size(), 998);
```
mreutegg: Please swap the two arguments for assertEquals(). The first argument is the expected value, which is 998 in this test.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/hybrid/LuceneJournalPropertyBuilderTest.java
```diff
@@ -53,7 +53,25 @@ public void nullOrEmptyJson() throws Exception{
...
+    }
 
+    @Test
+    public void addJsonBiggerThanMaxBuilderSize() throws Exception {
+        String a = null;
+        for (int i = 0; i < 502; i++) {
+            a = "{\"/var/eventing/jobs/foo/2022/4/19/14/27/af96fcfa9e32_8589" + i + "\" :[\"/oak:index/foo\",\"/oak:index/bar\"]}";
+            builder.addSerializedProperty(a);
+        }
+        assertEquals(createdIndexPathMap((IndexedPaths)builder.build()).size(), 1000);
```
mreutegg: Same as above. Please swap the two arguments to assertEquals().

mreutegg (changes requested):

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/hybrid/LuceneJournalPropertyBuilder.java
```diff
@@ -93,6 +93,9 @@ public void addSerializedProperty(@Nullable String json) {
                     }
                     if (sizeWithinLimits()) {
                         indexedNodes.put(path, reader.readString());
+                    } else {
+                        // return if max limit reached for builder to avoid overflow exception
+                        return;
```
thomasmueller: I can see why the old code is broken: if the size is exceeded, we don't call readString(), but continue and try to read a comma...

But I also don't like the new code: we wouldn't detect any errors in the Json format afterwards... Which I would expect we do.

Instead of return, what about:

```
String x = reader.readString()
if (sizeWithinLimits()) {
    indexedNodes.put(path, x);
}
```

Sure, one could argue it is more efficient to return, but I think the side effect of _not_ parsing the rest in this case is worse than the performance hit (which is unlikely to be measurable).

thomasmueller (changes requested):

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/hybrid/LuceneJournalPropertyBuilder.java
mreutegg: The same thought crossed my mind as well when I reviewed the change but I didn't point it out.

+1 for parsing the entire JSON.

thomasmueller:
BTW, thanks a lot for the PR! It is great that you have found and fixed the issue!

mreutegg (approved):
Looks good to me.

I think the failed test in oak-search-elastic is a known issue. It was already reported in https://issues.apache.org/jira/browse/OAK-9141

PR #551, opened 2022-04-26
OAK-9754 increase default dump threshold to 16 MB and expose as system property

Ewocker:


badvision (approved):
Makes sense to me -- are there any tests that need modification in regards to this change?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -101,7 +102,7 @@ public class FlatFileNodeStoreBuilder {
     private long entryCount = 0;
     private File flatFileStoreDir;
     private final MemoryManager memoryManager;
-    private long dumpThreshold = DEFAULT_DUMP_THRESHOLD;
+    private long dumpThreshold = Integer.getInteger(OAK_INDEXER_DUMP_THRESHOLD_IN_MB, OAK_INDEXER_DUMP_THRESHOLD_IN_MB_DEFAULT) * FileUtils.ONE_MB;
```
thomasmueller, line 105: That works, and there is no risk of integer overflow, as FileUtils.ONE_MB is a long.

thomasmueller (approved):

PR #555, opened 2022-04-30
Fix indeterminate map ordering that may causes Sweep2Test to fail

faaple:
The test `org.apache.jackrabbit.oak.plugins.document.Sweep2Test#testSweep2Uncommitted` can fail based on the order of iteration in `HashMap`. I found the issue using [NonDex](https://github.com/TestingResearchIllinois/NonDex):
```
mvn install -DskipTests -pl oak-store-document -am
mvn -pl oak-store-document edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=org.apache.jackrabbit.oak.plugins.document.Sweep2Test#testSweep2Uncommitted
```
The proposed fix changes several `HashMap` to `LinkedHashMap`, which has determinate iteration order.


mreutegg:
@MarcyGO please create a JIRA issue for this PR.

Preferably the PR should also have a unit test that reproduces the issue reliably.

I'm not sure the fix is done at the appropriate place. The proposed changes are in module oak-store-spi. Does the current implementation of those two classes really require a determinate iteration order or is it rather Sweep2Test that makes assumptions about an implementation detail? Maybe just the test should be fixed.

PR #557, opened 2022-05-05
OAK-9753: ES connection parameters should be easily accessible from E…

tihom88:
…lasticConnectionRule

### DIFF
--- a/oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnectionRule.java
+++ b/oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnectionRule.java
@@ -29,6 +29,10 @@ import org.testcontainers.elasticsearch.ElasticsearchContainer;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 /*
 To be used as a @ClassRule
@@ -41,6 +45,8 @@ public class ElasticConnectionRule extends ExternalResource {
     private static boolean useDocker = false;
 
     private final String elasticConnectionString;
+
+    private ElasticConnectionModel elasticConnectionModel;
 
     public ElasticConnectionRule(String elasticConnectionString) {
         this.elasticConnectionString = elasticConnectionString;
@@ -55,16 +61,19 @@ public class ElasticConnectionRule extends ExternalResource {
     @Override
     public Statement apply(Statement base, Description description) {
         Statement s = super.apply(base, description);
-        if (elasticConnectionString == null || getElasticConnectionFromString() == null) {
+        if (!isValidUri(elasticConnectionString)) {
             elastic = ElasticTestServer.getESTestServer();
             setUseDocker(true);
+            initializeElasticConnectionModel(elastic);
+        } else {
+            initializeElasticConnectionModel(elasticConnectionString);
         }
         return s;
     }
 
     @Override
     protected void after() {
-        ElasticConnection esConnection = useDocker() ? getElasticConnectionForDocker() : getElasticConnectionFromString();
+        ElasticConnection esConnection = getElasticConnection();
         if (esConnection != null) {
             try {
                 esConnection.getClient().indices().delete(new DeleteIndexRequest(esConnection.getIndexPrefix() + "*"), RequestOptions.DEFAULT);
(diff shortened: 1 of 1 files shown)


oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnectionRule.java
```diff
@@ -75,6 +80,46 @@ protected void after() {
...
+            URI uri = new URI(elasticConnectionString);
+            String host = uri.getHost();
+            String scheme = uri.getScheme();
+            int port = uri.getPort();
+            String query = uri.getQuery();
+
+            String apiKey = null;
+            String apiSecret = null;
+            if (query != null) {
+                apiKey = query.split(",")[0].split("=")[1];
```
thomasmueller: As far as I understand, this assumes that the first parameter is always the API key, and the second always the secret... Example: "key=abc,secret=def". But if someone manually changes that to "secret=def,key=abc"?

What about using https://stackoverflow.com/questions/14768171/convert-string-representing-key-value-pairs-to-map to convert the key-value pair to a map and then retrieve the key and secret from the map?

thomasmueller (changes requested):
I know the problem was there before... But still a good time to fix this.

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnectionRule.java
tihom88: This was already implemented and in comment we had defined url so that this logic will work.
`// Set this connection string as
    // <scheme>://<hostname>:<port>?key_id=<>,key_secret=<>
    // key_id and key_secret are optional in case the ES server
    // needs authentication
    // Do not set this if docker is running and you want to run the tests on docker instead.`

But yes we can change the logic to make it to be more generic.

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnectionRule.java
```diff
@@ -75,6 +80,46 @@ protected void after() {
...
+            elasticConnectionModel.indexPrefix = indexPrefix + System.currentTimeMillis();
+        } catch (URISyntaxException e) {
+            LOG.error("Provided elastic connection string  is not valid ", e);
+        }
+    }
+
+    private void initializeElasticConnectionModel(ElasticsearchContainer elastic) {
+        this.elasticConnectionModel = new ElasticConnectionModel();
+        elasticConnectionModel.scheme = ElasticConnection.DEFAULT_SCHEME;
+        elasticConnectionModel.elasticHost = elastic.getContainerIpAddress();
```
fabriziofortino: `getContainerIpAddress` is deprecated in favour of `getHost`

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnectionRule.java
```diff
@@ -83,33 +128,62 @@ public ElasticConnection getElasticConnectionFromString() {
...
 
+    private boolean isValidUri(String connectionString) {
+        if (connectionString == null) {
+            return false;
+        }
+        try {
+            new URI(connectionString);
+            return true;
+        } catch (URISyntaxException e) {
+            LOG.debug("Provided elastic connection string  is not valid ", e);
```
fabriziofortino: extra space. Could it be useful to print `connectionString` as part of the log message?
```suggestion
            LOG.debug("Provided elastic connection string is not valid ", e);
```

fabriziofortino (approved):
LGTM. Just a couple of minors.

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnectionRule.java
tihom88: Exception also logs the malformed uri. So didn't logged it again in message.

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnectionRule.java
tihom88: Done

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnectionRule.java
tihom88: Done

thomasmueller (approved):

PR #560, opened 2022-05-05
OAK-9757 : increased node name limit for mongo 4.2 version

rishabhdaim:


reschke (changes requested):
Wow. I would say that if we need to modify 109 files, there's something very wrong with this change.

reschke:
Idea: Move (or copy) the util functions to the DocumentStore API, where they have access to metadata.

rishabhdaim:
Thanks @reschke, that's a good idea to move util function to DocumentStore and that would help us avoiding in calling metadata multiple times. 
Let me make changes and raise the PR again, but I doubt that would restrict the code changes to less file since most of the changes are in tests cases and the changes are to pass the parameters to utils class.

mreutegg:
I suggest we take a slightly different approach. To me it feels wrong that `Utils.isLongPath()` is currently responsible for detecting when a node name is too long. I think this is the reason why so many other code changes are necessary. E.g. the method is then used in `Utils.getIdFromPath()`, which in turn is used in many places. But in many cases the check for the node name length is not actually needed. I my view it would be better if the node name length check is moved out of `Utils.isLongPath()` to a place were it is really needed. That is, when a new node is added through the `DocumentNodeStore` or created with an `UpdateOp` on the `DocumentStore`.

mreutegg:
Also note that checks fails because two new files for classes `MongoVersion` and `MongoVersionTest` do not have a license header.

rishabhdaim:
@mreutegg 

> To me it feels wrong that Utils.isLongPath() is currently responsible for detecting when a node name is too long

Yes, I think it should be other way  around i.e. **node name too long** should check whether path is long or not.
I think, we should move out the logic to check **node name too long** to a separate method and call `Utils.isLongPath()` from it since **node name too long** make sense only for long paths. 
`
/*** The maximum size a node name, in bytes. This is only a problem for long path.*/
           int NODE_NAME_LIMIT = Integer.getInteger("oak.nodeNameLimit", 150);
`
 As mentioned by you, we should move whole logic for **node name too long** here https://github.com/apache/jackrabbit-oak/blob/trunk/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeState.java#L417 and then we should be able to reduce the number of changed files significantly.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersion.java
```diff
@@ -0,0 +1,89 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Objects;
+
+/**
+ * Util Class to represent mongo version
+ *
+ * @author daim
+ */
+public final class MongoVersion implements Comparable<MongoVersion>{
```
mreutegg: The MongoDB Java driver library has a class ServerVersion that also implements Comparable. I suggest to use that class. This will also avoid introducing a new dependency on commons-lang3.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStore.java
```diff
@@ -456,4 +456,18 @@ <T extends Document> T findAndUpdate(Collection<T> collection,
...
+     * The maximum size a node name, in bytes. This is only a problem for long path.
+     */
+    int NODE_NAME_LIMIT = Integer.getInteger("oak.nodeNameLimit", 150);
+
+    /**
+     * Return the size limit for node name based on the document store implementation
+     *
+     * @return node name size limit
+     */
+    default int getSizeLimit() {
```
mreutegg: I think this method should be renamed to better indicate where this limit applies. E.g. `getNodeNameLimit()`.

mreutegg (changes requested):

oak-store-document/pom.xml
```diff
@@ -152,6 +152,10 @@
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+    </dependency>
```
mreutegg: Please introduce new dependencies only when really necessary.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeState.java
```diff
@@ -414,6 +414,11 @@ public String toString() {
      * @param revision the revision this node is created.
      */
     UpdateOp asOperation(@NotNull Revision revision) {
+
+        if (Utils.isNodeNameLong(path, store.getDocumentStore().getSizeLimit())) {
+            throw new IllegalArgumentException("Node name is too long: " + path);
```
mreutegg: I'm aware the current code throws an IllegalArgumentException, but it somehow looks wrong. The path is not an argument when `asOperation(Revision)` is called. I think a `DocumentStoreException` is more appropriate.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -114,10 +114,7 @@
 import static org.apache.jackrabbit.oak.plugins.document.NodeDocument.SD_MAX_REV_TIME_IN_SECS;
 import static org.apache.jackrabbit.oak.plugins.document.NodeDocument.SD_TYPE;
 import static org.apache.jackrabbit.oak.plugins.document.UpdateOp.Condition.newEqualsCondition;
-import static org.apache.jackrabbit.oak.plugins.document.mongo.MongoUtils.createIndex;
-import static org.apache.jackrabbit.oak.plugins.document.mongo.MongoUtils.createPartialIndex;
-import static org.apache.jackrabbit.oak.plugins.document.mongo.MongoUtils.getDocumentStoreExceptionTypeFor;
-import static org.apache.jackrabbit.oak.plugins.document.mongo.MongoUtils.hasIndex;
+import static org.apache.jackrabbit.oak.plugins.document.mongo.MongoUtils.*;
```
mreutegg: Please do not use wildcard imports.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoUtils.java
```diff
@@ -190,4 +190,15 @@ static Type getDocumentStoreExceptionTypeFor(Throwable t) {
...
         return type;
     }
+
+    /**
+     * Util method to get node size limit for current mongo version
+     *
+     * @param version version of current mongo db
+     * @return size limit based on mongo db version
+     */
+    static int getSizeLimit(String version) {
```
mreutegg: I think the method name should be more explicit about where the size limit applies.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -365,14 +361,29 @@ private static boolean isLongPath(String path) {
...
+    }
+
+    /**
+     * Checks whether Node name is too long or not based on underlining document store
+     * @param path node path
+     * @param sizeLimit sizeLimit for node name
+     * @return true if node name is long else false
+     */
+
+    public static boolean isNodeNameLong(String path, int sizeLimit) {
```
mreutegg: This variant of `isNodeNameLong()` is only used by two tests. I think it's better to remove this method.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -794,12 +799,7 @@ public StringValue apply(String input) {
      * Transforms the given paths into ids using {@link #getIdFromPath(String)}.
      */
     public static Iterable<String> pathToId(@NotNull Iterable<String> paths) {
-        return transform(paths, new Function<String, String>() {
-            @Override
-            public String apply(String input) {
-                return getIdFromPath(input);
-            }
-        });
+        return transform(paths, input -> getIdFromPath(input));
```
mreutegg, line 794: This is an unrelated change, but I think it's fine.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentSplitTest.java
```diff
@@ -732,21 +732,21 @@ public <T extends Document> T find(Collection<T> collection,
 
         ns.dispose();
     }
-    
+
```
mreutegg: Changes in this test are just whitespace. Please revert.

mreutegg (changes requested):

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MongoDocumentStoreTest.java
```diff
@@ -241,6 +241,26 @@ public void queryWithLimit() throws Exception {
...
+            DocumentNodeState n = new DocumentNodeState(store, Path.fromString(DocumentMK.LONG_PATH + i),
+                    new RevisionVector(rev));
+            inserts.add(n.asOperation(rev));
+        }
+        docStore.create(Collection.NODES, inserts);
+        List<NodeDocument> docs = docStore.query(Collection.NODES,
+                Utils.getKeyLowerLimit(Path.ROOT),  Utils.getKeyUpperLimit(Path.ROOT),
+                DocumentMK.MANY_CHILDREN_THRESHOLD);
+        assertEquals(DocumentMK.MANY_CHILDREN_THRESHOLD, docs.size());
+        store.dispose();
```
mreutegg: Should call `store.dispose()` in a finally block or use JUnit rule `DocumentMKBuilderProvider`.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MongoDocumentStoreTest.java
```diff
@@ -241,6 +241,26 @@ public void queryWithLimit() throws Exception {
...
+        for (int i = 0; i < 2; i++) {
+            DocumentNodeState n = new DocumentNodeState(store, Path.fromString(DocumentMK.LONG_PATH + i),
+                    new RevisionVector(rev));
+            inserts.add(n.asOperation(rev));
+        }
+        docStore.create(Collection.NODES, inserts);
+        List<NodeDocument> docs = docStore.query(Collection.NODES,
+                Utils.getKeyLowerLimit(Path.ROOT),  Utils.getKeyUpperLimit(Path.ROOT),
+                DocumentMK.MANY_CHILDREN_THRESHOLD);
+        assertEquals(DocumentMK.MANY_CHILDREN_THRESHOLD, docs.size());
```
mreutegg: This is unreachable. `n.asOperation(rev)` is expected to throw the `IllegalArgumentException`. The test does not seem to exercise what it intends to do.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/Sweep2TestHelper.java
```diff
@@ -36,8 +36,6 @@
 import org.apache.jackrabbit.oak.plugins.document.memory.MemoryDocumentStore;
 import org.apache.jackrabbit.oak.plugins.document.util.LeaseCheckDocumentStoreWrapper;
 import org.apache.jackrabbit.oak.plugins.document.util.Utils;
-import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
```
mreutegg: Changes in this file unrelated. Can you please revert?

mreutegg (changes requested):

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/util/UtilsTest.java
```diff
@@ -16,10 +16,7 @@
  */
 package org.apache.jackrabbit.oak.plugins.document.util;
 
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.Random;
+import java.util.*;
```
mreutegg: Please do not use wildcard imports.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -114,10 +114,10 @@
 import static org.apache.jackrabbit.oak.plugins.document.NodeDocument.SD_MAX_REV_TIME_IN_SECS;
 import static org.apache.jackrabbit.oak.plugins.document.NodeDocument.SD_TYPE;
 import static org.apache.jackrabbit.oak.plugins.document.UpdateOp.Condition.newEqualsCondition;
-import static org.apache.jackrabbit.oak.plugins.document.mongo.MongoUtils.createIndex;
 import static org.apache.jackrabbit.oak.plugins.document.mongo.MongoUtils.createPartialIndex;
 import static org.apache.jackrabbit.oak.plugins.document.mongo.MongoUtils.getDocumentStoreExceptionTypeFor;
 import static org.apache.jackrabbit.oak.plugins.document.mongo.MongoUtils.hasIndex;
+import static org.apache.jackrabbit.oak.plugins.document.mongo.MongoUtils.createIndex;
```
stefan-egli: This seems to have been at the right place where it was before, I'd move it back to reduce number of changes

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoUtils.java
```diff
@@ -190,4 +195,15 @@ static Type getDocumentStoreExceptionTypeFor(Throwable t) {
...
+
+    /**
+     * Util method to get node size limit for current mongo version
+     *
+     * @param version version of current mongo db
+     * @return size limit based on mongo db version
+     */
+    static int getNodeNameLimit(final String version) {
+        final ServerVersion sv = new ServerVersion(Arrays.stream(version.split("\\.")).map(Integer::new).collect(Collectors.toList()));
+        return sv.compareTo(new ServerVersion(Lists.newArrayList(4,0,0))) > 0 ? Integer.MAX_VALUE : DocumentStore.NODE_NAME_LIMIT;
```
stefan-egli: I guess you could also do
```suggestion
        return sv.compareTo(new ServerVersion(4, 0) > 0 ? Integer.MAX_VALUE : DocumentStore.NODE_NAME_LIMIT;
```

but that's a detail

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -92,11 +92,6 @@ public class Utils {
      */
     public static final int PATH_LONG = Integer.getInteger("oak.pathLong", 350);
 
-    /**
-     * The maximum size a node name, in bytes. This is only a problem for long path.
-     */
-    public static final int NODE_NAME_LIMIT = Integer.getInteger("oak.nodeNameLimit", 150);
-
```
stefan-egli: Was thinking about this a bit. Currently you're moving it to `DocumentStore` as that's where there's now a new `getNodeNameLimit`. But actually there's also (test) code that still refers to the static field. So in light of reducing the number of changes of this PR even further, but also not change if it's really necessary, what do you think about leaving `NODE_NAME_LIMIT` in here in Utils - and have `DocumentStore.getNodeNameLimit` refer to this one? It would reduce the diff and separate one concern: keep the static field here as the 'hardcoded default' vs allow subclasses of `DocumentStore` overwriting it (without necessarily also keeping this field there)

oak-upgrade/src/main/java/org/apache/jackrabbit/oak/upgrade/nodestate/NameFilteringNodeState.java
```diff
@@ -77,10 +78,10 @@ protected PropertyState decorateProperty(@NotNull final PropertyState delegatePr
     private boolean isNameTooLong(@NotNull String name) {
         // OAK-1589: maximum supported length of name for DocumentNodeStore
         // is 150 bytes. Skip the sub tree if the the name is too long
-        if (name.length() <= Utils.NODE_NAME_LIMIT / 3) {
+        if (name.length() <= DocumentStore.NODE_NAME_LIMIT / 3) {
             return false;
         }
-        if (name.getBytes(Charsets.UTF_8).length <= Utils.NODE_NAME_LIMIT) {
+        if (name.getBytes(Charsets.UTF_8).length <= DocumentStore.NODE_NAME_LIMIT) {
```
stefan-egli: (this class could remain unchanged if the static is kept in `Utils` - see earlier comment )

oak-upgrade/src/main/java/org/apache/jackrabbit/oak/upgrade/RepositoryUpgrade.java
```diff
@@ -1022,10 +1023,10 @@ void assertNoLongNames() throws RepositoryException {
     }
 
     private boolean nameMayBeTooLong(String name) {
-        if (name.length() <= Utils.NODE_NAME_LIMIT / 3) {
+        if (name.length() <= DocumentStore.NODE_NAME_LIMIT / 3) {
             return false;
         }
-        if (name.getBytes(Charsets.UTF_8).length <= Utils.NODE_NAME_LIMIT) {
+        if (name.getBytes(Charsets.UTF_8).length <= DocumentStore.NODE_NAME_LIMIT) {
```
stefan-egli: (this class could remain unchanged if the static is kept in `Utils` - see earlier comment )

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/util/UtilsTest.java
```diff
@@ -75,6 +76,7 @@
 public class UtilsTest {
 
     private static final long TIME_DIFF_WARN_THRESHOLD_MILLIS = 2000L;
+    private static final String LONG_PATH = "/foo/barbar/qwerty/asdfgh/zxcvbnm/bu~pcsdb_2_c_cm~2020_idea_pad_idea_centre_holiday_campaign_cv~all_the_right_answers_idea_centre_aio_amd_carousel_ct~fb_cs~1_x_1_lg~en_pr~idea_centre_aio_5_fs~amd_ot~_cr~the_woo_5_jpg/bu~pcsdb_2_c_cm~2020_idea_pad_idea_centre_holiday_campaign_cv~all_the_right_answers_idea_centre_aio_amd_carousel_ct~fb_cs~1_x_1_lg~en_pr~idea_centre_aio_5_fs~amd_ot~_cr~the_woo_5_jpg_1/bu~pcsdb_2_c_cm~2020_idea_pad_idea_centre_holiday_campaign_cv~all_the_right_answers_idea_centre_aio_amd_carousel_ct~fb_cs~1_x_1_lg~en_pr~idea_centre_aio_5_fs~amd_ot~_cr~the_woo_5_jpg_2";
```
stefan-egli: I would further obfuscate the path name. It still looks too similar to an actual name used out in the field. What about making up the whole deep path structure from foo bar bazes..?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/Sweep2TestHelper.java
```diff
@@ -85,7 +85,7 @@ static MemoryDocumentStore getMemoryDocumentStore(DocumentNodeStore ns) {
      * <li>if simulatePreFixUpgrade then simulate an upgrade to a version post 1.8 but without a fix for the branch commit problem</li>
      * <li>upgrade to a fixed version</li>
      * </ol>
-     * @param ns
+     * @param memStore
```
stefan-egli: seems unrelated from this PR and might complicate backporting work unnecessarily, even if just very slightly

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentSplitTest.java
stefan-egli: +1, this comment still applied, I'd too suggest to revert

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentMK.java
```diff
@@ -52,6 +52,9 @@ public class DocumentMK {
      * The threshold where special handling for many child node starts.
      */
     static final int MANY_CHILDREN_THRESHOLD = DocumentNodeStoreBuilder.MANY_CHILDREN_THRESHOLD;
+
+    static final String LONG_PATH = "/foo/barbar/qwerty/asdfgh/zxcvbnm/bu~pcsdb_2_c_cm~2020_idea_pad_idea_centre_holiday_campaign_cv~all_the_right_answers_idea_centre_aio_amd_carousel_ct~fb_cs~1_x_1_lg~en_pr~idea_centre_aio_5_fs~amd_ot~_cr~the_woo_5_jpg/bu~pcsdb_2_c_cm~2020_idea_pad_idea_centre_holiday_campaign_cv~all_the_right_answers_idea_centre_aio_amd_carousel_ct~fb_cs~1_x_1_lg~en_pr~idea_centre_aio_5_fs~amd_ot~_cr~the_woo_5_jpg_1/bu~pcsdb_2_c_cm~2020_idea_pad_idea_centre_holiday_campaign_cv~all_the_right_answers_idea_centre_aio_amd_carousel_ct~fb_cs~1_x_1_lg~en_pr~idea_centre_aio_5_fs~amd_ot~_cr~the_woo_5_jpg_2";
```
stefan-egli: I would further obfuscate the path name. It still looks too similar to an actual name used out in the field. What about making up the whole deep path structure from foo bar bazes..?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -365,14 +361,17 @@ private static boolean isLongPath(String path) {
...
-        if (parent.length < PATH_LONG) {
-            return false;
-        }
-        String name = PathUtils.getName(path);
-        if (name.getBytes(UTF_8).length > NODE_NAME_LIMIT) {
-            throw new IllegalArgumentException("Node name is too long: " + path);
-        }
-        return true;
+        return parent.length >= PATH_LONG;
+    }
```
stefan-egli, line 369: can you explain why the node name check is obsolete here? I understand `DocumentNodeState.asOperation` is now the place where the node name length check is done. But is that (enough) justification to remove this name check here?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -338,6 +333,7 @@ public static StringBuilder encodeHexString(byte[] data, StringBuilder sb) {
      *     <li>If id is for root path</li>
      *     <li>If id is for an invalid path</li>
      * </ul>
+     *
```
stefan-egli: I'd remove this unrelated one line change

stefan-egli (changes requested):
Added a couple comments, some are cosmetic but might be worth while as it would reduce the PR change set further (and make a final review easier). The main comment might be about where `NODE_NAME_LIMIT` should live and the question whether `isLongPath` really doesn't need the node name check anymore.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
rishabhdaim: From @mreutegg earlier comment:

> I suggest we take a slightly different approach. To me it feels wrong that Utils.isLongPath() is currently responsible for detecting when a node name is too long. I think this is the reason why so many other code changes are necessary. E.g. the method is then used in Utils.getIdFromPath(), which in turn is used in many places. But in many cases the check for the node name length is not actually needed. I my view it would be better if the node name length check is moved out of Utils.isLongPath() to a place were it is really needed. That is, when a new node is added through the DocumentNodeStore or created with an UpdateOp on the DocumentStore.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
rishabhdaim: Sure, it make sense to keep it where it was originally present.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeState.java
```diff
@@ -414,6 +414,11 @@ public String toString() {
      * @param revision the revision this node is created.
      */
     UpdateOp asOperation(@NotNull Revision revision) {
+
+        if (Utils.isNodeNameLong(path, store.getDocumentStore().getNodeNameLimit())) {
+            throw new DocumentStoreException("Node name is too long: " + path);
+        }
+
```
stefan-egli: Actually I still think this does not cover all cases, now that `Utils.getIdFromPath` doesn't do the check anymore.

What about moving the check to `UpdateOp` 's constructor, where it initializes the `id`?

Any add/modify write to DocumentStore goes through an `UpdateOp` - so if you can avoid creating a faulty (too-long node-name) `UpdateOp` you'd be fine. 
Alternatively it could be done in the DocumentStore's implementations - but that would multiply the changes again and perhaps open doors for missing it in some. So `UpdateOp` sounds like a central piece where this could be applied.

Wdyt?

rishabhdaim: For a Node Name to be long, there must be parent path which is long as well. 

Since `UpdateOp` doesn't have path (parent for newly added node), so I don't we can move this check to `UpdateOp`.

stefan-egli: Good point, `UpdateOp` doesnt have the `path` .. and `asOperation` is used as part of `addNode`, so it does make sense. I'm just wondering if that covers 100% of the cases.. hmm

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoUtils.java
mreutegg: Hmm, is this really correct? Doesn't this return Integer.MAX_VALUE when e.g. the server is on version 4.0.27?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CommitBuilderTest.java
```diff
@@ -150,6 +150,18 @@ public void addNodeState() {
...
+    public void addLongNameNodeState() {
+        Path path = Path.fromString(DocumentMK.LONG_PATH);
+        CommitBuilder builder = new CommitBuilder(ns, null);
+        builder.addNode(path);
+        Commit c = builder.build(ns.newRevision());
+        UpdateOp up = c.getUpdateOperationForNode(path);
+        UpdateOp.Operation op = up.getChanges().get(
+                new UpdateOp.Key("_deleted", c.getRevision()));
+        assertNotNull(op);
+    }
```
stefan-egli: builder.addNode is expected to throw an exception here - what about removing the code below and replacing it with
```suggestion
        fail("should not reach this point");
    }
```

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MongoDocumentStoreTest.java
```diff
@@ -241,6 +244,19 @@ public void queryWithLimit() throws Exception {
...
+    public void queryWithNodeNameTooLong() {
+        DocumentStore docStore = openDocumentStore();
+        DocumentNodeStore store = builderProvider.newBuilder().setDocumentStore(docStore).setAsyncDelay(0).getNodeStore();
+        Revision rev = Revision.newRevision(0);
+        List<UpdateOp> inserts = Lists.newArrayList();
+        for (int i = 0; i < 2; i++) {
+            DocumentNodeState n = new DocumentNodeState(store, Path.fromString(DocumentMK.LONG_PATH + i),
+                    new RevisionVector(rev));
+            inserts.add(n.asOperation(rev));
+        }
```
stefan-egli: Same thing as with the other test : it expects to fail but doesn't assert that explicitly (just implicitly that the test itself should throw an exception). Might be matter of taste, but this would be an alternative:
```suggestion
        DocumentNodeState n = new DocumentNodeState(store, Path.fromString(DocumentMK.LONG_PATH + i),
                new RevisionVector(rev));
        inserts.add(n.asOperation(rev));
        fail("should not reach this point");
```

stefan-egli: (actually my example doesn't compile I just noticed - the `i` has to be defined)

stefan-egli: (but outside of the existance of a loop that `i` maybe doesn't make much sense neither so could also be removed i guess)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeState.java
stefan-egli: What I was wondering, what about moving this check from the somewhat implicit `asOperaiton` location to the a bit more explicit `CommitBuilder.addNode(Path)`? It should have the same effect, as `addPath` does eventually call `asOperation` .. but having it in `addNode(Path)` would seem a bit more logical for people reading the code.

But the question would still be somewhat open: is this the only place where long node names can be added .. maybe @mreutegg / @reschke can comment too. The only suspicious place I found was `SplitOperations.createIntermediateDocs` - but that only adds a few characters to the id and therefore is probably fine.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoUtils.java
```diff
@@ -190,4 +194,15 @@ static Type getDocumentStoreExceptionTypeFor(Throwable t) {
...
+
+    /**
+     * Util method to get node size limit for current mongo version
+     *
+     * @param version version of current mongo db
+     * @return size limit based on mongo db version
+     */
+    static int getNodeNameLimit(final String version) {
+        final ServerVersion sv = new ServerVersion(Arrays.stream(version.split("\\.")).map(Integer::new).collect(Collectors.toList()));
+        return sv.compareTo(new ServerVersion(4,2)) >= 0 ? Integer.MAX_VALUE : Utils.NODE_NAME_LIMIT;
```
mreutegg: This can be further simplified. I noticed MongoDB versions can also have non-digit characters. Though, these versions should be limited to release candidates.

https://www.mongodb.com/docs/manual/reference/versioning/#release-candidate--rc--releases

I think it would still be nice if Oak can also run on a release candidate.

The lines could be replaced with a call to MongoStatus.isVersion(4, 2).
https://github.com/apache/jackrabbit-oak/blob/c5eb75abce6cac9fd9e9244d0fd34b1345a285e8/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoStatus.java#L160

I'll create a PR for your feature branch.

mreutegg: See https://github.com/rishabhdaim/jackrabbit-oak/pull/1

mreutegg: @rishabhdaim I noticed you updated this PR with changes I proposed. However, some changes are missing. The tests you added to MongoUtilsTest and I updated are now missing in this PR. Can you please add them again?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeState.java
mreutegg: @stefan-egli , do you mean these lines?

https://github.com/apache/jackrabbit-oak/blob/c6ddcc55bee3de915459af01e91edad32d538f3d/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/SplitOperations.java#L317-L318

That should be fine because the ID of an intermediate document always ends with a sequence of digits that indicate the height of the split document tree. This is usually just one digit.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoUtils.java
rishabhdaim: @mreutegg I have removed those tests because I added your proposed changes to use `MongoStatus.isVersion` in the `MongoDocumentStore` class.

I made this change to further reduce the change set in order to make it easier to backport (if required).

mreutegg: I see. On the other hand there are no more tests that check if the name length limit is handled correctly for different MongoDB versions.

rishabhdaim: I got it. I have reverted my change and merged your PR in my local fork.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeState.java
stefan-egli: yes those lines.

Ok, I'll mark this as resolved. I've double checked again and found no other cases calling `isLongPath` (that would now get the name too-long check removed).

mreutegg (changes requested):
The changes look pretty good now but then I noticed a problem.

The new default method on the DocumentStore interfaces is not overridden in wrapper classes like LeaseCheckDocumentStoreWrapper. This means the increased node name capability of a MongoDocumentStore may be hidden behind one of those wrapper classes.

stefan-egli (approved):
+1, lgtm now. 
(I agree, the other `DocumentStore` classes in `/src/test` do not necessarily have to implement `getNodeNameLimit` - so that's fine to leave those out)

mreutegg (approved):

PR #562, opened 2022-05-06
OAK-9760 Oak run index purge command active index check is in correct

FrancoisZhang:


oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -86,42 +92,65 @@ public static List<IndexVersionOperation> generateIndexVersionOperationList(Node
...
 
         if (!reverseSortedIndexNameList.isEmpty()) {
-            IndexName activeIndexNameObject = reverseSortedIndexNameList.remove(0);
-            NodeState activeIndexNode = indexDefParentNode.getChildNode(PathUtils.getName(activeIndexNameObject.getNodeName()));
-            boolean isActiveIndexLongEnough = isIndexPurgeReady(activeIndexNameObject, activeIndexNode, purgeThresholdMillis);
-            int activeProductVersion = activeIndexNameObject.getProductVersion();
-            indexVersionOperationList.add(new IndexVersionOperation(activeIndexNameObject));
+            IndexName activeIndexNameObject = getActiveIndex(reverseSortedIndexNameList, parentPath, rootNode);
+            if (activeIndexNameObject == null) {
+                LOG.warn("Cannot find any active index from the list: {}", reverseSortedIndexNameList);
```
thomasmueller, line 97: I think it's fine to log a warning currently... I just wonder, how would we remove an OOTB index from AEM. The first such case is the generic Lucene index (/oak:index/lucene). We can remove all versions of that index in GraniteContent. So there would never be this case that he have indexes, but none is active (the old state is: active; the new state is: removed).

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -86,42 +92,65 @@ public static List<IndexVersionOperation> generateIndexVersionOperationList(Node
...
+                            // the check hidden oak mount logic only works when passing through the proper composite store
+                            if (isHiddenOakMountExists(indexNode)) {
+                                LOG.info("Found hidden oak mount node for: '{}', disable it but no index definition deletion", indexName);
+                                indexVersionOperation.setOperation(Operation.DELETE_HIDDEN_AND_DISABLE);
+                            } else {
+                                indexVersionOperation.setOperation(Operation.DELETE);
+                            }
                         }
+                        // if the index product version is larger than active index, leave it as is
+                        // for instance: if there is active damAssetLucene-7, the inactive damAssetLucene-8 will be leave there as is
```
thomasmueller: Shouldn't we add an "else" case where, where we also log this case (info level), and set the operation? (Operation.NOOP). I know, Operation.NOOP is the default (and that's good), so setting this to NOOP is a no-op :-) But when reading this code here, that's not obvious.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -86,42 +92,65 @@ public static List<IndexVersionOperation> generateIndexVersionOperationList(Node
...
+            IndexName activeIndexNameObject = getActiveIndex(reverseSortedIndexNameList, parentPath, rootNode);
+            if (activeIndexNameObject == null) {
+                LOG.warn("Cannot find any active index from the list: {}", reverseSortedIndexNameList);
+            } else {
+                NodeState activeIndexNode = indexDefParentNode.getChildNode(PathUtils.getName(activeIndexNameObject.getNodeName()));
+                boolean isActiveIndexLongEnough = isIndexPurgeReady(activeIndexNameObject, activeIndexNode, purgeThresholdMillis);
+                int activeProductVersion = activeIndexNameObject.getProductVersion();
+                indexVersionOperationList.add(new IndexVersionOperation(activeIndexNameObject));
 
             // for rest indexes except active index
```
thomasmueller: I would make this comment a bit clearer:

```
// the reverseSortedIndexNameList will now contain the remaining indexes;
// the active index was removed from that list
```

thomasmueller (changes requested):

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
FrancoisZhang: @thomasmueller good points, yeah makes sense just to leave a warning message as is for now. we can check further later based on the log analysis.
For now, it will be too risky to simply remove all indexes when no active index.

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
FrancoisZhang: @thomasmueller makes sense, let me refine this a bit to have a log here

FrancoisZhang: @thomasmueller just push a commit to reflect this, can you check it again? thanks!

oak-run/src/main/java/org/apache/jackrabbit/oak/indexversion/IndexVersionOperation.java
```diff
@@ -86,42 +92,69 @@ public static List<IndexVersionOperation> generateIndexVersionOperationList(Node
...
-                    } else {
-                        // the check hidden oak mount logic only works when passing through the proper composite store
-                        if (isHiddenOakMountExists(indexNode)) {
-                            LOG.info("Found hidden oak mount node for: '{}', disable it but no index definition deletion", indexName);
+            IndexName activeIndexNameObject = getActiveIndex(reverseSortedIndexNameList, parentPath, rootNode);
+            if (activeIndexNameObject == null) {
+                LOG.warn("Cannot find any active index from the list: {}", reverseSortedIndexNameList);
+            } else {
+                NodeState activeIndexNode = indexDefParentNode.getChildNode(PathUtils.getName(activeIndexNameObject.getNodeName()));
+                boolean isActiveIndexLongEnough = isIndexPurgeReady(activeIndexNameObject, activeIndexNode, purgeThresholdMillis);
```
thomasmueller: Not sure about the name "isActiveIndexLongEnough". Indexes are not "long" or "short". But indexes can be active for a long time / short time.

What about one of the following:
* isIndexActiveLongEnough (is the index active long enough?), or
* isActiveIndexOldEnough (is the active index old enough?)

FrancoisZhang: @thomasmueller thanks, just pushed a commit to change the method name

thomasmueller (approved):

PR #563, opened 2022-05-10
OAK-9756 statistics for the IndexSanityChecker

joerghoh:
I made the experience that the IndexSanityChecker can take quite a bit of time, especially if the index binaries need to be downloaded first, and that these downloads can cause massive delays in the startup of the application. Right now it's not possible to determine exactly how much time is actually spent on these activities.

The IndexSanityChecker is invoked for each index indepedently, so there is no way to tell "this was the last index, now dump the timing information". So on every successful check the accumulated statistics are added to the already existing log message including timing information and the size of the indexes already checked.

With this additional logging in place it's now possible to determine if this index download is a bottleneck in terms of startup time.

### DIFF
--- a/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/directory/IndexSanityChecker.java
+++ b/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/directory/IndexSanityChecker.java
@@ -48,8 +48,9 @@ public class IndexSanityChecker {
         this.indexPath = indexPath;
     }
 
-    public boolean check() throws IOException {
+    public boolean check(IndexSanityStatistics stats) throws IOException {
         boolean allFine = true;
+        long start = System.currentTimeMillis();
         //TODO Add support for checksum based checks
         if (isThereASizeMismatch()){
             //In case of any mismatch just purge all local files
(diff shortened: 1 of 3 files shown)


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/directory/IndexSanityChecker.java
```diff
@@ -118,4 +122,54 @@ private static void deleteAllFiles(Directory dir) throws IOException {
...
+         */
+        public long getAccumulatedDuration() {
+            return totalDurationInMs;
+        }
+        
+        /**
+         * Record the size of an index considered
+         * @param bytes
+         */
+        public void addIndexSize (long bytes) {
```
thomasmueller, line 156: Nit: why a space before '('?

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/directory/IndexSanityChecker.java
```diff
@@ -118,4 +122,54 @@ private static void deleteAllFiles(Directory dir) throws IOException {
...
     }
+    
+    
+    /**
+     * Accumulated statistics across multiple invocations of the IndexSanityChecker
+     *
+     */
+    public static class IndexSanityStatistics {
+        
+        long totalDurationInMs = 0;
```
thomasmueller: Nit: = 0 is not needed.

thomasmueller (approved):

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/directory/IndexSanityChecker.java
```diff
@@ -118,4 +122,54 @@ private static void deleteAllFiles(Directory dir) throws IOException {
...
+         */
+        public long getAccumulatedIndexSize() {
+            return totalIndexSize;
+        }
+        
+        @Override
+        public String toString() {
+            return String.format("[duration: %d ms, "
+                    + "checked index size: %d bytes (%s)]", totalDurationInMs,
+                    totalIndexSize,IOUtils.humanReadableByteCount(totalIndexSize));
```
fabriziofortino: minor: missing space
```suggestion
                    totalIndexSize, IOUtils.humanReadableByteCount(totalIndexSize));
```

fabriziofortino (approved):

PR #564, opened 2022-05-11
OAK-9769: PathPredicate not being used properly when building FlatFileStore

Ewocker:
PathPredicate not being used properly when building FlatFileStore

This is a bug introduced in [OAK-9747](https://issues.apache.org/jira/browse/OAK-9747) where PathPredicate and hidden node check are separated out from NodeTraverser.

The bug will affect the previous mode performance such as StoreAndSortStrategy and TraverseWithSortStrategy, where we might be indexing more than what we want.

The intention in [OAK-9747](https://issues.apache.org/jira/browse/OAK-9747) was to move the pathPredicate and hidden node check out from the Traverser so that we can save the progress more often for download retry, but this bug with missing configuration and old mode does not honor new way of pathPredicate was created as a side affect.

### DIFF
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
@@ -158,7 +158,7 @@ public abstract class DocumentStoreIndexerBase implements Closeable{
                         .withBlobStore(indexHelper.getGCBlobStore())
                         .withPreferredPathElements((preferredPathElements != null) ? preferredPathElements : indexer.getRelativeIndexedNodeNames())
                         .addExistingDataDumpDir(indexerSupport.getExistingDataDumpDir())
-                        .withPathPredicate((pathPredicate != null) ? pathPredicate : indexer::shouldInclude)
+                        .withPathPredicate(pathPredicate)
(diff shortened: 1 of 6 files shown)


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -251,7 +238,7 @@ public void reindex() throws CommitFailedException, IOException {
 
         closer.register(indexer);
 
-        FlatFileStore flatFileStore = buildFlatFileStore(checkpointedState, indexer);
+        FlatFileStore flatFileStore = buildFlatFileStore(checkpointedState, indexer, indexer::shouldInclude, null);
```
nit0906, line 241: just a nit pick.
this seems redundant since we are already using the same predicate in line 161 too in case the pathPredicate passed to that method is null.
I think either one of these could be removed (I don't have a strong opinion around this though.)

nit0906, line 241: just a nit pick.
this seems redundant since we are already using the same predicate in line 161 too in case the pathPredicate passed to that method is null.
I think either one of these could be removed (I don't have a strong opinion around this though.)

nit0906 (approved):
lgtm , just a minor comment.

thomasmueller (approved):
It looks good, but yes it would be good to remove the redundant code

thomasmueller (approved):

PR #568, opened 2022-05-16
OAK-9686: replace Elastic RHLC with Elastic Java Client

AngelaFabregues:
Query migration to the new Elasticsearch-Java client library.


oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticTestServer.java
```diff
@@ -16,18 +16,19 @@
...
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClients;
 import org.apache.jackrabbit.oak.commons.IOUtils;
-import org.elasticsearch.Version;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testcontainers.DockerClientFactory;
 import org.testcontainers.containers.Network;
 import org.testcontainers.elasticsearch.ElasticsearchContainer;
+import org.testcontainers.shaded.org.bouncycastle.asn1.x509.qualified.ETSIQCObjectIdentifiers;
```
fabriziofortino: unused
```suggestion
```

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticTestServer.java
```diff
@@ -141,4 +142,20 @@ private void checkIfDockerClientAvailable() {
         assumeNotNull(client);
     }
 
+    /**
+     * Launches an Elasticsearch Test Server to re-use among several tests.
+     * @param args
```
fabriziofortino: remove since no args are supported
```suggestion
```

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticTestServer.java
```diff
@@ -141,4 +142,20 @@ private void checkIfDockerClientAvailable() {
...
+     */
+    public static void main(String[] args){
+        ElasticsearchContainer esContainer = ElasticTestServer.getESTestServer();
+        try {
+            System.out.println("Docker container with Elasticsearch launched at \""+esContainer.getHttpHostAddress()+
```
fabriziofortino:
```suggestion
            System.out.println("Docker container with Elasticsearch launched at \"" + esContainer.getHttpHostAddress() +
```

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticTestServer.java
```diff
@@ -141,4 +142,20 @@ private void checkIfDockerClientAvailable() {
...
+                "\". Please PRESS ENTER to stop it...");
+            System.in.read();
+            esContainer.stop();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
```
fabriziofortino: since you are just wrapping the exception in a RuntimeException, you could remove the try-catch block and add the throws in the method declaration

fabriziofortino (changes requested):

oak-search-elastic/pom.xml
```diff
@@ -33,8 +33,8 @@
   <description>Oak Elasticsearch integration subproject</description>
 
   <properties>
-    <elasticsearch.version>7.16.3</elasticsearch.version>
-    <lucene.version>8.10.1</lucene.version>
+    <elasticsearch.version>7.17.3</elasticsearch.version>
+    <jackson.version>2.12.3</jackson.version>
```
nit0906: If possible can we keep this in sync with the jackson dependency in the parent pom - https://github.com/apache/jackrabbit-oak/blob/trunk/oak-parent/pom.xml#L66. ? Either downgrade this or upgrade that.

nit0906: If possible can we keep this in sync with the jackson dependency in the parent pom - https://github.com/apache/jackrabbit-oak/blob/trunk/oak-parent/pom.xml#L66. ? Either downgrade this or upgrade that.

nit0906: ah -  I suppose initially it was picking up the parent pom version - do we specifically need this higher version here ?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexStatistics.java
```diff
@@ -164,77 +157,59 @@ static <K, V> LoadingCache<K, V> setupCache(long maxSize, long expireMin, long r
...
-
-                if (!indices.isEmpty()) {
-                    // we ask for a specific index, so we can get the first entry
-                    Map<String, String> indexProps = indices.get(0);
-                    String size = indexProps.get("store.size");
-                    String creationDate = indexProps.get("creation.date");
-                    String luceneDocsCount = indexProps.get("docs.count");
-                    String luceneDocsDeleted = indexProps.get("docs.deleted");
-                    return new StatsResponse(
+            IndicesRecord record = crd.connection.getClient().cat().indices(i -> i
```
nit0906: Could this be null (similar to the response earlier on ?)
. If yes, then we could possibly have a potential NPE

oak-search-elastic/pom.xml
AngelaFabregues: Checking it

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexStatistics.java
AngelaFabregues: According to the documentation, nulls are prevented by the new library. I will double-check just in case.

AngelaFabregues: You are right, this is a potential null pointer exception. Line 198 ends with getting the first element of a list which can potentially be empty, the index has been deleted.

oak-search-elastic/pom.xml
AngelaFabregues: I've changed the jackson-databind version to the one in the parent module (using the same variable). I've executed mvn clean install in the module and executed the tests. All tests passed so... I suppose we can use the old version instead of the version provided by the documentation https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/7.17/installation.html#maven
@fabriziofortino , is it ok?

AngelaFabregues: @fabriziofortino , do you know why we removed the Lucene version. Or even better, why was it there originally? 
oak-search-elastic tests work well without it.

AngelaFabregues:
@fabriziofortino , I've checked tests while reusing the Elasticsearch instance among test executions. In this case, some tests failed. I skipped two tests from ElasticIndexProviderServiceTest with assumeTrue to avoid null pointer exception. And I left the ElasticReliabilityTest tests even though the assertion fails. Is it enough for this case?

oak-search-elastic/pom.xml
fabriziofortino: The lucene dependency was required by some methods invoked with the RHLC. These methods have been replaced by the new Java client.

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexProviderServiceTest.java
```diff
@@ -95,6 +96,8 @@ public void defaultSetup() {
 
     @Test
     public void withElasticSetup() throws Exception {
+        // Skips this test when a single Elasticsearch cluster is re-used among test executionsgit , see the ElasticTestServer main.
+        assumeTrue(elasticRule.useDocker());
```
fabriziofortino: That's not correct: the tests need to work also with single ES clusters. Please revert these changes. To fix the problem lines 105 and 124 need to be replaced with (I have tried to push the commit myself but I have problem s with this branch):
```java
props.put(PROP_ELASTIC_PORT, elasticRule.getElasticConnectionModel().getElasticPort());
```

fabriziofortino (changes requested):

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexProviderServiceTest.java
nit0906: The assumeTrue(elasticRule.useDocker()) statement should also be removed from both the tests in which it was added. Otherwise they will only run on a docker based setup.

nit0906 (changes requested):
Minor change requested

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexProviderServiceTest.java
AngelaFabregues: I removed it

fabriziofortino (approved):

AngelaFabregues:
> 

I don't know how it got fixed but glad that it is.

nit0906 (approved):

PR #569, opened 2022-05-16
OAK-9758-followup - update log level and log message

Ewocker:


oak-run/src/main/java/org/apache/jackrabbit/oak/plugins/tika/TextExtractor.java
```diff
@@ -272,10 +272,11 @@ public InputStream get() {
...
             // ignore these errors.
-            String format = "Failed to extract text from a binary property: {}."
-                            + " This often happens when some media types are disabled by configuration."
-                            + " The stack trace is included to flag some 'unintended' failures";
-            log.warn(format, linkageErrorFound ? path : new Object[]{path, e});
+            String infoMsgFmt = "Did not extract text from a binary property: {}.";
+            String debugMsgFmt = "This often happens when some media types are disabled by configuration."
+                    + " The stack trace is included to flag some 'unintended' failures. {}";
+            log.info(infoMsgFmt, path);
+            log.debug(debugMsgFmt, linkageErrorFound ? "Stacktrace suppressed due to multiple occurrence of same error." : e);
```
thomasmueller, line 279: There is no need to suppress the stack trace for log.debug, because log.debug is disabled by default. That means, linkageErrorFound is not needed.

thomasmueller (changes requested):

thomasmueller:
I openen an alternative PR: https://github.com/apache/jackrabbit-oak/pull/570

PR #572, opened 2022-05-18
OAK-9771: IndexWriter config should be customisable from oak

tihom88:


thomasmueller (commented):
I think we don't need to document these settings as of now. Not documenting them has the advantage that people don't use them, and hence don't rely on them.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexConfigTest.java
```diff
@@ -0,0 +1,69 @@
...
+        System.setProperty("oak.index.lucene.maxBufferedDeleteTerms", "1000");
+        System.setProperty("oak.index.lucene.ramPerThreadHardLimitMB", "100");
+    }
+
+    @After
+    public void after() {
+        System.setProperties(properties);
+        for (File d : dirs) {
+            FileUtils.deleteQuietly(d);
+        }
```
fabriziofortino: `dirs` does not get ever updated. Can we remove it or some code is missing?

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexConfigTest.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Test
+    public void testIndexWriterConfig() throws Exception {
```
fabriziofortino: not needed
```suggestion
    public void testIndexWriterConfig() {
```

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexConfigTest.java
```diff
@@ -0,0 +1,69 @@
...
+        System.setProperty("oak.index.lucene.ramPerThreadHardLimitMB", "100");
+    }
+
+    @After
+    public void after() {
+        System.setProperties(properties);
+        for (File d : dirs) {
+            FileUtils.deleteQuietly(d);
+        }
+    }
```
fabriziofortino: Could we use this Junit4 extension https://stefanbirkner.github.io/system-rules/#ProvideSystemProperty to safely deal with system properties?

fabriziofortino (approved):
LGTM. Left a few comments on the test class.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexConfigTest.java
tihom88: Done

PR #573, opened 2022-05-18
OAK-9772 Elastic and Lucene tokenizer difference

thomasmueller:


oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
```diff
@@ -228,11 +231,21 @@ protected String getDefaultFunctionName() {
     /**
      * Returns {@code true} if original terms need to be preserved at indexing analysis phase
      */
-    public boolean indexOriginalTerms() {
+    public boolean analyzerConfigIndexOriginalTerms() {
         NodeState analyzersTree = definition.getChildNode(ANALYZERS);
         return getOptionalValue(analyzersTree, INDEX_ORIGINAL_TERM, false);
```
tihom88, line 236: For lucene this is true. I think we should change it's default value to true, so that index definition for lucene and ES remain same WRT analyzers.

https://github.com/apache/jackrabbit-oak/blob/4f99576158aa9b190879291d79002e20aef28ee4/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/OakAnalyzer.java#L69

fabriziofortino (approved):

tihom88 (approved):

PR #574, opened 2022-05-20
OAK-9775 : ACEs with unsupported restrictions must be cleared upon editing, OAK-9782 : CompositeRestrictionProvider must call validate on aggregated providers

anchela:


anchela:
@joerghoh , as discussed in private. don't know why i can't send you an review-invite through the 'reviewers' field. but your feedback would be very much appreciated as always.

mreutegg:
> don't know why i can't send you an review-invite through the 'reviewers' field

I think this is because Joerg is not a Jackrabbit committer.

anchela:
@mreutegg , @reschke , @joerghoh , unless you have any review findings, i would like to merge the changes.
one general question: oak-blob-plugins  and oak-webapp failed in the checks below and i am pretty sure it is unrelated to my changes. it is ok to ignore that?

anchela:
@mreutegg , @reschke , additionally included a fix for OAK-9782 and additional tests verifying that OAK-9775 is also addressed in case of multiple restriction providers are aggregated.
will leave the PR open until start of next week to give you another chance to take a look.

mreutegg:
I won't have time to reviw the changes this week and will be away next week. Maybe some other committer can review?

PR #578, opened 2022-05-24
OAK-9783: useInExcerpt support

AngelaFabregues:
Making the ExcerptTest portable from oak-search to the LuceneExcerptTest in oak-lucene.
Porting it to oak-search-elastic.
Enabling the ingestion of fields to use in excerpts into elastic.
Implementing the excerpts:
- keeping the original test code
- skipping the relative excerpt test because of https://github.com/apache/jackrabbit-oak/blob/a0d1d8c8e511139299252a3ab4aef7d08c1a289c/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java#L147
- ignoring allProps because they are not supported yet.
- using assert eventually
- keeping a single excerpt value (fragment) because even though the Lucene implementation seems to support multiple values (https://github.com/apache/jackrabbit-oak/blob/a87a66322f536832863fce2615fd55d80c9d0f75/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java#L711) they will never be executed because the same implementation is requesting a single fragment (https://github.com/apache/jackrabbit-oak/blob/a87a66322f536832863fce2615fd55d80c9d0f75/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java#L684)

Whenever we merge this PR, the fields to be used in excerpts that weren't indexed already will be indexed now using the default dynamic mapping strategy (keyword + text). Consider optimising the mapping property definition before merging.


AngelaFabregues:
This is still a draft. Better merge it once all Elastic excerpt tests are passed.

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/ExcerptTest.java
```diff
@@ -0,0 +1,340 @@
...
+        contentRoot.setProperty("baz", "fox ifoxing");
+        root.commit();
+
+        List<String> columns = newArrayList("rep:excerpt", "rep:excerpt(.)", "rep:excerpt(foo)", "rep:excerpt(bar)");
+        String selectColumns = Joiner.on(",").join(
+                columns.stream().map(col -> "[" + col + "]").collect(Collectors.toList())
+        );
+        assertEventually(() -> {
+            String query = "SELECT " + selectColumns + " FROM [nt:base] WHERE CONTAINS(*, 'fox')";
+            Result result = null;
```
fabriziofortino: null init is redundant

```suggestion
            Result result;
```

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/ExcerptTest.java
```diff
@@ -0,0 +1,340 @@
...
+    @Test
+    public void nodeExcerpt() throws Exception {
+        Tree contentRoot = root.getTree("/").addChild("testRoot");
+        contentRoot.setProperty("foo", "is fox ifoxing");
+        contentRoot.setProperty("bar", "ifoxing fox");
+        root.commit();
+
+        String query = "SELECT [rep:excerpt],[rep:excerpt(.)] FROM [nt:base] WHERE CONTAINS(*, 'fox')";
+        assertEventually(() -> {
+            Result result = null;
```
fabriziofortino:
```suggestion
            Result result;
```

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/ExcerptTest.java
```diff
@@ -0,0 +1,340 @@
...
+    @Test
+    public void nonIndexedRequestedPropExcerpt() throws Exception {
+        Tree contentRoot = root.getTree("/").addChild("testRoot");
+        contentRoot.setProperty("foo", "fox");
+        contentRoot.setProperty("baz", "is fox ifoxing");
+        root.commit();
+
+        assertEventually(() -> {
+            String query = "SELECT [rep:excerpt(baz)] FROM [nt:base] WHERE CONTAINS(*, 'fox')";
+            Result result = null;
```
fabriziofortino:
```suggestion
            Result result;
```

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/ExcerptTest.java
```diff
@@ -0,0 +1,340 @@
...
+
+    @Test
+    public void propExcerpt() throws Exception {
+        Tree contentRoot = root.getTree("/").addChild("testRoot");
+        contentRoot.setProperty("foo", "is fox ifoxing");
+        root.commit();
+
+        String query = "SELECT [rep:excerpt(foo)] FROM [nt:base] WHERE CONTAINS(*, 'fox')";
+        assertEventually(() -> {
+            Result result = null;
```
fabriziofortino:
```suggestion
            Result result;
```

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/ExcerptTest.java
```diff
@@ -0,0 +1,340 @@
...
+                ResultRow firstRow = resultIter.next();
+
+                PropertyValue nodeExcerpt;
+                String excerpt;
+
+                nodeExcerpt = firstRow.getValue("rep:excerpt(foo)");
+                assertNotNull("rep:excerpt(foo) not evaluated", nodeExcerpt);
+                excerpt = nodeExcerpt.getValue(STRING);
+                assertTrue("rep:excerpt(foo) didn't evaluate correctly - got '" + excerpt + "'",
+                        "is <strong>fox</strong> ifoxing".equals(excerpt));
```
fabriziofortino: this can be simplified 

```suggestion
                assertEquals("rep:excerpt(foo) didn't evaluate correctly - got '" + excerpt + "'", 
                        "is <strong>fox</strong> ifoxing", excerpt);
```

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/ExcerptTest.java
```diff
@@ -0,0 +1,340 @@
...
+
+    @Test
+    public void indexedNonRequestedPropExcerpt() throws Exception {
+        Tree contentRoot = root.getTree("/").addChild("testRoot");
+        contentRoot.setProperty("foo", "is fox ifoxing");
+        root.commit();
+
+        String query = "SELECT [rep:excerpt] FROM [nt:base] WHERE CONTAINS(*, 'fox')";
+        assertEventually(() -> {
+            Result result = null;
```
fabriziofortino:
```suggestion
            Result result;
```

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/ExcerptTest.java
```diff
@@ -0,0 +1,340 @@
...
+                String excerpt;
+
+                nodeExcerpt = firstRow.getValue("rep:excerpt(foo)");
+                assertNotNull("rep:excerpt(foo) not evaluated", nodeExcerpt);
+                excerpt = nodeExcerpt.getValue(STRING);
+                assertTrue("rep:excerpt(foo) didn't evaluate correctly - got '" + excerpt + "'",
+                        excerpt.contains("<strong>fox</strong>"));
+
+                assertTrue("rep:excerpt(foo) highlighting inside words - got '" + excerpt + "'",
+                        !excerpt.contains("i<strong>fox</strong>ing"));
```
fabriziofortino:
```suggestion
                assertFalse("rep:excerpt(foo) highlighting inside words - got '" + excerpt + "'", 
                        excerpt.contains("i<strong>fox</strong>ing"));
```

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/ExcerptTest.java
```diff
@@ -0,0 +1,340 @@
...
+    @Test
+    public void nonIndexedNonRequestedPropExcerpt() throws Exception {
+        Tree contentRoot = root.getTree("/").addChild("testRoot");
+        contentRoot.setProperty("foo", "fox");
+        contentRoot.setProperty("baz", "is fox ifoxing");
+        root.commit();
+
+        String query = "SELECT [rep:excerpt] FROM [nt:base] WHERE CONTAINS(*, 'fox')";
+        assertEventually(() -> {
+            Result result = null;
```
fabriziofortino:
```suggestion
            Result result;
```

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/ExcerptTest.java
```diff
@@ -0,0 +1,340 @@
...
+                String excerpt;
+
+                nodeExcerpt = firstRow.getValue("rep:excerpt(baz)");
+                assertNotNull("rep:excerpt(baz) not evaluated", nodeExcerpt);
+                excerpt = nodeExcerpt.getValue(STRING);
+                assertTrue("rep:excerpt(foo) didn't evaluate correctly - got '" + excerpt + "'",
+                        excerpt.contains("<strong>fox</strong>"));
+
+                assertTrue("rep:excerpt(baz) highlighting inside words - got '" + excerpt + "'",
+                        !excerpt.contains("i<strong>fox</strong>ing"));
```
fabriziofortino:
```suggestion
                assertFalse("rep:excerpt(baz) highlighting inside words - got '" + excerpt + "'", 
                        excerpt.contains("i<strong>fox</strong>ing"));
```

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/ExcerptTest.java
```diff
@@ -0,0 +1,340 @@
...
+            } catch (ParseException e) {
+                fail(e.getMessage());
+            }
+        });
+    }
+
+    //We 'intentionally' are indexing node names only on root state as we don't support indexing relative or
+    //regex for node name indexing. Comment taken from FultextDocumentMaker #148. Test skipped
```
fabriziofortino: typo
```suggestion
    //regex for node name indexing. Comment taken from FulltextDocumentMaker #148. Test skipped
```

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/ExcerptTest.java
```diff
@@ -0,0 +1,340 @@
...
+    //@Test
+    public void relativePropExcerpt() throws Exception {
+        Tree contentRoot = root.getTree("/").addChild("testRoot");
+        contentRoot.addChild("relative").setProperty("baz", "is fox ifoxing");
+        root.commit();
+
+        String query = "SELECT [rep:excerpt(relative/baz)] FROM [nt:base] WHERE CONTAINS([relative/baz], 'fox')";
+        assertEventually(() -> {
+            Result result = null;
```
fabriziofortino:
```suggestion
            Result result;
```

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/ExcerptTest.java
```diff
@@ -0,0 +1,340 @@
...
+                ResultRow firstRow = resultIter.next();
+
+                PropertyValue nodeExcerpt;
+                String excerpt;
+
+                nodeExcerpt = firstRow.getValue("rep:excerpt(relative/baz)");
+                assertNotNull("rep:excerpt(relative/baz) not evaluated", nodeExcerpt);
+                excerpt = nodeExcerpt.getValue(STRING);
+                assertTrue("rep:excerpt(relative/baz) didn't evaluate correctly - got '" + excerpt + "'",
+                        "is <strong>fox</strong> ifoxing".equals(excerpt));
```
fabriziofortino:
```suggestion
                assertEquals("rep:excerpt(relative/baz) didn't evaluate correctly - got '" + excerpt + "'", 
                        "is <strong>fox</strong> ifoxing", excerpt);
```

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/ExcerptTest.java
```diff
@@ -0,0 +1,340 @@
...
+                        !excerpt.contains("i<strong>fox</strong>ing"));
+            } catch (ParseException e) {
+                fail(e.getMessage());
+            }
+        });
+    }
+
+    //We 'intentionally' are indexing node names only on root state as we don't support indexing relative or
+    //regex for node name indexing. Comment taken from FultextDocumentMaker #148. Test skipped
+    //@Test
```
fabriziofortino:
```suggestion
    @Test
    @Ignore
```

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/ExcerptTest.java
```diff
@@ -0,0 +1,340 @@
...
+        Tree contentRoot = root.getTree("/").addChild("testRoot");
+
+        String binaryText = "is fox foxing as a fox cub";
+        Blob blob = new ArrayBasedBlob(binaryText.getBytes());
+        TestUtil.createFileNode(contentRoot, "binaryNode", blob, "text/plain");
+        root.commit();
+
+        String query = "SELECT [rep:excerpt] FROM [nt:base] WHERE CONTAINS(*, 'fox')";
+        assertEventually(() -> {
+            Result result = null;
```
fabriziofortino:
```suggestion
            Result result;
```

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticDocument.java
```diff
@@ -147,13 +155,18 @@ public String build() {
                 if (!similarityTags.isEmpty()) {
                     builder.field(ElasticIndexDefinition.SIMILARITY_TAGS, similarityTags);
                 }
+                for (Map.Entry<String, Set<Object>> prop : toStoreInSource.entrySet()) {
+                    if (!properties.containsKey(prop.getKey())) {
+                        builder.field(prop.getKey(), prop.getValue());
+                    }
+                }
```
fabriziofortino: I guess this was made to support regex properties. Although it would work, we should refrain to do that because it will go against the strict mapping we need to soon support.

This is not just a problem related to the excerpt feature but a more general one (eg: regex properties, function properties).

My proposal is to simplify this PR by adding excerpt support for specific properties only (we need to document the difference) and have a separate PR to cover the dynamic properties and aim for a NON-dynamic mapping.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticDocument.java
```diff
@@ -28,12 +28,7 @@
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
fabriziofortino: do not use * imports

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
```diff
@@ -138,13 +140,38 @@ public void on(Hit<ObjectNode> searchHit) {
...
+    /**
+     * Reads excerpts from elasticsearch response.
+     * rep:excerpt and rep:excerpt(.) keys are used for :fulltext
+     * rep:excerpt(PROPERTY) for other fields.
+     * Note: properties to get excerpt from must be included in the _source, which means ingested,
+     * not necessarily Elasticsearch indexed, neither included in the mapping properties.
+     *
+     * @param searchHit
+     * @return a map with the excerpts
+     * @throws IOException
```
fabriziofortino: the method does not throw any IOException and the other stuff does not bring additional info. I would remove it
```suggestion
```

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
```diff
@@ -138,13 +140,38 @@ public void on(Hit<ObjectNode> searchHit) {
...
+     */
+    private Map<String, String> readExcerptsFromResponse(Hit<ObjectNode> searchHit) {
+        Map<String, String> excerpts = Maps.newHashMap();
```
fabriziofortino: Guava does not bring any value here. We are trying to minimize its use. Please remove it
```suggestion
        Map<String, String> excerpts = new HashMap<>();
```

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMaker.java
```diff
@@ -88,7 +88,7 @@ public LuceneDocumentMaker(@Nullable FulltextBinaryTextExtractor textExtractor,
     @Override
     protected void indexAnalyzedProperty(Document doc, String pname, String value, PropertyDefinition pd) {
         String analyzedPropName = constructAnalyzedPropertyName(pname);
-        doc.add(newPropertyField(analyzedPropName, value, !pd.skipTokenization(pname), pd.stored));
+        doc.add(newPropertyField(analyzedPropName, value, !pd.skipTokenization(pname), pd.useInExcerpt));
```
fabriziofortino: not sure about this change

fabriziofortino (changes requested):
I reviewed it partially. I think we should simplify the PR (see my comment ElasticDocument). I will complete the review after discussing that point.

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/ExcerptTest.java
AngelaFabregues: This was the original test code.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticDocument.java
AngelaFabregues: Consider dynamic:false instead of dynamic:strict. With dynamic:false the ingested fields are only stored in _source. Which is enough for excerpt.

AngelaFabregues: Aligned, regex are excluded and a next PR is setting the fields dynamicly created to be only stored. if they are useInExcerpt only.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMaker.java
AngelaFabregues: I've reverted even thought I still don't see the why stored is even ok here :)

AngelaFabregues: The pd.stored is being populated from useInExcerpt JCR property name. The PropertyDefinition class as a concept represents the JCR properties of a JCR node which represents an oak index rule property.
stored is not a valid JCR node property for an oak index rule property. Stored is a Lucene/Elastic internal thing. The class and the methods become harder to read with stored instead of useInExcerpt.
Also, I think that backward compatibility is provided at oak index rule property configuration, not at class param (code) level, right? <--- not sure @fabriziofortino

fabriziofortino: `oak-search`  could be potentially used by an external project as a dependency. If you want to change a public member, you have to a) deprecate it b) remove it in a successive major release. My suggestion is to use `stored`.

fabriziofortino (approved):

nit0906 (commented):
A general query here - in case of lucene - we use useInExcerpt property to enable/disable if a value of a given property would be used in excerpts (since this would increase the index size).

But I couldn't find dependency on this property in this PR - so is it that elastic supports excerpts by default ? And we just need to handle for query time ?

fabriziofortino:
@nit0906 for Elastic, excerpts (aka highlight) get extracted from the source. We have decided to always store the complete source. Disabling it could save some space but it would have some consequences. See `Think before disabling the _source field` in https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-source-field.html

nit0906 (approved):

PR #582, opened 2022-06-01
Deeper representation of the trees with properties and in json format.

AngelaFabregues:
AbstractTree's previous toString() method was calling a toString(int depth) method that wasn't working really. It did only represent the first level of the tree. Also the properties were represented together with the child nodes.
This implementation goes as deep as necessary and represents the node in JSON format. Therefore, making it queryable with JSON tools like jq.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
@@ -41,6 +41,7 @@ import org.apache.jackrabbit.oak.api.PropertyState;
 import org.apache.jackrabbit.oak.api.Tree;
 import org.apache.jackrabbit.oak.api.Type;
 import org.apache.jackrabbit.oak.commons.PathUtils;
+import org.apache.jackrabbit.oak.commons.json.JsopBuilder;
 import org.apache.jackrabbit.oak.plugins.index.IndexConstants;
 import org.apache.jackrabbit.oak.plugins.index.reference.NodeReferenceConstants;
 import org.apache.jackrabbit.oak.plugins.tree.TreeConstants;
@@ -175,18 +176,18 @@ public abstract class AbstractTree implements Tree {
                 str.append("[ ");
                 for (int i = 0; i < ps.count(); i++) {
                     try {
-                        str.append(quote(escapeJsonString(ps.getValue(Type.STRING, i))) + ",");
+                        str.append(JsopBuilder.encode(ps.getValue(Type.STRING, i)) + ",");
                     } catch (Exception e) {
-                        str.append(quote("ERROR:" + escapeJsonString(e.getMessage())) + ",");
+                        str.append(quote("ERROR:" + JsopBuilder.encode(e.getMessage())) + ",");
                     }
                 }
                 str.deleteCharAt(str.length() - 1); //removing the space or the ,
                 str.append("],");
             } else {
                 try {
-                    str.append(quote(escapeJsonString(ps.getValue(Type.STRING))) + ",");
+                    str.append(JsopBuilder.encode(ps.getValue(Type.STRING)) + ",");
                 } catch (Exception e) {
-                    str.append(quote("ERROR:" + escapeJsonString(e.getMessage())) + ",");
+                    str.append(quote("ERROR:" + JsopBuilder.encode(e.getMessage())) + ",");
                 }
             }
         }
@@ -199,10 +200,6 @@ public abstract class AbstractTree implements Tree {
         str.deleteCharAt(str.length() - 1); //removing the ,
         str.append("}");
         return str.toString();
-    }
-
-    private String escapeJsonString(String value) {
-        return value.replaceAll("\"", "\\/");
     }
 
     private String quote(String toQuote) {
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
@@ -31,7 +31,6 @@ import static org.apache.jackrabbit.oak.api.Tree.Status.UNCHANGED;
 import static org.apache.jackrabbit.oak.api.Type.NAMES;
 import static org.apache.jackrabbit.oak.plugins.tree.TreeConstants.OAK_CHILD_ORDER;
 
-import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 
@@ -166,39 +165,40 @@ public abstract class AbstractTree implements Tree {
         if(depth == 0){
             return quote("...");
         }
-        String str = "{";
-        str += quote("_properties_")+":{ ";
+        StringBuilder str = new StringBuilder();
+        str.append("{");
+        str.append(quote("_properties_")+":{ ");
         for (PropertyState ps : this.getProperties()){
-            str+=quote(ps.getName())+":";
+            str.append(quote(ps.getName())+":");
 
             if(ps.getType().isArray()){
-                str += "[ ";
+                str.append("[ ");
                 for(int i=0; i<ps.count(); i++){
                     try {
-                        str += quote(jsonStringEscaper(ps.getValue(Type.STRING, i)))+",";
+                        str.append(quote(jsonStringEscaper(ps.getValue(Type.STRING, i)))+",");
                     }catch (Exception e){
-                        str += quote("ERROR:" + jsonStringEscaper(e.getMessage()))+",";
+                        str.append(quote("ERROR:" + jsonStringEscaper(e.getMessage()))+",");
                     }
                 }
-                str = str.substring(0,str.length()-1); //removing the space or the ,
-                str += "],";
+                str.deleteCharAt(str.length()-1); //removing the space or the ,
+                str.append("],");
             }else {
                 try{
-                    str += quote(jsonStringEscaper(ps.getValue(Type.STRING))) + ",";
+                    str.append(quote(jsonStringEscaper(ps.getValue(Type.STRING))) + ",");
                 }catch (Exception e){
-                    str += quote("ERROR:" + jsonStringEscaper(e.getMessage())) + ",";
+                    str.append(quote("ERROR:" + jsonStringEscaper(e.getMessage())) + ",");
                 }
             }
         }
-        str = str.substring(0,str.length()-1); //removing the space or the ,
-        str+="},";
+        str.deleteCharAt(str.length()-1);  //removing the space or the ,
+        str.append("},");
         for (Tree child : this.getChildren()){
-            str+=quote(child.getName())+":";
-            str+=((AbstractTree)child).toJsonString(depth-1)+",";
+            str.append(quote(child.getName())+":");
+            str.append(((AbstractTree)child).toJsonString(depth-1)+",");
         }
-        str = str.substring(0,str.length()-1); //removing the ,
-        str+="}";
-        return str;
+        str.deleteCharAt(str.length()-1); //removing the ,
+        str.append("}");
+        return str.toString();
     }
 
     private String jsonStringEscaper(String value){
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
@@ -177,7 +177,6 @@ public abstract class AbstractTree implements Tree {
                     try {
                         str += "\"" + jsonStringEscaper(ps.getValue(Type.STRING, i)) + "\",";
                     }catch (Exception e){
-                        e.printStackTrace();
                         str += "\"ERROR:" + jsonStringEscaper(e.getMessage()) + "\",";
                     }
                 }
@@ -187,7 +186,6 @@ public abstract class AbstractTree implements Tree {
                 try{
                     str += "\"" + jsonStringEscaper(ps.getValue(Type.STRING)) + "\",";
                 }catch (Exception e){
-                    e.printStackTrace();
                     str += "\"ERROR:" + jsonStringEscaper(e.getMessage()) + "\",";
                 }
             }
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
@@ -161,51 +161,51 @@ public abstract class AbstractTree implements Tree {
      * @param depth tree depth to represent. Use -1 for an unlimited depth.
      * @return json object representation of the tree as a string
      */
-    public String toJsonString(int depth){
-        if(depth == 0){
+    public String toJsonString(int depth) {
+        if (depth == 0) {
             return quote("...");
         }
         StringBuilder str = new StringBuilder();
         str.append("{");
-        str.append(quote("_properties_")+":{ ");
-        for (PropertyState ps : this.getProperties()){
-            str.append(quote(ps.getName())+":");
+        str.append(quote("_properties_") + ":{ ");
+        for (PropertyState ps : this.getProperties()) {
+            str.append(quote(ps.getName()) + ":");
 
-            if(ps.getType().isArray()){
+            if (ps.getType().isArray()) {
                 str.append("[ ");
-                for(int i=0; i<ps.count(); i++){
+                for (int i = 0; i < ps.count(); i++) {
                     try {
-                        str.append(quote(jsonStringEscaper(ps.getValue(Type.STRING, i)))+",");
-                    }catch (Exception e){
-                        str.append(quote("ERROR:" + jsonStringEscaper(e.getMessage()))+",");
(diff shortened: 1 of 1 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -150,34 +151,60 @@ protected Iterable<String> getChildNames() {
...
+        for (PropertyState ps : this.getProperties()){
+            str+="\""+ps.getName()+"\":";
+
+            if(ps.getType().isArray()){
+                str += "[ ";
+                for(int i=0; i<ps.count(); i++){
+                    try {
+                        str += "\"" + jsonStringEscaper(ps.getValue(Type.STRING, i)) + "\",";
+                    }catch (Exception e){
+                        e.printStackTrace();
```
joerghoh: e.printStackTrace will not be logged, but always printed to stderr, which is not always captured, and misses context like a timestamp. 
Please use proper logging.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -150,34 +151,60 @@ protected Iterable<String> getChildNames() {
...
+        if(depth == 0){
+            return "\"...\"";
         }
-
-        if (names.hasNext()) {
-            sb.append(" ...");
+        String str = "{";
+        str += "\"_properties_\":{ ";
+        for (PropertyState ps : this.getProperties()){
+            str+="\""+ps.getName()+"\":";
```
joerghoh: I don't feel comfortable with all these string operations here, especially when it's about quoting.

can you wrap these string operations into meaningful static methods like this?
```
private static String quote (String toQuote) {
  return "\"" + toQuote + "\"";
}

[...]
str+=quote(ps.get.Name());
[...]
```
That would make the code easier to understand.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -150,34 +151,60 @@ protected Iterable<String> getChildNames() {
...
-            sb.append(' ').append(names.next()).append(" = { ... },");
+    /**
+     * Represents an OAK tree as a JSON object string
+     * When the representation is limited in depth, {"...":"..."} represents the limit.
+     * Note that, try-catch sentences are used to add "ERROR: "+e.getMessage() values in case of representation error.
+     * Note that, this is a recursive method.
+     * @param depth tree depth to represent. Use -1 for an unlimited depth.
+     * @return json object representation of the tree as a string
+     */
+    public String toJsonString(int depth){
```
joerghoh: I think that we need to have some tests for this.

joerghoh (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
AngelaFabregues: True!! and it is neither needed. I will delete the e.printStackTrace();

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
AngelaFabregues: Cool!

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
AngelaFabregues: I need your help here. I've tested it with the ExcerptTest repositories created with initial data for elastic and lucene. The methods seems to work like a charm.
Feel free to debug any code and evaluate, for instance, trees from the root with `((AbstractTree)root.getTree("/")).toJsonString(-1)`.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -150,34 +151,62 @@ protected Iterable<String> getChildNames() {
...
-            sb.append(' ').append(p).append(',');
+    /**
+     * Represents an OAK tree as a JSON object string
+     * When the representation is limited in depth, {"...":"..."} represents the limit.
+     * Note that, try-catch sentences are used to add "ERROR: "+e.getMessage() values in case of representation error.
+     * Note that, this is a recursive method.
+     * @param depth tree depth to represent. Use -1 for an unlimited depth.
+     * @return json object representation of the tree as a string
+     */
+    public String toJsonString(int depth){
```
thomasmueller: Nit: please add spaces before "{" (standard code style)

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -150,34 +151,62 @@ protected Iterable<String> getChildNames() {
...
         }
-
-        Iterator<String> names = this.getChildNames().iterator();
-        int count = 0;
-        while (names.hasNext() && ++count <= childNameCountLimit) {
-            sb.append(' ').append(names.next()).append(" = { ... },");
+        String str = "{";
+        str += quote("_properties_")+":{ ";
+        for (PropertyState ps : this.getProperties()){
+            str+=quote(ps.getName())+":";
```
thomasmueller: spaces should be added around "+=" and "+", and after "if" and so on.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -150,34 +151,62 @@ protected Iterable<String> getChildNames() {
...
+
+            if(ps.getType().isArray()){
+                str += "[ ";
+                for(int i=0; i<ps.count(); i++){
+                    try {
+                        str += quote(jsonStringEscaper(ps.getValue(Type.STRING, i)))+",";
+                    }catch (Exception e){
```
thomasmueller: Spaces after "}" and before "{"

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -150,34 +151,62 @@ protected Iterable<String> getChildNames() {
...
-        while (names.hasNext() && ++count <= childNameCountLimit) {
-            sb.append(' ').append(names.next()).append(" = { ... },");
+        String str = "{";
+        str += quote("_properties_")+":{ ";
+        for (PropertyState ps : this.getProperties()){
+            str+=quote(ps.getName())+":";
+
+            if(ps.getType().isArray()){
+                str += "[ ";
+                for(int i=0; i<ps.count(); i++){
```
thomasmueller: Spaces around "=" and "<" and after "for" and before "{"

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -150,34 +151,62 @@ protected Iterable<String> getChildNames() {
...
+                    try {
+                        str += quote(jsonStringEscaper(ps.getValue(Type.STRING, i)))+",";
+                    }catch (Exception e){
+                        str += quote("ERROR:" + jsonStringEscaper(e.getMessage()))+",";
+                    }
+                }
+                str = str.substring(0,str.length()-1); //removing the space or the ,
```
thomasmueller: I would use the following style instead:

    for (int i = 0; i < ps.count(); i++) {
        if (i > 0) {
            buff.append(", ");
        }
        buff.append(...)

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -150,34 +151,62 @@ protected Iterable<String> getChildNames() {
...
+    public String toJsonString(int depth){
+        if(depth == 0){
+            return quote("...");
         }
-
-        Iterator<String> names = this.getChildNames().iterator();
-        int count = 0;
-        while (names.hasNext() && ++count <= childNameCountLimit) {
-            sb.append(' ').append(names.next()).append(" = { ... },");
+        String str = "{";
```
thomasmueller: Please use StringBuilder instead of concatenating strings.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -150,34 +151,62 @@ protected Iterable<String> getChildNames() {
...
-            sb.deleteCharAt(sb.length() - 1);
-        }
-        sb.append('}');
+    private String jsonStringEscaper(String value){
+        return value.replaceAll("\"","\\/");
+    }
 
-        return sb.toString();
+    private String quote (String toQuote) {
+        return "\"" + toQuote + "\"";
```
thomasmueller, line 206: What about escaping?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -150,34 +151,62 @@ protected Iterable<String> getChildNames() {
...
-        if (sb.charAt(sb.length() - 1) == ',') {
-            sb.deleteCharAt(sb.length() - 1);
-        }
-        sb.append('}');
+    private String jsonStringEscaper(String value){
+        return value.replaceAll("\"","\\/");
+    }
 
-        return sb.toString();
+    private String quote (String toQuote) {
```
thomasmueller: Actually there should be no space before "("

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
AngelaFabregues: Done!

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
AngelaFabregues: Done!

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
AngelaFabregues: Done

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
AngelaFabregues: Done

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
AngelaFabregues: Done

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
AngelaFabregues: Done!

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
AngelaFabregues: done

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
AngelaFabregues: it would be` if ( i < ps.count()-1 ) { `at the end of the for and I cannot use it for `for (Tree child : this.getChildren()) {`.
I think it will be more confusing to have two ways of doing the same. Any alternatives?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -150,34 +150,63 @@ protected Iterable<String> getChildNames() {
...
+                }
+            }
         }
-
-        if (names.hasNext()) {
-            sb.append(" ...");
+        str.deleteCharAt(str.length() - 1);  //removing the space or the ,
+        str.append("},");
+        for (Tree child : this.getChildren()) {
+            str.append(quote(child.getName()) + ":");
```
thomasmueller, line 197: I think it needs to be:

    str.append(quote(escapeJsonString(child.getName())) + ":");

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -150,34 +150,63 @@ protected Iterable<String> getChildNames() {
...
-        if (sb.charAt(sb.length() - 1) == ',') {
-            sb.deleteCharAt(sb.length() - 1);
-        }
-        sb.append('}');
+    private String escapeJsonString(String value) {
+        return value.replaceAll("\"","\\/");
+    }
 
-        return sb.toString();
+    private String quote(String toQuote) {
```
thomasmueller: private static

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -150,34 +150,63 @@ protected Iterable<String> getChildNames() {
...
+        str.append("}");
+        return str.toString();
+    }
 
-        if (sb.charAt(sb.length() - 1) == ',') {
-            sb.deleteCharAt(sb.length() - 1);
-        }
-        sb.append('}');
+    private String escapeJsonString(String value) {
+        return value.replaceAll("\"","\\/");
```
thomasmueller: This doesn't correctly escape... It's probably better to use oak-commons org.apache.jackrabbit.oak.commons.json 
JsopBuilder.escape

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
thomasmueller: so:

    StringBuilder buff = new StringBuilder();
    JsopBuilder.escape(s, s.length(), buff);
    return buff.toString();

AngelaFabregues: JsopBuilder.encode is perfect. Thanks for spotting it out.
Done

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
AngelaFabregues: Done

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
AngelaFabregues: Applied with JsopBuilder.encode()

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -150,34 +151,59 @@ protected Iterable<String> getChildNames() {
...
+            return quote("...");
         }
-
-        if (names.hasNext()) {
-            sb.append(" ...");
+        StringBuilder str = new StringBuilder();
+        str.append("{");
+        str.append(quote("_properties_") + ":{ ");
+        for (PropertyState ps : this.getProperties()) {
+            str.append(quote(ps.getName()) + ":");
```
thomasmueller, line 173: To replace the trailing comma with a prefix comma:

    int count = 0;
    for (PropertyState ps : this.getProperties()) {
        if (count++ > 0) {
            buff.append(", ");
        }

reschke: or

~~~
String sep = "";
for (PropertyState ps : this.getProperties()) {
    buf.append(sep);
    sep = ", ";
~~~

AngelaFabregues:
Hi, we've merged the code into a branch so more people can work on improving it before merging to trunk.
This is the merged PR https://github.com/apache/jackrabbit-oak/pull/583 towards the branch apache:OAK-9789.

PR #583, opened 2022-06-01
Abstract tree to string fixed as json

AngelaFabregues:


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -150,34 +151,59 @@ protected Iterable<String> getChildNames() {
...
+                    str.append(JsopBuilder.encode(ps.getValue(Type.STRING)) + ",");
+                } catch (Exception e) {
+                    str.append(quote("ERROR:" + JsopBuilder.encode(e.getMessage())) + ",");
+                }
+            }
         }
-
-        if (sb.charAt(sb.length() - 1) == ',') {
-            sb.deleteCharAt(sb.length() - 1);
+        str.deleteCharAt(str.length() - 1);  //removing the space or the ,
```
reschke, line 194: I would avoid that approach. Just start with an empty separator string, and set it to ", " after first use.

AngelaFabregues: Yeah! We had a discussion with @thomasmueller and @joerghoh about this. We thought that better merge it to a branch so more people can contribute before it is PR towards the trunk.

AngelaFabregues:
Let's merge it to this branch so more people can test it.

thomasmueller:
Notice we merge to a branch only

PR #590, opened 2022-06-10
OAK-9773: DefaultSyncContext#syncMembership() compares external ids c…

mbaedke:
…ase-sensitively.

Fixed.


oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/basic/DefaultSyncContext.java
```diff
@@ -522,7 +522,7 @@ protected void syncMembership(@NotNull ExternalIdentity external, @NotNull Autho
         while (grpIter.hasNext()) {
             Group grp = grpIter.next();
             if (isSameIDP(grp)) {
-                declaredExternalGroups.put(grp.getID(), grp);
+                declaredExternalGroups.put(grp.getID().toLowerCase(), grp);
```
anchela, line 527: it is an implementation detail that the default user management in oak is not case-sensitive.
therefore, at the very least this code should log a huge warning if the map already contained a previous value which would be replaced by this call.

anchela (changes requested):
see comments.
apart from that: please also add tests to this PR.

oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/basic/DefaultSyncContext.java
mbaedke: Done. Whoever has this issue will be spammed with warnings, but I guess they have to live with that.

anchela: hi @mbaedke , i am not sure i can follow you on this one..... what is the new group-ids list for? and why would someone get spammed with warnings? sorry.... i don't get it.

what i had in mind is something along the following lines:
```
Group previous = declaredExternalGroups.put(grp.getID().toLowerCase(), grp);
if (previous != null) {
     log.warn("Replacing group '{}' with same case-insensitive name '{}'", previous.getId(), grp.getID());
}

```

mbaedke: Hi @anchela, I don't understand. Your code would do nothing, because declaredExternalGroups is empty initially.
Re spamming: the warnings will stay, because the ids do not change in the repo or in the LDAP.

anchela (approved):
misunderstanding regarding comment sorted out in private discussion

PR #592, opened 2022-06-10
OAK-9800 update Maven plugins to most recent version

kwin:


oak-parent/pom.xml
```diff
@@ -868,7 +828,7 @@
 
    <!--  - - - - - - - - - - - - - - - - - - - will run all the required tests before a release -->
    <profile>
-      <id>release</id>
```
kwin, line 872: which profile has been used in the past? The one from here or the one from the reactor root?

mreutegg: You mean while creating a release? I think the one in the reactor root. When I created the current 1.44.0 release I did not specify the profile `release`. I just ran the maven release commands without setting any profile explicitly.

This release profile was created with https://issues.apache.org/jira/browse/OAK-4136

I still don't quite understand why it was created, but I would rather refrain from changing the profile name to `apache-release`. I think it might mess up the release process.

kwin: Ok, even the formerly used ASF Parent 18 had https://github.com/apache/maven-apache-parent/blob/aad3c292f623f798936bf8daeb5237a53282eaa4/pom.xml#L208. Therefore this profile seems to be legacy and can be removed. It was never automatically activated and the release process at https://jackrabbit.apache.org/jcr/creating-releases.html doesn't mention it either, so I don't think anyone ever activated it manually.

mreutegg: Well, according to https://lists.apache.org/thread/10f7mfj31pt7hx4yg27k8swtoovfx0q4 Davide did use it ;-)

I'm fine removing it, but it doesn't necessarily need to happen in this PR.

mreutegg: And just to properly close the loop here. Davide created the profile to run tests and checks before the actual release. The `release` profile is not intended to be used while creating the release.

kwin: I will add a comment making the intent more clear. Then later on we can decide if we want to remove it.

oak-parent/pom.xml
```diff
@@ -34,6 +34,8 @@
   <packaging>pom</packaging>
 
   <properties>
+    <minimalMavenBuildVersion>3.3.9</minimalMavenBuildVersion>
+    <maven.compiler.target>1.8</maven.compiler.target>
```
rombert: The value is duplicated with the java.version property. Shouldn't the value be set to `${java.version}`?

rombert (commented):
LGTM overall, only a question regarding duplication.

oak-parent/pom.xml
kwin: Yes, good point, although `java.version` is a poor name for a Maven Property as there is also the same named [Java system property](https://docs.oracle.com/javase/tutorial/essential/environment/sysprop.html): https://maven.apache.org/pom.html#properties but that is a different discussion....

kwin: Done in https://github.com/apache/jackrabbit-oak/pull/592/commits/27f0db6327475abf02c2c5357e5c9f3b2d32c25d.

oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/checks/NodeTypeDefinitionNodeStoreChecker.java
```diff
@@ -75,6 +75,11 @@ public void onConstraintViolation(String path, List<String> typeNames, int code,
         NodeBuilder rootBuilder = new ReadOnlyBuilder(root); // prevent accidental changes
         
         String primary = root.getName(JCR_PRIMARYTYPE);
+        // workaround until https://issues.apache.org/jira/browse/OAK-9863 is fixed
```
kwin, line 78: @rombert Do you plan on working on OAK-9863 soon or are you fine with this temporary workaround for now?

rombert: Let's go with this one for now.

rombert (approved):

PR #594, opened 2022-06-14
OAK-9801 improve error messages

joerghoh:
provide better error messages in case a null value is provided as parameter

### DIFF
--- a/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionImpl.java
+++ b/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionImpl.java
@@ -147,8 +147,7 @@ public class SessionImpl implements JackrabbitSession {
     }
 
     @NotNull
-    private String getOakPathOrThrow(@NotNull String absPath)
-            throws RepositoryException {
+    private String getOakPathOrThrow(@NotNull String absPath) throws RepositoryException {
         String p = sessionContext.getOakPathOrThrow(absPath);
         if (!PathUtils.isAbsolute(p)) {
             throw new RepositoryException("Not an absolute path: " + absPath);
@@ -157,14 +156,12 @@ public class SessionImpl implements JackrabbitSession {
     }
 
     @NotNull
-    private String getOakPathOrThrowNotFound(@NotNull String absPath)
-            throws PathNotFoundException {
+    private String getOakPathOrThrowNotFound(@NotNull String absPath) throws PathNotFoundException {
         return sessionContext.getOakPathOrThrowNotFound(absPath);
     }
 
     @Nullable
-    private ItemImpl<?> getItemInternal(@NotNull String oakPath)
-            throws RepositoryException {
+    private ItemImpl<?> getItemInternal(@NotNull String oakPath) throws RepositoryException {
         checkAlive();
         ItemDelegate item = sd.getItem(oakPath);
         if (item instanceof NodeDelegate) {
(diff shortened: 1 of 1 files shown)


mreutegg (commented):
Looks useful to me. Can you please insert a space between the comma and the second parameter?

reschke (changes requested):
Rather than adding more use of Guava, shouldn't we use https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-java.lang.String- ?

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionImpl.java
```diff
@@ -223,7 +220,7 @@ public Property performNullable() {
     @Override
     @Nullable
     public Item getItemOrNull(final String absPath) throws RepositoryException {
```
anchela, line 222: could we also mark the param as @NotNull in the jackrabbit API and here?

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionImpl.java
```diff
@@ -179,7 +176,7 @@ private ItemImpl<?> getItemInternal(@NotNull String oakPath)
     @Override
     @Nullable
     public Node getNodeOrNull(final String absPath) throws RepositoryException {
```
anchela, line 178: could we also mark the param as NotNull in the jackrabbit-api and here?

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionImpl.java
```diff
@@ -197,7 +194,7 @@ public Node performNullable() throws RepositoryException {
     @Nullable
     public Property getPropertyOrNull(final String absPath) throws RepositoryException {
```
anchela, line 195: could we mark the param as NotNull in the jackrabbit-api and here? that might actually spare us the null check altogether..... couldn't it?

anchela (approved):
hi @joerghoh the changes look sensible to me.
what i would do though (probably under a different JIRA ticket) is mark the parameters for the getXOrNull methods as @NotNull to make the intent clear. then we might even omit the null-check altogether....

joerghoh:
@anchela I agree on adding the annotations, but in the case of a public API I would do the checks nevertheless ...

joerghoh:
I cannot merge, can someone from the Oak team perform that for me? Thanks!

anchela:
hi @joerghoh , did you address the concern raised by @reschke ? once that is sorted out i will merge right away.

joerghoh:
> Rather than adding more use of Guava, shouldn't we use https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-java.lang.String- ?

Good idea, switched over.

reschke (approved):
It would probably be good to undo white-space only changes...

anchela:
hi @reschke , i usually agree that whitespace-only changes should be tracked separately. but since this PR is about a minor cleanup in the error-msg, i don't find them as distracting or making the PR harder to read. unless you object, i would merge it anyway.

anchela:
@joerghoh , i merged it. for a next PR i would like to encourage you to follow the advice from @reschke to keep whitespace-only changes separately. it makes the PR harder to review as it distracts from the real changes you want to make.

PR #608, opened 2022-06-28
describe the process of executing a query

joerghoh:
Describe the execution of a query

### DIFF
--- a/oak-doc/src/site/markdown/query/query-engine.md
+++ b/oak-doc/src/site/markdown/query/query-engine.md
@@ -139,7 +139,7 @@ If a query requests an ordered result set, the Query Engine tries to get an alre
 
 #### Iterating the Result Set
 Query results are implemented as lazy iterators, and the result set is only read if needed. When the next result is requested, the result iterator seeks the potential nodes to find the next node matching the query. 
-During this seek process the Query Engine reads and filters the potential nodes until if finds a match. Even if the query is handled completely by an index, the Query Engine needs to check if the requesting session is allowed to read the nodes.
+During this seek process the Query Engine reads and filters the potential nodes until it finds a match. Even if the query is handled completely by an index, the Query Engine needs to check if the requesting session is allowed to read the nodes.
 
 That means that during this final step every potential node must be loaded from the node store, thus counting towards the read limit (see [Slow Queries and Read Limits](#slow-queries-and-read-limits)).
 
--- a/oak-doc/src/site/markdown/query/query-engine.md
+++ b/oak-doc/src/site/markdown/query/query-engine.md
@@ -22,36 +22,41 @@ grep "^#.*$" src/site/markdown/query/query-engine.md | sed 's/#/    /g' | sed 's
 
 ## The Query Engine
 
-* [Overview](#Overview)
-    * [Query Processing](#Query_Processing)
-        * [Cost Calculation](#Cost_Calculation)
-    * [Query Options](#Query_Options)
-        * [Query Option Traversal](#Query_Option_Traversal)
-        * [Query Option Index Tag](#Query_Option_Index_Tag)
-        * [Query Option Offset / Limit](#Query_Option_Offset__Limit)
-        * [Index Selection Policy](#Index_Selection_Policy)
-    * [Compatibility](#Compatibility)
-        * [Result Size](#Result_Size)
-        * [Quoting](#Quoting)
-        * [Equality for Path Constraints](#Equality_for_Path_Constraints)
-    * [Slow Queries and Read Limits](#Slow_Queries_and_Read_Limits)
-    * [Keyset Pagination](#Keyset_Pagination)
-    * [Full-Text Queries](#Full-Text_Queries)
-    * [Excerpts and Highlighting](#Excerpts_and_Highlighting)
-    * [Native Queries](#Native_Queries)
-    * [Similarity Queries](#Similarity_Queries)
-    * [Spellchecking](#Spellchecking)
-    * [Suggestions](#Suggestions)
-    * [Facets](#Facets)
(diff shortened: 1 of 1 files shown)


oak-doc/src/site/markdown/query/query-engine.md
```diff
@@ -122,6 +127,23 @@ so the method should be reasonably fast (not read any data itself, or at least n
...
+In this case the Query Engine tries to let the index handle as much constraints as possible and later executes all remaining constraints on its own, accessing the node store and doing all necessary operations there, which can result in a traversal. This means that despite the use of an index an additional traversal is required.
+
+If no matching index is determined in the previous step, the Query Engine executes this query solely based on a traversal.
+
+#### Ordering
+If a query requests an ordered result set, the Query Engine tries to get an already ordered result from the index; in case the index definition does not support the requested ordering or in case of a traversal, the Query Engine must execute the ordering itself. To achieve this the entire result set is read into memory and then sorted which consumes memory and takes time.
+
+#### Iterating the result set
```
thomasmueller: Iterating the Result Set (title case)

thomasmueller (commented):
Looks good!

oak-doc/src/site/markdown/query/query-engine.md
joerghoh: done

oak-doc/src/site/markdown/query/query-engine.md
```diff
@@ -122,6 +127,23 @@ so the method should be reasonably fast (not read any data itself, or at least n
 
 If an index implementation can not query the data, it has to return `Infinity` (`Double.POSITIVE_INFINITY`).
 
+#### Identifying Nodes
+
+If an index is selected, the query is executed against the index. The translation from the JCR Query syntax into the query language supported by the index includes as many constraints as possible which are supported by the index. Depending on the index definition this can mean that not all constraints can be resolved by the index itself. 
+In this case the Query Engine tries to let the index handle as much constraints as possible and later executes all remaining constraints on its own, accessing the node store and doing all necessary operations there, which can result in a traversal. This means that despite the use of an index an additional traversal is required.
```
klcodanr: Not 100% sure on the format of the anchor, but I think this ties the two identification and iteration together. 

```suggestion
In this case the Query Engine tries to let the index handle as much constraints as possible and later executes all remaining constraints on its own, accessing the node store and doing all necessary operations there, which can result in a traversal. This means that despite the use of an index an additional traversal is required.
```
```suggestion
In this case, the Query Engine tries to let the index handle as much constraints as possible and later executes all remaining constraints while [Iterating the Result Set](#iterating-the-result-set) by retrieving the nodes from the node store and evaluating the nodes against the constraints. Each retrieval, including non-matching nodes is counted as a traversal. This means that despite the use of an index an additional traversal can be required if not all constraints in a query are executed against the index.
```

oak-doc/src/site/markdown/query/query-engine.md
```diff
@@ -122,6 +127,23 @@ so the method should be reasonably fast (not read any data itself, or at least n
...
+
+If no matching index is determined in the previous step, the Query Engine executes this query solely based on a traversal.
```
klcodanr:
```suggestion
If no matching index is determined in the previous step, the Query Engine executes this query solely based on a traversing the Node Store and evaluating the nodes against the constraints.
```

oak-doc/src/site/markdown/query/query-engine.md
```diff
@@ -122,6 +127,23 @@ so the method should be reasonably fast (not read any data itself, or at least n
...
+
+#### Ordering
+If a query requests an ordered result set, the Query Engine tries to get an already ordered result from the index; in case the index definition does not support the requested ordering or in case of a traversal, the Query Engine must execute the ordering itself. To achieve this the entire result set is read into memory and then sorted which consumes memory and takes time.
```
klcodanr:
```suggestion
If a query requests an ordered result set, the Query Engine tries to get an already ordered result from the index; in case the index definition does not support the requested ordering or in case of a traversal, the Query Engine must execute the ordering itself. To achieve this the entire result set is read into memory and then sorted. This consumes memory, takes time and requires the Query Engine to read the full result set even in the case where a limit setting would otherwise limit the number of results traversed.
```

klcodanr (commented):
Some small tweaks but looks great and explains the process significantly more clearly!

oak-doc/src/site/markdown/query/query-engine.md
```diff
@@ -122,6 +127,23 @@ so the method should be reasonably fast (not read any data itself, or at least n
...
+
+#### Iterating the Result Set
+Query results are implemented as lazy iterators, and the result set is only read if needed. When the next result is requested, the result iterator seeks the potential nodes to find the next node matching the query. 
+During this seek process the Query Engine reads and filters the potential nodes until if finds a match. Even if the query is handled completely by an index, the Query Engine needs to check if the requesting session is allowed to read the nodes.
+
+That means that during this final step every potential node must be loaded from the node store, thus counting towards the read limit (see [Slow Queries and Read Limits](#slow-queries-and-read-limits)).
```
fabriziofortino: typo

```suggestion
Query results are implemented as lazy iterators, and the result set is only read if needed. When the next result is requested, the result iterator seeks the potential nodes to find the next node matching the query. 
During this seek process the Query Engine reads and filters the potential nodes until it finds a match. Even if the query is handled completely by an index, the Query Engine needs to check if the requesting session is allowed to read the nodes.

That means that during this final step every potential node must be loaded from the node store, thus counting towards the read limit (see [Slow Queries and Read Limits](#slow-queries-and-read-limits)).
```

fabriziofortino (approved):

PR #609, opened 2022-06-28
OAK-9819 : Improve Node.hasProperty/getProperty and Node.hasNode/getNode combinations

anchela:
@joerghoh , @reschke , @mreutegg , would you have time to take a look this improvement? thanks

### DIFF
--- a/oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/JackrabbitNode.java
+++ b/oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/JackrabbitNode.java
@@ -16,6 +16,11 @@
  */
 package org.apache.jackrabbit.api;
 
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import javax.jcr.Node;
+import javax.jcr.Property;
 import javax.jcr.RepositoryException;
 import javax.jcr.lock.LockException;
 import javax.jcr.nodetype.ConstraintViolationException;
@@ -26,7 +31,7 @@ import javax.jcr.version.VersionException;
  * The Jackrabbit Node interface. This interface contains the
  * Jackrabbit-specific extensions to the JCR {@link javax.jcr.Node} interface.
  */
-public interface JackrabbitNode {
+public interface JackrabbitNode extends Node {
 
     /**
      * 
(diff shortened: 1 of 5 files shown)


reschke (commented):
Note that equivalent methods already exist in JackrabbitSession. Maybe it would be better to just delegate there. In which case we might be able to do all of this in a default impl in the interface...

I'm also not convinced that getNodeOrNull can return null although the node exists (but is not a JackrabbitNode); maybe it would be better to change the return value to a "regular" node?

anchela:
hi @reschke , thanks for looking into the PR.

> Note that equivalent methods already exist in JackrabbitSession. Maybe it would be better to just delegate there. 

not quite. `JackrabbitSession.getNodeOrNull `takes an absolute path where as `Node.getNode` and `Node.getNodeOrNull `take a relative path. So, i don't think we can delegate it there.

Regarding:
> I'm also not convinced that getNodeOrNull can return null although the node exists (but is not a JackrabbitNode); 
> maybe it would be better to change the return value to a "regular" node?

if the node exists but is not accessible it would also return null.
in the first draft, i had the method return a `Node` but then I wondered if there is any reasonable scenario where a Node is a JackrabbitNode but one of it's child nodes isn't.....  and i concluded that this is very unlikely and thus return null instead of having code that might through classcasteexception was fine for the default implementation.
but i don't have a strong preference.... return just a `Node` would likely require API consumers to cast it to JackrabbitNode again or do some extra check.... but as I said... no strong preference.

reschke:
Ok...

But can we have a default impl in the interface (that just catches Exceptions and returns null)?

reschke (approved):
All ok; I misread the diffs and did not realize I was indeed looking at the default impl in the interface.

anchela:
@mreutegg , unless you have any concerns i would go ahead a merge the improvement.

mreutegg (commented):
`NodeImpl` is currently defined as `implements Node, JackrabbitNode`. I think this could be simplified into just `implements JackrabbitNode`.

anchela:
@mreutegg , good catch! i fixed it.

anchela:
@joerghoh , just wanted to see if i can invite you for reviews now.... :-) *works*. please approve or request changes.... or merge to give it a try....

PR #610, opened 2022-07-06
OAK-9834 : clarify node names of ACL nodes

kwin:


oak-doc/src/site/markdown/security/accesscontrol/default.md
```diff
@@ -129,8 +129,8 @@ found in section [Restriction Management](../authorization/restriction.html).
 <a name="representation"></a>
 ### Representation in the Repository
 
-All access control policies defined with an Oak repository are stores child of
-the node they are bound to. The node type definition used to represent access control content:
+All access control policies defined with an Oak repository are stored as child of
```
anchela: hi @kwin , thanks for spotting the mistake!
given that the statement is only true for the default authorization model, it might be more accurate and a bit less prone to misunderstandings if we would make this clear in this sentence.... e.g. something like

`Access control policies created and modified using the default authorization model are stored as child of`

because the statement that this applies for _all policies defined with an oak repository_ is quite obviously not accurate. btw. it is not even accurate for the default model which also defines 'named policies' for the configured readable paths..... the statement only applies for those policies that are applicabe/modifiable through JCR ac management API.

wdyt

anchela (changes requested):
hi @kwin thanks a lot for this.... the clarification makes sense.
re-reading the sentences i also feel that we should clarify that the repository representation does only reflect the details for the default authorization model (and not for every other model that may be used or plugged into the security setup).
i guess the documentation predates composite authorization so having that improved would definitely be valuable.

anchela:
@kwin , one more thing: can you please create a dedicated jira ticket for this improvement. there should IMHO be no changes made without jira reference.

kwin:
> here should IMHO be no changes made without jira reference.

There are quite some commits without JIRA references in https://github.com/apache/jackrabbit-oak/commits/trunk/oak-doc.
Also there should be almost no obstacles to apply clarifications and fix typos, I would strongly propose to be less formal with those, as IMHO Jackrabbit has a problem with getting contributions at all, rather than getting low quality contributions.

anchela:
hi @kwin  if it was just about fixing a typo, i might agree but it's not just that but changing phrasing that in the current context actually introduces a mistunderstanding because the start of the segment sounds as if that was true for all policies which is not correct.... and therefore i think asking for a dedicated ticket is not too much.

oak-doc/src/site/markdown/security/accesscontrol/default.md
kwin: Done

anchela (approved):
hi @kwin , thanks a lot for the additional adjustments. the patch looks good to me.

PR #612, opened 2022-07-06
OAK-9679 support testing with DynamoDBLocal on Apple Silicon

kwin:


oak-segment-aws/pom.xml
```diff
@@ -309,6 +308,14 @@
             <type>dll</type>
             <scope>test</scope>
         </dependency>
+        <!-- add native implementation for Apple silicon from fork -->
+        <dependency>
+            <groupId>io.github.ganadist.sqlite4java</groupId>
```
thomasmueller, line 314: according to https://github.com/ganadist/sqlite4java
"sqlite4java is licensed under Apache License 2.0."

oak-segment-aws/pom.xml
```diff
@@ -120,7 +120,6 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-dependency-plugin</artifactId>
-                <version>3.1.1</version>
```
thomasmueller: How is this change related?

oak-segment-aws/pom.xml
```diff
@@ -309,6 +308,14 @@
             <type>dll</type>
             <scope>test</scope>
         </dependency>
+        <!-- add native implementation for Apple silicon from fork -->
+        <dependency>
+            <groupId>io.github.ganadist.sqlite4java</groupId>
+            <artifactId>libsqlite4java-osx-aarch64</artifactId>
+            <version>${sqlite4java.version}</version>
+            <type>dylib</type>
+            <scope>test</scope>
```
thomasmueller, line 318: There is no test yet, just adding the dependency... what about we first add a test case before we add a dependency (in the same PR)?

oak-segment-aws/pom.xml
kwin: sqlite4java was always a dependency, the fork uses the same license, so this should be fine.

oak-segment-aws/pom.xml
kwin: not related, gonna remove

oak-segment-aws/pom.xml
kwin: This is not true, almost all existing tests fail on Apple M1 without this dependency, they succeed with that. Unfortunately neither GitHub Actions nor ASF Jenkins supports (testing with) that platform.

kwin: Also notice that this is only a test dependency!

mreutegg (approved):
Looks good to me. Tests still succeed on my machine with Intel CPU.

@thomasmueller does this work for you? I remember you are also on an Apple M1.

PR #587, opened 2022-06-03
OAK-9790 - Implement parallel indexing for speeding up oak run indexing command

Ewocker:
[OAK-9790 - Implement parallel indexing for speeding up oak run indexing command
](https://issues.apache.org/jira/browse/OAK-9790)

Since indexing was single threads, which is slow for large repository. In order to improve the indexing performance we need to implement parallel indexing.

The work is cover for both lucene and elastic indexing. In order to support parallel indexing, it need to split the big flat file store file ahead, which add a big overhead, but make parallel index possible and much faster.

Another change together is support the LZ4 compression since which is much faster compare to gzip.

PR #615, opened 2022-07-11
OAK-9790 Implement parallel indexing for speeding up oak run indexing command

thomasmueller:
(This is the PR https://github.com/apache/jackrabbit-oak/pull/587 but merged into a branch of Oak)

Since indexing was single threads, which is slow for large repository. In order to improve the indexing performance we need to implement parallel indexing.

The work is cover for both lucene and elastic indexing. In order to support parallel indexing, it need to split the big flat file store file ahead, which add a big overhead, but make parallel index possible and much faster.

Another change together is support the LZ4 compression since which is much faster compare to gzip.


oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
```diff
@@ -60,6 +60,7 @@ class ElasticBulkProcessorHandler {
     private static final Logger LOG = LoggerFactory.getLogger(ElasticBulkProcessorHandler.class);
     private final int FAILED_DOC_COUNT_FOR_STATUS_NODE = Integer.getInteger("oak.failedDocStatusLimit", 10000);
 
+    private static final int BULK_PROCESSOR_CONCURRENCY = 4;
```
thomasmueller, line 63: BULK_PROCESSOR_CONCURRENCY should be configurable and not set to 4 as default as this would impact async indexing as well -- see https://github.com/apache/jackrabbit-oak/pull/587#pullrequestreview-1022678685

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticConnection.java
```diff
@@ -118,6 +120,9 @@ private Clients getClients() {
                         Header[] headers = new Header[]{new BasicHeader("Authorization", "ApiKey " + apiKeyAuth)};
                         builder.setDefaultHeaders(headers);
                     }
+                    builder.setRequestConfigCallback(
+                            requestConfigBuilder -> requestConfigBuilder.setSocketTimeout(ES_CLIENT_SOCKET_TIMEOUT).setSocketTimeout(ES_CLIENT_SOCKET_TIMEOUT));
```
thomasmueller, line 124: Why do we need to change the socket timeout?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
nit0906, line 63: This PR should not be changing ElasticBulkProcessorHandler like this, since this would directly impact async indexing as well.

If we need to have a BP concurrency here, I suggest we make this configurable and set it via the oak-run OutOfIndexing flow (which this PR targets) and keep the default concurrency to 1 (i.e no concurrency).

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/progress/IndexingProgressReporter.java
```diff
@@ -229,10 +231,12 @@ public IndexUpdateState(String indexPath, boolean reindex, long estimatedCount)
...
 
         public void indexUpdate() throws CommitFailedException {
-            updateCount++;
-            if (updateCount % 10000 == 0) {
-                log.info("{} => Indexed {} nodes in {} ...", indexPath, updateCount, watch);
-                watch.reset().start();
+            long count = updateCount.incrementAndGet();
+            if (count % 10000 == 0) {
+                synchronized (this) {
+                    log.info("{} => Indexed {} nodes in {} ...", indexPath, updateCount, watch);
```
joerghoh:
```suggestion
                    log.info("{} => Indexed {} nodes in {} ...", indexPath, count, watch);
```

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/DefaultIndexWriter.java
```diff
@@ -19,12 +19,15 @@
 
 package org.apache.jackrabbit.oak.plugins.index.lucene.writer;
 
+import static com.google.common.base.Preconditions.checkNotNull;
```
joerghoh: It's safe to use Java 8's ```Objects.requireNonNull()``` here. We should avoid adding even more references to guava.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/DefaultIndexWriter.java
```diff
@@ -180,25 +187,26 @@ IndexWriter getWriter() throws IOException {
...
-        } finally {
-            closer.close();
+        synchronized (this) {
+            final Closer closer = Closer.create();
+
+            NodeBuilder suggesterStatus = definitionBuilder.child(suggestDirName);
+            DirectoryReader reader = closer.register(DirectoryReader.open(writer, false));
+            final Directory suggestDirectory = directoryFactory.newInstance(definition, definitionBuilder, suggestDirName, false);
+            // updateSuggester would close the directory (directly or via lookup)
+            // closer.register(suggestDirectory);
```
joerghoh, line 197: is this a comment (that means helpful to understand the code) or just commented code, which has been created during development but which is not used here? In the later case, please remove it.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileSplitter.java
```diff
@@ -0,0 +1,261 @@
...
+
+        // return original if file too small or split size equals 1
+        if (splitThreshold < MINIMUM_SPLIT_THRESHOLD || splitSize <= 1) {
+            LOG.info("split is not necessary, skip splitting");
+            return returnOriginalFlatFile();
+        }
+
+        LOG.info("original flat file size: ~{}",
+                FileUtils.byteCountToDisplaySize(fileSizeInBytes));
+        LOG.info("split threshold is ~{} bytes, estimate split size >={} files",
```
joerghoh: it would make sense to merge these 2 INFO messages into a single one.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileSplitter.java
```diff
@@ -0,0 +1,261 @@
...
+                    writer = createWriter(currentFile, algorithm);
+                    splitFlatFiles.add(currentFile);
+                    LOG.info("split position found at line {}, creating new split file {}", lineCount, currentFile.getAbsolutePath());
+                }
+                writer.append(line);
+                writer.newLine();
+                readPos += line.length() + 1;
+                lineCount++;
+            }
+            writer.close();
```
joerghoh: this should also live in a ```finally```  clause, otherwise it might left open.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileSplitter.java
```diff
@@ -0,0 +1,261 @@
...
+            LOG.info("created split flat file {} with size {}", currentFile.getAbsolutePath(), FileUtils.byteCountToDisplaySize(currentFile.length()));
+
+            LOG.info("split total line count: {}", lineCount);
```
joerghoh: please merge these 2 INFO messages into a single statement.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileSplitter.java
```diff
@@ -0,0 +1,261 @@
...
+            return property.getValue(Type.NAME);
+        }
+        return "";
+    }
+
+    private boolean canSplit(Set<String> nodeTypes, Stack<String> nodeTypeNameStack) {
+        if (nodeTypeNameStack.contains("")) {
+            return false;
+        }
+        for (String parentNodeTypeName : nodeTypeNameStack.subList(0, nodeTypeNameStack.size()-1)) {
```
joerghoh: this loop is a bit inconsistent to the previous check (```nodeTypeNameStack.contains("")```), because it does not iterate over the last element in the stack, but the ```contains()``` does); is this intendend and valid?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreUtils.java
```diff
@@ -30,46 +32,52 @@
...
-
-import static com.google.common.base.Charsets.UTF_8;
 
+/**
+ * This class provides common utility functions for building FlatFileStore.
+ */
 class FlatFileStoreUtils {
 
+    /**
+     * This function by default uses GNU zip as compression algorithm for backward compatibility.
```
joerghoh:
```suggestion
     * This function by default uses gzip as compression algorithm for backward compatibility.
```

The term ```gzip``` is  more widely used than ```GNU zip```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreUtils.java
```diff
@@ -30,46 +32,52 @@
...
-            }
-            return br;
+            return new BufferedReader(new InputStreamReader(algorithm.getInputStream(in)));
         } catch (IOException e) {
             throw new RuntimeException("Error opening file " + file, e);
         }
     }
 
+    /**
+     * This function by default uses GNU zip as compression algorithm for backward compatibility.
```
joerghoh:
```suggestion
     * This function by default uses gzip as compression algorithm for backward compatibility.
```

joerghoh (changes requested):

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/DefaultIndexWriter.java
thomasmueller: Jun didn't change the old code here (from Chetan / Vikas / Tommaso probably), he just synchronized the method. So I think we should keep it as is. Unless if you can find out if the comment is better kept or removed... I can't, looking at the code.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileSplitter.java
thomasmueller: The code itself is correct (at least this line is correct), but it is not properly documented.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileSplitter.java
```diff
@@ -0,0 +1,308 @@
...
+     * @param parentNodeTypeNames the stack of node types processed so far
+     * @param line the path of the current node
+     */
+    private void updateNodeTypeStack(Stack<String> parentNodeTypeNames, String line) {
+        NodeStateHolder ns = new SimpleNodeStateHolder(line);
+        List<String> pathElements = ns.getPathElements();
+        int currentLineDepth = pathElements.size();
+        int parentTypesDepth = parentNodeTypeNames.size();
+        if (currentLineDepth > parentTypesDepth) {
+            // TODO we also need to consider mixins
```
thomasmueller, line 196: (I added this TODO on purpose: we can't merge the PR currently, because this still needs to be implemented.)

The code only deals with primary types (e.g. acme:Asset), but doesn't deal with mixing types.
We need to work on this, before we can merge the patch.
Using a stack alone isn't sufficient: we need a stack of sets :-/

thomasmueller (commented):
The PR is NOT ready to be merged. But the review can continue.

amit-jain:
Superseded by PR https://github.com/apache/jackrabbit-oak/pull/715

PR #617, opened 2022-07-11
OAK-9831: Update spotbugs plugin to 4.7.1.0

mreutegg:


mreutegg:
Test failures in oak-segment-azure are most likely not caused by spotbugs plugin update, but rather related to docker setup.

PR #619, opened 2022-07-11
OAK-9833: UpgradeIT fails on Java 17

mreutegg:
Disable test when running on Java version higher than 11. The Groovy version included in oak-run 1.6 does not work well on Java 17.


mreutegg:
Test failures in oak-search-elastic are unrelated to changes in this PR.

PR #622, opened 2022-07-14
OAK-9780: Prefetch node states

mreutegg:
This is a new pull request based on the work started by @stefan-egli. The code has been cleaned up and in my view is in a state where it can be included in Oak.


mreutegg:
@thomasmueller, I just noticed you did some refactoring and moved cursor implementations into a dedicated package. I think the package also needs to be excluded from the baseline check in oak-parent/pom.xml. I'm going to add it. Let me know what you think.

thomasmueller (approved):

PR #623, opened 2022-07-15
OAK-9840: wait for all required MountInfoConfigs

kwin:
also consider later configs by making the reference greedy

### DIFF
--- a/oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/MountInfoProviderService.java
+++ b/oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/MountInfoProviderService.java
@@ -19,9 +19,10 @@
 
 package org.apache.jackrabbit.oak.composite;
 
-import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


jelmini:
Thanks for catching this.

jelmini (approved):

oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/MountInfoProviderService.java
```diff
@@ -125,7 +126,7 @@ private void addMountInfoConfigFromProperties(Props mountInfoProps) {
         }
     }
 
-    @Reference(cardinality = ReferenceCardinality.MULTIPLE)
+    @Reference(cardinality = ReferenceCardinality.MULTIPLE, policyOption = ReferencePolicyOption.GREEDY)
```
kwin: Probably this was thought to be dynamic (not static) as it is threadsafe (CopyOnWriteArrayList) and leads to less reactivation of this service. Also dynamic makes the greedy part obsolete: https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#d0e30480

PR #629, opened 2022-07-18
OAK-9846: Emit origin for each checker report

kwin:


oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/checks/ErrorHolder.java
```diff
@@ -27,16 +27,16 @@ class ErrorHolder {
     private static final int FAIL_IMMEDIATELY_THRESHOLD = 100;
     private final List<String> errors = new ArrayList<>();
     
-    public void report(MountedNodeStore mountedStore, String path, String error) {
-        errors.add(String.format("For NodeStore mount %s, path %s, encountered the following problem: '%s'", mountedStore.getMount().getName(), path, error));
+    public void report(MountedNodeStore mountedStore, String path, String error, MountedNodeStoreChecker<?> checker) {
```
kwin, line 34: @rombert This package is not versioned but still exported. Do we need to keep backwards compatibility? I don't think it should actually be exported at all!

kwin:
@rombert Any feedback, otherwise I will merge tomorrow?

oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/checks/ErrorHolder.java
rombert: I think 'not versioned but exported' is the general Oak policy. I don't agree, but this is what we settled on.

For this particular change I have no backwards compatibility concerns since I did not see any exported APIs changes.

rombert (approved):
LGTM, thanks.

oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/checks/ErrorHolder.java
kwin: > I did not see any exported APIs changes.

Well, ErrorHolder is exported, and the signature of one of its public methods changed with this PR!

kwin: Sorry for the noise, ErrorHolder is package -private

PR #635, opened 2022-07-21
OAK-9851 : added support for throttling in mongo oak client based on …

rishabhdaim:
…oplog replication window


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -100,6 +100,12 @@
     static final int MANY_CHILDREN_THRESHOLD = Integer.getInteger(
             "oak.documentMK.manyChildren", 50);
 
+    public static final int DEFAULT_THROTTLING_THRESHOLD = Integer.getInteger(
```
mreutegg: Can you please add some comments what this threshold is about?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -100,6 +100,12 @@
...
+            "oak.documentMK.throttlingThreshold", 2);
+
+    public static final long DEFAULT_THROTTLING_TIME = Long.getLong(
```
mreutegg: Please some comment what this is and add time unit to name. E.g. throttlingTimeMs for milliseconds.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
```diff
@@ -267,4 +267,13 @@
...
             @Option(label = "LENIENT", value = "LENIENT")})
     String leaseCheckMode() default "STRICT";
+
+    @AttributeDefinition(
+            name = "Document Node Store throttling",
+            description = "Boolean value indicating whether throttling should be " +
+                    "enabled for document node store or not. The Default value is " +
+                    DocumentNodeStoreService.DEFAULT_THROTTLE_DOCUMENT_STORE +
+                    "Note that this value can be overridden via framework " +
+                    "property 'oak.documentstore.throttleDocumentStore'")
```
mreutegg: I'm not too happy with the name, because it suggests the DocumentStore is always throttled and _documentstore_ appears twice in the property name.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
```diff
@@ -137,6 +137,7 @@ public class DocumentNodeStoreService {
     static final int DEFAULT_BLOB_CACHE_SIZE = 16;
     static final String DEFAULT_DB = "oak";
     static final boolean DEFAULT_SO_KEEP_ALIVE = true;
+    static final boolean DEFAULT_THROTTLE_DOCUMENT_STORE = true;
```
mreutegg: Since this is a new feature, I would rather disable it by default.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/SynchronizingDocumentStoreWrapper.java
```diff
@@ -157,6 +158,16 @@ public synchronized int getNodeNameLimit() {
         return store.getNodeNameLimit();
     }
 
+    /**
+     * Return the @{@link ThrottlingMetrics} for the underlying document store
+     *
+     * @return throttling metric for document store
+     */
+    @Override
+    public ThrottlingMetrics throttlingMetrics() {
```
mreutegg: Method should be synchronized.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStoreThrottling.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package org.apache.jackrabbit.oak.plugins.document;
+
+/**
+ * Marker interface to check whether underlying document store needs to throttle the incoming requests or not
+ *
+ * Concrete implementation needs to provide the logic for implementing the throttling based on their internal usages/statistics
+ */
+public interface DocumentStoreThrottling {
```
mreutegg: This interface is not used as a marker. The only implementation is MongoDocumentStoreThrottling. I don't see the value of this interface and would rather remove it.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStoreThrottling.java
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * Check whether the underlying document store need to throttle the system or not
+     */
+    void updateMetrics();
```
mreutegg: The name of this method and JavaDoc are misleading. The actual implementation schedules a task that runs periodically, whereas the method suggests a one-time update of metrics.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/ThrottlingMetrics.java
```diff
@@ -0,0 +1,51 @@
...
+ * Concrete implementations of this interface are required to provide implementations
+ * of this class expose their throttling metrics
+ */
+public interface ThrottlingMetrics {
+
+    int threshold();
+
+    double currValue();
+
+    long throttlingTime();
```
mreutegg: Can you please add JavaDoc to these three methods and explain what those values are?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottling.java
```diff
@@ -0,0 +1,99 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Objects;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+public class MongoDocumentStoreThrottling implements DocumentStoreThrottling {
```
mreutegg: I think this class should have a different name. AFAICS it doesn't do any throttling. Its main purpose is to calculate the oplog window.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottling.java
```diff
@@ -0,0 +1,99 @@
...
+                        oplogWindow.set(updateOplogWindow(maxSizeGb, usedSizeGb, first, last));
+                    }
+                }
+            }
+        }, 10, 30, SECONDS);
+    }
+
+    // helper methods
+    @VisibleForTesting
+    double updateOplogWindow(final double maxSize, final double usedSize, final @NotNull Document first,
```
mreutegg: I think it's perfectly fine to have this package private. The method can even be static. It does not access any state of MongoDocumentStoreThrottling.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottling.java
```diff
@@ -0,0 +1,99 @@
...
+                                   final @NotNull Document last) {
+        final BsonTimestamp startTime = first.get(TS_TIME, BsonTimestamp.class);
+        final BsonTimestamp lastTime = last.get(TS_TIME, BsonTimestamp.class);
+        long timeDiffSec = Math.abs(lastTime.getTime() - startTime.getTime());
+        double timeDiffHr = Math.ceil(((double)timeDiffSec/(60*60)) * 100000)/100000;
+        double currentOplogHourRate = usedSize /timeDiffHr;
```
mreutegg: I think this may result in a division by zero when when the oplog just has a single entry. Add a test and fix?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottling.java
```diff
@@ -0,0 +1,99 @@
...
+        double timeLeft = maxSize /currentOplogHourRate;
```
mreutegg: Similarly when `usedSize` is zero, then `currentOplogHourRate` is zero as well. It's probably an edge case never called that way, because it implies the oplog is empty and then `updateMetrics()` does not call `updateOplogWindow()`. I think it would still be good to cover that case in `updateOplogWindow()`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottling.java
```diff
@@ -0,0 +1,99 @@
...
+    private static final String MAX_SIZE = "maxSize";
+    private static final String OPLOG_RS = "oplog.rs";
+    public static final String SIZE = "size";
+    private final ScheduledExecutorService throttlingExecutor;
+    private final AtomicDouble oplogWindow;
+    private final MongoDatabase localDb;
+
+    public MongoDocumentStoreThrottling(final @NotNull MongoDatabase localDb, final @NotNull MongoThrottlingMetrics mongoThrottlingMetrics) {
+        this.throttlingExecutor = Executors.newSingleThreadScheduledExecutor();
+        this.oplogWindow = mongoThrottlingMetrics.oplogWindow;
```
mreutegg: Wouldn't it be nicer if the constructor just gets the `oplogWindow` `AtomicDouble`? It doesn't need anything else from `MongoThrottlingMetrics`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -295,15 +322,38 @@ public MongoDocumentStore(MongoClient connection, MongoDatabase db,
         this.nodeLocks = new StripedNodeDocumentLocks();
         this.nodesCache = builder.buildNodeDocumentCache(this, nodeLocks);
 
+        // if throttling is enabled
+        if (throttleDocumentStore) {
+            this.localDb = connection.getDatabase("local");
```
mreutegg: I think `localDb` can be turned into a local variable.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -295,15 +322,38 @@ public MongoDocumentStore(MongoClient connection, MongoDatabase db,
...
+            final MongoIterable<String> collectionNames = localDb.listCollectionNames();
+            String ol = null;
+            for (String e: collectionNames) {
+                if (Objects.equals(e, OPLOG_RS)) {
+                    ol = OPLOG_RS;
+                    break;
+                }
+            }
```
mreutegg: Could use Guava instead.
```suggestion
            String ol = Iterables.tryFind(localDb.listCollectionNames(), 
                    s -> Objects.equals(s, OPLOG_RS)).orNull();
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoThrottlingMetrics.java
```diff
@@ -0,0 +1,58 @@
...
+package org.apache.jackrabbit.oak.plugins.document.mongo;
+
+import com.google.common.util.concurrent.AtomicDouble;
+import org.apache.jackrabbit.oak.plugins.document.ThrottlingMetrics;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Java Object to represent Mongo throttling metrics
+ */
+public class MongoThrottlingMetrics implements ThrottlingMetrics {
```
mreutegg: There is nothing MongoDB specific in this class. Maybe this is just a default implementation that can be moved to the parent package?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottling.java
```diff
@@ -0,0 +1,99 @@
...
+
+public class MongoDocumentStoreThrottling implements DocumentStoreThrottling {
+
+    private static final Logger LOG = LoggerFactory.getLogger(MongoDocumentStoreThrottling.class);
+    static final String TS_TIME = "ts";
+    private static final String NATURAL = "$natural";
+    private static final String MAX_SIZE = "maxSize";
+    private static final String OPLOG_RS = "oplog.rs";
+    public static final String SIZE = "size";
+    private final ScheduledExecutorService throttlingExecutor;
```
mreutegg: The executor service should be shutdown when not used anymore. I think this means `MongoDocumentStoreThrottling` should implement `Closeable` and shutdown the executor in `close()`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ThrottlingDocumentStoreWrapper.java
```diff
@@ -0,0 +1,241 @@
...
+import java.util.Map;
+
+import static java.lang.Thread.sleep;
+
+/**
+ * Wrapper of another DocumentStore that does a throttling check on any method
+ * invocation (create or update) and throttled the system if under high load.
+ * <p>
+ *     TODO update issue
+ * @see "https://issues.apache.org/jira/browse/OAK-2739 for more details"
```
mreutegg: This looks unrelated. Remove?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ThrottlingDocumentStoreWrapper.java
```diff
@@ -0,0 +1,241 @@
...
+        final double currValue = metrics.currValue();
+
+        if (DoubleMath.fuzzyCompare(currValue,threshold/8,  0.001) <= 0) {
+            throttleTime = throttleTime * 8;
+        } else if (DoubleMath.fuzzyCompare(currValue,threshold/4, 0.001) <= 0) {
+            throttleTime = throttleTime * 4;
+        } else if (DoubleMath.fuzzyCompare(currValue, threshold/2, 0.001) <= 0) {
+            throttleTime = throttleTime * 2;
+        } else if (DoubleMath.fuzzyCompare(currValue, threshold,0.001) <= 0) {
+            throttleTime = metrics.throttlingTime();
```
mreutegg: I don't understand this logic. Why is throttleTime only multiplied around certain thresholds and not when the value is between thresholds?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/TimingDocumentStoreWrapper.java
```diff
@@ -443,6 +444,23 @@ public int getNodeNameLimit() {
...
+     * Return the @{@link ThrottlingMetrics} for the underlying document store
+     *
+     * @return throttling metric for document store
+     */
+    @Override
+    public ThrottlingMetrics throttlingMetrics() {
+        try {
+            long start = now();
+            final ThrottlingMetrics result = base.throttlingMetrics();
+            updateAndLogTimes("getNodeNameLimit", start, 0, 0);
```
mreutegg: Wrong operation name.
```suggestion
            updateAndLogTimes("throttlingMetrics", start, 0, 0);
```

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoThrottlingMetricsTest.java
```diff
@@ -0,0 +1,42 @@
...
+public class MongoThrottlingMetricsTest {
+
+    @Test
+    public void testCurrValue() {
+        MongoThrottlingMetrics mongoThrottlingMetrics = of(new AtomicDouble(MAX_VALUE), 10, 100);
+        assertEquals(MAX_VALUE, mongoThrottlingMetrics.currValue(), 0.001);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testNonNullOplog() {
```
mreutegg: This test currently fails. See also Jenkins failure in this PR.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStoreThrottling.java
rishabhdaim: @mreutegg I have created this interface in case we need to add throttling for the RDB store in the future.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoThrottlingMetrics.java
rishabhdaim: @mreutegg this class derives its value from oplogWindow variable specific to mongo.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ThrottlingDocumentStoreWrapper.java
rishabhdaim: @mreutegg I am updating the throttling time in multiple of 2 after in conjunction with every drop (in multiples of 2) in the current value is compared to the threshold.

rishabhdaim:
@stefan-egli could you please review the PR and share your feedback.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottling.java
rishabhdaim: @mreutegg @stefan-egli wouldn't it be better if we use a `MoreExecutors.getExitingScheduledExecutorService` executor from guava that attaches a shutdown hook and shuts down itself on the JVM shutdown call?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStoreThrottling.java
stefan-egli: I agree with what @mreutegg  wrote : it's not a marker. Plus it's not used anywhere except the implementation.

Maybe removing this is more straight forward and doesn't prevent it from being re-added in the future if there's a use case then.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/ThrottlingMetrics.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+package org.apache.jackrabbit.oak.plugins.document;
+
+/**
+ * Interface to expose throttling metrics.
+ *
+ * Concrete implementations of this interface are required to provide implementations
+ * of this class expose their throttling metrics
+ */
+public interface ThrottlingMetrics {
```
stefan-egli: I was wondering if this class could be split along the different concerns it seems to currently combine: config (threshold/throttlingTime) and metric (currValue).

What came to mind as an alternative would be something like
```
public interface Throttler {
  long throttlingTime();
}
```
that could replace the `ThrottlingMetrics`.

then you could move the actual algorithm that calculates the throttling time into eg a `ExponentialThrottler` that works based on an AtomicLong (for value updates) and thresholds (these 2 would move here from the current `ThrottlingMetrics`) - and similarly we could come up with different such algorithms for Throttler implementations (which could mean perhaps an interface as its input...)

and `MongoThrottlingMetrics` might either not be needed anymore or become the factory that creates an `ExponentialThrottler` with the correct inputs, taken from oplog metric etc.

This way essentially the `metrics` part is hidden from the throttling logic and become a factory/implementation detail.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/ThrottlingMetrics.java
```diff
@@ -0,0 +1,51 @@
...
+        @Override
+        public double currValue() {
+            return 0;
+        }
+
+        @Override
+        public long throttlingTime() {
+            return 0;
+        }
+    };
```
stefan-egli: Based on the `Throttler` interface sketched above, this could become sth like
```suggestion
    Throttling NO_THROTTLING = new Throttling() {

        @Override
        public long throttlingTime() {
            return 0;
        }
    };
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottling.java
stefan-egli: or we could consider using [sling scheduler via whiteboardutils](https://github.com/apache/jackrabbit-oak/blob/c6ddcc55bee3de915459af01e91edad32d538f3d/oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/whiteboard/WhiteboardUtils.java#L54) - to be discussed I guess...

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ThrottlingDocumentStoreWrapper.java
```diff
@@ -0,0 +1,216 @@
...
+     */
+    @Override
+    public Throttler throttler() {
+        return store.throttler();
+    }
+
+    // helper methods
+
+    private void performThrottling() {
+
```
stefan-egli: I think we should not throttle the lease updates - even though currently the throttling times are lowish and probably don't interfere with the 2+ min lease durations. If the lease cannot be maintained, the throttling is equivalent to a shutdown. And that's probably not the intention. So, even though not throttling lease updates increases the use of oplog window while that is low, I'd argue lease updates are a high priority kind of data and could be excempted.

One way to do that - with least line changes - could for example be to change `performThrottling` like so:
```
    private void performThrottling(final Collection<T> collection) {
        if (Collection.CLUSTER_NODES == collection) {
            // no throttling of CLUSTER_NODES collection
            return;
        }
   ...
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottlingMetricsUpdater.java
```diff
@@ -0,0 +1,112 @@
...
+
+    public MongoDocumentStoreThrottlingMetricsUpdater(final @NotNull MongoDatabase localDb, final @NotNull AtomicDouble oplogWindow) {
+        // exiting scheduled executor, will exit when we call to shut down jvm
+        this.throttlingMetricsExecutor = MoreExecutors.getExitingScheduledExecutorService(
+                (ScheduledThreadPoolExecutor) Executors.newScheduledThreadPool(1), 40, SECONDS);
+        this.oplogWindow = oplogWindow;
+        this.localDb = localDb;
+    }
+
+    public void updateMetrics() {
```
stefan-egli: Was wondering about the 3 warn cases : while it's unclear why those problems exist, it might be worth setting the `oplogWindow` to max_integer in that case - as basically we don't know how long such error situations last and what the impact is. So the fallback would be to say : "unknown oplog window", thus max_integer
(currently it leaves the `oplogWindow` unchanged in warn cases, which might mean throttling even though perhaps in the meantime the oplog window grew - of course the opposite is also true, but given this is an exceptional situation I'd steer to the infinite side)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -100,6 +100,20 @@
...
+    public static final int DEFAULT_THROTTLING_THRESHOLD = Integer.getInteger(
+            "oak.documentMK.throttlingThreshold", 2);
+
+    /**
+     * The default throttling time (in millis) when throttling is enabled. This is the time for
+     * which we block any data modification operation when system has been throttled.
+     */
+    public static final long DEFAULT_THROTTLING_TIME_MS = Long.getLong(
+            "oak.documentMK.throttlingTime", 20);
+
```
stefan-egli: I think these could be moved to MongoDocumentStore (and then consequently renamed to `"oak.mongo..."`) as they are now specific to mongo only

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ThrottlingDocumentStoreWrapper.java
```diff
@@ -0,0 +1,216 @@
...
+        final Throttler throttler = throttler();
+        long throttleTime = throttler.throttlingTime();
+
+        if (throttleTime == 0) {
+            return; // no throttling
+        }
+
+        try {
+            LOG.info("Throttling the system for {} ms", throttleTime);
```
stefan-egli: I agree that we should log at INFO if throttling is happening. I'm wondering though how noisy this one could become. Could we reduce the frequency of how much we log at INFO, for example only every 10 sec (without causing too much synchronization between the different threads calling `performThrottle`..)?

stefan-egli (commented):
Some more late comments that unfortunately slipped my first review.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ThrottlingDocumentStoreWrapper.java
stefan-egli: on 2nd thought, perhaps this even qualifies for a WARN?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ThrottlingDocumentStoreWrapper.java
rishabhdaim: I think it makes sense to exempt lease updates and other high-priority updates (if any).

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ThrottlingDocumentStoreWrapper.java
```diff
@@ -0,0 +1,227 @@
...
+        if (throttleTime == 0) {
+            return; // no throttling
+        }
+
+        try {
+            // log message every 10 secs once to reduce noise
+            if (now().getSecond() - lastLogTime >= 10) {
+                LOG.warn("Throttling the system for {} ms for {} collection", throttleTime, collection);
+                lastLogTime = now().getSecond();
+            }
```
stefan-egli: now().getSecond() seems to only return `0..59` though, so might not work as expected?
What about using decaseconds? 

PS: the initialization could then be `lastLogTime = 0`

```suggestion
            // (decaseconds since 1970 - overflows roughly in year 2650)
            final int currentDecaseconds = (int) (System.currentTimeMillis() / 10000);
            if (currentDecaseconds - lastLogTime >= 1) {
                lastLogTime = currentDecaseconds;
                LOG.warn("Throttling the system for {} ms for {} collection", throttleTime, collection);
            }
```

stefan-egli (commented):
just one more comment - the other aspects now look good for me, thx for addressing them!

stefan-egli (approved):
+1, lgtm now

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottling.java
joerghoh: @stefan-egli I wouldn't be super happy about that, first beause of the additional dependency and secondly of the additional configuration. I don't see any value in adding that and favor the approach @mreutegg  suggested.

stefan-egli (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -132,11 +139,29 @@ public class MongoDocumentStore implements DocumentStore {
 
     private static final Bson BY_ID_ASC = new BasicDBObject(Document.ID, 1);
 
+    private static final String OPLOG_RS = "oplog.rs";
+
+    /**
+     * The threshold value after which the document store should start (if enabled) throttling.
+     */
+    // For mongo based document store this value is threshold for the oplog replication window.
+    public static final int DEFAULT_THROTTLING_THRESHOLD = Integer.getInteger("oak.mongo.throttlingThreshold", 2);
```
mreutegg, line 151: Can you please merge the two comments into a single block comment? I assume the unit for this value is hours, but I'm not sure. Can you please clarify with a sentence in the comment?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottling.java
mreutegg: I think this is still a problem. The code now checks if `startTime` and `lastTime` are equal and returns early in that case. However, `timeDiffSec` is calculated from `lastTime.getTime()` and `startTime.getTime()`. These values can be the same even when the two `BsonTimestamp` are not equal. Method `getTime()` returns just the seconds since the epoch part of the timestamp and ignores the increment value.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottling.java
mreutegg: I'm still not happy with the MoreExecutors.getExitingScheduledExecutorService() solution. It means scheduled executors will accumulate e.g. when running tests. I would rather rename `MongoDocumentStoreThrottlingMetricsUpdater.updateMetrics()` to `MongoDocumentStoreThrottlingMetricsUpdater.scheduleMetricsUpdate()` and return a `Closable`. The `MongoDocumentStore` could then close it in `dispose()`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ThrottlingDocumentStoreWrapper.java
mreutegg: OK, I see. Somehow I was confused by the threshold with the comparison. What is the reason for not using `Double.compare()` instead?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottling.java
mreutegg: Actually, I'm wrong. This won't throw an exception when `timeDiffSec` and `timeDiffHr` are zero, because `timeDiffHr` is of type `double`. In that case `currentOplogHourRate` will be `Infinity` and `timeLeft` zero.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceConfigurationTest.java
```diff
@@ -92,6 +94,14 @@ public void presetMongoURI() throws Exception {
         assertEquals(uri, config.mongouri());
     }
 
+    @Test
+    public void throttleEnabled() throws Exception {
+        boolean throttleDocStore = false;
```
mreutegg: This is the default. Shouldn't we test the opposite?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottlingMetricsUpdaterTest.java
```diff
@@ -0,0 +1,82 @@
...
+import static org.apache.jackrabbit.oak.plugins.document.mongo.MongoDocumentStoreThrottlingMetricsUpdater.updateOplogWindow;
+import static org.junit.Assert.assertEquals;
+
+public class MongoDocumentStoreThrottlingMetricsUpdaterTest {
+
+    @Test
+    public void testUpdateOplogWindow() {
+
+        double oplogWindow = updateOplogWindow(1024, 512,
+                new Document(TS_TIME, new BsonTimestamp((int) currentTimeMillis(), 0)),
```
mreutegg: All these `BsonTimestamp` created with `(int) currentTimeMillis()` look suspicious to me. The first parameter is supposed to be seconds since the epoch, while currentTimeMillis() gives you milliseconds since the epoch.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottlingMetricsUpdaterTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+        double oplogWindow = updateOplogWindow(1024, 0,
+                new Document(TS_TIME, new BsonTimestamp((int) currentTimeMillis(), 0)),
+                new Document(TS_TIME, new BsonTimestamp((int) currentTimeMillis() + 3600, 0)));
+
+        assertEquals(Integer.MAX_VALUE, oplogWindow, 0.001);
+    }
+
+    @Test
+    public void testUpdateOplogWindow_OneOplogEntry() {
```
mreutegg, line 77: Please add a test similar to this but with BsonTimestamps that differ only in their increment values.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ThrottlingDocumentStoreWrapper.java
rishabhdaim: `Double.compare()` doesn't take tolerance into account while comparing 2 doubles. So I decided to use `DoubleMath.fuzzyCompare()`.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceConfigurationTest.java
rishabhdaim: Yes, it makes sense to test the opposite of default, will update the test case.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottlingMetricsUpdaterTest.java
rishabhdaim: Sure, will update the code to use `Instant.now().getEpochSeconds`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ThrottlingDocumentStoreWrapper.java
mreutegg: > Double.compare() doesn't take tolerance into account while comparing 2 doubles

But why is this needed here?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottlingMetricsUpdater.java
```diff
@@ -0,0 +1,138 @@
...
+                // Wait a while for tasks to respond to being cancelled
+                if (!throttlingMetricsExecutor.awaitTermination(10, SECONDS)){
+                    LOG.warn("throttlingMetricsExecutor did not terminate");
+                }
+            }
+        } catch (InterruptedException ex) {
+            throttlingMetricsExecutor.shutdownNow();
+            // Preserve interrupt status
+            Thread.currentThread().interrupt();
+        }
```
mreutegg: Can you please use `ExecutorCloser` from `oak-commons` module instead?

mreutegg (changes requested):

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottlingMetricsUpdaterTest.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Test
+    public void testUpdateOplogWindow_OplogEntries_InSameSecond_WithZeroUsedSize() {
+
+        double oplogWindow = updateOplogWindow(1024, 0,
+                new Document(TS_TIME, new BsonTimestamp((int) EPOCH_SECOND, 0)),
+                new Document(TS_TIME, new BsonTimestamp((int) EPOCH_SECOND, 1)));
+
+        // expected value should be very close to zero, since we are filling oplog window in same second
```
mreutegg: This comment does not match the assert.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ThrottlingDocumentStoreWrapper.java
rishabhdaim: Using `DoubleMath.fuzzyCompare()` helps in removing rounding off errors. Rounding errors can also result in flaky tests.

I agree that we can go with `Double.compare()` but, then we would only be throttling after reaching the threshold, any value which is very close to the threshold value would be ignored, so I wanted to handle those cases as well.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottlingMetricsUpdater.java
rishabhdaim: thanks, will update the code to use `ExecutorCloser`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ThrottlingDocumentStoreWrapper.java
mreutegg: OK. I still think `Double.compare()` is sufficient, but I don't mind using `DoubleMath.fuzzyCompare()`.

mreutegg (approved):
Changes look good to me now.

PR #637, opened 2022-07-22
OAK-9854: Remove version information from Oak pages

mreutegg:


mreutegg:
@kwin looks like others aren't interested in this topic. I'm aware you would rather keep the snapshot version on the pages. How strongly do you oppose this change? I'm fine if you really want to keep the information, but if you can live without it, then I'd like to remove the snapshot version.

kwin:
I am also fine with removing it, maybe clarifies that this is a living document and may change independent of releases.

kwin (approved):

PR #638, opened 2022-07-22
OAK-9855 - Enable for Lucene the tests common to Lucene and Elastic that are failing only with Elastic

nfsantos:
In the common tests that are passing on Lucene but failing on ES, override the test on the Elastic test class and mark it as ignored there, so that the Lucene test is executed.

Additionally, addressed some compilation warnings.

### DIFF
--- a/oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDescendantSpellcheckCommonTest.java
+++ b/oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDescendantSpellcheckCommonTest.java
@@ -40,21 +40,19 @@ public class ElasticIndexDescendantSpellcheckCommonTest extends IndexDescendantS
         return jcr.createRepository();
     }
 
-    //TODO ES Failing: this seems to be similar to
+    //TODO this seems to be similar to
     //https://discuss.elastic.co/t/elasticsearch-suggestion-completes-return-incorrect-number-of-suggestions/214522
     //Even increasing the suggester query size the number of max results seems to be fixed to 5
     @Test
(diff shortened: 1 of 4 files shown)


oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDescendantSpellcheckCommonTest.java
```diff
@@ -38,4 +40,23 @@ protected Repository createJcrRepository() {
         return jcr.createRepository();
     }
 
+    //TODO ES Failing: this seems to be similar to
+    //https://discuss.elastic.co/t/elasticsearch-suggestion-completes-return-incorrect-number-of-suggestions/214522
+    //Even increasing the suggester query size the number of max results seems to be fixed to 5
+    @Test
+    @Ignore("Failing on ES")
```
thomasmueller: We often use the Jira issue id in the ignore annotation. So what about opening a new Oak Jira issue to fix this particular problem, and then use the id here? E.g. `@Ignore("OAK-123456")`

thomasmueller (changes requested):

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDescendantSpellcheckCommonTest.java
nfsantos: I have created issues for all the failing tests and referenced them on the `@Ignore` tag.

thomasmueller (approved):

fabriziofortino (approved):

PR #639, opened 2022-07-26
OAK-9856 | Log query and caller stack trace in case of ES search fail…

nit0906:
…ures

### DIFF
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
@@ -341,8 +341,8 @@ public class ElasticResultRowAsyncIterator implements Iterator<FulltextResultRow
                 LOG.warn("Error reference for async iterator was previously set to {}. It has now been reset to new error {}", error.getMessage(), t.getMessage());
             }
 
-            LOG.error("Error retrieving data from Elastic for jcr query [{}] :: Corresponding ES query {} : closing scanner, notifying listeners"
-                    , indexPlan.getFilter(), ElasticIndexUtils.toString(query), t);
+            LOG.error("Error retrieving data for jcr query [{}] :: Corresponding ES query {} : closing scanner, notifying listeners",
+                indexPlan.getFilter(), ElasticIndexUtils.toString(query), t);
             // closing scanner immediately after a failure avoiding them to hang (potentially) forever
             close();
         }
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
@@ -120,7 +120,7 @@ public class ElasticResultRowAsyncIterator implements Iterator<FulltextResultRow
         Throwable error = errorRef.getAndSet(null);
         if (error != null) {
             error.fillInStackTrace();
-            LOG.error("Error while fetching Results from Elastic for [{}]", indexPlan.getFilter().toString(), error);
+            LOG.error("Error while fetching results from Elastic for [{}]", indexPlan.getFilter(), error);
         }
         return !POISON_PILL.path.equals(nextRow.path);
     }
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
@@ -341,7 +341,8 @@ public class ElasticResultRowAsyncIterator implements Iterator<FulltextResultRow
                 LOG.warn("Error reference for async iterator was previously set to {}. It has now been reset to new error {}", error.getMessage(), t.getMessage());
             }
 
-            LOG.error("Error retrieving data from Elastic for query [{}] : closing scanner, notifying listeners", indexPlan.getFilter().toString(), t);
+            LOG.error("Error retrieving data from Elastic for jcr query [{}] :: Corresponding ES query {} : closing scanner, notifying listeners"
+                    , indexPlan.getFilter(), ElasticIndexUtils.toString(query), t);
             // closing scanner immediately after a failure avoiding them to hang (potentially) forever
             close();
         }
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
@@ -119,12 +119,8 @@ public class ElasticResultRowAsyncIterator implements Iterator<FulltextResultRow
 
         Throwable error = errorRef.getAndSet(null);
         if (error != null) {
-            Exception e = new Exception();
-            e.setStackTrace(Thread.currentThread().getStackTrace());
-            String exceptionMsg = error.getMessage();
-            LOG.error("Error while fetching Results from Elastic for [{}]. Exception from Elastic : {}, Caller Stack Trace :"
-                    , indexPlan.getFilter().toString(), exceptionMsg, e);
-
+            error.fillInStackTrace();
+            LOG.error("Error while fetching Results from Elastic for [{}]", indexPlan.getFilter().toString(), error);
         }
         return !POISON_PILL.path.equals(nextRow.path);
     }
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
@@ -49,6 +49,7 @@ import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 
@@ -75,7 +76,7 @@ public class ElasticResultRowAsyncIterator implements Iterator<FulltextResultRow
     private final ElasticRequestHandler elasticRequestHandler;
     private final ElasticResponseHandler elasticResponseHandler;
     private final ElasticFacetProvider elasticFacetProvider;
-    private final BlockingQueue<Throwable> errorQueue = new LinkedBlockingQueue<>();
+    private final AtomicReference<Throwable> errorRef = new AtomicReference();
 
     private FulltextResultRow nextRow;
 
@@ -111,24 +112,20 @@ public class ElasticResultRowAsyncIterator implements Iterator<FulltextResultRow
             }
         }
 
-        // Check if there are any exception traces filled from onFailure Callback in the errorQueue
+        // Check if there are any Throwable filled from onFailure Callback in the errorReference
         // Any exception (such as ParseException) during the prefetch (init scanner) via the async call to ES would be available here
         // when the cursor is actually being traversed.
         // This is being done so that we can log the caller stack trace in case of any exception from ES and not just the trace of the async query thread.
-        if (!errorQueue.isEmpty()) {
+
+        Throwable error = errorRef.get();
+        if (error != null) {
             Exception e = new Exception();
             e.setStackTrace(Thread.currentThread().getStackTrace());
-            String exceptionMsg = null;
-            try {
-                exceptionMsg = errorQueue.take().getMessage();
-            } catch (InterruptedException interruptedException) {
-                LOG.debug("Error while trying to read ES exception from error queue");
-            }
+            String exceptionMsg = error.getMessage();
             LOG.error("Error while fetching Results from Elastic for [{}]. Exception from Elastic : {}, Caller Stack Trace :"
                     , indexPlan.getFilter().toString(), exceptionMsg, e);
+
         }
-
-
         return !POISON_PILL.path.equals(nextRow.path);
     }
 
@@ -341,13 +338,8 @@ public class ElasticResultRowAsyncIterator implements Iterator<FulltextResultRow
         public void onFailure(Throwable t) {
             metricHandler.measureFailedQuery(indexNode.getDefinition().getIndexPath(),
                     System.currentTimeMillis() - searchStartTime);
-            try {
-                errorQueue.put(t);
-            } catch (InterruptedException e) {
-                LOG.debug("Error while adding ES exception to error queue. Caller stack - trace wouldn't be logged.");
-                // Log just the ES exception with the async trace and the query only in this case.
-                LOG.error("Error retrieving data from Elastic for query [{}] : closing scanner, notifying listeners", indexPlan.getFilter().toString(), t);
-            }
+            errorRef.set(t);
+            LOG.error("Error retrieving data from Elastic for query [{}] : closing scanner, notifying listeners", indexPlan.getFilter().toString(), t);
             // closing scanner immediately after a failure avoiding them to hang (potentially) forever
             close();
         }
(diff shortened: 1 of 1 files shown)


oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
```diff
@@ -71,11 +71,11 @@ public class ElasticResultRowAsyncIterator implements Iterator<FulltextResultRow
     private final Predicate<String> rowInclusionPredicate;
     private final ElasticMetricHandler metricHandler;
     private final LMSEstimator estimator;
-
     private final ElasticQueryScanner elasticQueryScanner;
     private final ElasticRequestHandler elasticRequestHandler;
     private final ElasticResponseHandler elasticResponseHandler;
     private final ElasticFacetProvider elasticFacetProvider;
+    private final BlockingQueue<Throwable> errorQueue = new LinkedBlockingQueue<>();
```
nfsantos: Is it possible for `onFailure()` to be called more than once? My understanding is that it is called at most once, so maybe we could use a simpler data structure, like an `AtomicReference`.
Or as an alternative, use the same queue as the one used for retrieving the results, doing something similar to the poison pill. This may simplify the logic as well because there will be a single queue to check. Or maybe even use the poison pill and change it to optionally include an exception, so it can be used for both the successful and unsuccessful case.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
```diff
@@ -111,6 +110,25 @@ public boolean hasNext() {
...
+        // This is being done so that we can log the caller stack trace in case of any exception from ES and not just the trace of the async query thread.
+        if (!errorQueue.isEmpty()) {
+            Exception e = new Exception();
+            e.setStackTrace(Thread.currentThread().getStackTrace());
+            String exceptionMsg = null;
+            try {
+                exceptionMsg = errorQueue.take().getMessage();
+            } catch (InterruptedException interruptedException) {
+               LOG.debug("Error while trying to read ES exception from error queue");
+            }
```
nfsantos: Is the stack trace of the original exception logged somewhere? Here we are creating a new stack trace from the callers's thread, which is for sure useful to log. But we are only taking the message from the original exception, not the stack trace. The original stack trace is also not logged inside `onFailure()`. I think we should log both stack traces.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
```diff
@@ -111,6 +110,25 @@ public boolean hasNext() {
...
+        // Check if there are any exception traces filled from onFailure Callback in the errorQueue
+        // Any exception (such as ParseException) during the prefetch (init scanner) via the async call to ES would be available here
+        // when the cursor is actually being traversed.
+        // This is being done so that we can log the caller stack trace in case of any exception from ES and not just the trace of the async query thread.
+        if (!errorQueue.isEmpty()) {
+            Exception e = new Exception();
+            e.setStackTrace(Thread.currentThread().getStackTrace());
+            String exceptionMsg = null;
+            try {
+                exceptionMsg = errorQueue.take().getMessage();
```
nfsantos: The call to `take()` has no timeout, so it can block for an unbounded time. This should not be a problem in this case, but as a general practice I think it's best to always have a timeout. Otherwise, there is a chance that a bug may cause the caller thread to block here forever, which can be very hard to debug as the system will seem to be hang without any useful information on the logs. So often, in these cases, the only way to figure out what is happening is to get a thread dump, which may not be easy to get. If all blocking operations have a timeout, then after a while the operation fails and we get an exception on the logs. 

In this case, as we checked that the queue has something in it, we could use `poll()` instead, failing immediately if the queue is empty.

nfsantos (commented):
I think this code is logging only the stack trace of the caller thread, not of the async thread. And both are important for debugging any issue.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
nit0906: @nfsantos  - there could be possible multiple calls to onFailure (since we get the results from ES in batched - there could possibly be a case where the query is fine and returns well on the first fetch but there are some connectivity issues during the second fetch - I am not entirely sure about such a case though.)

But agreed, I can possibly try and use the existing result queue somehow.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
nit0906: I can log that back again. I was just not finding it much useful. Apart from the exception message from ES - the rest of the trace is just mostly the ES async call handling. 

But I think it wouldn't be any harm to log the original stack trace too. I will do that.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
nfsantos: Ok, I see how `onFailure()` can potentially be called multiple times, although it's not totally clear to me how that would happen. `onFailure()` calls `close()`, which suggests that the intention is to stop any more retrieval of results or of errors. And there is a semaphore to ensure that there is just one request sent at a time. But the way that synchronisation is done with the semaphore, it seems possible that we send another request even after calling close. Then again, I'm not totally sure about the flow of this code.

nit0906: Thanks @nfsantos  - I think using an AtomicReference here should be ok - made the changes accordingly.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
nit0906: This should now be redundant since now I am using AtomicReference instead of queue

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
```diff
@@ -111,6 +111,21 @@ public boolean hasNext() {
                 throw new IllegalStateException("Error reading next result from Elastic", e);
             }
         }
+
+        // Check if there are any Throwable filled from onFailure Callback in the errorReference
+        // Any exception (such as ParseException) during the prefetch (init scanner) via the async call to ES would be available here
+        // when the cursor is actually being traversed.
+        // This is being done so that we can log the caller stack trace in case of any exception from ES and not just the trace of the async query thread.
+
+        Throwable error = errorRef.get();
```
nfsantos: The `errorRef` is not cleared by this call to `get`. Is there the risk that this method will be called again and it will log the same exception several times? Maybe use `getAndSet(null)` to ensure we do not log more than once.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
```diff
@@ -323,7 +338,8 @@ public void onSuccess(SearchResponse<ObjectNode> searchResponse) {
         public void onFailure(Throwable t) {
             metricHandler.measureFailedQuery(indexNode.getDefinition().getIndexPath(),
                     System.currentTimeMillis() - searchStartTime);
-            LOG.error("Error retrieving data from Elastic: closing scanner, notifying listeners", t);
+            errorRef.set(t);
```
nfsantos: This is not checking if errorRef is already set. In that case, the exception stored there will be lost. If we assume that onFailure is called only once, then it's fine (although we may want to log a warning in case errorRef is already set, `getAndSet()` could once again be useful here). But if the assumption is that it can be called multiple times, then this can lead to losing exceptions.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
nit0906: Thanks. Good catch.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
nit0906: Done.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
```diff
@@ -111,6 +111,21 @@ public boolean hasNext() {
...
+
+        // Check if there are any Throwable filled from onFailure Callback in the errorReference
+        // Any exception (such as ParseException) during the prefetch (init scanner) via the async call to ES would be available here
+        // when the cursor is actually being traversed.
+        // This is being done so that we can log the caller stack trace in case of any exception from ES and not just the trace of the async query thread.
+
+        Throwable error = errorRef.getAndSet(null);
+        if (error != null) {
+            Exception e = new Exception();
+            e.setStackTrace(Thread.currentThread().getStackTrace());
```
nfsantos: Does this line have any effect? The constructor of `Exception` (actually, of `Throwable`) calls `fillInStackTrace()`, which should be very similar to what `Thread.currentThread().getStackTrace()` returns. What do you think about using the original exception but replacing the stack trace with the caller's trace by calling `error.fillInStackTrace()`?

nfsantos (approved):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
nfsantos: In fact, calling `Thread.currentThread().getStackTrace()` boils down to this:

```
public StackTraceElement[] getStackTrace() {
    ....
    return (new Exception()).getStackTrace();
  }
}
```
So it's an indirect way of calling `fillInStackTrace()`

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
```diff
@@ -111,6 +111,17 @@ public boolean hasNext() {
...
+
+        // Check if there are any Throwable filled from onFailure Callback in the errorReference
+        // Any exception (such as ParseException) during the prefetch (init scanner) via the async call to ES would be available here
+        // when the cursor is actually being traversed.
+        // This is being done so that we can log the caller stack trace in case of any exception from ES and not just the trace of the async query thread.
+
+        Throwable error = errorRef.getAndSet(null);
+        if (error != null) {
+            error.fillInStackTrace();
+            LOG.error("Error while fetching Results from Elastic for [{}]", indexPlan.getFilter().toString(), error);
```
thomasmueller: Nit: Results => results

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
```diff
@@ -323,7 +334,14 @@ public void onSuccess(SearchResponse<ObjectNode> searchResponse) {
...
                     System.currentTimeMillis() - searchStartTime);
-            LOG.error("Error retrieving data from Elastic: closing scanner, notifying listeners", t);
+            // Check in case errorRef is already set - this seems unlikely since we close the scanner once we hit failure.
+            // But still, in case this do happen, we will log a warn.
+            Throwable error = errorRef.getAndSet(t);
+            if (error != null) {
+                LOG.warn("Error reference for async iterator was previously set to {}. It has now been reset to new error {}", error.getMessage(), t.getMessage());
+            }
+
+            LOG.error("Error retrieving data from Elastic for query [{}] : closing scanner, notifying listeners", indexPlan.getFilter().toString(), t);
```
thomasmueller: I think the "toString()" is not necessary. Removing it also protects against NullPointerException (which is likely not possible... but still)

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
```diff
@@ -111,6 +111,17 @@ public boolean hasNext() {
                 throw new IllegalStateException("Error reading next result from Elastic", e);
```
thomasmueller, line 111: I wonder if we should log this...?

nit0906: As in log and return false (let it return silently ?) (btw, this is not part of the PR) 
I am not entirely sure why this was thrown initially...

nit0906: As in log and return false (let it return silently ?) (This is not part of the PR) 
I am not entirely sure why this was thrown initially ?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
fabriziofortino: Remove `toString`. We should log the Elastic query too.

```suggestion
            LOG.error("Error retrieving data from Elastic for query [{}] : closing scanner, notifying listeners", indexPlan.getFilter(), t);
```

fabriziofortino (changes requested):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
fabriziofortino: This will be thrown only in case the thread gets interrupted. I don't think it's an issue leaving it like that.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
nit0906: done. Added the ES query as well

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
```diff
@@ -323,7 +334,15 @@ public void onSuccess(SearchResponse<ObjectNode> searchResponse) {
...
-            LOG.error("Error retrieving data from Elastic: closing scanner, notifying listeners", t);
+            // Check in case errorRef is already set - this seems unlikely since we close the scanner once we hit failure.
+            // But still, in case this do happen, we will log a warn.
+            Throwable error = errorRef.getAndSet(t);
+            if (error != null) {
+                LOG.warn("Error reference for async iterator was previously set to {}. It has now been reset to new error {}", error.getMessage(), t.getMessage());
+            }
+
+            LOG.error("Error retrieving data from Elastic for jcr query [{}] :: Corresponding ES query {} : closing scanner, notifying listeners"
+                    , indexPlan.getFilter(), ElasticIndexUtils.toString(query), t);
```
fabriziofortino:
```suggestion
            LOG.error("Error retrieving data for jcr query [{}] :: Corresponding ES query {} : closing scanner, notifying listeners",
                indexPlan.getFilter(), ElasticIndexUtils.toString(query), t);
```

fabriziofortino (approved):
Fixed formatting. LGTM.

PR #640, opened 2022-07-26
OAK-9849: implement handling when oak-run is terminated with an OS signal

joerghoh:
When oak-run is terminated with a signal (e.g. because it runs under the supervision of the unix command "timeout"), it will print on stderr a message just indicating this. Also it can be configured to print a stacktrace, which helps to identify the cause why such a situation occurred.

This is implemented by a shutdownHook, which will print the message and do the threaddump, unless the execution flow is returning from the command execution.

While implementing this I had to change the ```Command``` interface to return a int, which is considered as a existcode and returned to the caller (typically the shell). Before that many Commands called ```System.exit``` by themselves; and in this case the flow never returned to the Main class; in this case the shutdown hook is always under the impression, that the command was not returning properly, but instead the JVM is shutdown by a different mean.

 I found that in general the handling of exit codes is quite inconsistent and sometimes not even existing. I created https://issues.apache.org/jira/browse/OAK-9861 for it.

### DIFF
--- a/oak-benchmarks-elastic/src/main/java/org/apache/jackrabbit/oak/benchmark/ElasticBenchmarkRunner.java
+++ b/oak-benchmarks-elastic/src/main/java/org/apache/jackrabbit/oak/benchmark/ElasticBenchmarkRunner.java
@@ -25,7 +25,11 @@ public class ElasticBenchmarkRunner extends BenchmarkRunner {
 
     private static ElasticConnection connection;
 
-    public static int main(String[] args) throws Exception {
+    public static void main(String[] args) throws Exception {
+        System.exit(runBenchmark(args));
+    }
+
+    public static int runBenchmark(String[] args) throws Exception {
         initOptionSet(args);
         statsProvider = options.has(benchmarkOptions.getMetrics()) ? getStatsProvider() : StatisticsProvider.NOOP;
         // Create an Elastic Client Connection here and pass down to all the tests
@@ -80,7 +84,7 @@ public class ElasticBenchmarkRunner extends BenchmarkRunner {
 
                     )
             );
-            return BenchmarkRunner.main(args);
+            return BenchmarkRunner.runBenchmark(args);
         } finally {
             connection.close();
         }
--- a/oak-benchmarks-elastic/src/main/java/org/apache/jackrabbit/oak/benchmark/ElasticBenchmarkRunner.java
+++ b/oak-benchmarks-elastic/src/main/java/org/apache/jackrabbit/oak/benchmark/ElasticBenchmarkRunner.java
@@ -25,7 +25,7 @@ public class ElasticBenchmarkRunner extends BenchmarkRunner {
 
     private static ElasticConnection connection;
 
-    public static void main(String[] args) throws Exception {
+    public static int main(String[] args) throws Exception {
         initOptionSet(args);
         statsProvider = options.has(benchmarkOptions.getMetrics()) ? getStatsProvider() : StatisticsProvider.NOOP;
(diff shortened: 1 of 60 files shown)


oak-benchmarks-elastic/src/main/java/org/apache/jackrabbit/oak/benchmark/ElasticBenchmarkRunner.java
```diff
@@ -25,7 +25,7 @@ public class ElasticBenchmarkRunner extends BenchmarkRunner {
 
     private static ElasticConnection connection;
 
-    public static void main(String[] args) throws Exception {
+    public static int main(String[] args) throws Exception {
```
thomasmueller: You can run Java programs with "java ... <className>" but only if they have public static *void* main method. So this change would have the (unintended?) consequence that this is no longer possible.

I think public static void main, and then System.exit(code) wouldn't cause this incompatibility. Are you sue System.exit doesn't work?

thomasmueller: Actually... With System.exit(code), behaviour of existing code would change it it calls the main method from some other method... so instead of System.exit(code > 0), maybe throwing an exception would be the right way?

joerghoh: ```ElasticBenchmarkRunner``` is called from the BenchmarkElasticCommand, and it does not integrate into ```oak-run```, but nevertheless implements ```Command```.

I agree, you should still have the chance to start the  ```ElasticBenchmarkRunner``` as a standalone command, but in that case I would move the logic out of the ```main``` method into a ```runBenchmark``` method, which gets the same params as ```main``` but has a return value. 

Will post an update to this PR.

joerghoh: Whenever a command executes ```System.exit(statuscode);``` directly, a stacktrace will be written during termination. For the shutdownHook in ```org.apache.jackrabbit.oak.run.Main```  it's not possible to find out if the JVM is terminated by the OS or by triggered by a ```System.exit()``` call.
Therefor all commands need to return to this main method.

thomasmueller (changes requested):
My concerns in the PR don't seem to be addressed.
"Actually... With System.exit(code), behaviour of existing code would change it it calls the main method from some other method... so instead of System.exit(code > 0), maybe throwing an exception would be the right way?"
I don't like to add System.exit to all main methods. I also don't think it is necessary to resolve the problem described in this jira issue

oak-benchmarks-elastic/src/main/java/org/apache/jackrabbit/oak/benchmark/ElasticBenchmarkRunner.java
joerghoh: Regarding the "throw an exception instead of returning a int as exitcode": This could be an alternative, but I don't see any improvement over my approach. Because if you just throw an exception and let the JVM handle it, the JVM will implicitly set the exit code to "1". 
This would allow a binary state: Success or fail. But looking at the variety of calls to ```System.exit()``` in the various commands, I see more exit codes than just "0" and "1". I believe we even have "-1". 
Converting these cases all into throwing exceptions, we would also get changes in the behavior. Not to mention that java terminating a program with an exception is likely to change the output of the tool.

PR #642, opened 2022-07-28
OAK-9873 Prefetch node states: composite node store support

thomasmueller:
To simplify testing (both automated and manual testing), I also extended support on the query engine side:

    /jcr:root//*[@foo = 'bar'] option(prefetches 10)

This allows to enable prefetch on a per-query basis, even if the JMX query engine settings bean doesn't have prefetch enabled.

I also added logging to CacheWarming to simplify testing.

The main challenge was adding a test case. I used the existing oak-lucene composite node store tests.

The main change was in CompositionContext, where the node state needs to be un-wrapped.

Documentation is missing still, but that's on purpose: we don't want people to use this feature yet, before we have validated that it actually helps.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryOptions.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryOptions.java
@@ -39,6 +39,7 @@ public class QueryOptions {
     public Optional<Long> limit = Optional.empty();
     public Optional<Long> offset = Optional.empty();
     public List<String> prefetch = Collections.emptyList();
+    public Optional<Integer> prefetchCount = Optional.empty();
     
     public enum Traversal {
         // traversing without index is OK for this query, and does not fail or log a warning
@@ -61,6 +62,7 @@ public class QueryOptions {
         limit = defaultValues.limit;
         offset = defaultValues.offset;
         prefetch = defaultValues.prefetch;
+        prefetchCount = defaultValues.prefetchCount;
     }
 
     QueryOptions(JsonObject json) {
@@ -83,6 +85,14 @@ public class QueryOptions {
                 limit = Optional.of(Long.parseLong(x));
             } catch (NumberFormatException e) {
                 LOG.warn("Invalid limit {}", x);
+            }
+        }
+        x = map.get("prefetches");
+        if (x != null) {
+            try {
+                prefetchCount = Optional.of(Integer.parseInt(x));
+            } catch (NumberFormatException e) {
+                LOG.warn("Invalid prefetch count {}", x);
             }
         }
         x = map.get("prefetch");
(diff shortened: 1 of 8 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryOptions.java
```diff
@@ -39,6 +39,7 @@ public class QueryOptions {
     public Optional<Long> limit = Optional.empty();
     public Optional<Long> offset = Optional.empty();
     public List<String> prefetch = Collections.emptyList();
+    public Optional<Integer> prefetchCount = Optional.empty();
```
rishabhdaim, line 42: I think we can use `java.util.OptionalInt` here and avoid unnecessary boxes.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/prefetch/CacheWarming.java
```diff
@@ -48,6 +52,9 @@ public void prefetch(@NotNull Iterable<String> paths,
                 ids.add(id);
             }
         }
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("Prefetch {} nodes", ids.size());
+        }
```
nfsantos, line 57: Just out of curiosity, does calling `LOG.isDebugEnabled()` actually helps performance in this case? My understanding is that `LOG.debug()` will only do a check if debug is enabled and returns at once if so, so the cost of calling that method is only the cost of evaluating the arguments. In this case, `ids.size()` simply returns an int, so it's very cheap (probably it's even inlined by the JIT). Or is there any other reason for the if guard?

thomasmueller: > does calling LOG.isDebugEnabled() actually helps performance in this case?

Probably it doesn't help currently, except if ids.size() is slow (you are right, it is not slow, for LinkedList -- for ConcurrentLinkedDeque it can be slow).
I just wanted to avoid any kind of risks, including the risk that someone might later add more data, that ids is null, things like that.
It's more following the pattern that all LOG.debug are behind LOG.isDebugEnabled()

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryOptions.java
thomasmueller: Good point! there's also OptionalLong for limit and offset! For this case, memory usage / performance isn't an issue, but I'm fine changing it -- maybe in another PR?

mreutegg (commented):
Looks good to me. I would add a simple test like the following to `CompositionContextTest`.
```
    @Test
    public void prefetch() {
        MountInfoProvider mip = Mounts.newBuilder()
                .mount("libs","/libs")
                .build();
        MyStore ns = spy(MyStore.class);
        CompositionContext ctx = new CompositionContext(mip, ns, Collections.emptyList(), CompositeNodeStoreMonitor.EMPTY_INSTANCE, CompositeNodeStoreMonitor.EMPTY_INSTANCE);
        NodeState root = EmptyNodeState.EMPTY_NODE;
        ctx.prefetch(Collections.singletonList("/foo"), ctx.createRootNodeState(root));
        verify(ns, times(1)).prefetch(any(), same(root));
    }

    interface MyStore extends NodeStore, PrefetchNodeStore {
    }

```

See https://github.com/apache/jackrabbit-oak/pull/647

PR #643, opened 2022-07-29
OAK-9870 | Escaping brackets to avoid query parse exceptions

nit0906:


fabriziofortino (approved):

thomasmueller:
Could you please revert it? I find it dangerous, because it affects Lucene indexes and it's not behind a feature flag.

PR #644, opened 2022-07-29
OAK-9874 (unit test only) - Add a (failing and @Ignored) unit test to check that queries return same result with and without index.

nfsantos:
The unit tests in this PR reproduces the issue in [OAK-9874](https://issues.apache.org/jira/browse/OAK-9874)

### DIFF
--- a/oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexAndTraversalQueriesSimilarResultsCommonTest.java
+++ b/oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexAndTraversalQueriesSimilarResultsCommonTest.java
@@ -19,17 +19,17 @@ package org.apache.jackrabbit.oak.plugins.index.elastic;
 import com.google.common.collect.ImmutableList;
 import org.apache.jackrabbit.oak.InitialContentHelper;
 import org.apache.jackrabbit.oak.api.ContentRepository;
-import org.apache.jackrabbit.oak.plugins.index.SameQueryResultsWithAndWithoutIndexTest;
+import org.apache.jackrabbit.oak.plugins.index.IndexAndTraversalQueriesSimilarResultsCommonTest;
 import org.apache.jackrabbit.oak.plugins.memory.MemoryNodeStore;
 import org.junit.ClassRule;
 
-public class ElasticSameQueryResultsWithAndWithoutIndexTest extends SameQueryResultsWithAndWithoutIndexTest {
+public class ElasticIndexAndTraversalQueriesSimilarResultsCommonTest extends IndexAndTraversalQueriesSimilarResultsCommonTest {
 
     @ClassRule
     public static final ElasticConnectionRule elasticRule =
             new ElasticConnectionRule(ElasticTestUtils.ELASTIC_CONNECTION_STRING);
 
-    public ElasticSameQueryResultsWithAndWithoutIndexTest() {
+    public ElasticIndexAndTraversalQueriesSimilarResultsCommonTest() {
         indexOptions = new ElasticIndexOptions();
         super.passingQueries = ImmutableList.of(
                 // Full-text queries
--- a/oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexAndTraversalQueriesSimilarResultsCommonTest.java
+++ b/oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexAndTraversalQueriesSimilarResultsCommonTest.java
@@ -16,7 +16,6 @@
  */
 package org.apache.jackrabbit.oak.plugins.index.elastic;
 
-import com.google.common.collect.ImmutableList;
 import org.apache.jackrabbit.oak.InitialContentHelper;
 import org.apache.jackrabbit.oak.api.ContentRepository;
 import org.apache.jackrabbit.oak.plugins.index.IndexAndTraversalQueriesSimilarResultsCommonTest;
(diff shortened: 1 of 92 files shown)


thomasmueller (approved):

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/SameQueryResultsWithAndWithoutIndexTest.java
```diff
@@ -0,0 +1,131 @@
...
+    protected void createTestIndexNode() throws Exception {
+        // This is necessary so that queries without indexes will return results
+        setTraversalEnabled(true);
+    }
+
+    protected Tree setIndex(IndexDefinitionBuilder builder, String idxName) {
+        return builder.build(root.getTree("/").addChild(INDEX_DEFINITIONS_NAME).addChild(idxName));
+    }
+
+    final private static List<String> testQueries = ImmutableList.of(
```
tihom88: I think most of the queries should have correct results for elastic and lucene. I think we should create 2 methods:
1. one for passing queries  and enable this method. 
2. Separate method for failing queries and annotate it with ignore.
With this we will have one active test.

tihom88 (commented):
nitpicking: May be we should add common in SameQueryResultsWithAndWithoutIndexTest, to indicate that it is a common test for both lucene and elastic.

nfsantos:
> nitpicking: May be we should add common in SameQueryResultsWithAndWithoutIndexTest, to indicate that it is a common test for both lucene and elastic.

@thomasmueller suggested removing the `Common` from the names of the tests. I'm fine with either option, but let's agree on one naming pattern.

nit0906:
> > nitpicking: May be we should add common in SameQueryResultsWithAndWithoutIndexTest, to indicate that it is a common test for both lucene and elastic.
> 
> @thomasmueller suggested removing the `Common` from the names of the tests. I'm fine with either option, but let's agree on one naming pattern.

Given that a lot of tests are there that still need to be ported - I would suggest to have some identifier that would suggest that these tests have already been ported simply by looking at the class name itself.

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/SameQueryResultsWithAndWithoutIndexTest.java
nfsantos: Done. Please check if this is what you suggested.

nfsantos:
> > > nitpicking: May be we should add common in SameQueryResultsWithAndWithoutIndexTest, to indicate that it is a common test for both lucene and elastic.
> > 
> > 
> > @thomasmueller suggested removing the `Common` from the names of the tests. I'm fine with either option, but let's agree on one naming pattern.
> 
> Given that a lot of tests are there that still need to be ported - I would suggest to have some identifier that would suggest that these tests have already been ported simply by looking at the class name itself.

After talking with Thomas, I have renamed the test to include Common in the name.

nit0906 (approved):

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexAndTraversalQueriesSimilarResultsCommonTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+public class ElasticIndexAndTraversalQueriesSimilarResultsCommonTest extends IndexAndTraversalQueriesSimilarResultsCommonTest {
+
+    @ClassRule
+    public static final ElasticConnectionRule elasticRule =
+            new ElasticConnectionRule(ElasticTestUtils.ELASTIC_CONNECTION_STRING);
+
+    public ElasticIndexAndTraversalQueriesSimilarResultsCommonTest() {
+        indexOptions = new ElasticIndexOptions();
+        super.passingQueries = ImmutableList.of(
```
thomasmueller: This and LuceneIndexAndTraversalQueriesSimilarResultsCommonTest seem to have the same values for passingQueries and failingQueries... so could we move this to IndexAndTraversalQueriesSimilarResultsCommonTest ? Maybe I'm missing some aspect that wouldn't allow this... or do you think this will change in the future?

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/IndexAndTraversalQueriesSimilarResultsCommonTest.java
```diff
@@ -0,0 +1,131 @@
...
+        setTraversalEnabled(true);
+    }
+
+    protected Tree setIndex(IndexDefinitionBuilder builder, String idxName) {
+        return builder.build(root.getTree("/").addChild(INDEX_DEFINITIONS_NAME).addChild(idxName));
+    }
+
+    protected List<String> passingQueries;
+
+    protected List<String> failingQueries;
```
thomasmueller: Nitpicking here a bit... for me the term "queriesFailed" means that there was an exception when running the query... What about "sameResultQueries" / "differentResultQueries" or "queriesWithSameResult" / "queriesWithDifferentResult"?

nfsantos: Good point, I changed the names to `.queriesWitSameResults`.

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexAndTraversalQueriesSimilarResultsCommonTest.java
nfsantos: I moved these lists to the base class to remove duplication. In the future, I expect that they may be different, but then we can move the lists to the derived classes. For the time being, I agree it's better to avoid duplication.

tihom88 (approved):

PR #646, opened 2022-08-02
OAK-9850: ConcurrentPrefetchAndUpdateIT.cacheConsistency fails occasi…

mreutegg:
…onally

Keys collection passed to registerTracker must not be modified later.
NodeDocumentCache.putNonConflictingDocs() must always notify other trackers.
Enable test again.

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
@@ -1411,7 +1411,7 @@ public class MongoDocumentStore implements DocumentStore {
         List<String> resultKeys = new ArrayList<>(keys.size());
         CacheChangesTracker tracker = null;
         if (collection == Collection.NODES) {
-            tracker = nodesCache.registerTracker(keys);
+            tracker = nodesCache.registerTracker(new HashSet<>(keys));
         }
         Throwable t;
         try {
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
@@ -1411,6 +1411,8 @@ public class MongoDocumentStore implements DocumentStore {
         List<String> resultKeys = new ArrayList<>(keys.size());
         CacheChangesTracker tracker = null;
         if (collection == Collection.NODES) {
+            // keys set is modified later. create a copy of the keys set
+            // owned by the cache changes tracker.
             tracker = nodesCache.registerTracker(new HashSet<>(keys));
         }
         Throwable t;
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/cache/NodeDocumentCache.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/cache/NodeDocumentCache.java
@@ -438,13 +438,23 @@ public class NodeDocumentCache implements Closeable {
             Lock lock = locks.acquire(id);
             try {
                 NodeDocument cachedDoc = getIfPresent(id);
-                // if an old document is present in the cache, we can simply update it
                 if (cachedDoc != null && isNewer(cachedDoc, d)) {
+                    // if an old document is present in the cache, we can simply update it
                     putInternal(d, tracker);
-                // if the document hasn't been invalidated or added during the tracker lifetime,
-                // we can put it as well
                 } else if (cachedDoc == null && !tracker.mightBeenAffected(id)) {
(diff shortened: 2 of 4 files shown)


stefan-egli (commented):
Besides fixing the test, and besides it being related to a new feature being involved (prefetch), would there also have been a possibility for outdated cache state previously?

mreutegg:
> Besides fixing the test

The test was not broken and this PR does not make any changes to the test except enable it again.

> would there also have been a possibility for outdated cache state previously?

I think that's possible. The added else clause in NodeDocumentCache.putNonConflictingDocs() is not specific to prefetching.

stefan-egli:
Wsa wondering if it would make sense to try and come up with a test case that would have failed without this fix. Maybe it could help finding or explaining other issues (like some stray OakMerge0004 conflicts..).

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1411,7 +1411,7 @@ public <T extends Document> void prefetch(Collection<T> collection,
         List<String> resultKeys = new ArrayList<>(keys.size());
         CacheChangesTracker tracker = null;
         if (collection == Collection.NODES) {
-            tracker = nodesCache.registerTracker(keys);
+            tracker = nodesCache.registerTracker(new HashSet<>(keys));
```
thomasmueller, line 1416: Is this to remove duplicates? Using HashSet will re-order the entries, but I guess that's fine. 

Possibly it makes sense to sort the entries so that MongoDB can retrieve them faster (a MongoDB specific optimization); I could imagine that MongoDB doesn't do that itself. Or so that compression over the wire can work better (I _think_ LZ4 is used?)

mreutegg: > Is this to remove duplicates?

No, it just registers the given keys with a `CacheChangesTracker`. Order is not important here. The `CacheChangesTracker` just needs an efficient lookup whether a key is in this set.

> Possibly it makes sense to sort the entries so that MongoDB can retrieve them faster

I don't know if that helps and I would only sort if there is evidence that it is faster.

> I think LZ4 is used?

It _can_ be used. Compression over the wire is a connection [option](https://www.mongodb.com/docs/drivers/java/sync/current/fundamentals/connection/network-compression/#specify-compression-algorithms).

mreutegg:
> Was wondering if it would make sense to try and come up with a test case that would have failed without this fix.

Running `ConcurrentPrefetchAndUpdateIT` in a loop eventually failed on my machine without the fix. But I agree, it would be better if there was a deterministic test that fails immediately without the fix.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
thomasmueller: > it just registers the given keys

Yes, but we had "registerTracker(keys)" before and you changed it to "registerTracker(new HashSet<>(keys))" and I wanted to understand the reason for this change

```
-            tracker = nodesCache.registerTracker(keys);
+            tracker = nodesCache.registerTracker(new HashSet<>(keys));

```

> I would only sort if there is evidence that it is faster.

Sure.

> Compression over the wire is a connection option

Thanks! So no LZ4, but there's Snappy. (which is a bit slower than LZ4 I think)

thomasmueller: Please ignore my comments about compression and ordering... I see now this doesn't apply here.

But "keys" is already "Set<String> keys = new HashSet<>();", and you wrapped that inside a new HashSet, what is the reason?

thomasmueller: Ah, is it because "Keys collection passed to registerTracker must not be modified later"  and you wanted to ensure the set is not modified later? I think it's fine... It might make sense to add a comment in the code

mreutegg: Yes, that's exactly the reason. Good point about the comment. I'll add that.

mreutegg:
Test failure in oak-blob-plugins seems unrelated. I can reproduce the failure locally and created https://issues.apache.org/jira/browse/OAK-9879

mreutegg:
> a test case that would have failed without this fix

@stefan-egli there is now a simple test that illustrates why changes to NodeDocumentCache are necessary. `NodeDocumentCacheTest` fails deterministically without changes to `NodeDocumentCache.putNonConflictingDocs()`.

stefan-egli (approved):
+1, nice test

PR #647, opened 2022-08-02
OAK-9873: Prefetch node states: composite node store support

mreutegg:


thomasmueller (approved):

mreutegg:
oak-core-spi failure looks like an infra glitch.
```
builds41 was marked offline: Connection was broken: java.io.IOException: Pipe closed after 0 cycles
	at org.apache.sshd.common.channel.ChannelPipedInputStream.read(ChannelPipedInputStream.java:126)
```
All other modules were successful.

PR #648, opened 2022-08-03
OAK-9803 : Extend DynamicSyncHandler to allow for dynamic groups

anchela:


oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/DefaultSyncConfigImpl.java
```diff
@@ -272,6 +272,23 @@ public class DefaultSyncConfigImpl extends DefaultSyncConfig {
...
+    /**
+     * @see DefaultSyncConfig.Group#getDynamicGroups()
+     */
+    public static final boolean PARAM_GROUP_DYNAMIC_GROUPS_DEFAULT = false;
+    
+    /**
+     * @see DefaultSyncConfig.Group#getDynamicGroups()
+     */
+    @Property(
+            label = "Group 'Dynamic Groups'",
```
telegrapher: This label seems difficult to understand.

It comes to my mind "Dynamic Group Membership" since the group membership gets dynamically calculated. Or maybe "Dynamically Resolved Group Membership" if we want to be more precise?

anchela: agree.... dynamic membership is already used but i will think about a better name.

anchela:
@telegrapher , i am going to merge the PR to avoid running into conflicts. i will keep the jira ticket open and will address additional review-findings as they appear. hope that is ok with you.

PR #649, opened 2022-08-04
OAK-9875 - Prefix query on a Long, analyzed field fails when executed over Elastic

nfsantos:
In ES, for properties with `analyzed=true`, use multi-fields for the full-text index:

```json
"properties": {
  "propa": {
    "type": "long",
    "ignore_malformed" : true,
    "fields" : {
      "text" : {
        "type" : "text",
        "analyzer" : "oak_analyzer"
      }
    }
  }
}
```
This fixes a bug where if a property had a non-string type and was marked as analyzed, the ES plugin would not create a field in ES for full text search, so full-text searches would fail.

## Additional changes
-  Use the Elastic option `ignore_malformed` to make Elastic accept documents with values that cannot be coerced to the type of the field. Without this option, if the user would provide an invalid value for long, then Elastic would discard the field and would also not index it as full-text. With ignore_malformed, the top level field will be ignored but the nested full-text index will still be updated. This is in line with the Lucene plugin, which always indexes a property, even if its value cannot be converted to the type. 
- Previously, the Elastic plugin was trying to convert the value to the type of the mapping before sending it to Elastic. This was done as a workaround to a bug where adding a string that did not represent a valid date to a field of type Date would result in Elastic throwing an exception and discarding the document (See [OAK-9665](https://issues.apache.org/jira/browse/OAK-9665)). In the previous implementation, if the Elastic plugin failed to convert the value to the type of the property, it would log a warning and discard the property. But we can no longer ignore the property, because even if it does not have a valid value, we must send it to Elastic to index as full-text in the nested field. Therefore, this PR relies on setting `ignore_malformed` in Elastic and sends to Elastic the value as received from the user, relying on Elastic to convert it.
- The only exception is for boolean properties, because Elastic does not support `ignore_malformed` for this type. Here the PR preserves the current behavior of trying to convert in OAK the value to a boolean and not indexing the field if that fails. This means that boolean properties do not support `analyzed=true` (we log an error when the index is created).  

Update: there is an Elasticsearch issue to add support for `ignore_malformed` to boolean fields: https://github.com/elastic/elasticsearch/issues/89542.

## Minor fixes:
- Fix: `AbstractQueryTest#assertResult` was not comparing the expected and actual lists correctly, it was only checking that all elements of the expected list are also in the actual list. For instance, these two lists would pass the assertion: `expected=['a', 'a']`, `actual=['a', 'b']`.
- Improved logging of errors in requests to Elastic.
- Correct a few spelling mistakes.


oak-core/src/test/java/org/apache/jackrabbit/oak/query/AbstractQueryTest.java
```diff
@@ -344,13 +345,21 @@ protected static void assertResultSorted(@NotNull List<String> expected, @NotNul
...
+        checkNotNull(actual);
+        assertEquals("Result set size is different. Expected: " + expected + ", actual: " + actual, expected.size(), actual.size());
+
+        String[] sortedActual = actual.toArray(new String[0]);
+        Arrays.sort(sortedActual);
+
+        String[] sortedExpected = expected.toArray(new String[0]);
+        Arrays.sort(sortedExpected);
+
+        assertTrue("Invalid result set. Expected: " + expected + ", Actual: " + actual, Arrays.equals(sortedExpected, sortedActual));
```
fabriziofortino: this can be simplified
```suggestion
        assertArrayEquals("Invalid result set. Expected: " + expected + ", Actual: " + actual, sortedExpected, sortedActual);
```

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexHelper.java
```diff
@@ -207,21 +222,57 @@ private static void mapIndexRules(ElasticIndexDefinition indexDefinition, XConte
...
+             *  field if propertyIndex=false.
+             */
             mappingBuilder.startObject(name);
             {
                 // https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-types.html
                 if (Type.BINARY.equals(type)) {
                     mappingBuilder.field("type", "binary");
+                    // ignore_malformed not allowed for binary fields but also not needed as there is no type conversion
+                    if (indexDefinition.isAnalyzed(propertyDefinitions)) {
+                        createNestedAnalyzed(mappingBuilder);
```
thomasmueller: I wonder if we could move the code to _after_ the if .. else if ... list, to avoid repetition:

```
                    if (indexDefinition.isAnalyzed(propertyDefinitions)) {
                        createNestedAnalyzed(mappingBuilder);
                    }

```

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexHelper.java
```diff
@@ -207,21 +222,57 @@ private static void mapIndexRules(ElasticIndexDefinition indexDefinition, XConte
...
+                    }
                 } else if (Type.BOOLEAN.equals(type)) {
                     mappingBuilder.field("type", "boolean");
+                    // ES does not support "ignore_malformed" on boolean properties, so all values added to the index
+                    // must parse as valid booleans or else the document is rejected. We therefore also do not support
+                    // the nested full text field, because to do so we would need to provide a value for the top level
+                    // field, which must be a valid value. As in this case we cannot set ignore_malformed=true to still
+                    // index as full-text the invalid values, we cannot have a full-text index. This breaks compatibility
+                    // with Lucene, which accepts any value, even if it is not a valid boolean, but this is unlikely to
+                    // have much impact because setting analyzed=true in a boolean property should have no real world
```
thomasmueller, line 267: Yes, I think we don't have any such cases currently, by running

`    grep -B 2 -A 2 -R --include="*.json" "Boolean" data | grep -B 2 -A 2 "analyzed" `

But we should probably document this as an incompatibility in https://jackrabbit.apache.org/oak/docs/query/elastic.html

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexHelper.java
```diff
@@ -207,21 +222,57 @@ private static void mapIndexRules(ElasticIndexDefinition indexDefinition, XConte
...
+                    // use case.
+                    if (indexDefinition.isAnalyzed(propertyDefinitions)) {
+                        LOG.warn("Property {} of type boolean does not support analyzed=true.", name);
+                    }
                 } else {
+                    // OAK-9875 - The ES field with the same name as the property should not be used for the full text
+                    // index, it should be used only for the keyword index. The full text index is done separately below.
+                    // We preserve the old mapping in case we find some problem in production with the new mapping.
```
thomasmueller: Not sure if I understand... so we keep this code to provide backward compatibility for old oak versions, during the upgrade of the existing indexes (to support blue-green deployments), plus for the case that there are problems? That sounds reasonable... We can then remove it later if everything goes fine.

If that's our plan, then what adding a new oak issue to remove it later, and in this patch add a constant "static final boolean BACKWARD_COMPATIBLE_MAPPING_OAK_... = true", so that we can track removal of this code? Hiding the code behind "if (BACKWARD_COMPATIBLE_MAPPING_OAK_...)" to make it obvious that this code will go away soon...

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
```diff
@@ -341,8 +343,16 @@ public void onFailure(Throwable t) {
                 LOG.warn("Error reference for async iterator was previously set to {}. It has now been reset to new error {}", error.getMessage(), t.getMessage());
             }
 
+            if (t instanceof ElasticsearchException) {
+                ElasticsearchException esException = (ElasticsearchException) t;
+                StringBuffer errorMessage = new StringBuffer("Status: " + esException.status() + ", Message: " + esException.getMessage());
+                for (ErrorCause c : esException.response().error().rootCause()) {
```
thomasmueller: The javadoc https://javadoc.io/doc/co.elastic.clients/elasticsearch-java/latest/co/elastic/clients/elasticsearch/_types/ElasticsearchException.html doesn't say, but could response() return null? For error() the javadoc says "Required - API name: error", so it shouldn't be possible (but then, there is no guarantee it might change), and root cause might also return null? ("API name: root_cause")

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticDocumentMaker.java
```diff
@@ -142,32 +148,39 @@ protected boolean isFulltextValuePersistedAtNode(PropertyDefinition pd) {
...
-                f = property.getValue(Type.DOUBLE, i);
-            } else if (tag == Type.BOOLEAN.tag()) {
+            Object f;
+            if (pdTypeTag == Type.BOOLEAN.tag()) {
+                // Try to convert to boolean here, as ES does not support ignore_malformed in boolean fields
                 f = property.getValue(Type.BOOLEAN, i).toString();
             } else {
-                f = property.getValue(Type.STRING, i);
+                // Let ES convert the property and rely on ignore_malformed=true to skip if the value is not valid
+                int indexTypeTag = property.getType().tag();
```
fabriziofortino: cant we still use `pdTypeTag` here?

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticAbstractQueryTest.java
```diff
@@ -140,7 +145,7 @@ protected ContentRepository createRepository() {
         ElasticIndexTracker indexTracker = new ElasticIndexTracker(esConnection, getMetricHandler());
         ElasticIndexEditorProvider editorProvider = new ElasticIndexEditorProvider(indexTracker, esConnection,
                 new ExtractedTextCache(10 * FileUtils.ONE_MB, 100));
-        ElasticIndexProvider indexProvider = new ElasticIndexProvider(indexTracker);
+        ElasticIndexProvider indexProvider = new ElasticIndexProvider(indexTracker, separateFullTextSearchESFieldFeature);
```
thomasmueller: If we want to add feature flags (now or later), then I think we will end up with many of them at some point. So, when we have more than one, we should probably have a wrapper class "ElasticConfiguration" that wraps the configuration (all feature flags). That way, when we add one more, we don't need to change the constructor again - and will never get 10 boolean parameters :-)

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticDocumentMaker.java
```diff
@@ -142,32 +148,39 @@ protected boolean isFulltextValuePersistedAtNode(PropertyDefinition pd) {
...
+                if (indexTypeTag == Type.LONG.tag()) {
+                    f = property.getValue(Type.LONG, i);
```
fabriziofortino, line 172: the comment above says `Let ES convert the property ...` but here (and also for the other types below) we do `property.getValue(Type.LONG, i);`. Doesn't this call perform a conversion? If not, we should at least put a comment here because I find it confusing.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticDocumentMaker.java
```diff
@@ -142,32 +148,39 @@ protected boolean isFulltextValuePersistedAtNode(PropertyDefinition pd) {
...
+                } else if (indexTypeTag == Type.DATE.tag()) {
+                    f = property.getValue(Type.DATE, i);
+                } else if (indexTypeTag == Type.DOUBLE.tag()) {
+                    f = property.getValue(Type.DOUBLE, i);
+                } else if (indexTypeTag == Type.BOOLEAN.tag()) {
+                    f = property.getValue(Type.BOOLEAN, i).toString();
+                } else {
```
fabriziofortino: is this needed?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
```diff
@@ -220,6 +222,35 @@ public String getElasticKeyword(String propertyName) {
...
+
+        Type<?> type = null;
+        for (PropertyDefinition pd : propertyDefinitions) {
+            type = Type.fromTag(pd.getType(), false);
+        }
+
+        if (isAnalyzed(propertyDefinitions)) {
+            // The full text index for String properties is <propertyName>, while for non-string properties is part of
+            // the multi-field and is called <propertyName>.text
+            if (Type.BINARY.equals(type) ||
```
thomasmueller, line 238: Isn't BINARY also indexed as text?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexHelper.java
```diff
@@ -194,6 +197,18 @@ private static void mapInternalProperties(XContentBuilder mappingBuilder) throws
         // .endObject();
     }
 
+    private static void createNestedAnalyzed(XContentBuilder mappingBuilder) throws IOException {
```
fabriziofortino: I would change the name to something like `createAnalyzedSubfield`. We should not use `Nested` since it can be confused with the nested type in elasticsearch that we use widely but not in this case.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexHelper.java
nfsantos: I added this comment to the Elastic documentation:

```
* Properties of type Boolean do not support `analyzed=true`.
```

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexHelper.java
nfsantos: The comment was out of date. I replaced it by the following:
```
// OAK-9875 - For String properties, the full-text index is stored at the top-level field. This is
// the opposite of what is done for non-String properties, where the full text field is nested, as
// propa.text. We do this to keep backwards compatibility with pre OAK=9875 indexes.
```
The comment about backwards compatibility is still valid. However, if we want to clean the code by breaking backwards compatibility, there will not be a better time than now when this module is still not in widespread usage. In the future it will be more costly and we will likely not do it. Or do you see any plan where we could change the format of the indexes without causing much disruption in the future?

I think adding a flag indicating that this is just for backwards compatibility is in principle a good idea, but I'm afraid that we will never get around to remove this behaviour and this logic will remain forever.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexHelper.java
nfsantos: I will try to simplify the logic on this method once we have a final version of the mappings. This logic is needed for all cases except for booleans, so it is not so straightforward to move it to a common place.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
nfsantos: I added null checks for all the intermediate calls. In the current version of the library that we use, they never return null, but as you say, this is not stated in the Javadocs so it can change in the future, so better to safeguard against that possibility.

oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticAbstractQueryTest.java
nfsantos: Yes. I was thinking the same when I was adding the future flag, that we would need to pass some kind of context object. But we likely will not need a FT in this case, so we can leave it for the future.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
nfsantos: In Elastic, Binary is a Base64 encoded string, which by default is not indexed.
https://www.elastic.co/guide/en/elasticsearch/reference/7.17/binary.html

I have not changed the way the JCR binary type is mapped, even before this PR, the mapping logic handles Binary, Long, Double, Date and Boolean all the in the same way, and text in a different way. But Binary does seem to require special treatment, as there are no good use cases for it to be indexed. That is, are there any situations where it may make sense to run either term-level or full-text queries on Binary fields?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticDocumentMaker.java
nfsantos: There are two types, the one defined in the index and the one of the property that is being indexed. `pdType` is the type in the index, while the second one is the type of the property. The variable names are wrong, I fixed that and improved the comments.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticDocumentMaker.java
nfsantos: It does not convert, because it is getting the value in the same type as the property. I added a comment to explain this better.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticDocumentMaker.java
nfsantos: The `toString()` conversion is in the current trunk. 
I don't think it's needed. With `toString()`, the value indexed is a string, with either "false" or "true". Elastic will coerce this value to a Boolean internally. Without `toString()`, Oak will send the boolean values false/true. The value indexed in Elastic is the same, the only difference is in the source document stored in Elastic, which I don't think is important. I don't know what is the reason why the `toString()` was there in the first place. 
I have removed the `toString()` cast.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexHelper.java
nfsantos: Done.

fabriziofortino (approved):
The fix LGTM. I still think we should not include the FT logic in this PR. If needed, it should be moved in a separate PR.

PR #651, opened 2022-08-04
OAK-9880 : simplify rgc query

stefan-egli:


stefan-egli:
one concern that just came to mind with this: obviously this would increase the number of queries executed - and that scales with the number of clusterNodeIds in use...

stefan-egli:
Maybe to compensate for that, do less frequent RGC, currently it runs every 5sec by default - changing that to every 1min might not have too much of an impact

mreutegg:
> one concern that just came to mind with this: obviously this would increase the number of queries executed - and that scales with the number of clusterNodeIds in use...

We could address this to some degree by remembering successful operations on minMaxRevTimeInSecs per clusterId. An inactive cluster node will not have its sweep revision updated and minMaxRevTimeInSecs will not change. This would limit operations to the number of active clusterIds.

> do less frequent RGC, currently it runs every 5sec by default - changing that to every 1min might not have too much of an impact

I would prefer to keep the interval small. This ensures RGC activity is stretched out more evenly and avoids spikes that impact an application. Every 5 seconds what an initial default set for this kind of continuous RGC. We can certainly discuss a change. Maybe in a separate issue?

mreutegg (changes requested):
Changes look good to me. This change does not come with a test. Could you at least add a test that checks an expected number of deleteMany operations are issued for a given set of clusterIds?

stefan-egli:
* added logic to avoid calling `deleteMany` for idle clusterNodeIds - plus tests for this
* added a test for number of `deleteMany` calls for the above and some of the existing tests

stefan-egli:
@mreutegg, as I have modified this PR from your earlier review would appreciate a second (final?) review, thx!

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
```diff
@@ -100,6 +105,11 @@ public MongoVersionGCSupport(MongoDocumentStore store) {
         }
     }
 
+    /** accessor for testing only */
+    protected RevisionVector getLastDefaultNoBranchDeletionRevs() {
```
mreutegg: I cannot find any usages of this method. Remove?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MongoVersionGCSupportDefaultNoBranchTest.java
```diff
@@ -0,0 +1,431 @@
...
+        if (MONGO.isAvailable()) {
+            fixtures.add(new MongoFixture() {
+                @Override
+                public DocumentStore createDocumentStore(Builder builder) {
+                    try {
+                        MongoConnection connection = MongoUtils.getConnection();
+                        connections.add(connection);
+                        MongoDatabase db = connection.getDatabase();
+                        final AtomicReference<String> noEx = new AtomicReference<>(null);
+                        MongoTestDatabase testDb = new MongoTestDatabase(db, noEx, noEx, noEx) {
```
mreutegg: Could this also go to `CountingMongoDatabase` and `CountingMongoCollection`?

mreutegg (approved):
Other than above, looks good to me.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
stefan-egli: agree, a left over from a previous test version. removed now

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MongoVersionGCSupportDefaultNoBranchTest.java
stefan-egli: yes, can (and without interfering with existing tests). done now in https://github.com/apache/jackrabbit-oak/pull/651/commits/d82ba32226e09ec545e348bc57367167052bfdbb

PR #652, opened 2022-08-04
OAK-9879: CacheLIRS statistics may return incorrect load count

mreutegg:
Synchronize access to loadSuccessCount
Enable test


oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/CacheLIRS.java
```diff
@@ -1025,7 +1025,9 @@ V load(K key, int hash, Callable<? extends V> valueLoader) throws ExecutionExcep
             long start = System.nanoTime();
             try {
                 value = valueLoader.call();
-                loadSuccessCount++;
+                synchronized (this) {
+                    loadSuccessCount++;
```
thomasmueller, line 1029: I think AtomicLong for all the counts would be even better... if we want to rely on the numbers

thomasmueller:
See alternative patch https://github.com/apache/jackrabbit-oak/pull/673

mreutegg:
Closing this PR in favour of #673

PR #653, opened 2022-08-05
OAK-9878 | OAK-9858 | Fixing failing(Ignored) tests for ES

nit0906:
This PR fixes multiple issues - both in tests as well as implementation code.

* https://github.com/apache/jackrabbit-oak/pull/653/commits/5ee6730f50171c657c40bbed986270a511c48229 - Make ES indexing sync using commit info map for IndexQueryCommonTests (this fixed a lot of test failures - behaviour is now similar to lucene)
* https://github.com/apache/jackrabbit-oak/pull/653/commits/4808d85e2bf33dacdf33cb2791a92dedf3b5c781 - Re-enable travesal for queries because there are some join queries in the tests where one part of the join requires traversal , this commit also fixes some expectations in tests that have been ignored for a while during which new tests added more collateral which changed the test expectations.
* https://github.com/apache/jackrabbit-oak/pull/653/commits/6c04951ecbd00b9f280872cc127ed49128dfdedf - Fixing sort for ES in case of sort property is not explicitly defined in the index definition but covered via a regex property.
* https://github.com/apache/jackrabbit-oak/pull/653/commits/5f161a2501342188f412241bf00489021e10d5eb - Fixing like queries in ES in case of fulltext/analyzed properties (needed to use fieldName.keyword instead of fieldName)
* https://github.com/apache/jackrabbit-oak/pull/653/commits/ce86979395ff1d1d632e2d18b11aceb0da41be1c - Fixing queries on nodeName constraint (:nodeName)
* https://github.com/apache/jackrabbit-oak/pull/653/commits/fa3dc375a0b83900eb5ae726553806ec5011b14b - Added missing test for path() function

### DIFF
--- a/oak-core/src/test/java/org/apache/jackrabbit/oak/query/AbstractQueryTest.java
+++ b/oak-core/src/test/java/org/apache/jackrabbit/oak/query/AbstractQueryTest.java
@@ -29,6 +29,7 @@ import java.nio.charset.StandardCharsets;
 import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -227,7 +228,16 @@ public abstract class AbstractQueryTest {
                     w.println(line);
                     line = line.substring("commit".length()).trim();
                     apply(root, line);
-                    root.commit();
+                    // This part of code is used by both lucene and elastic tests.
+                    // The index definitions in these tests don't have async property set
+                    // So lucene, in this case behaves in a sync manner. But the tests fail on Elastic,
+                    // since ES indexing is always async.
+                    // The below commit info map (sync-xmode = rt) would make Elastic use RealTimeBulkProcessHandler.
+                    // This would make ES indexes also sync.
+                    // This will NOT have any impact on the lucene tests.
+                    Map<String, Object> map = new HashMap<>();
+                    map.put("sync-mode", "rt");
+                    root.commit(map);
                 }
                 w.flush();
             }
--- a/oak-core/src/test/java/org/apache/jackrabbit/oak/query/AbstractQueryTest.java
+++ b/oak-core/src/test/java/org/apache/jackrabbit/oak/query/AbstractQueryTest.java
@@ -232,12 +232,10 @@ public abstract class AbstractQueryTest {
                     // The index definitions in these tests don't have async property set
                     // So lucene, in this case behaves in a sync manner. But the tests fail on Elastic,
                     // since ES indexing is always async.
-                    // The below commit info map (sync-xmode = rt) would make Elastic use RealTimeBulkProcessHandler.
+                    // The below commit info map (sync-mode = rt) would make Elastic use RealTimeBulkProcessHandler.
                     // This would make ES indexes also sync.
                     // This will NOT have any impact on the lucene tests.
-                    Map<String, Object> map = new HashMap<>();
-                    map.put("sync-mode", "rt");
-                    root.commit(map);
+                    root.commit(Collections.singletonMap("sync-mode", "rt"));
                 }
                 w.flush();
             }
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
@@ -736,7 +736,7 @@ public class ElasticRequestHandler {
         String first = pr.first != null ? pr.first.getValue(Type.STRING) : null;
         if (pr.first != null && pr.first.equals(pr.last) && pr.firstIncluding && pr.lastIncluding) {
             // [property]=[value]
-            return Query.of(q -> q.term(t -> t.field(FieldNames.NODE_NAME).value(FieldValue.of(first))));
+            return Query.of(q -> q.term(t -> t.field(elasticIndexDefinition.getElasticKeyword(FieldNames.NODE_NAME)).value(FieldValue.of(first))));
         }
 
         if (pr.isLike) {
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
@@ -228,7 +228,7 @@ public class ElasticRequestHandler {
             } else if (JCR_SCORE.equals(sortPropertyName)) {
                 fieldName = "_score";
             } else if (indexProperties.containsKey(sortPropertyName) || elasticIndexDefinition.getDefinedRules()
-                    .stream().map(rule -> Optional.ofNullable(rule.getConfig(sortPropertyName))).anyMatch(Optional::isPresent)) {
+                    .stream().anyMatch(rule -> rule.getConfig(sortPropertyName) != null)) {
                 // There are 2 conditions in this if statement -
                 // First one returns true if sortPropertyName is one of the defined indexed properties on the index
                 // Second condition returns true if sortPropertyName might not be explicitly defined but covered by a regex property
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
@@ -41,6 +41,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.BiPredicate;
@@ -227,11 +228,13 @@ public class ElasticRequestHandler {
             } else if (JCR_SCORE.equals(sortPropertyName)) {
                 fieldName = "_score";
             } else if (indexProperties.containsKey(sortPropertyName) || elasticIndexDefinition.getDefinedRules()
-                    .stream().map(rule -> rule.getConfig(sortPropertyName))
+                    .stream().map(rule -> Optional.ofNullable(rule.getConfig(sortPropertyName)))
+                    .filter(Optional::isPresent)
                     .collect(Collectors.toList()).size() > 0) {
                 // There are 2 conditions in this if statement -
                 // First one returns true if sortPropertyName is one of the defined indexed properties on the index
-                // Second condition returns true if sortPropertyName might not be explicitly defined but covered by a regex property.
+                // Second condition returns true if sortPropertyName might not be explicitly defined but covered by a regex property
+                // in any of the defined index rules.
                 fieldName = elasticIndexDefinition.getElasticKeyword(sortPropertyName);
             } else {
                 LOG.warn("Unable to sort by {} for index {}", sortPropertyName, elasticIndexDefinition.getIndexName());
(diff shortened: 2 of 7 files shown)


oak-core/src/test/java/org/apache/jackrabbit/oak/query/AbstractQueryTest.java
```diff
@@ -227,7 +228,16 @@ protected void test(String file) throws Exception {
                     w.println(line);
                     line = line.substring("commit".length()).trim();
                     apply(root, line);
-                    root.commit();
+                    // This part of code is used by both lucene and elastic tests.
+                    // The index definitions in these tests don't have async property set
+                    // So lucene, in this case behaves in a sync manner. But the tests fail on Elastic,
+                    // since ES indexing is always async.
+                    // The below commit info map (sync-xmode = rt) would make Elastic use RealTimeBulkProcessHandler.
```
fabriziofortino:
```suggestion
                    // The below commit info map (sync-mode = rt) would make Elastic use RealTimeBulkProcessHandler.
```

oak-core/src/test/java/org/apache/jackrabbit/oak/query/AbstractQueryTest.java
```diff
@@ -227,7 +228,16 @@ protected void test(String file) throws Exception {
...
+                    // This would make ES indexes also sync.
+                    // This will NOT have any impact on the lucene tests.
+                    Map<String, Object> map = new HashMap<>();
+                    map.put("sync-mode", "rt");
+                    root.commit(map);
```
fabriziofortino: more concise
```suggestion
                    root.commit(Collections.singletonMap("sync-mode", "rt"));
```

oak-core/src/test/java/org/apache/jackrabbit/oak/query/AbstractQueryTest.java
fabriziofortino: another alternative would be to create a method (eg `syncIndex`) in `AbstractQueryTest` overridden in `ElasticAbstractQueryTest` to pass the `sync-mode` option.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
```diff
@@ -225,7 +226,12 @@ public Query baseQuery() {
                 hasTieBreaker = true;
             } else if (JCR_SCORE.equals(sortPropertyName)) {
                 fieldName = "_score";
-            } else if (indexProperties.containsKey(sortPropertyName)) {
+            } else if (indexProperties.containsKey(sortPropertyName) || elasticIndexDefinition.getDefinedRules()
+                    .stream().map(rule -> rule.getConfig(sortPropertyName))
+                    .collect(Collectors.toList()).size() > 0) {
```
fabriziofortino: this should produce the same result and should be more robust (better handling of nulls). I have not tested it
```suggestion
                    .stream().map(rule -> Optional.ofNullable(rule.getConfig(sortPropertyName))).findAny().isPresent()) {
```

fabriziofortino (approved):
LGTM. Just a few potential improvements.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
nit0906: `.stream().map(rule -> Optional.ofNullable(rule.getConfig(sortPropertyName))).findAny().isPresent()`
This seems to be always returning true. Looks like `map(rule -> Optional.ofNullable(rule.getConfig(sortPropertyName)))` returns an Optional[Optional.Empty] so the element as result of findAny() is actually an Optional of another Optional  (Optional.Empty) and isPresent returns true in this case.
I think the issue is that findAny() also returns an Optional 

> Returns an Optional describing some element of the stream, or an empty Optional if the stream is empty.


Looking at this again - 
`.stream().map(rule -> rule.getConfig(sortPropertyName))
                    .collect(Collectors.toList()).size() > 0)` this too would always return true since the size of this list would always be greater than 0 even if all the elements are null. 

Also, since an index definition could possibly have multiple indexRules and we should return true if sortPropertyName mathces a namePattern in either of the rules - using findAny() could also be problematic. I think we need to filter out the nulls from the map and then Collect to list and check the size.

@fabriziofortino  how about something like 
`elasticIndexDefinition.getDefinedRules()
                    .stream().map(rule -> Optional.ofNullable(rule.getConfig(sortPropertyName)))
                    .filter(Optional::isPresent)
                    .collect(Collectors.toList()).size() > 0)`

oak-core/src/test/java/org/apache/jackrabbit/oak/query/AbstractQueryTest.java
nit0906: Leaving it in AbstractQueryTest for now, since the additional commit info doesn't really impact lucene.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
```diff
@@ -225,7 +227,14 @@ public Query baseQuery() {
                 hasTieBreaker = true;
             } else if (JCR_SCORE.equals(sortPropertyName)) {
                 fieldName = "_score";
-            } else if (indexProperties.containsKey(sortPropertyName)) {
+            } else if (indexProperties.containsKey(sortPropertyName) || elasticIndexDefinition.getDefinedRules()
+                    .stream().map(rule -> Optional.ofNullable(rule.getConfig(sortPropertyName)))
+                    .filter(Optional::isPresent)
+                    .collect(Collectors.toList()).size() > 0) {
```
fabriziofortino: this can be optimized with (no need to transform it into a list):
```suggestion
                    .stream().map(rule -> Optional.ofNullable(rule.getConfig(sortPropertyName))).anyMatch(Optional::isPresent)) {
```

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
```diff
@@ -225,7 +227,12 @@ public Query baseQuery() {
                 hasTieBreaker = true;
             } else if (JCR_SCORE.equals(sortPropertyName)) {
                 fieldName = "_score";
-            } else if (indexProperties.containsKey(sortPropertyName)) {
+            } else if (indexProperties.containsKey(sortPropertyName) || elasticIndexDefinition.getDefinedRules()
+                    .stream().map(rule -> Optional.ofNullable(rule.getConfig(sortPropertyName))).anyMatch(Optional::isPresent)) {
```
nfsantos: I think this can be simplified to: 
```
elasticIndexDefinition
    .getDefinedRules()
    .stream()
    .anyMatch(rule -> rule.getConfig(sortPropertyName) != null) 
```

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
```diff
@@ -740,27 +747,32 @@ private static Query nodeName(Filter.PropertyRestriction pr) {
...
+
         if (indexOfWS == len || indexOfWC == len) {
             // remove trailing "*" for prefix query
             first = first.substring(0, first.length() - 1);
             if (JCR_PATH.equals(name)) {
                 return newPrefixPathQuery(first);
             } else {
-                return newPrefixQuery(name, first);
+                return newPrefixQuery(field, first);
             }
```
nfsantos, line 770: FYI, this is currently unreachable code:
https://issues.apache.org/jira/browse/OAK-9881
We can handle it in another PR.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
```diff
@@ -740,27 +747,32 @@ private static Query nodeName(Filter.PropertyRestriction pr) {
...
 
         int indexOfWS = first.indexOf(WildcardQuery.WILDCARD_STRING);
         int indexOfWC = first.indexOf(WildcardQuery.WILDCARD_CHAR);
         int len = first.length();
 
+        // Non full text (Non analyzed) properties are keyword types in ES. For those field would be equal to name.
+        // Analyzed properties, however are of text type on which we can't perform wildcard or prefix queries so we use the keyword (sub) field
+        // by appending .keyword to the name here.
+        String field = elasticIndexDefinition.getElasticKeyword(name);
+
```
nfsantos, line 762: My current proposal to deal with full-text queries on non-string fields(https://github.com/apache/jackrabbit-oak/pull/649) changes the field used for full text queries: it creates a new field named `full:<pname>' just for full text queries. The current field `<pname>` would be used for non-full text queries, including queries that require keyword index. So with my PR, this logic here would have to change and likely would be simplified, as it would become something like: `if (full text query) then "full:<pname>" else <pname>`.

But currently my PR is just a proposal open for discussion, I'm not fully aware of the consequences of the change I propsosed. So I'm leaving this comment here just for information.

oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/IndexQueryCommonTest.java
```diff
@@ -167,6 +177,26 @@ public void descendantTestWithIndexTag() throws Exception {
...
 
+    @Test
+    public void descendantTestWithIndexTagExplain() throws Exception {
+        Tree test = root.getTree("/").addChild("test");
+        test.addChild("a");
+        test.addChild("b");
+        root.commit();
+
+        String query = "explain select [jcr:path] from [nt:base] where isdescendantnode('/test') option (index tag x)";
+        assertEventually(getAssertionForExplain(query, Query.JCR_SQL2, getExplainValueForDescendantTestWithIndexTagExplain(), true));
```
nfsantos, line 188: Unrelated to this PR, a general comment.
I wonder if there is a better way to check that the index was created. The way we are using assertEventually(), is testing for two conditions: 

1) Index was created - Must retry until this is true by polling.
2) the test condition passes - If it fails, the test can be aborted immediately. 

But currently we are retrying for failures in 1 (correct) and 2 (not needed), which will slow down the tests in case of failures of the test condition, as it will keep retrying a test that will not pass until reaching the retry limit. 

Maybe there is a way of separating the two conditions, with a generic polling test to check if the index was updated, separate from the condition under test? The slowdown occurs only when there are failures, but this is the common case when we are developing, so fixing this issue would be a nice quality of life improvement for us while we are in the edit-compile-test loop. And it would also make the intention of the code more clean, by separating logic from the test framework from the logic specific to the current test.

nfsantos (approved):
I have only some minor comments, probably you should not count my review for the aprovals needed to merge, as I'm still don't know the code very well. 

I have this PR which adds a few more common tests, so it's related and may even be affected by this one. It can be merged after this. Currently, it's waiting for a second review: https://github.com/apache/jackrabbit-oak/pull/644/files

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
nit0906: Yes, I was thinking of handling it here - but thought it might be too many (unrelated) changes in the same PR

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
nit0906: makes sense. Thanks

PR #659, opened 2022-08-09
OAK-9720: Update Oak 1.22 to Jackrabbit 2.20.5

reschke:


oak-upgrade/pom.xml
```diff
@@ -134,12 +139,12 @@
     <dependency>
       <groupId>org.apache.jackrabbit</groupId>
       <artifactId>jackrabbit-core</artifactId>
-      <version>${jackrabbit.version}</version>
+      <version>${jackrabbit.test.version}</version>
```
mreutegg: This and the next change are not on test dependencies. I think this is confusing.

oak-upgrade/pom.xml
```diff
@@ -31,6 +31,11 @@
   <name>Oak upgrade</name>
   <description>Tooling for upgrading Jackrabbit repositories to Oak</description>
 
+  <properties>
+    <!-- Version of Jackrabbit used for testing, see OAK-9890 -->
+    <jackrabbit.test.version>2.20.4</jackrabbit.test.version>
```
anchela: hi @reschke , not sure i get the description of OAK-9890 :-(
why does the test-version needs to be 2.20.4 when the regular dependency is 2.20.5?

would it make sense to fix the tests that cause the issue?

reschke: The issue is that you can't deploy jackrabbit-core 2.20.5 alongside oak 1.22.x, as Jackrabbit 2.20.x is on a newer version of jackrabbit-api.

The test has a jackrabbit dependency as it compares Oak's with Jackrabbit's behavior. For *that* purpose, using an older version should be fine. Or we get rid of the test, but that would be something for you, it's related to authentication.

oak-upgrade/pom.xml
reschke: It is; I'll adjust the comment and the property name.

oak-upgrade/pom.xml
anchela: @reschke , i see.... what test is it? afaik i was not involved with the upgrade from jackrabbit to oak. i believe tomek worked on that. but i can take a look of needed.

anchela (approved):

oak-upgrade/pom.xml
reschke: it's in oak-jcr. See https://issues.apache.org/jira/browse/OAK-9889

PR #660, opened 2022-08-09
OAK-9868 : Introduce interface to extract tree from users

anchela:
@joerghoh , i would appreciate if you had time to take a look at this PR

### DIFF
--- a/oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/DynamicGroupUtil.java
+++ b/oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/DynamicGroupUtil.java
@@ -22,6 +22,7 @@ import org.apache.jackrabbit.api.security.user.Group;
 import org.apache.jackrabbit.oak.api.PropertyState;
 import org.apache.jackrabbit.oak.api.Root;
 import org.apache.jackrabbit.oak.api.Tree;
+import org.apache.jackrabbit.oak.plugins.tree.TreeAware;
 import org.apache.jackrabbit.oak.plugins.tree.TreeUtil;
 import org.apache.jackrabbit.oak.spi.security.user.AuthorizableType;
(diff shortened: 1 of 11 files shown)


oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/DynamicGroupUtil.java
```diff
@@ -65,7 +66,7 @@ static boolean isMemberProperty(@NotNull PropertyState propertyState) {
 
     @NotNull
     static Tree getTree(@NotNull Authorizable authorizable, @NotNull Root root) throws RepositoryException {
-        return root.getTree(authorizable.getPath());
+        return (authorizable instanceof TreeAware) ? ((TreeAware) authorizable).getTree() : root.getTree(authorizable.getPath());
```
joerghoh, line 69: I see this expression quite often in this patch, but spread across multiple modules. Do you see a sensible location where we could centralize this code and just reference it? Just like the ```Utils.getTree()``` method below.

joerghoh (approved):

oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/DynamicGroupUtil.java
anchela: hi @joerghoh , good point. i actually was tempted to add it to `UserUtil` in _oak-security-spi_ the reason why i decided against it: the path of a User/Group is a JCR path. so to it 'right' it in a generic way the call would also need a `NamePathMapper` to convert the JCR path to an Oak path as it is required for the` Root.getTree` call. But for all usages (except maybe for one) i know that the user/group has. been retrieved using Oak API which doesn't have any namespace-remapping in place and thus decided to live with the duplication instead.

PR #662, opened 2022-08-09
OAK-9893 Prefetch for queries

thomasmueller:
Details are in https://issues.apache.org/jira/browse/OAK-9893


oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
```diff
@@ -576,6 +577,8 @@ public Oak with(@NotNull Whiteboard whiteboard) {
             newSettings.setLimitReads(this.queryEngineSettings.getLimitReads());
             this.queryEngineSettings = new AnnotatedQueryEngineSettings(newSettings);
         }
+        Feature prefetchFeature = Feature.newFeature(QueryEngineSettings.FT_NAME_PREFETCH_FOR_QUERIES, whiteboard);
+        queryEngineSettings.setPrefetchFeature(prefetchFeature);
```
nit0906: This seems to be causing an NPE in one of the tests in oak-core.

nit0906 (changes requested):
I think we should also add a higher level test where we validate the result set of a few queries in case of prefetch enabled especially a case where the result set size would be greater than the prefetch count.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
```diff
@@ -59,7 +61,9 @@ public class QueryEngineSettings implements QueryEngineSettingsMBean, QueryLimit
 
     public static final String OAK_QUERY_PREFETCH_COUNT = "oak.prefetchCount";
 
-    public static final int DEFAULT_PREFETCH_COUNT = Integer.getInteger(OAK_QUERY_PREFETCH_COUNT, 0);
+    public static final String FT_NAME_PREFETCH_FOR_QUERIES = "FT_PREFETCH_OAK-9893";
```
nit0906: I wonder if it might make sense to have all the Feature Toggles listed in a separate class (just to have a readily available list of feature toggles available in oak).

Just a thought - not a strong suggestion.

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
joerghoh: Why do you set this feature here? Wouldn't it be sufficient to centralize this handling in the ```QueryEngine``` itself and just there?

joerghoh (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
thomasmueller: > Why do you set this feature here?

Because "whiteboard" is not available in the QueryEngine.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
thomasmueller: I expect that we add more and more feature toggles in various components. It would limit modularity if we list them in a central place I assume.

thomasmueller:
> I think we should also add a higher level test where we validate the result set of a few queries in case of prefetch enabled especially a case where the result set size would be greater than the prefetch count.

@nit0906  I have added the test you described now, and a few more. The code coverage of the PrefetchCursor was nearly 90% before these tests, and is now >90%. Actually I don't think the code coverage can be increased further in a reasonable way.

nit0906:
> I have added the test you described now, and a few more. The code coverage of the PrefetchCursor was nearly 90% before these tests, and is now >90%. Actually I don't think the code coverage can be increased further in a reasonable way.

@thomasmueller  - Thanks. I was actually thinking on the lines of adding a test in oak-search where Oak might be initialised with system property to enable the prefetch feature for queries. But yes, you are right, that might not increase the line coverage as such.

I think the new added tests should also give good coverage.

nit0906 (approved):

stefan-egli (approved):
can't speak to the use of `Feature` (as I haven't played with that from downstream's point of view), but other than that lgtm

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
thomasmueller: @nit0906 NPE: you are right! It is now resolved.

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
```diff
@@ -576,6 +577,10 @@ public Oak with(@NotNull Whiteboard whiteboard) {
             newSettings.setLimitReads(this.queryEngineSettings.getLimitReads());
             this.queryEngineSettings = new AnnotatedQueryEngineSettings(newSettings);
         }
+        if (queryEngineSettings != null) {
+            Feature prefetchFeature = Feature.newFeature(QueryEngineSettings.FT_NAME_PREFETCH_FOR_QUERIES, whiteboard);
```
mreutegg, line 581: Should register feature with closer to make sure it is removed from whiteboard when the ContentRepository is closed.
```suggestion
            Feature prefetchFeature = Feature.newFeature(QueryEngineSettings.FT_NAME_PREFETCH_FOR_QUERIES, whiteboard);
            closer.register(prefetchFeature);
```

mreutegg (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
```diff
@@ -576,6 +577,11 @@ public Oak with(@NotNull Whiteboard whiteboard) {
             newSettings.setLimitReads(this.queryEngineSettings.getLimitReads());
             this.queryEngineSettings = new AnnotatedQueryEngineSettings(newSettings);
         }
+        if (queryEngineSettings != null) {
```
Joscorbe, line 580: Could this be a mistake? We have an environment where the FT gets never loaded and I think it could be because of this check. It seems there is a local variable that was initialized above:

`QueryEngineSettings queryEngineSettings = WhiteboardUtils.getService(whiteboard, QueryEngineSettings.class);`

But probably this condition should check if the field is null, instead of that local variable:

`if (this.queryEngineSettings != null) {`

Could somebody confirm this? And, in case this is expected and I'm wrong, could you help me to know what config would be needed to enable the FT in an environment?

/cc @thomasmueller @nit0906 @joerghoh @mreutegg @stefan-egli 
PD: Sorry for pinging directly, but as the PR is already closed I'm not sure if notifications will arrive otherwise.

stefan-egli: Agree, there is a path where the local variable queryEngineSettings is null but this.queryEngineSettings is not. 
But would the change then have to be
```
        if (this.queryEngineSettings != null) {
            Feature prefetchFeature = Feature.newFeature(QueryEngineSettings.FT_NAME_PREFETCH_FOR_QUERIES, whiteboard);
            this.queryEngineSettings.settings.setPrefetchFeature(prefetchFeature);
        }
``` 
?

PR #664, opened 2022-08-10
OAK-9896 - Running unit-tests in IntelliJ dos not work

jsedding:


oak-parent/pom.xml
```diff
@@ -1316,5 +1297,53 @@
         </pluginManagement>
       </build>
     </profile>
+    <profile>
+      <id>intellij-build-workaround</id>
```
kwin, line 1320: I find it a bit nasty that this workaround profile is actually containing the default configuration when it is not running in IntelliJ. Can we rather leave the default in place without any profile, and override only for IntelliJ (with this profile)?

jsedding: I agree with you in principle. The problem is that I don't know how to remove the `_plugin.scr` instruction once it's configured. Do you know of a way to achieve that?

kwin: Just look at https://maven.apache.org/guides/introduction/introduction-to-profiles.html#profile-order and https://maven.apache.org/pom.html#advanced_configuration_inheritance

jsedding: Thank you for the pointer! Please have a look at the revised change.

oak-parent/pom.xml
```diff
@@ -1316,5 +1316,37 @@
...
+      <!-- workaround for IntelliJ issue https://youtrack.jetbrains.com/issue/IDEA-141732/OSMORC-unable-to-load-bnd-plugin -->
+      <!-- IntelliJ does not add the plugin's extra dependency to the plugin's class-path, which causes the build to fail. -->
+      <!-- This means unit-tests cannot be run from within the IDE. Adding the relevant configuration only in the absence  -->
```
kwin: this comment is outdated now.

jsedding: Thank's again :) Should be ok now.

kwin (approved):
LGTM now.

PR #665, opened 2022-08-10
Oak 9897 split persistence gc

jsedding:
A `SplitPersistence` is backed by two other persistences, to which it delegates. One of these persistences is read-only.

When running compaction and cleanup on a `SplitPersistence`, no archives must be deleted from the read-only persistence. Furthermore, it also makes no sense to mark and sweep the read-only persistence.

Currently, the `FileReaper` fails and retries when it attempts to delete archives from the read-only part of the persistence, making compaction and cleanup impossible on such setups.

This PR addresses this issue by introducing the method `SegmentArchivemanager#isReadOnly(String archiveName)`, which allows the `TarReader` to skip read-only these archives during mark and sweep.

### DIFF
--- a/oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/spi/persistence/split/AzureOnTarBaseSplitPersistenceTest.java
+++ b/oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/spi/persistence/split/AzureOnTarBaseSplitPersistenceTest.java
@@ -1,0 +1,189 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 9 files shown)


jsedding:
@dulceanu I would appreciate it very much if you could find the time for a review!

jsedding:
Could someone please take a look at this PR? It has an API change. Namely it adds `SegmentArchiveManager#isReadOnly(String archiveName)` with a default implementation.

oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/spi/persistence/split/AzureOnTarBaseSplitPersistenceTest.java
```diff
@@ -0,0 +1,201 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.jackrabbit.oak.segment.spi.persistence.split;
+
+import com.microsoft.azure.storage.StorageException;
```
ierandra: please use azure sdk 12

jsedding: Good catch! I removed it.

PR #667, opened 2022-08-15
OAK-9885 - Elastic does not handle correctly escaped wildcards in like queries

nfsantos:
Fix handling of escaped characters in like queries with Elastic indexes, using the algorithm that was already being used for Lucene (which is working correctly).

### DIFF
--- a/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/util/QueryUtils.java
+++ b/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/util/QueryUtils.java
@@ -1,0 +1,36 @@
+package org.apache.jackrabbit.oak.plugins.index.search.util;
+
+public class QueryUtils {
+
+    public static final char WILDCARD_STRING = '*';
+    public static final char WILDCARD_CHAR = '?';
+    public static final char WILDCARD_ESCAPE = '\\';
+
+    private static void replaceWildcard(StringBuilder sb, int position, char oldWildcard, char newWildcard) {
+        if (sb.charAt(position) == oldWildcard) {
+            int escapeCount = 0;
+            for (int m = position - 1; m >= 0 && sb.charAt(m) == WILDCARD_ESCAPE; m--, escapeCount++) ;
+            if (escapeCount % 2 == 0) {
+                sb.setCharAt(position, newWildcard);
+            }
+        }
+    }
+
+    /**
+     * Converts an SQL2 like pattern to a Lucene/Elastic pattern for wildcard queries. In particular, this method
+     * converts any wildcard in the SQL2 like pattern (% and _) to Lucene wildcards (* and ?), handling escaped
+     * characters (as in \% and \_).
+     *
+     * @param likePattern The match pattern operator in SQL syntax
+     * @return The match pattern for Lucene/Elastic wildcard queries.
+     */
+    public static String sqlLikeToLuceneWildcardQuery(String likePattern) {
+        StringBuilder firstBuilder = new StringBuilder(likePattern);
+        for (int k = 0; k < firstBuilder.length(); k++) {
+            replaceWildcard(firstBuilder, k, '%', WILDCARD_STRING);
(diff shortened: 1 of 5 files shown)


nit0906 (approved):
lgtm - just one nitpick - if possible - probably we can move the replaceWildCard method to some common util class in oak-search to avoid code duplication ?

nfsantos:
> lgtm - just one nitpick - if possible - probably we can move the replaceWildCard method to some common util class in oak-search to avoid code duplication ?

Done, created a new class QueryUtils in oak-search.

fabriziofortino (approved):

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/util/QueryUtils.java
```diff
@@ -0,0 +1,55 @@
...
+
+package org.apache.jackrabbit.oak.plugins.index.search.util;
+
+public class QueryUtils {
+
+    public static final char WILDCARD_STRING = '*';
+    public static final char WILDCARD_CHAR = '?';
+    public static final char WILDCARD_ESCAPE = '\\';
+
+    private static void replaceWildcard(StringBuilder sb, int position, char oldWildcard, char newWildcard) {
```
thomasmueller, line 28: Even thought this is an internal method, I think adding Javadocs would make sense.
Plus maybe adding comments in the code itself.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/util/QueryUtils.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Converts an SQL2 like pattern to a Lucene/Elastic pattern for wildcard queries. In particular, this method
+     * converts any wildcard in the SQL2 like pattern (% and _) to Lucene wildcards (* and ?), handling escaped
+     * characters (as in \% and \_).
+     *
+     * @param likePattern The match pattern operator in SQL syntax
+     * @return The match pattern for Lucene/Elastic wildcard queries.
+     */
+    public static String sqlLikeToLuceneWildcardQuery(String likePattern) {
```
thomasmueller, line 46: One of the nice things about static methods is that we can easily add unit tests. Could we add unit tests for this class as well? We can likely get high code coverage in this way.

thomasmueller (commented):
I'm not satisfied with that it's already merged...

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/util/QueryUtils.java
thomasmueller: My recommendation would be:

Replace a wildcard character, unless if it's escaped.
There are multiple cases:
- it is not escaped
- it is escaped
- there are two escape characters before the wildcard, in which case the escape character is escaped and not the wildcard
- there are even more escape characters before the wildcard, in which case it depends on whether the number of escape characters is even or odd

nfsantos: The algorithm in this class was taken verbatim from LucenePropertyIndex:

https://github.com/apache/jackrabbit-oak/blob/27a7a9ffa0e78e5b258626aef24066eb58efc559/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java#L1161-L1177

This PR only updates Elastic to handle escaped wildcards in the same way as Lucene is doing and it assumes that Lucene's behavior is correct. I have moved the algorithm to this common class to avoid code duplication between Elastic and Lucene. Otherwise, I have not changed anything in the algorithm itself.

I don't see a reason to change the algorithm, unless we have a reason to believe that Lucene's behavior is incorrect. And if that is the case, probably it should be done as a separate issue.

That's also the reason I did not add any unit tests or documentation trying to explain the algorithm, as I just assumed it was correct and treated it as a black box. I did not try to understand it.

PR #673, opened 2022-08-19
OAK-9879: CacheLIRS statistics may return incorrect load count

thomasmueller:
Alternative for https://github.com/apache/jackrabbit-oak/pull/652


mreutegg (approved):
I like this better than the added synchronization proposed in #652.

fabriziofortino (approved):

PR #675, opened 2022-08-22
OAK-9908: Recovery may revert committed changes

mreutegg:
Ensure suspect documents are either up-to-date in the cache or invalidated
Enable test


mreutegg:
@Joscorbe, can you please review as well?

stefan-egli (approved):
+1 for this fix.

What I was wondering if we should have another test case. Currently it tests adding a child shortly before a crash. What about also testing removing a chlid shortly before crash (to verify how that case is affected and if it would be equally fixed by fix)?

Joscorbe (approved):
+2 to the fix.

As this invalidates the cache I would say it should work in every scenario, but I agree it would be a good idea to have more comprehensive tests. We could have the child removal proposed by @stefan-egli, and I wonder if would make sense to change a property too to see if it's affected.

mreutegg:
Thanks for the reviews. I added tests you suggested with https://github.com/apache/jackrabbit-oak/pull/675/commits/8a603d5f4e13304310aa619a56951109e373da4d

PR #682, opened 2022-09-01
Azure archive recovery for read-only Oak process

smiroslav:
https://issues.apache.org/jira/browse/OAK-9914


jsedding (commented):
As I have explained in the JIRA, I don't think that the proposed fix correctly addresses the issue. I also don't understand why two different behaviours (system property) would be desirable.

smiroslav:
@jsedding, thanks. Let's further discuss this in JIRA.

PR #686, opened 2022-09-02
OAK-9914 read-only mode for Azure segment store does not start recovery if archive is not closed

smiroslav:
https://issues.apache.org/jira/browse/OAK-9914


smiroslav:
Closing in favor of https://github.com/apache/jackrabbit-oak/pull/685/files

PR #689, opened 2022-09-05
OAK-9899 - Optimize in(x, y, z) constraints to use the terms ES query instead of a union of range queries

nfsantos:
Change the way that the `in()` property restriction in SQL2 is mapped to an Elastic query: instead of using a union of range queries, one for each value in the set, use the [`terms`](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-terms-query.html) query of Elastic.


fabriziofortino (approved):

nit0906 (approved):
lgtm

PR #690, opened 2022-09-06
OAK-9913 : added metrics for mongo document store throttling feature

rishabhdaim:


rishabhdaim:
@Joscorbe @stefan-egli @ashokpanghal could you please review the PR

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/CreateStatsConsumer.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package org.apache.jackrabbit.oak.plugins.document.util;
+
+import java.util.List;
+
+/**
+ * {@link FunctionalInterface} to consume Metric Stats for create/upsert operation
+ * @param <T>
+ * @param <U>
+ * @param <V>
```
Joscorbe: Could you add some description to those parameters? I think that would improve understandability of the interface and help in future cases whether it could be useful to implement it.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/StatsConsumer.java
```diff
@@ -0,0 +1,28 @@
...
+ * under the License.
+ */
+package org.apache.jackrabbit.oak.plugins.document.util;
+
+/**
+ * {@link FunctionalInterface} to consume Metric Stats for update/remove operation
+ * @param <T>
+ * @param <U>
+ */
+public interface StatsConsumer<T, U> {
```
Joscorbe: Same here

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStoreStats.java
```diff
@@ -115,6 +152,29 @@ public class DocumentStoreStats implements DocumentStoreStatsCollector, Document
...
+    private final MeterStats updateNodeWithThrottlingMeter;
+    private final TimerStats updateNodeWithThrottlingTimer;
+    private final MeterStats createNodeUpsertWithThrottlingMeter;
+    private final TimerStats createNodeUpsertWithThrottlingTimer;
+    private final MeterStats createNodeWithThrottlingMeter;
+    private final TimerStats createNodeWithThrottlingTimer;
+    private final MeterStats createJournalWithThrottling;
+    private final TimerStats createJournalWithThrottlingTimer;
+    private final MeterStats removeNodesWithThrottling;
+    private final TimerStats removeNodesWithThrottlingTimer;
```
mreutegg: These and a few other fields can be turned into local variables.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/CreateMetricUpdater.java
```diff
@@ -0,0 +1,99 @@
...
+        requireNonNull(journalStatsConsumer);
+
+        if (isNodesCollectionUpdated.test(collection, ids.size()) && insertSuccess) {
+            createStatsConsumer.accept(createNodeMeter, createSplitNodeMeter, createNodeTimer, ids, timeTakenNanos);
+        } else if (isJournalCollection.test(collection)) {
+            journalStatsConsumer.accept(createJournal, createJournalTimer, ids.size(), timeTakenNanos);
+        }
+    }
+
+    public static class CreateMetricUpdaterWithoutThrottling extends CreateMetricUpdater {
```
mreutegg: Why are there `CreateMetricUpdaterWithoutThrottling` and `CreateMetricUpdaterWithThrottling`? There seems to be no difference between the two, except for a different name. Similar for other updater related classes.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ModifyMetricUpdater.java
```diff
@@ -0,0 +1,116 @@
...
+                       final ObjIntConsumer<MeterStats> retryNodesConsumer,
+                       final Consumer<MeterStats> failureNodesConsumer) {
+
+        requireNonNull(isNodesCollection);
+        requireNonNull(createStatsConsumer);
+        requireNonNull(updateStatsConsumer);
+        requireNonNull(retryNodesConsumer);
+        requireNonNull(failureNodesConsumer);
+
+        if (isNodesCollection.negate().test(collection)) {return;}
```
mreutegg: Can you please reformat and put return on a new line?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/RemoveMetricUpdater.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public void update(final Collection<? extends Document> collection, final int removeCount, final long timeTakenNanos,
+                       final BiPredicate<Collection<? extends Document>, Integer> isNodesCollectionUpdated,
+                       final StatsConsumer<MeterStats, TimerStats> removeStatsConsumer) {
+
+        requireNonNull(isNodesCollectionUpdated);
+        requireNonNull(removeStatsConsumer);
+
+        if (isNodesCollectionUpdated.negate().test(collection, removeCount)) {return;}
```
mreutegg: Same as above.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/UpsertMetricUpdater.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    public void update(final Collection<? extends Document> collection, final long timeTakenNanos,
+                       final List<String> ids, final BiPredicate<Collection<? extends Document>, Integer> isNodesCollectionUpdated,
+                       final CreateStatsConsumer<MeterStats, MeterStats, TimerStats> upsertStatsConsumer) {
+
+        requireNonNull(isNodesCollectionUpdated);
+        requireNonNull(upsertStatsConsumer);
+
+        if (isNodesCollectionUpdated.negate().test(collection, ids.size())) {return;}
```
mreutegg: Same as above.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentStoreStatsTest.java
```diff
@@ -49,6 +74,10 @@ public void shutDown(){
         statsProvider.close();
         new ExecutorCloser(executor).close();
     }
+    @Before
+    public void startUp() {
+        stats = new DocumentStoreStats(statsProvider);
+    }
```
mreutegg: Why is this needed? The instance variable `stats` already gets initialized like this.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/CreateMetricUpdater.java
rishabhdaim: I have created 2 implementations for each updater (abstract) class, one with & other without throttling.

Their implementations only differ in metrics that are passed into their respective constructors.
So the one with throttling has metrics with throttling and the other without (as is the case with their respective names). Rest everything is identical.

In this way, I can reuse the code in the base updater class, while using different metrics based on whether or not throttling is going on.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ModifyMetricUpdater.java
rishabhdaim: Sure, will update the code.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentStoreStatsTest.java
rishabhdaim: I will revert this change, I agree this is not required.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/CreateMetricUpdater.java
mreutegg: > I can reuse the code in the base updater class, while using different metrics based on whether or not throttling is going on.

But you can do this with just `CreateMetricUpdater` if it wasn't abstract. I still don't see a need for the two sub classes.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/CreateStatsConsumer.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.List;
+
+/**
+ * {@link FunctionalInterface} to consume Metric Stats for create/upsert operation
+ * @param <T> the type of the first argument to the consumer
+ * @param <U> the type of the second argument to the consumer
+ * @param <V> the type of the third argument to the consumer
+ */
+public interface CreateStatsConsumer<T, U, V> {
+    void accept(T t, U u, V v, List<String> ids, long tTN);
```
mreutegg: I would not use generics here and in the other consumer interfaces, but simply define the parameter types explicitly in accept(). I think it would make it easier to read when all the parameters would have an explicit type and intended use is documented. Currently it just says first, second and third argument and it could be anything.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStoreStats.java
```diff
@@ -157,6 +203,41 @@ public DocumentStoreStats(StatisticsProvider provider) {
 
         prefetchNodes = provider.getMeter(NODES_PREFETCH, StatsOptions.DEFAULT);
         prefetchNodesTimer = provider.getTimer(NODES_PREFETCH_TIMER, StatsOptions.METRICS_ONLY);
+
+        // metrics for throttling
+        MeterStats createSplitNodeWithThrottlingMeter = provider.getMeter(NODES_CREATE_SPLIT_WITH_THROTTLING, StatsOptions.DEFAULT);
```
mreutegg: I'm not sure, but I would rather keep the operation and throttling time separate. That is, no matter if throttling is on or off, we always have a metric for the operation with the underlying DocumentStore implementation. When throttling is on and in effect, then we'd see it in a newly introduced metric. For this case here, I would name it `NODES_CREATE_SPLIT_THROTTLING`. This metric would only reflect/cover the throttling part of the operation.
Another advantage of this design is backward compatibility. With the proposed change the current metric would not get any updates as soon as throttling kicks in. It will look as if there were no operations at all while throttling was in effect.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/CreateMetricUpdater.java
rishabhdaim: I get your point, will update the PR.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/CreateStatsConsumer.java
rishabhdaim: Sure, will make the change.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStoreStats.java
rishabhdaim: Do you want me to update the names of all metrics with throttling?

mreutegg: Yes, I would change the naming.

mreutegg:
I only noticed now that Stefan also [mentioned](https://issues.apache.org/jira/browse/OAK-9913?focusedCommentId=17598459&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17598459) he would keep the two metrics separate. The existing one that gives us timings with the DocumentStore implementation and the throttling metrics.

There is also good [comment](https://issues.apache.org/jira/browse/OAK-9913?focusedCommentId=17600332&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17600332) from Jose in the JIRA issue.

This is a bit late in the review and I feel bad that I didn't point this out earlier, but wouldn't it be easier and more decoupled if there was a `ThrottlingStatsCollector`? It would be similar to `DocumentStoreStatsCollector`, but without methods for read operations. The `DocumentNodeStoreBuilder` would provide a `ThrottlingStatsCollector` that can be used in the constructor of `DocumentNodeStore` when a `ThrottlingDocumentStoreWrapper` created. The `ThrottlingStatsCollector` is passed as an argument to `ThrottlingDocumentStoreWrapper`. Whenever the `ThrottlingDocumentStoreWrapper` throttles an operation it reports it to the `ThrottlingStatsCollector`. As mentioned earlier, the existing metrics for write operations would be untouched and decoupled from the new throttling metrics.

rishabhdaim:
I think we discussed on these lines earlier to create a separate `StatsCollector` for throttling operation but It would essentially mean to duplicate the stats collection logic and its related APIs at 2 different places i.e. `DocumentStoreStatsCollector` & `ThrottlingStatsCollector`. 

The only thing we achieve from this is having 2 different `StatsCollectors` on basis of whether throttling is enabled or not.

If we look closely, then these 2 classes are essentially the same barring the method to calculate the time taken to complete the operation and the actual metrics to store the stats (both of which can be parameterized).

IMHO, the exiting APIs suit perfectly well for throttling after adding the time taken logic and throttling metrics.

stefan-egli:
Regarding the "two separate metrics" question: I think there's two different layers to this question: a code design and a metric design question. I've only had a quick look at the PR now, so my comment is only about the metric design.

So only regarding that: IIUC then the state of the PR now already does have the two metrics separate. Except that it has the throttling time representing the sum of mongo time plus throttling time. And that could be something to discuss: do we want the metric to represent the sum, or only the throttling time. I think it might be easier if it was only the throttling time (but actually, we might provide all 3 :)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
```diff
@@ -599,7 +599,7 @@ public DocumentNodeStore(DocumentNodeStoreBuilder<?> builder) {
         this.clusterId = clusterNodeInfo.getId();
 
         if (isThrottlingEnabled(builder)) {
-            s = new ThrottlingDocumentStoreWrapper(s);
+            s = new ThrottlingDocumentStoreWrapper(s, builder.getThrottlingStatsCollector());
```
mreutegg, line 602: The second parameter is annotated with `NotNull` but `builder.getThrottlingStatsCollector()` can return null.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -145,6 +145,7 @@
     private BlobStoreStats blobStoreStats;
     private CacheStats blobStoreCacheStats;
     private DocumentStoreStatsCollector documentStoreStatsCollector;
+    private ThrottlingStatsCollector throttlingStatsCollector;
```
mreutegg, line 149: I would initialize this with a NOP ThrottlingStatsCollector to ensure this is always non-null.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -526,6 +527,15 @@ public T setDocumentStoreStatsCollector(DocumentStoreStatsCollector documentStor
         return thisBuilder();
     }
 
+    public ThrottlingStatsCollector getThrottlingStatsCollector() {
```
mreutegg, line 532: Annotate with NotNull?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -526,6 +527,15 @@ public T setDocumentStoreStatsCollector(DocumentStoreStatsCollector documentStor
...
+        return throttlingStatsCollector;
+    }
+
+    public T setThrottlingStatsCollector(final ThrottlingStatsCollector throttlingStatsCollector) {
+        this.throttlingStatsCollector = throttlingStatsCollector;
+        return thisBuilder();
+    }
```
mreutegg: I would annotate the parameter with NotNull and then check this before setting `this.throttlingStatsCollector`. Together with above suggested changes, this ensures getThrottlingStatsCollector() always return a non-null value.

rishabhdaim: how about the following code change in the `getThrottlingStatsCollector`:
```@NotNull
    public ThrottlingStatsCollector getThrottlingStatsCollector() {
        if (isNull(throttlingStatsCollector)) {
            throttlingStatsCollector = new ThrottlingStatsCollectorImpl(statisticsProvider);
        }
        return throttlingStatsCollector;
    }
```

This would ensure that we always get non-null throttlingStatsCollector.
We have followed the same approach in `getDocumentStoreStatsCollector` as well.

mreutegg (approved):
Looks good to me.

PR #691, opened 2022-09-06
OAK-9898 expire PRs older than 2 years

joerghoh:
Let's start with all PRs older than 2 years; when we dealt with them (the list is long!) we can reduce it in a second step to all PRs older than 1 year.


kwin (approved):
lgtm

fabriziofortino (approved):

PR #692, opened 2022-09-07
OAK-9922 Parallel Compaction

lweitzendorf:
Parallelize compaction over subtrees

### DIFF
--- a/oak-doc/src/site/markdown/nodestore/segment/overview.md
+++ b/oak-doc/src/site/markdown/nodestore/segment/overview.md
@@ -288,6 +288,22 @@ TarMK GC #2: compacting checkpoints/5c45ca7b-5863-4679-a7c5-6056a999a6cd/root.
 TarMK GC #2: compacting root.
 ```
 
+##### <a name="how-does-compaction-make-use-of-multithreading"/> How does compaction make use of multithreading?
+
+The parallel compactor adds an initial exploration phase to the compaction process, which scans and splits the content tree
+into multiple parts to be processed simultaneously. For this to be efficient, the tree is only expanded until a certain 
+number of nodes is reached, which is defined relative to the number of threads (main thread + compaction workers).
+
+```
+TarMK GC #2: compacting with 8 threads.
+TarMK GC #2: exploring content tree to find subtrees for parallel compaction.
+TarMK GC #2: target node count for expansion is 7000, based on 7 available workers.
+TarMK GC #2: Found 3 nodes at depth 1, target is 7000.
+TarMK GC #2: Found 48 nodes at depth 2, target is 7000.
+TarMK GC #2: Found 663 nodes at depth 3, target is 7000.
+TarMK GC #2: Found 66944 nodes at depth 4, target is 7000.
+```
+
 ##### <a name="how-does-compaction-works-with-concurrent-writes"/> How does compaction work with concurrent writes?
 
 When compaction runs as part of online garbage collection, it has to work concurrently with the rest of the system.
@@ -807,24 +823,25 @@ This option is optional and is disabled by default.
 ### <a name="compact"/> Compact
 
 ```
-java -jar oak-run.jar compact [--force] [--mmap] [--compactor] SOURCE [--target-path DESTINATION] [--persistent-cache-path PERSISTENT_CACHE_PATH] [--persistent-cache-size-gb <PERSISTENT_CACHE_SIZE_GB>]
+java -jar oak-run.jar compact [--force] [--mmap] [--compactor] [--threads] SOURCE [--target-path DESTINATION] [--persistent-cache-path PERSISTENT_CACHE_PATH] [--persistent-cache-size-gb <PERSISTENT_CACHE_SIZE_GB>]
 ```
 
 The `compact` command performs offline compaction of the local/remote Segment Store at `SOURCE`. 
 `SOURCE` must be a valid path/uri to an existing Segment Store. Currently, Azure Segment Store and AWS Segment Store the supported remote Segment Stores. 
 Please refer to the [Remote Segment Stores](#remote-segment-stores) section for details on how to correctly specify connection URIs.
 
-If the optional `--force [Boolean]` argument is set to `true` the tool ignores a non 
-matching Segment Store version. *CAUTION*: this will upgrade the Segment Store to the 
+If the optional `--force [Boolean]` argument is set to `true` the tool ignores a non-matching Segment Store version. *CAUTION*: this will upgrade the Segment Store to the 
 latest version, which is incompatible with older versions. *There is no way to downgrade 
 an accidentally upgraded Segment Store*.  
 
 The optional `--mmap [Boolean]` argument can be used to control the file access mode. Set
 to `true` for memory mapped access and `false` for file access. If not specified, memory 
-mapped access is used on 64 bit systems and file access is used on 32 bit systems. On
+mapped access is used on 64-bit systems and file access is used on 32-bit systems. On
 Windows, regular file access is always enforced and this option is ignored.
 
-The optional `--compactor [String]` argument can be used to pick the compactor type to be used. Valid choices are *classic* and *diff*. While the former is slower, it might be more stable, due to lack of optimisations employed by the *diff* compactor which compacts the checkpoints on top of each other. If not specified, *diff* compactor is used.
+The optional `--compactor [String]` argument can be used to pick the compactor type to be used. Valid choices are *classic*, *diff* and *parallel*. While *classic* is slower, it might be more stable, due to lack of optimisations employed by the *diff* compactor which compacts the checkpoints on top of each other and the *parallel* compactor, which additionally divides the repository into multiple parts to process in parallel. If not specified, *parallel* compactor is used.
+
+The optional `--threads [Integer]` argument specifies the number of threads to use for compaction. This is only applicable to the *parallel* compactor. If not specified, this defaults to the number of available processors.
 
 In order to speed up offline compaction for remote Segment Stores, three new options were introduced for configuring the destination segment store where compacted archives will be written and also to configure a persistent disk cache for speeding up segments reading during compaction. All three options detailed below **apply only for remote Segment Stores**.
 
--- a/oak-run/src/main/java/org/apache/jackrabbit/oak/run/CompactCommand.java
+++ b/oak-run/src/main/java/org/apache/jackrabbit/oak/run/CompactCommand.java
@@ -18,16 +18,14 @@
 package org.apache.jackrabbit.oak.run;
 
 import java.io.File;
-
 import com.google.common.base.StandardSystemProperty;
 import joptsimple.OptionParser;
 import joptsimple.OptionSet;
 import joptsimple.OptionSpec;
 import org.apache.jackrabbit.oak.run.commons.Command;
-import org.apache.jackrabbit.oak.segment.azure.tool.AzureCompact;
-import org.apache.jackrabbit.oak.segment.azure.tool.AzureCompact.Builder;
-import org.apache.jackrabbit.oak.segment.compaction.SegmentGCOptions.CompactorType;
 import org.apache.jackrabbit.oak.segment.aws.tool.AwsCompact;
+import org.apache.jackrabbit.oak.segment.azure.tool.AzureCompact;
+import org.apache.jackrabbit.oak.segment.compaction.SegmentGCOptions.CompactorType;
 import org.apache.jackrabbit.oak.segment.tool.Compact;
 
 class CompactCommand implements Command {
@@ -56,11 +54,16 @@ class CompactCommand implements Command {
                 .withOptionalArg()
                 .ofType(Boolean.class);
         OptionSpec<String> compactor = parser.accepts("compactor",
-                "Allow the user to control compactor type to be used. Valid choices are \"classic\" and \"diff\". " +
-                        "While the former is slower, it might be more stable, due to lack of optimisations employed " +
-                        "by the \"diff\" compactor which compacts the checkpoints on top of each other. If not " +
-                        "specified, \"diff\" compactor is used.")
+                "Allow the user to control compactor type to be used. Valid choices are \"classic\", \"diff\", \"parallel\". " +
+                        "While \"classic\" is slower, it might be more stable, due to lack of optimisations employed " +
+                        "by the \"diff\" compactor which compacts the checkpoints on top of each other and \"parallel\" compactor, which splits " +
+                        "the repository into smaller parts and compacts them concurrently. If not specified, \"parallel\" compactor is used.")
                 .withRequiredArg().ofType(String.class);
+        OptionSpec<Integer> nThreads = parser.accepts("threads", "Specify the number of threads used" +
+                "for compaction. This is only applicable to the \"parallel\" compactor. Defaults to the number of available processors.")
+                .withRequiredArg()
+                .ofType(Integer.class)
+                .defaultsTo(-1);
         OptionSpec<String> targetPath = parser.accepts("target-path", "Path/URI to TAR/remote segment store where " +
                 "resulting archives will be written")
                 .withRequiredArg()
@@ -72,8 +75,9 @@ class CompactCommand implements Command {
         OptionSpec<Integer> persistentCacheSizeGb = parser.accepts("persistent-cache-size-gb", "Size in GB (defaults to 50 GB) for "
                 + "the persistent disk cache")
                 .withRequiredArg()
-                .defaultsTo("50")
-                .ofType(Integer.class);
+                .ofType(Integer.class)
+                .defaultsTo(50);
+
 
         OptionSet options = parser.parse(args);
 
@@ -85,7 +89,7 @@ class CompactCommand implements Command {
             System.exit(-1);
         }
 
-        int code = 0;
+        int code;
 
         if (path.startsWith("az:")) {
             if (targetPath.value(options) == null) {
@@ -100,45 +104,48 @@ class CompactCommand implements Command {
                 System.exit(-1);
             }
 
-            Builder azureBuilder = AzureCompact.builder()
+            AzureCompact.Builder azureBuilder = AzureCompact.builder()
                     .withPath(path)
                     .withTargetPath(targetPath.value(options))
                     .withPersistentCachePath(persistentCachePath.value(options))
                     .withPersistentCacheSizeGb(persistentCacheSizeGb.value(options))
                     .withForce(isTrue(forceArg.value(options)))
-                    .withGCLogInterval(Long.getLong("compaction-progress-log", 150000));
+                    .withGCLogInterval(Long.getLong("compaction-progress-log", 150000))
+                    .withConcurrency(nThreads.value(options));
 
             if (options.has(compactor)) {
                 azureBuilder.withCompactorType(CompactorType.fromDescription(compactor.value(options)));
             }
 
-            code = azureBuilder
-                    .build()
-                    .run();
+            code = azureBuilder.build().run();
         } else if (path.startsWith("aws:")) {
-            code = AwsCompact.builder()
+            AwsCompact.Builder awsBuilder = AwsCompact.builder()
                     .withPath(path)
                     .withForce(isTrue(forceArg.value(options)))
                     .withSegmentCacheSize(Integer.getInteger("cache", 256))
                     .withGCLogInterval(Long.getLong("compaction-progress-log", 150000))
-                    .build()
-                    .run();
+                    .withConcurrency(nThreads.value(options));
+
+            if (options.has(compactor)) {
+                awsBuilder.withCompactorType(CompactorType.fromDescription(compactor.value(options)));
+            }
+
+            code = awsBuilder.build().run();
         } else {
-            org.apache.jackrabbit.oak.segment.tool.Compact.Builder tarBuilder = Compact.builder()
+            Compact.Builder tarBuilder = Compact.builder()
                     .withPath(new File(path))
                     .withForce(isTrue(forceArg.value(options)))
                     .withMmap(mmapArg.value(options))
                     .withOs(StandardSystemProperty.OS_NAME.value())
                     .withSegmentCacheSize(Integer.getInteger("cache", 256))
-                    .withGCLogInterval(Long.getLong("compaction-progress-log", 150000));
+                    .withGCLogInterval(Long.getLong("compaction-progress-log", 150000))
+                    .withConcurrency(nThreads.value(options));
 
             if (options.has(compactor)) {
                 tarBuilder.withCompactorType(CompactorType.fromDescription(compactor.value(options)));
             }
 
-            code = tarBuilder
-                    .build()
-                    .run();
+            code = tarBuilder.build().run();
         }
 
         System.exit(code);
--- a/oak-run/src/main/java/org/apache/jackrabbit/oak/run/CompactCommand.java
+++ b/oak-run/src/main/java/org/apache/jackrabbit/oak/run/CompactCommand.java
@@ -60,10 +60,10 @@ class CompactCommand implements Command {
                         "the repository into smaller parts and compacts them concurrently. If not specified, \"parallel\" compactor is used.")
                 .withRequiredArg().ofType(String.class);
         OptionSpec<Integer> nThreads = parser.accepts("threads", "Specify the number of threads used" +
-                "for compaction. This is only applicable to the \"parallel\" compactor. Defaults to the number of available processors.")
+                "for compaction. This is only applicable to the \"parallel\" compactor. Defaults to 1.")
                 .withRequiredArg()
                 .ofType(Integer.class)
-                .defaultsTo(-1);
+                .defaultsTo(1);
         OptionSpec<String> targetPath = parser.accepts("target-path", "Path/URI to TAR/remote segment store where " +
                 "resulting archives will be written")
                 .withRequiredArg()
--- a/oak-segment-aws/src/main/java/org/apache/jackrabbit/oak/segment/aws/tool/AwsCompact.java
+++ b/oak-segment-aws/src/main/java/org/apache/jackrabbit/oak/segment/aws/tool/AwsCompact.java
@@ -75,7 +75,7 @@ public class AwsCompact {
 
         private CompactorType compactorType = CompactorType.PARALLEL_COMPACTOR;
 
-        private int concurrency = -1;
+        private int concurrency = 1;
 
         private Builder() {
             // Prevent external instantiation.
--- a/oak-segment-aws/src/main/java/org/apache/jackrabbit/oak/segment/aws/tool/AwsCompact.java
+++ b/oak-segment-aws/src/main/java/org/apache/jackrabbit/oak/segment/aws/tool/AwsCompact.java
@@ -34,6 +34,7 @@ import com.google.common.io.Files;
 
 import org.apache.jackrabbit.oak.segment.SegmentCache;
 import org.apache.jackrabbit.oak.segment.aws.tool.AwsToolUtils.SegmentStoreType;
+import org.apache.jackrabbit.oak.segment.compaction.SegmentGCOptions.CompactorType;
 import org.apache.jackrabbit.oak.segment.file.FileStore;
 import org.apache.jackrabbit.oak.segment.file.JournalReader;
 import org.apache.jackrabbit.oak.segment.spi.monitor.FileStoreMonitorAdapter;
@@ -71,6 +72,10 @@ public class AwsCompact {
         private long gcLogInterval = 150000;
 
         private int segmentCacheSize = DEFAULT_SEGMENT_CACHE_MB;
+
+        private CompactorType compactorType = CompactorType.PARALLEL_COMPACTOR;
+
+        private int concurrency = -1;
 
         private Builder() {
             // Prevent external instantiation.
@@ -129,6 +134,27 @@ public class AwsCompact {
         }
 
         /**
+         * The compactor type to be used by compaction. If not specified it defaults to
+         * "parallel" compactor
+         * @param compactorType the compactor type
+         * @return this builder
+         */
+        public Builder withCompactorType(CompactorType compactorType) {
+            this.compactorType = compactorType;
+            return this;
+        }
+
+        /**
+         * The number of threads to be used for compaction. This only applies to the "parallel" compactor
+         * @param concurrency the number of threads
+         * @return this builder
+         */
+        public Builder withConcurrency(int concurrency) {
+            this.concurrency = concurrency;
+            return this;
+        }
+
+        /**
          * Create an executable version of the {@link Compact} command.
          *
          * @return an instance of {@link Runnable}.
@@ -147,11 +173,17 @@ public class AwsCompact {
 
     private final long gcLogInterval;
 
+    private final CompactorType compactorType;
+
+    private final int concurrency;
+
     private AwsCompact(Builder builder) {
         this.path = builder.path;
         this.segmentCacheSize = builder.segmentCacheSize;
         this.strictVersionCheck = !builder.force;
         this.gcLogInterval = builder.gcLogInterval;
+        this.compactorType = builder.compactorType;
+        this.concurrency = builder.concurrency;
     }
 
     public int run() throws IOException {
@@ -173,7 +205,7 @@ public class AwsCompact {
         System.out.printf("    -> compacting\n");
 
         try (FileStore store = newFileStore(persistence, Files.createTempDir(), strictVersionCheck, segmentCacheSize,
-                gcLogInterval)) {
+                gcLogInterval, compactorType, concurrency)) {
             if (!store.compactFull()) {
                 System.out.printf("Compaction cancelled after %s.\n", printableStopwatch(watch));
                 return 1;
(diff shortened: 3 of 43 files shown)


reschke:
Please only use OAK JIRA tickets in PR titles and commit messages,

oak-doc/src/site/markdown/nodestore/segment/overview.md
```diff
@@ -807,24 +823,25 @@ This option is optional and is disabled by default.
...
 to `true` for memory mapped access and `false` for file access. If not specified, memory 
-mapped access is used on 64 bit systems and file access is used on 32 bit systems. On
+mapped access is used on 64-bit systems and file access is used on 32-bit systems. On
 Windows, regular file access is always enforced and this option is ignored.
 
-The optional `--compactor [String]` argument can be used to pick the compactor type to be used. Valid choices are *classic* and *diff*. While the former is slower, it might be more stable, due to lack of optimisations employed by the *diff* compactor which compacts the checkpoints on top of each other. If not specified, *diff* compactor is used.
+The optional `--compactor [String]` argument can be used to pick the compactor type to be used. Valid choices are *classic*, *diff* and *parallel*. While *classic* is slower, it might be more stable, due to lack of optimisations employed by the *diff* compactor which compacts the checkpoints on top of each other and the *parallel* compactor, which additionally divides the repository into multiple parts to process in parallel. If not specified, *parallel* compactor is used.
```
jelmini, line 842: It says here that `if not specified, *parallel* compactor is used`. I think it's safer to keep the previous behaviour, which is to use the `diff` compactor. 
In my opinion the new parallel compactor should be considered in preview mode at first. Once we have more experience with it in production (validating it on a limited number of customers) and we become certain of its robustness, we may make it the default compactor.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CompactCommand.java
```diff
@@ -56,11 +54,16 @@ public void execute(String... args) throws Exception {
...
                 .ofType(Boolean.class);
         OptionSpec<String> compactor = parser.accepts("compactor",
-                "Allow the user to control compactor type to be used. Valid choices are \"classic\" and \"diff\". " +
-                        "While the former is slower, it might be more stable, due to lack of optimisations employed " +
-                        "by the \"diff\" compactor which compacts the checkpoints on top of each other. If not " +
-                        "specified, \"diff\" compactor is used.")
+                "Allow the user to control compactor type to be used. Valid choices are \"classic\", \"diff\", \"parallel\". " +
+                        "While \"classic\" is slower, it might be more stable, due to lack of optimisations employed " +
+                        "by the \"diff\" compactor which compacts the checkpoints on top of each other and \"parallel\" compactor, which splits " +
+                        "the repository into smaller parts and compacts them concurrently. If not specified, \"parallel\" compactor is used.")
```
jelmini, line 60: See the other comment about keeping `diff` as default compactor.

oak-segment-aws/src/main/java/org/apache/jackrabbit/oak/segment/aws/tool/AwsCompact.java
```diff
@@ -72,6 +73,10 @@ public static class Builder {
 
         private int segmentCacheSize = DEFAULT_SEGMENT_CACHE_MB;
 
+        private CompactorType compactorType = CompactorType.PARALLEL_COMPACTOR;
```
jelmini, line 76: Should default to `CompactorType.CHECKPOINT_COMPACTOR`

oak-segment-aws/src/main/java/org/apache/jackrabbit/oak/segment/aws/tool/AwsCompact.java
```diff
@@ -128,6 +133,27 @@ public Builder withGCLogInterval(long gcLogInterval) {
             return this;
         }
 
+        /**
+         * The compactor type to be used by compaction. If not specified it defaults to
+         * "parallel" compactor
```
jelmini, line 138: Should default to `diff`.

oak-segment-aws/src/main/java/org/apache/jackrabbit/oak/segment/aws/tool/AwsToolUtils.java
```diff
@@ -69,14 +70,26 @@ public String description(String pathOrUri) {
     }
 
     public static FileStore newFileStore(SegmentNodeStorePersistence persistence, File directory,
-            boolean strictVersionCheck, int segmentCacheSize, long gcLogInterval)
-            throws IOException, InvalidFileStoreVersionException, URISyntaxException {
+                                         boolean strictVersionCheck, int segmentCacheSize, long gcLogInterval)
+            throws IOException, InvalidFileStoreVersionException {
+        return newFileStore(persistence, directory, strictVersionCheck, segmentCacheSize,
+                gcLogInterval, CompactorType.PARALLEL_COMPACTOR, 1);
```
jelmini, line 76: Should default to CompactorType.CHECKPOINT_COMPACTOR

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/tool/AzureCompact.java
```diff
@@ -79,7 +79,9 @@ public static class Builder {
 
         private int segmentCacheSize = 2048;
 
-        private CompactorType compactorType = CompactorType.CHECKPOINT_COMPACTOR;
+        private CompactorType compactorType = CompactorType.PARALLEL_COMPACTOR;
```
jelmini, line 82: Should default to CompactorType.CHECKPOINT_COMPACTOR

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/tool/AzureCompact.java
```diff
@@ -159,7 +161,7 @@ public Builder withGCLogInterval(long gcLogInterval) {
 
         /**
          * The compactor type to be used by compaction. If not specified it defaults to
-         * "diff" compactor
+         * "parallel" compactor
```
jelmini, line 164: Should default to `diff`

oak-segment-tar/pom.xml
```diff
@@ -411,8 +411,6 @@
             <groupId>org.apache.jackrabbit</groupId>
             <artifactId>oak-core</artifactId>
             <version>${project.version}</version>
-            <classifier>tests</classifier>
```
jelmini: I don't think we should bring `oak-core` as a compile-time dependency. There was an effort to depend only on `oak-core-spi` and `oak-api`, related to [this Oak issue](https://issues.apache.org/jira/browse/OAK-6073) about allowing for independent releases.
As the objective of this dependency is only to bring a single class, `ApproximateCounter`, I would consider 2 options:
* Copy `ApproximateCounter` to `oak-store-spi`, modify all code to use it instead and deprecate the one in `oak-core`. `oak-store-spi` is the module where this class has the least dependencies to other modules. Moreover, it's depended upon by both `oak-core` and `oak-segment-tar`.
* If the above is too complex, simply copy `ApproximateCounter` to `oak-segment-tar`.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/CheckpointCompactor.java
```diff
@@ -18,13 +18,15 @@
...
 
+import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.collect.Lists.newArrayList;
 import static com.google.common.collect.Maps.newHashMap;
 import static com.google.common.collect.Maps.newLinkedHashMap;
 import static org.apache.jackrabbit.oak.commons.PathUtils.elements;
 import static org.apache.jackrabbit.oak.commons.PathUtils.getName;
 import static org.apache.jackrabbit.oak.commons.PathUtils.getParentPath;
 import static org.apache.jackrabbit.oak.plugins.memory.EmptyNodeState.EMPTY_NODE;
+import static org.apache.jackrabbit.oak.segment.ClassicCompactor.getStableIdBytes;
```
jelmini: I appreciate the idea of removing duplication by depending on a single implementation of `getStableIdBytes()`. However, this method does not really belong to `ClassicCompactor`, nor any other `Compactor` implementation, and now there is an arbitrary dependency to `ClassicCompactor` here. I suggest to move this method to a new class `CompactorUtils` (under `util` package).

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/DefaultSegmentWriterBuilder.java
```diff
@@ -29,10 +30,11 @@
 import org.apache.jackrabbit.oak.segment.file.tar.GCGeneration;
 import org.apache.jackrabbit.oak.segment.memory.MemoryStore;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * Builder for building {@link DefaultSegmentWriter} instances.
- * The returned instances are thread safe if {@link #withWriterPool()}
+ * The returned instances are thread safe if {@link #withWriterPool(PoolType)}
  * was specified and <em>not</em> thread sage if {@link #withoutWriterPool()}
```
jelmini: Typo: thread sage -> thread safe

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/DefaultSegmentWriterBuilder.java
```diff
@@ -106,8 +108,8 @@ public DefaultSegmentWriterBuilder withGeneration(@NotNull GCGeneration generati
      * The returned instance is thread safe.
      */
     @NotNull
-    public DefaultSegmentWriterBuilder withWriterPool() {
-        this.pooled = true;
+    public DefaultSegmentWriterBuilder withWriterPool(PoolType writerType) {
```
jelmini, line 116: This is changing the signature of a public method in a public class. If some third-party implemented a custom store using this class, then this change could break their code.
I suggest adding an override with the previous signature, setting a default value:
```java
public DefaultSegmentWriterBuilder withWriterPool() {
        this.poolType = PoolType.GLOBAL;
        return this;
}
```

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/ParallelCompactor.java
```diff
@@ -0,0 +1,373 @@
...
+            @Nullable BlobStore blobStore,
+            @NotNull GCNodeWriteMonitor compactionMonitor,
+            int nThreads) {
+        super(gcListener, reader, writer, blobStore, compactionMonitor);
+
+        int availableProcessors = Runtime.getRuntime().availableProcessors();
+        if (nThreads < 0) {
+            nThreads += availableProcessors + 1;
+        }
+        numWorkers = Math.max(0, nThreads - 1);
```
jelmini, line 101: Not sure I understand the logic to determine the number of workers. 
First, I think the `+=` is a mistake and should actually be a simple `=`. Otherwise, assuming `availableProcessors=8`, if someone passes for example `nThreads=-4` in the constructor, then `nThreads` would be set to `-4 + 8 + 1 = 5` which would be surprising. 
Second, we seem to support `numWorkers = 0`. What's the behaviour in this case? Does it degrade to the same behaviour than a single-threaded `CheckpointCompactor`?

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/ParallelCompactor.java
```diff
@@ -0,0 +1,373 @@
...
+import org.apache.jackrabbit.oak.spi.state.NodeStateDiff;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.*;
```
jelmini: [Code Style] We should not use * imports

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/RecordCache.java
```diff
@@ -19,29 +19,21 @@
...
-import static com.google.common.base.Preconditions.checkNotNull;
-
-import java.util.LinkedHashMap;
-import java.util.Map;
-
 import com.google.common.base.Supplier;
-import com.google.common.cache.CacheStats;
-import com.google.common.cache.Weigher;
-
+import com.google.common.cache.*;
```
jelmini: [Code Style] We should not use * imports

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/AbstractCompactionStrategy.java
```diff
@@ -29,12 +29,7 @@
 
 import com.google.common.base.Function;
 
-import org.apache.jackrabbit.oak.segment.CheckpointCompactor;
-import org.apache.jackrabbit.oak.segment.ClassicCompactor;
-import org.apache.jackrabbit.oak.segment.Compactor;
-import org.apache.jackrabbit.oak.segment.RecordId;
-import org.apache.jackrabbit.oak.segment.SegmentNodeState;
-import org.apache.jackrabbit.oak.segment.SegmentWriter;
+import org.apache.jackrabbit.oak.segment.*;
```
jelmini: [Code Style] We should not use * imports

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/FileStore.java
```diff
@@ -40,13 +40,7 @@
...
 
 import org.apache.jackrabbit.oak.commons.Buffer;
-import org.apache.jackrabbit.oak.segment.RecordId;
-import org.apache.jackrabbit.oak.segment.Segment;
-import org.apache.jackrabbit.oak.segment.SegmentId;
-import org.apache.jackrabbit.oak.segment.SegmentNodeState;
-import org.apache.jackrabbit.oak.segment.SegmentNotFoundException;
-import org.apache.jackrabbit.oak.segment.SegmentNotFoundExceptionListener;
-import org.apache.jackrabbit.oak.segment.SegmentWriter;
+import org.apache.jackrabbit.oak.segment.*;
```
jelmini: [Code Style] We should not use * imports

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/FileStore.java
```diff
@@ -198,7 +192,7 @@ private static GarbageCollectionStrategy newGarbageCollectionStrategy() {
                 defaultSegmentWriterBuilder("c")
                     .with(builder.getCacheManager().withAccessTracking("COMPACT", statsProvider))
                     .withGeneration(generation)
-                    .withoutWriterPool()
+                    .withWriterPool(SegmentBufferWriterPool.PoolType.THREAD_SPECIFIC)
```
jelmini, line 202: Why not keep using `withoutWriterPool()` here?

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/PriorityCache.java
```diff
@@ -156,13 +154,40 @@ public static long nextPowerOfTwo(int size) {
...
         checkArgument(bitCount(size) == 1);
         checkArgument(rehash >= 0);
         checkArgument(rehash < 32 - numberOfTrailingZeros(size));
         this.rehash = rehash;
         entries = new Entry<?,?>[size];
         fill(entries, Entry.NULL);
         this.weigher = checkNotNull(weigher);
+
+        numSegments = Math.min(numSegments, size);
+        checkArgument((size % numSegments) == 0);
```
jelmini: I suggest to add an error message explaining why this check failed and how to fix it, ie what are acceptables values of `numSegments` in relation to `size`. 
The Javadoc for `numSegments` should also explain this.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/PriorityCache.java
```diff
@@ -218,42 +259,45 @@ public synchronized boolean put(@NotNull K key, @NotNull V value, int generation
...
-                break;
             } else if (entry.cost < cheapest) {
                 // Candidate slot, keep on searching for even cheaper slots
                 cheapest = entry.cost;
                 index = i;
                 eviction = true;
+                if (k < rehash) {
+                    continue;
+                }
+            } else {
```
jelmini: Is there a reason other than stylistic for the use of `continue` instead of `break` in this loop?
I find the previous version using `break` more readable than this one.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/PriorityCache.java
```diff
@@ -218,42 +259,45 @@ public synchronized boolean put(@NotNull K key, @NotNull V value, int generation
...
-            weight += weighEntry(newE);
+            weight.addAndGet(weighEntry(newE));
+            lockedSegment.unlock();
             return true;
-        } else {
-            loadExceptionCount++;
-            return false;
         }
+
+        checkNotNull(lockedSegment).unlock();
```
jelmini: In case of exception, `unlock()` will not be called. It's safer to call it in a `finally` clause. The `try` block should start just after `lockedSegment` is defined. The `unlock()` call at line 294 can then be removed.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/PriorityCache.java
```diff
@@ -289,35 +338,42 @@ public synchronized V get(@NotNull K key, int generation) {
...
+            for (int i = 0; i < entriesPerSegment; i++) {
+                int j = i + s * entriesPerSegment;
+                Entry<?, ?> entry = entries[j];
+                if (entry != Entry.NULL && purge.apply(entry.generation)) {
+                    entries[j] = Entry.NULL;
+                    size.decrementAndGet();
+                    weight.addAndGet(-weighEntry(entry));
+                }
             }
+            segments[s].unlock();
```
jelmini: In case of exception, `unlock()` will not be called. It's safer to call it in a `finally` clause. The `try` block should start just after `lock()` is called.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/memory/MemoryStore.java
```diff
@@ -27,17 +27,7 @@
...
-import org.apache.jackrabbit.oak.segment.Segment;
-import org.apache.jackrabbit.oak.segment.SegmentId;
-import org.apache.jackrabbit.oak.segment.SegmentIdFactory;
-import org.apache.jackrabbit.oak.segment.SegmentIdProvider;
-import org.apache.jackrabbit.oak.segment.SegmentNotFoundException;
-import org.apache.jackrabbit.oak.segment.SegmentReader;
-import org.apache.jackrabbit.oak.segment.SegmentStore;
-import org.apache.jackrabbit.oak.segment.SegmentTracker;
-import org.apache.jackrabbit.oak.segment.SegmentWriter;
+import org.apache.jackrabbit.oak.segment.*;
```
jelmini: [Code Style] We should not use * imports

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/memory/MemoryStore.java
```diff
@@ -72,7 +62,9 @@ public SegmentId newSegmentId(long msb, long lsb) {
         });
         this.revisions = new MemoryStoreRevisions();
         this.segmentReader = new CachingSegmentReader(this::getWriter, null, 16, 2, NoopStats.INSTANCE);
-        this.segmentWriter = defaultSegmentWriterBuilder("sys").withWriterPool().build(this);
+        this.segmentWriter = defaultSegmentWriterBuilder("sys")
```
jelmini: Given my other comment about keeping `withWriterPool()`, I suggest reverting this changes.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/tool/Compact.java
```diff
@@ -164,7 +166,7 @@ public Builder withGCLogInterval(long gcLogInterval) {
 
         /**
          * The compactor type to be used by compaction. If not specified it defaults to
-         * "diff" compactor
+         * "parallel" compactor
```
jelmini, line 169: Should default to `diff`.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/PriorityCache.java
```diff
@@ -265,22 +309,27 @@ public synchronized boolean put(@NotNull K key, @NotNull V value, int generation
...
+                    costs[entry.cost - Byte.MIN_VALUE].incrementAndGet();
                 }
-                hitCount++;
-                return (V) entry.value;
+                hitCount.incrementAndGet();
+                V value = (V) entry.value;
+                segment.unlock();
+                return value;
             }
+            segment.unlock();
```
jelmini: In case of exception, `unlock()` will not be called. It's safer to call it in a `finally` clause. The `try` block should start just after `lock()` is called. The call to `unlock()` at line 327 can be removed.

oak-segment-tar/src/test/java/org/apache/jackrabbit/oak/segment/AbstractCompactorExternalBlobTest.java
```diff
@@ -0,0 +1,142 @@
...
+import static org.apache.jackrabbit.oak.plugins.memory.EmptyNodeState.EMPTY_NODE;
+import static org.apache.jackrabbit.oak.segment.CompactorTestUtils.addTestContent;
+import static org.apache.jackrabbit.oak.segment.CompactorTestUtils.assertSameRecord;
+import static org.apache.jackrabbit.oak.segment.CompactorTestUtils.assertSameStableId;
+import static org.apache.jackrabbit.oak.segment.CompactorTestUtils.checkGeneration;
+import static org.apache.jackrabbit.oak.segment.CompactorTestUtils.createBlob;
+import static org.apache.jackrabbit.oak.segment.CompactorTestUtils.getCheckpoint;
+import static org.apache.jackrabbit.oak.segment.file.FileStoreBuilder.fileStoreBuilder;
+import static org.apache.jackrabbit.oak.segment.file.tar.GCGeneration.newGCGeneration;
+import static org.junit.Assert.*;
```
jelmini: [Code style] We should not use * imports.

oak-segment-tar/src/test/java/org/apache/jackrabbit/oak/segment/AbstractCompactorTest.java
```diff
@@ -0,0 +1,118 @@
...
+    public void testCompact() throws Exception {
+        addTestContent("cp1", nodeStore, 42);
+        String cp1 = nodeStore.checkpoint(DAYS.toMillis(1));
+        addTestContent("cp2", nodeStore, 42);
+        String cp2 = nodeStore.checkpoint(DAYS.toMillis(1));
+
+        SegmentNodeState uncompacted1 = fileStore.getHead();
+        SegmentNodeState compacted1 = compactor.compact(EMPTY_NODE, uncompacted1, EMPTY_NODE, Canceller.newCanceller());
+        assertNotNull(compacted1);
+        assertFalse(uncompacted1 == compacted1);
```
jelmini: Can be simplified to:
```java
assertNotSame(uncompacted1, compacted1);
```

oak-segment-tar/src/test/java/org/apache/jackrabbit/oak/segment/AbstractCompactorTest.java
```diff
@@ -0,0 +1,118 @@
...
+        // Simulate a 2nd compaction cycle
+        addTestContent("cp3", nodeStore, 42);
+        String cp3 = nodeStore.checkpoint(DAYS.toMillis(1));
+        addTestContent("cp4", nodeStore, 42);
+        String cp4 = nodeStore.checkpoint(DAYS.toMillis(1));
+
+        SegmentNodeState uncompacted2 = fileStore.getHead();
+        SegmentNodeState compacted2 = compactor.compact(uncompacted1, uncompacted2, compacted1, Canceller.newCanceller());
+        assertNotNull(compacted2);
+        assertFalse(uncompacted2 == compacted2);
```
jelmini: Can be simplified to:
```java
assertNotSame(uncompacted2, compacted2);
```

oak-segment-tar/src/test/java/org/apache/jackrabbit/oak/segment/NodeRecordTest.java
```diff
@@ -105,7 +105,7 @@ public void baseNodeStateShouldBeReusedAcrossGenerations() throws Exception {
 
             SegmentWriter writer = defaultSegmentWriterBuilder("test")
                     .withGeneration(generation)
-                    .withWriterPool()
+                    .withWriterPool(SegmentBufferWriterPool.PoolType.GLOBAL)
```
jelmini: Given my other comment about keeping `withWriterPool()`, I suggest reverting this changes.

oak-segment-tar/src/test/java/org/apache/jackrabbit/oak/segment/RecordCacheStatsTest.java
```diff
@@ -37,18 +37,9 @@ public class RecordCacheStatsTest {
     private final Random rnd = new Random();
     private final MemoryStore store = new MemoryStore();
 
-    private final RecordCache<Integer> cache = newRecordCache(KEYS);
+    private final RecordCache<Integer> cache = newRecordCache(KEYS * 4 / 3);
```
jelmini: I've noticed that if I remove your change and initialise the `RecordCache` with only KEYS (=100) values like before, then some tests fail. This means the behaviour of `RecordCache` is not preserved. This could break other code relying on the previous behaviour. 
Can you explain why the stats are not the same? Can you restore the previous behaviour?

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/FileStore.java
```diff
@@ -146,7 +140,7 @@ private static GarbageCollectionStrategy newGarbageCollectionStrategy() {
 
         this.segmentWriter = defaultSegmentWriterBuilder("sys")
                 .withGeneration(() -> getGcGeneration().nonGC())
-                .withWriterPool()
+                .withWriterPool(SegmentBufferWriterPool.PoolType.GLOBAL)
```
jelmini: Given my other comment about keeping withWriterPool(), I suggest reverting this change.

jelmini (commented):
Great contribution overall!
As you are changing/adding some critical code, I suggest to add multi-threaded tests to prove its thread-safety. 
The classes that should be tested more thoroughly for thread-safety are `PriorityCache`, `RecordCache` & `ParallelCompactor`.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/RecordCache.java
```diff
@@ -19,29 +19,21 @@
...
 import org.jetbrains.annotations.NotNull;
 
+import java.util.concurrent.atomic.AtomicLong;
+
+import static com.google.common.base.Preconditions.checkNotNull;
```
reschke: Actually, we should avoid Guava methods everywhere we can use the JDK.

oak-segment-tar/src/test/java/org/apache/jackrabbit/oak/segment/RecordCacheStatsTest.java
lweitzendorf: The reason for this change in behavior is that the Guava cache may preemptively evict items for performance reasons. This means that filling the cache exactly to capacity no longer results in an eviction count of 0. I cannot change this behavior. What I can do is keep both implementations like with the SegmentBufferWriterPool and only use the concurrent cache for compaction.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/ParallelCompactor.java
lweitzendorf: This is actually not a mistake but intended behavior. Negative numbers are interpreted relative to the number of available processors. So specifying -1 means using all available cores, -2 all but one and then for your example -4 means all but 3 of the available processors (8-3 = 5). I've seen these semantics used in libraries such as [sklearn](https://scikit-learn.org/stable/glossary.html#term-n_jobs). Granted this is Python, but I think the semantics still make sense.

Yes you are correct, it degrades to the behavior of `CheckpointCompactor` (see first comment).

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/FileStore.java
lweitzendorf: A single `SegmentBufferWriter`, which is used if there is no writer pool, is not thread-safe. A writer pool for a single thread behaves the same as a `SegmentBufferWriter` and should only have minimal overhead with the thread-specific implementation.

oak-doc/src/site/markdown/nodestore/segment/overview.md
lweitzendorf: I agree that it's risky to default to the multithreaded `ParallelCompactor`. However, I also think defaulting to a single thread should be safe. If you have a look at  the `compactWithDelegate` method, it should be easy to verify that the behavior is in fact identical to `CheckpointCompactor` in this case. I will change the call for sequential compaction to `super.compactWithDelegate(before, after, onto, canceller)` to make this even more explicit. If you still have concerns, I will change all of the defaults to diff but keep the default thread count for parallel at -1 (if the semantics of thread count discussed in the other comment are acceptable).

oak-segment-tar/src/test/java/org/apache/jackrabbit/oak/segment/RecordCacheStatsTest.java
lweitzendorf: Actually, there is another option. The Guava cache uses a `ConcurrentLinkedHashMap` internally. It would be possible to just use it manually. Downside of this approach is that I would also need to keep track of the `CacheStats` manually.

lweitzendorf: Turns out using `ConcurrentLinkedHashMap` directly is significantly slower. The option of keeping both implementations is still there if necessary.

dulceanu: I also think that keeping both implementations of the record cache is the safest option here.

lweitzendorf: Nevermind the above. I restored the previous behavior by initializing the internal cache to 4/3 of the specified size. The old implementation also did this with its `LinkedHashMap`.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/SegmentBufferWriterPool.java
```diff
@@ -19,55 +19,35 @@
 
 package org.apache.jackrabbit.oak.segment;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-import static com.google.common.base.Preconditions.checkState;
-import static com.google.common.collect.Lists.newArrayList;
-import static com.google.common.collect.Maps.newHashMap;
-import static com.google.common.collect.Sets.newHashSet;
-import static java.lang.Thread.currentThread;
+import com.google.common.base.Supplier;
```
jelmini: Can be replaced by `java.util.function.Supplier`.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/SegmentBufferWriterPool.java
```diff
@@ -82,151 +62,292 @@ public class SegmentBufferWriterPool implements WriteOperationHandler {
...
+        private final String wid;
+        @NotNull
+        private final Supplier<GCGeneration> gcGeneration;
+
+        private SegmentBufferWriterPoolFactory(
+                @NotNull SegmentIdProvider idProvider,
+                @NotNull SegmentReader reader,
+                @NotNull String wid,
+                @NotNull Supplier<GCGeneration> gcGeneration) {
+            this.idProvider = checkNotNull(idProvider);
```
jelmini: Can be replaced by `Objects.requireNonNull`.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/SegmentBufferWriterPool.java
```diff
@@ -82,151 +62,292 @@ public class SegmentBufferWriterPool implements WriteOperationHandler {
...
         }
 
-        // Call flush from outside the pool monitor to avoid potential
-        // deadlocks of that method calling SegmentStore.writeSegment
-        for (SegmentBufferWriter writer : toFlush) {
-            writer.flush(store);
+        @NotNull
+        private SegmentBufferWriter getWriter(@NotNull Thread thread, @NotNull GCGeneration gcGeneration) {
+            SimpleImmutableEntry<?,?> key = new SimpleImmutableEntry<>(thread, gcGeneration);
+            SegmentBufferWriter writer = writers.get(key);
```
jelmini: This code can be replaced with:
```java
return writers.computeIfAbsent(key, k -> newWriter(gcGeneration));
```

jelmini (approved):

PR #705, opened 2022-09-13
OAK-9932 - Allow specifying the version of the ES docker image used for tests.

nfsantos:
Add the following configuration options using system properties:

- `elasticDockerImageVersion` - version of the Elasticsearch docker image used for the unit tests. If no value is provided, uses the same version as the Elasticsearch client in build definition.     
- `elasticKnnPluginUri` - URL from where to download the Elastiknn plugin. If this property is not defined, then we try to download the plugin from the oficial GitHub repo of Elastiknn.

For instance, to run the tests with a version 8.4.1 of Elasticsearch and with a custom build of Elastiknn plugin, one could pass the following arguments to Maven:

`-DelasticDockerImageVersion=8.4.1 -DelasticKnnPluginUri=file:///Users/joe/dev/libs/elastiknn-8.4.1.0.zip`

Other changes
- Set `action.destructive_requires_name: false` in the configuration of the test Elasticsearch container. Version 8.0 change the default value of this property to true, which prevents using wilcards in the name of indexes when deleting indexes. The unit tests use wildcards to delete the indexes at the end of the tests.
- Use [elasticsearch-plugins.yml to install](https://www.elastic.co/guide/en/elasticsearch/plugins/8.4/manage-plugins-using-configuration-file.html) the Elastiknn plugin instead of using an init script to install the plugin at container startup. The init script was not working with Elasticsearch 8.0. And the declarative way is cleaner and more robust, as it doesn't require tweaking  the startup sequence of the Elasticsearch docker image.
- Send output of Elasticsearch test container to the standard logs of Oak.


fabriziofortino (approved):
I don't see any issue in skipping the digest check for nonofficial knn releases

PR #706, opened 2022-09-14
OAK-9938: Fix initialization of BlobStore in GenerateVersionInconsistencyReport

Joscorbe:


oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateVersionInconsistencyReport.java
```diff
@@ -95,6 +95,8 @@ private void generateReport(String adminUser, String adminPwd, String...args) th
         DocumentNodeStore dns = null;
         try {
             DocumentNodeStoreBuilder<?> builder = Utils.createDocumentMKBuilder(args, closer, h);
+            // Getting blobstore to trigger it to use MemoryBlobStore
+            builder.getBlobStore();
```
mreutegg: I think it would be better to explicitly set a MemoryBlobStore. But I'm also fine with this change.

mreutegg (approved):

rishabhdaim (approved):

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateVersionInconsistencyReport.java
Joscorbe: I had to do a force-push to update the branch properly, as there were many changes in trunk.

PR #713, opened 2022-09-21
OAK-9951 - Improve support for running unit tests with Elasticsearch 8

nfsantos:
Add OOTB support for testing with Elasticsearch 8.3.3, 8.4.1 and 8.4.2.  Now it is enough to pass the property -DelasticDockerImageVersion=[8.3.3|8.4.1|8.4.2] to the Maven build.

In more detail: the tests need to install the Elastiknn plugin in the test Elasticsearch container, with a matching version. The plugin is downloaded from GitHub at the start of the tests. To validate that the right plugin was downloaded, the test infrastructure compares the digest of the file with the plugin with an expected digest. This PR adds the digests for versions 8.3.3, 8.4.1 and 8.4.2 of the Elastiknn plugin.


joerghoh (approved):

joerghoh:
Where did you take these digests from?

nfsantos:
> Where did you take these digests from?

I take them from a run of the unit tests. I run the tests with a version of the Elastiknn for which I don't have the digest and copy the digest from the error message that shows that the digest did not match with a dummy digest. This is just to make sure the released version of the plugin does not change.

fabriziofortino (approved):

PR #715, opened 2022-09-21
OAK-9790 - Implement parallel indexing for speeding up oak run indexing command

amit-jain:
- Parallel indexing enabled by splitting the FlatFileStore taking into account aggregation and then running multiple threads to do the indexing
- LZ4 compression enabled

Work over https://github.com/apache/jackrabbit-oak/pull/587


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -265,6 +271,55 @@ public void reindex() throws CommitFailedException, IOException {
...
+            futureList.add(future);
+        }
+
+        try {
+            for (Future future : futureList) {
+                future.get();
+            }
+            log.info("All {} indexing jobs are done", storeList.size());
+        } catch (InterruptedException | ExecutionException e) {
+            log.error("Failure getting indexing job result", e);
```
amit-jain: @fabriziofortino You had a comment in the original PR which I forgot to address. It might be   better to throw in case of exceptions as it affects correctness. @thomasmueller wdyt?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileSplitter.java
```diff
@@ -0,0 +1,255 @@
...
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileNodeStoreBuilder.OAK_INDEXER_USE_ZIP;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileStoreUtils.createReader;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileStoreUtils.createWriter;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileStoreUtils.getSortedStoreFileName;
+
+/**
+ * This class is being used when {@value IndexerConfiguration#PROP_OAK_INDEXER_PARALLEL_INDEX} is set to true.
+ * It will split a flat file safely by checking the index definitions. An entry is considered safe to split if only
+ * none of the parent directories contains nodes in indexRule and aggregate fields of the provided index definitions.
+ */
```
nit0906, line 60: An example would be beneficial here.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileSplitter.java
```diff
@@ -0,0 +1,255 @@
...
+
+        if (deleteOriginal) {
+            LOG.info("removing original flat file {} after splitting into {} files", flatFile.getAbsolutePath(), splitFlatFiles);
+            flatFile.delete();
+        }
+
+        return splitFlatFiles;
+    }
+
+    private void updateNodeTypeStack(Stack<String> parentNodeTypeNames, String line) {
```
nit0906: There are a few util methods in this class like this one - I wonder if it might make sense to make these static to be able to easily write test cases for these ?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
```diff
@@ -151,6 +153,7 @@ private BulkProcessor initBulkProcessor() {
         return BulkProcessor.builder(requestConsumer(),
                 new OakBulkProcessorListener(), this.indexName + "-bulk-processor")
                 .setBulkActions(indexDefinition.bulkActions)
+                .setConcurrentRequests(BULK_PROCESSOR_CONCURRENCY)
```
gnaresh19, line 156: @amit-jain : Can the concurrentRequests property be moved to index definition, similar to other properties like bulkSize, which would enable the caller to configure appropriate concurrentRequests in elastic index definition.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/IndexWriterUtils.java
```diff
@@ -33,12 +35,13 @@
...
 import org.apache.lucene.analysis.miscellaneous.PerFieldAnalyzerWrapper;
 import org.apache.lucene.analysis.shingle.ShingleAnalyzerWrapper;
+import org.apache.lucene.index.ConcurrentMergeScheduler;
 import org.apache.lucene.index.IndexWriterConfig;
 import org.apache.lucene.index.SerialMergeScheduler;
 
-import static org.apache.jackrabbit.oak.plugins.index.lucene.LuceneIndexConstants.VERSION;
-
 public class IndexWriterUtils {
+    private static final int INDEX_WRITER_MAX_MERGE = 8;
```
nit0906: this should be configurable imo

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/IndexWriterUtils.java
```diff
@@ -70,6 +73,10 @@ public static IndexWriterConfig getIndexWriterConfig(LuceneIndexDefinition defin
             IndexWriterConfig config = new IndexWriterConfig(VERSION, analyzer);
             if (remoteDir) {
```
nit0906, line 74: one reason why remoteDir could be true is because copyOnWrite is disabled (which usually isn't the case).

So effectively the new code change here (the else block) will impact normal indexing as well (not just the out of the band indexing from oak-run), from what I can tell - was that taken into account ?

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/IndexWriterUtils.java
```diff
@@ -70,6 +73,10 @@ public static IndexWriterConfig getIndexWriterConfig(LuceneIndexDefinition defin
...
                 config.setMergeScheduler(new SerialMergeScheduler());
+            } else {
```
nit0906, line 76: one reason why remoteDir could be true is because copyOnWrite is disabled (which usually isn't the case), which would imply we would most probably run into the else block here even during normal indexing.

So effectively the new code change here (the else block) will impact normal indexing as well (not just the out of the band indexing from oak-run), from what I can tell - was that taken into account ?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileSplitter.java
amit-jain: Examples are available as part of the tests, not sure it makes sense to list them here.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileSplitter.java
amit-jain: Possibly we can, but these steps should mostly be covered in the FlatFileSplitterTest. Do you think we are missing tests here.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/IndexWriterUtils.java
amit-jain: I am not sure. Can you suggest what change would be advisable here to account for that. The default is 1 thread so essentially the concurrent will default to serial ofcourse if the property is not overriden

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/IndexWriterUtils.java
amit-jain: Yes, makes sense.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
amit-jain: This config does not seem suitable to be included in the IndexDefinition as it is affecting the processing in general and not per index. @nit0906 wdyt?

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/IndexWriterUtils.java
nit0906: > The default is 1 thread so essentially the concurrent will default to serial ofcourse if the property is not overriden

How's the default 1 ? The max threads are hardcoded to 8 as of now - or did I miss something ?

amit-jain: As you suggested the properties should be configurable and if that's the case then the default should be 1. But question is what can be done if in future this default changes?

nit0906: ah - ok. I think it would be difficult to determine here whether this is being called from out-of-band indexing or regular async indexing. So I think the best we can do is to set the default to 1 and add a comment that the default value should be changed only after thorough analysis.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
nit0906: @amit-jain  - agreed.

the bulkProcessor instance is actually linked to index definition and all other configurations like bulk size threshold , flush interval etc are also defined/configured at index definition level. So if we want to keep configurations in the same line - then it would seem the concurrency can also go in the index definition. 

But this is a different case where we would want to distinguish this on the basis of whether indexing is running from oak-run or otherwise - so configuring this in index definition won't be possible (since the index def would remain the same in both cases but we would want the concurrency to be different - so a system property is the only solution here.)

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
fabriziofortino: I would propagate the exception and fail the indexing job.

amit-jain:
Alternate [changes](https://github.com/amit-jain/jackrabbit-oak/commit/9d0c53b62e8ff8a10438052796522afc4d6654c0) to remove dependency on the LZ4 for other modules and only introduce it for oak-run-commons/oak-run

oak-commons/pom.xml
```diff
@@ -58,6 +58,9 @@
               org.apache.jackrabbit.oak.commons.sort,
               org.apache.jackrabbit.oak.commons.properties
             </Export-Package>
+            <Embed-Dependency>
+              lz4-java;scope=compile|runtime
+            </Embed-Dependency>
```
thomasmueller: For this change, I think we should ask others in the Oak team... @reschke , @mreutegg , @anchela

thomasmueller (commented):
This is a big change, and I think we need to be careful to not break anything. I think we will need more reviewers.

oak-commons/pom.xml
amit-jain: > This is a big change, and I think we need to be careful to not break anything. I think we will need more reviewers.

This is a PR to incorporate changes from review of https://github.com/apache/jackrabbit-oak/pull/587. So, its been reviewed by many reviewers already

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/progress/IndexingProgressReporter.java
```diff
@@ -229,10 +231,12 @@ public IndexUpdateState(String indexPath, boolean reindex, long estimatedCount)
         }
 
         public void indexUpdate() throws CommitFailedException {
-            updateCount++;
-            if (updateCount % 10000 == 0) {
-                log.info("{} => Indexed {} nodes in {} ...", indexPath, updateCount, watch);
-                watch.reset().start();
+            long count = updateCount.incrementAndGet();
+            if (count % 10000 == 0) {
+                synchronized (this) {
```
joerghoh, line 236: Is there a need to synchronize this?

amit-jain: The Stopwatch class is not thread-safe so, potentially could result in instability otherwise

amit-jain:
LZ4 support moved to issue OAK-9968 & commit https://github.com/amit-jain/jackrabbit-oak/commit/e8b4a0617fc857f281d6dec967cd48f7e5a16450

thomasmueller (approved):

nit0906 (approved):

PR #719, opened 2022-09-23
OAK-9950|Update tika version in oak

nit0906:
BDSA-2021-0824 (CVE-2021-28657)

A carefully crafted or corrupt file may trigger an infinite loop in Tika's MP3Parser up to and including Tika 1.25. Apache Tika users should upgrade to 1.26 or later.

### DIFF
--- a/oak-parent/pom.xml
+++ b/oak-parent/pom.xml
@@ -617,7 +617,7 @@
       <dependency>
         <groupId>org.apache.commons</groupId>
         <artifactId>commons-lang3</artifactId>
-        <version>3.10</version>
+        <version>3.12.0</version>
       </dependency>
       <dependency>
         <groupId>commons-io</groupId>
--- a/oak-parent/pom.xml
+++ b/oak-parent/pom.xml
@@ -61,7 +61,7 @@
     <slf4j.version>1.7.32</slf4j.version> <!-- sync with logback version -->
     <logback.version>1.2.10</logback.version>
     <h2.version>2.0.206</h2.version>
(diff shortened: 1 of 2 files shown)


nit0906:
An update to commons-lang3 is also needed to update tika to 1.26

oak-parent/pom.xml
```diff
@@ -61,7 +61,7 @@
     <slf4j.version>1.7.32</slf4j.version> <!-- sync with logback version -->
     <logback.version>1.2.10</logback.version>
     <h2.version>2.0.206</h2.version>
-    <tika.version>1.24.1</tika.version>
+    <tika.version>1.26</tika.version>
```
reschke, line 64: I believe we should upgrade to the latest 1.x release (https://mvnrepository.com/artifact/org.apache.tika/tika-core/1.28.5)

fabriziofortino: +1

nit0906: I talked to @reschke . Seems like updating to tika 1.28.5 breaks TikaExtractionOsgiIT due to a major version change in poi dependency in tika (from 4.1.2 to 5.2.2). The changes to fix this are not trivial. 

So we would first get this change in to update to 1.26 to address the vulnerability first. Post that we can update to 1.28.5 - https://issues.apache.org/jira/browse/OAK-9957 

does that sound good - @reschke  @fabriziofortino  ?

fabriziofortino (approved):

reschke (approved):

PR #723, opened 2022-09-28
Oak-9958: Deprecate native queries

tihom88:


joerghoh (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/query/SQL2Parser.java
```diff
@@ -636,6 +636,7 @@ private ConstraintImpl parseConditionFunctionIf(String functionName) throws Pars
                 }
             }
         } else if ("NATIVE".equalsIgnoreCase(functionName)) {
+            LOG.warn("native queries are deprecated. query:{}", statement);
```
nfsantos: nitpick: 
```suggestion
            LOG.warn("native queries are deprecated. query: {}", statement);
```
Is it a convention that the log messages start with lower case? And even if that is the case, maybe the second sentence part should start with upper case.

tihom88: Done

nit0906 (approved):

PR #724, opened 2022-09-29
OAK-9959 allow to disable registering MBeans for every Session

kwin:


oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/Jcr.java
```diff
@@ -96,7 +98,7 @@ public class Jcr {
 
     public Jcr(Oak oak, boolean initialize) {
         this.oak = oak;
-
+        createSessionMBeans = true; // by default every (long-living) session will register an MBean
```
mreutegg: Looks good to me, but I would rather move this initialization up to where the field is defined.

kwin: done in https://github.com/apache/jackrabbit-oak/pull/724/commits/6249fe8188724d0e0e13d2469872d8c966e398de.

mreutegg (approved):

PR #725, opened 2022-09-29
OAK-9954 : Dynamic membership/group should spot conflicts with existing groups

anchela:
- avoid shortcut if dynamic group option is enabled
- check for conflicting principal names


telegrapher (commented):
LGTM!

PR #727, opened 2022-10-04
OAK-9964: Fix queryEngineSettings check.

Joscorbe:
This is a proposal to fix what I think is a mistake in this condition check. I added a comment in the original PR [0] but probably we could discuss here in case is needed.

If I'm wrong and this code is correct we can just close this PR.

I'm adding the same reviewers of the PR [1].

[0] https://github.com/apache/jackrabbit-oak/pull/662/files#r986917815
[1] https://github.com/apache/jackrabbit-oak/pull/662

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
@@ -564,24 +564,26 @@ public class Oak {
     @NotNull
     public Oak with(@NotNull Whiteboard whiteboard) {
         this.whiteboard = checkNotNull(whiteboard);
-        QueryEngineSettings queryEngineSettings = WhiteboardUtils.getService(whiteboard, QueryEngineSettings.class);
-        if (queryEngineSettings != null) {
-            this.queryEngineSettings = new AnnotatedQueryEngineSettings(queryEngineSettings);
+        QueryEngineSettings whiteboardSettings = WhiteboardUtils.getService(whiteboard, QueryEngineSettings.class);
+        if (whiteboardSettings != null) {
+            queryEngineSettings = new AnnotatedQueryEngineSettings(whiteboardSettings);
         }
         StatisticsProvider statisticsProvider = WhiteboardUtils.getService(whiteboard, StatisticsProvider.class);
         if (statisticsProvider != null) {
             QueryEngineSettings newSettings = new QueryEngineSettings(statisticsProvider);
-            newSettings.setFullTextComparisonWithoutIndex(this.queryEngineSettings.settings.getFullTextComparisonWithoutIndex());
-            newSettings.setFailTraversal(this.queryEngineSettings.getFailTraversal());
-            newSettings.setFastQuerySize(this.queryEngineSettings.isFastQuerySize());
-            newSettings.setLimitInMemory(this.queryEngineSettings.getLimitInMemory());
-            newSettings.setLimitReads(this.queryEngineSettings.getLimitReads());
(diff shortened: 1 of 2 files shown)


thomasmueller (changes requested):
It is probably better to open a new Oak issue, because OAK-9893 is already fixed...

I see the old code is confusing as we have queryEngineSettings which can be different from this.queryEngineSettings. It's probably best to address this as well... What about renaming the old queryEngineSettings to whiteboardSettings? Then we wouldn't need "this." later on (at all).

Are you sure the change fixes the issue? (Did you test it)

Joscorbe:
I have created [OAK-9964 ](https://issues.apache.org/jira/browse/OAK-9964)

Joscorbe:
Renaming the variable seems a cleaner way to achieve the same result, yes.

We have environments where the queryEngineSettings is not null but still the FT never gets initialized (so, enabling or disabling it doesn't have any effect because it's like never existed). Analyzing the code it seems the only way this could happen is because the queryEngineSettings returned from the Whiteboard is null, even when the config is set.

Checking if the field is set (instead of the return from Whiteboard) seems the correct way to do it.

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
```diff
@@ -974,6 +975,10 @@ public void setStrictPathRestriction(String strictPathRestriction) {
             settings.setStrictPathRestriction(strictPathRestriction);
         }
 
+        public void setPrefetchFeature(@Nullable Feature prefetch) {
```
rishabhdaim, line 980: It should be `@NotNull` since `Feature.newFeature(...)` always returns a non-null value.

stefan-egli (approved):
+1 lgtm (other than what was already commentted)

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
Joscorbe: This method is just a wrapper of this one, so that's why I think it is more appropriate to maintain the same definition, as it is public and could be used from other place at some point:

https://github.com/apache/jackrabbit-oak/blob/trunk/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java#L134

Joscorbe:
I have added the following lines to the original code:

```
LOG.info("Whiteboard queryEngineSettings is: " + queryEngineSettings);
LOG.info("Field this.queryEngineSettings is: " + this.queryEngineSettings);
```

And the output I got is this one:

> 11.10.2022 15:36:57.459 *INFO* [Apache Sling Repository Startup Thread #1] org.apache.jackrabbit.oak.Oak Whiteboard queryEngineSettings is: null
> 11.10.2022 15:36:57.459 *INFO* [Apache Sling Repository Startup Thread #1] org.apache.jackrabbit.oak.Oak Field this.queryEngineSettings is: QueryEngineSettings{limitInMemory=500000, limitReads=100000, failTraversal=false, fullTextComparisonWithoutIndex=false, sql2Optimisation=true, fastQuerySize=false, classNamesIgnoredInCallTrace=[]}

So it seems the queryEngineSettings returned by Whiteboard object is null, so the feature doesn't get registered.

Joscorbe:
After testing locally I have confirmed with @thomasmueller that the fix works and the feature gets registered properly.

rishabhdaim (approved):
+1 lgtm

thomasmueller (approved):

PR #732, opened 2022-10-14
OAK-9966 : Internal code calls Node.isCheckedOut and VersionManager.isCheckedOut

anchela:
hi @joerghoh , @mreutegg , @reschke , @Joscorbe , i would highly appreciate if you could take a careful look at this PR for the version-mgt implementation.

- as outlined in the ticket the implementation calling JCR API (and by doing so resolving the node again) is suboptiomal..... so the goal was to get rid of all JCR calls that create another operation-object inside a JCR call and avoid resolving the node again.
- the check for the tree.existing in ReadOnlyVersionManager is IMHO redundant (and not relevant. i added an additional line in the comment illustrating how it could be fixed if really needed and added a test illustrating the mismatch that we had/have. i verified that the test both passed before and after my change)

please pay attention to the following subtle changes and let me know if you see any issue with that:
- ItemImpl.getVersionManager no longer calls JCR API on Workspace but the internal variant that returns the impl and does not perform a check if the session is still alive. that is handled in preconditions of the individual JCR calls)
- all internal checks for the node being checkedout should now call the internal method taking a nodedelegate without creating a separate operation object

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/ReadOnlyVersionManager.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/ReadOnlyVersionManager.java
@@ -112,18 +112,16 @@ public abstract class ReadOnlyVersionManager {
      * @return whether the tree is checked out or not.
      */
     public boolean isCheckedOut(@NotNull Tree tree) {
-        if (checkNotNull(tree).exists()) {
-            PropertyState p = tree.getProperty(VersionConstants.JCR_ISCHECKEDOUT);
-            if (p != null) {
-                return p.getValue(Type.BOOLEAN);
-            }
-        } else {
-            // FIXME: this actually means access to the tree is restricted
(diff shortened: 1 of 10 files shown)


anchela:
@mreutegg, i haven't seen `org.apache.jackrabbit.oak.plugins.document.VersionGCWithSplitTest` failing when i ran the build locally. does that sound like being related to my patch?

mreutegg:
> i haven't seen `org.apache.jackrabbit.oak.plugins.document.VersionGCWithSplitTest` failing when i ran the build locally. does that sound like being related to my patch?

The test failure is most likely unrelated.

anchela:
> The test failure is most likely unrelated.

thanks, @mreutegg

reschke (approved):
Looks reasonable. I'll trust you on the details.

joerghoh (approved):
LGTM

anchela:
@reschke ..... uuuuhhhhh thanks..... not sure it is justified :-)

PR #738, opened 2022-10-27
OAK-9977 The query recorder lists the same query many times

thomasmueller:


oak-core/src/test/java/org/apache/jackrabbit/oak/query/stats/QueryRecorderTest.java
```diff
@@ -18,12 +18,48 @@
...
+                break;
+            }
+            String s = QueryRecorder.simplify(l);
+            set.add(s);
+        }
+        for (String x : set) {
+            System.out.println(x);
+        }
+    }
+
```
fabriziofortino: is this a leftover? Should we remove it?

oak-core/src/test/java/org/apache/jackrabbit/oak/query/stats/QueryRecorderTest.java
```diff
@@ -18,12 +18,48 @@
...
+import java.io.IOException;
+import java.util.TreeSet;
+
 import org.junit.Test;
 
+import com.google.common.io.LineReader;
+
 public class QueryRecorderTest {
     
+    public static void main(String... args) throws IOException {
```
amit-jain: This looks an unintended change

thomasmueller: Well... the "/Users/mueller/temp/queryRecorder.txt" is unintended... but I'll remove the whole thing, it's better

fabriziofortino (approved):

PR #741, opened 2022-11-01
OAK-9980: Index Purging Logic fails when trying to delete :oak:mount-…

tihom88:
…libs* (read-only) node under index definition for composite nodestore

### DIFF
--- a/oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/IndexUtils.java
+++ b/oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/IndexUtils.java
@@ -18,11 +18,17 @@
  */
 package org.apache.jackrabbit.oak.composite.blueGreen;
 
-import static org.apache.jackrabbit.oak.plugins.index.IndexConstants.INDEX_DEFINITIONS_NODE_TYPE;
-import static org.apache.jackrabbit.oak.plugins.index.IndexConstants.TYPE_PROPERTY_NAME;
-
-import java.util.ArrayList;
-import java.util.Arrays;
+import org.apache.jackrabbit.oak.api.CommitFailedException;
+import org.apache.jackrabbit.oak.api.Type;
+import org.apache.jackrabbit.oak.plugins.index.IndexConstants;
+import org.apache.jackrabbit.oak.plugins.index.lucene.LuceneIndexConstants;
+import org.apache.jackrabbit.oak.plugins.index.search.FulltextIndexConstants;
+import org.apache.jackrabbit.oak.plugins.index.search.IndexDefinition;
+import org.apache.jackrabbit.oak.plugins.index.search.IndexFormatVersion;
+import org.apache.jackrabbit.oak.spi.state.NodeState;
+import org.apache.jackrabbit.oak.spi.state.NodeStateUtils;
+import org.apache.jackrabbit.oak.spi.state.NodeStore;
+import org.junit.Assert;
 
 import javax.jcr.Node;
 import javax.jcr.NodeIterator;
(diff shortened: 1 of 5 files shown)


oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/IndexUtils.java
```diff
@@ -147,4 +155,38 @@ public static void checkLibsIsReadOnly(Persistence p) throws RepositoryException
         }
     }
 
+    public static boolean isIndexDisabledAndHiddenNodesDeleted(NodeStore store, String path) throws CommitFailedException {
+        NodeState nodeState = NodeStateUtils.getNode(store.getRoot(), path);
+        if (!nodeState.getProperty("type").getValue(Type.STRING).equals("disabled")) {
```
thomasmueller, line 161: Comment: getProperty could return null, but as this is just a test, I think it's fine and we don't need to check for null. But in production code, we would need to verify. In the test case, I actually prefer shorter / simpler code, even if it's slightly dangerous. The only problem is when someone moves the code to /src/main. What about adding a comment

// could throw NPE, but we don't care as it's a test

So this won't be forgotten when the code is moved

oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/Persistence.java
```diff
@@ -138,15 +140,22 @@ public static Persistence openComposite(File globalDir, File libsDir, Config con
...
                 globalStore).addMount("libs", libsStore).
                 setPartialReadOnly(true).build();
-        result.repository = openRepsitory(cn, config.indexDir);
+        result.repository = openRepsitory(result.compositeNodestore, config.indexDir);
         result.session = result.repository.login(createCredentials());
         return result;
     }
 
+    public NodeStore getCompositeNodestore() throws Exception {
+        if (compositeNodestore == null){
```
thomasmueller: Nitpick: space before { is missing

thomasmueller (approved):

oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/IndexUtils.java
```diff
@@ -147,4 +155,38 @@ public static void checkLibsIsReadOnly(Persistence p) throws RepositoryException
         }
     }
 
+    public static boolean isIndexDisabledAndHiddenNodesDeleted(NodeStore store, String path) throws CommitFailedException {
```
fabriziofortino: `CommitFailedException` is not thrown here. Please remove it.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/Persistence.java
```diff
@@ -138,15 +140,22 @@ public static Persistence openComposite(File globalDir, File libsDir, Config con
...
+            throw new Exception("persistence object was not opened in composite mode");
```
fabriziofortino: would it be better to throw an `IllegalStateException` here?

oak-run/src/test/java/org/apache/jackrabbit/oak/indexversion/PurgeOldIndexVersionIT.java
```diff
@@ -0,0 +1,129 @@
...
+        purgeOldIndexVersionLogger.starting();
+        PurgeOldIndexVersion purgeOldIndexVersion = new PurgeOldIndexVersion();
+        Persistence p = Persistence.openComposite(globalDir, libsDir, config);
+        NodeStore n = p.getCompositeNodestore();
+        purgeOldIndexVersion.execute(n, true, 1, Arrays.asList("/oak:index"));
+
+        String purgeLog = "Found some index need to be purged over base'/oak:index/test':" +
+                " '[/oak:index/test-1 base=/oak:index/test versioned product=1 custom=0 operation:DELETE_HIDDEN_AND_DISABLE]'";
+        Assert.assertEquals(1, purgeOldIndexVersionLogger.getLogs().size());
+        Assert.assertEquals(purgeLog, purgeOldIndexVersionLogger.getLogs().get(0).toString());
```
fabriziofortino: no need to call toString here
```suggestion
        Assert.assertEquals(purgeLog, purgeOldIndexVersionLogger.getLogs().get(0));
```

fabriziofortino (changes requested):

fabriziofortino (approved):

fabriziofortino (approved):

PR #745, opened 2022-11-07
OAK-9984: FlatFileSplitterTest reliably failing

amit-jain:
- Replace file length comparisons with file content comparison

### DIFF
--- a/oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileSplitterTest.java
+++ b/oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileSplitterTest.java
@@ -561,7 +561,7 @@ public class FlatFileSplitterTest {
     
     private static Set<String> readAllFilesAsSet(List<File> files) {
         Set<String> set = files.stream().flatMap(file -> {
-            try ( InputStream is = Files.newInputStream(file.toPath())) {
+            try (InputStream is = Files.newInputStream(file.toPath())) {
                 return FileIOUtils.readStringsAsSet(is, false).stream();
             } catch (IOException e) {
                 throw new RuntimeException(e);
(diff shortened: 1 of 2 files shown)


oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileSplitterTest.java
```diff
@@ -533,4 +548,15 @@ public void uncompress(File src, File dest) throws IOException {
             }
         }
     }
+    
+    private static Set<String> readAllFilesAsSet(List<File> files) {
+        Set<String> set = files.stream().flatMap(file -> {
+            try ( InputStream is = Files.newInputStream(file.toPath())) {
```
reschke:
```suggestion
            try (InputStream is = Files.newInputStream(file.toPath())) {
```

reschke (approved):
This indeed seems to eliminate test failures.

I still see leftover files in TMPDIR, but we can handle that as a less pressing, separate issue.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileSplitterTest.java
amit-jain: > I still see leftover files in TMPDIR, but we can handle that as a less pressing, separate issue.

This should be fixed with my latest commits to
- Creating TemporaryFolder into the target directory
- Setting the system property java.tmpdir under the temporary folder

reschke: Hm. My main concern is not leftover files from tests, but unclosed files in general. These may not manifest themselves as leftover files on non-Windows systems, but would still be a problem in production.

amit-jain: Yes good point. I am checking that too.

PR #750, opened 2022-11-14
OAK-9992: Introduce limit for oak.removeCollisions()

mreutegg:


oak-run/src/main/js/oak-mongo.js
```diff
@@ -401,9 +401,10 @@ var oak = (function(global){
      * @method removeCollisions
      * @param {string} path the path of a document
      * @param {number} clusterId collision markers for this clusterId will be removed.
+     * @param {number} (optional) limit maximum number of collision markers to remove.
```
stefan-egli:
```suggestion
     * @param {number} (optional) limit maximum number of collision markers to remove, default is 1000000.
```

oak-run/src/main/js/oak-mongo.js
```diff
@@ -433,6 +437,9 @@ var oak = (function(global){
                 unset["_collisions." + r] = "";
                 num++;
             }
+            if (num >= limit) {
+                break;
```
stefan-egli, line 441: What about adding a message when the limit is hit? eg
```
                print("limit of " + limit + " collissions hit");
```

stefan-egli (approved):
+1, just added 2 more minor improvement ideas

oak-run/src/main/js/oak-mongo.js
mreutegg: Hmm, not sure this is useful. The default limit is very high an unlikely to be hit. When you give it a limit, then you know it may not clean up everything and you have to run the command again. If there is nothing to clean up then there is a message that says: "No collisions found for clusterId X".

Maybe rather a log message in any case that tells how many were removed?

stefan-egli: > Maybe rather a log message in any case that tells how many were removed?

+1, sounds like a good option

PR #751, opened 2022-11-14
OAK-9994: avoid leaking out transitive dependencies to Guava

reschke:


anchela (approved):
hi @reschke looks good to me.

PR #760, opened 2022-11-18
OAK-10006 Prevent writes to Azure storage if lock is not held

jelmini:
If the lock is lost to another instance, all writes attempts should fail with an exception.
If the lock fails to be renewed, all writes are suspended (blocked) until the renewal succeed.

### DIFF
--- a/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureRepositoryLock.java
+++ b/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureRepositoryLock.java
@@ -21,13 +21,14 @@ import com.microsoft.azure.storage.StorageErrorCodeStrings;
 import com.microsoft.azure.storage.StorageException;
 import com.microsoft.azure.storage.blob.BlobRequestOptions;
 import com.microsoft.azure.storage.blob.CloudBlockBlob;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.function.Consumer;
 import org.apache.jackrabbit.oak.segment.spi.persistence.RepositoryLock;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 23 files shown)


oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureRepositoryLock.java
```diff
@@ -121,6 +132,21 @@ private void refreshLease() {
...
+        try {
+            lockStatusChangedCallback.accept(renewal);
+        } catch (RuntimeException e) {
+            log.warn("Exception in lock status change callback", e);
+        }
+    }
+
+    void doRenewLease() throws StorageException {
+        BlobRequestOptions blobRequestOptions = new BlobRequestOptions();
+        blobRequestOptions.setMaximumExecutionTimeInMs(Math.max(INTERVAL - renewalInterval - 1, INTERVAL / 2 - 1));
```
ahanikel: Could you add a comment here that explains this?

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/FileStore.java
```diff
@@ -237,6 +237,28 @@ private static GarbageCollectionStrategy newGarbageCollectionStrategy() {
         log.debug("TAR files: {}", tarFiles);
     }
 
+    private void onLockStatusChanged(SegmentNodeStorePersistence.LockStatus lockStatus) {
+        switch (lockStatus) {
+            case LOST:
+                try {
+                    if (tarFiles != null) {
+                        tarFiles.close();
```
ahanikel, line 246: This will write the graph and references to the tar file (`archive.close()` in `TarWriter`), shouldn't we avoid that? Also it will try to flush the write queue (`q.flush()` in `AbstractRemoteSegmentArchiveWriter`), so if it's not empty, it will wait there forever, right?

jelmini: Right. Thanks for spotting this, @ahanikel !
I suppose we should add an `unsafeShutdown()` method to `TarFiles`, `TarWriter` and `AbstractRemoteSegmentArchiveWriter`, so that we can shut them down without attempting any finalisation writes to the remote storage, which would anyway fail or corrupt the data.
What do you think?

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureRepositoryLock.java
jelmini: Ok, I will add a comment.
In summary, previously `renewLease()` was called with the default `BlobRequestOptions`. However, those defaults  can be set globally for the client instance and may be not a good fit for calling `renewLease()` as it could block for too long. As the lease is acquired for 60 seconds and `renewLease()` is called by default every 30 seconds, I think it should timeout after at most 29 seconds (giving 1 sec margin), so that we can call `notifyLockStatusChange(LockStatus.RENEWAL_FAILED)` and suspend writes before the lease becomes available to be acquired by another instance. 
BTW, I have just noticed that values should be in millis and thus multiplied by 1000...

PR #763, opened 2022-11-23
OAK-10010 : Fix Flaky Test in org.apache.jackrabbit.oak.plugins.document.RandomizedClusterTest.addRemoveSet

ycliu28:
Test 
`org.apache.jackrabbit.oak.plugins.document.RandomizedClusterTest.addRemoveSet`
is flaky as mentioned in [OAK-10010](https://issues.apache.org/jira/browse/OAK-10010) and it's due to comparisons between Json Strings in the `boolean get(int maxOp, String node, String head)` method.

However, JsonObject does not guarantee entry orders, its object is an unordered set of name/value pairs.

A simple fix for this problem is to use the GSON library JsonParser to convert the string back to JSON Object and compare them.


mreutegg (approved):
Looks good to me.

PR #766, opened 2022-11-24
OAK-9971: Migrate Oak maintenance builds from Travis CI to Jenkins

mreutegg:


mreutegg:
oak-segment-tar failed with 'no space left on device'. Looks like an infra isssue.

PR #771, opened 2022-11-29
OAK-9975: [DSGC] Report cummulative size of referenced blobs during Mark phase

amit-jain:
- Introduce new metrics NUM_BLOB_REFERENCES, BLOB_REFERENCES_SIZE
- calculate and push metrics in check consistency when aggregating all the references from all repositories
- calculate and push metrics on the mark phase and checkConsistencyAfterGc as well
- update tests to assert values for new metrics

### DIFF
--- a/oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
+++ b/oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
@@ -407,7 +407,7 @@ public class MarkSweepGarbageCollector implements BlobGarbageCollector {
         iterateNodeTree(fs, false);
         
         // Get size
-        sizeBlobStoreReferences(fs, stats);
+        getBlobReferencesSize(fs, stats);
         
         // Move the marked references file to the data store meta area if applicable
         GarbageCollectionType.get(blobStore).addMarked(blobStore, fs, repoId, uniqueSuffix);
@@ -415,7 +415,7 @@ public class MarkSweepGarbageCollector implements BlobGarbageCollector {
         LOG.debug("Ending mark phase of the garbage collector");
     }
 
-    private static void sizeBlobStoreReferences(GarbageCollectorFileState fs, GarbageCollectionOperationStats stats)
+    private static void getBlobReferencesSize(GarbageCollectorFileState fs, GarbageCollectionOperationStats stats)
         throws IOException {
         try (LineIterator lineIterator = new LineIterator(new FileReader(fs.getMarkedRefs()))) {
             lineIterator.forEachRemaining(line -> {
@@ -426,12 +426,12 @@ public class MarkSweepGarbageCollector implements BlobGarbageCollector {
                 stats.getCollector().updateNumBlobReferences(1);
 
                 if (length != -1) {
-                    stats.getCollector().updateSizeBlobReferences(length);
+                    stats.getCollector().updateBlobReferencesSize(length);
                 }
             });
         }
-        LOG.info("Blob references found : {} with size : {}", stats.numBlobReferences(),
-            stats.sizeBlobReferences());
+        LOG.info("Blob references found : {} with size : {}", stats.getNumBlobReferences(),
+            stats.getBlobReferencesSize());
     }
 
     /**
@@ -548,7 +548,7 @@ public class MarkSweepGarbageCollector implements BlobGarbageCollector {
         }
 
         if (checkConsistencyAfterGc) {
-            BlobCollectionType.get(blobStore).checkConsistencyAfterGC(blobStore, fs, consistencyStats, new File(root));
+            BlobCollectionType.get(blobStore).checkConsistencyAfterGC(blobStore, fs, consistencyStats);
         }
         BlobCollectionType.get(blobStore).handleRemoves(blobStore, fs.getGarbage(), fs.getMarkedRefs());
 
@@ -759,7 +759,7 @@ public class MarkSweepGarbageCollector implements BlobGarbageCollector {
             }
 
             // Get size
-            sizeBlobStoreReferences(fs, consistencyStats);
+            getBlobReferencesSize(fs, consistencyStats);
 
             LOG.trace("Starting difference phase of the consistency check");
             FileLineDifferenceIterator iter = new FileLineDifferenceIterator(
@@ -1119,7 +1119,7 @@ public class MarkSweepGarbageCollector implements BlobGarbageCollector {
         }
 
         void checkConsistencyAfterGC(GarbageCollectableBlobStore blobStore, GarbageCollectorFileState fs,
-            GarbageCollectionOperationStats stats, File root) throws IOException {
+            GarbageCollectionOperationStats stats) throws IOException {
             stats.getCollector().start();
             Stopwatch sw = Stopwatch.createStarted();
 
@@ -1159,7 +1159,7 @@ public class MarkSweepGarbageCollector implements BlobGarbageCollector {
                 }
                 
                 // Update the size of the referenced blobs
-                sizeBlobStoreReferences(fs, stats);
+                getBlobReferencesSize(fs, stats);
             } finally {
                 sw.stop();
                 stats.getCollector().updateDuration(sw.elapsed(TimeUnit.MILLISECONDS), TimeUnit.MILLISECONDS);
@@ -1267,7 +1267,7 @@ public class MarkSweepGarbageCollector implements BlobGarbageCollector {
                 }
 
                 @Override
-                public void updateSizeBlobReferences(long size) {
+                public void updateBlobReferencesSize(long size) {
                     blobReferencesSizeCounter.inc(size);
                 }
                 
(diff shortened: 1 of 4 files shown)


oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
```diff
@@ -405,13 +405,35 @@ protected void mark(GarbageCollectorFileState fs) throws IOException, DataStoreE
...
         GarbageCollectionType.get(blobStore).addMarked(blobStore, fs, repoId, uniqueSuffix);
 
         LOG.debug("Ending mark phase of the garbage collector");
     }
 
+    private static void sizeBlobStoreReferences(GarbageCollectorFileState fs, GarbageCollectionOperationStats stats)
+        throws IOException {
+        try (LineIterator lineIterator = new LineIterator(new FileReader(fs.getMarkedRefs()))) {
+            lineIterator.forEachRemaining(line -> {
+                String id = line.split(DELIM)[0];
```
thomasmueller, line 422: I'm not sure about exception handling. If the file is corrupt (eg. empty line, corruption length,...), what would happen? It looks like it would throw an exception. Instead, a warning should be logged, and then it should continue (I assume).

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
```diff
@@ -405,13 +405,35 @@ protected void mark(GarbageCollectorFileState fs) throws IOException, DataStoreE
...
+                long length = DataStoreBlobStore.BlobId.of(id).getLength();
+                LOG.info("Blob {} has size {}", id, length);
+
+                stats.getCollector().updateNumBlobReferences(1);
```
thomasmueller, line 426: what about "addTotalBlobCount" instead of "updateNumBlobReferences"?

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
```diff
@@ -405,13 +405,35 @@ protected void mark(GarbageCollectorFileState fs) throws IOException, DataStoreE
...
+
+                if (length != -1) {
+                    stats.getCollector().updateSizeBlobReferences(length);
```
thomasmueller: what about "addTotalBlobSize"?

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
```diff
@@ -1076,8 +1119,8 @@ void handleRemoves(GarbageCollectableBlobStore blobStore, File removedIds, File
         }
 
         void checkConsistencyAfterGC(GarbageCollectableBlobStore blobStore, GarbageCollectorFileState fs,
-            OperationStatsCollector consistencyStatsCollector, File root) throws IOException {
-            consistencyStatsCollector.start();
+            GarbageCollectionOperationStats stats, File root) throws IOException {
```
fabriziofortino: `root` param is never used

fabriziofortino (approved):
LGTM

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
```diff
@@ -689,6 +711,9 @@ public long checkConsistency() throws Exception {
 
             // Mark all used blob references
             iterateNodeTree(fs, true);
+            // Move the marked references file to the data store meta area if applicable
+            String uniqueSuffix = UUID.randomUUID().toString();
+            GarbageCollectionType.get(blobStore).addMarked(blobStore, fs, repoId, uniqueSuffix);
```
thomasmueller, line 716: I'm not sure... it doesn't seem to be related to reporting the size? Not saying it is wrong... Just that I would expect it in a different PR.

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
```diff
@@ -703,6 +728,21 @@ public long checkConsistency() throws Exception {
...
+                    ((SharedDataStore) blobStore).getAllMetadataRecords(SharedStoreRecordType.REPOSITORY.getType());
+                LOG.info("Repositories registered {}", repoFiles);
+
+                // Retrieve repos for which reference files have not been created
+                Set<String> unAvailRepos =
+                    SharedDataStoreUtils.refsNotAvailableFromRepos(repoFiles, refFiles);
+                LOG.info("Repositories with unavailable references {}", unAvailRepos);
+                
+                if (!unAvailRepos.isEmpty()) {
+                    throw new NotAllRepositoryMarkedException("Not all repositories have marked references available");
```
thomasmueller, line 743: Same as above: it doesn't seem to be related to reporting the size?

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
```diff
@@ -1156,6 +1202,10 @@ class GarbageCollectionOperationStats implements OperationsStatsMBean {
         static final String NUM_BLOBS_DELETED = "NUM_BLOBS_DELETED";
         static final String TOTAL_SIZE_DELETED = "TOTAL_SIZE_DELETED";
         static final String NUM_CANDIDATES = "NUM_CANDIDATES";
+        
+        static final String NUM_BLOB_REFERENCES = "NUM_BLOB_REFERENCES";
```
thomasmueller, line 1206: I would rename the fields...

thomasmueller (changes requested):

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
amit-jain: This method uses the same file that Mark phase uses where its essential to bail out if an error. Here though not very essential but not sure if a good idea to report inaccurate size. In rare case (as havent seen that kind of corrption in practice) that happens i think its ok to throw an error.

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
amit-jain: ack

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
amit-jain: actually would have to differentiate between total blob count and total references so maybe `addTotalBlobReferencesCount`

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
amit-jain: similar maybe`addTotalBlobReferencesSize`?

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
amit-jain: Yes, the process only looked at the files generated from local, but neither the consistency nor the size is accurate if not accounting all references available (as in sweep).

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
amit-jain: Would rename as per the above approach for methods.

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
amit-jain: I pushed a few changes to the naming but left these as I see the earlier ones are also named update.

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
amit-jain: renamed to updateBlobReferencesSize, as above since all other methods have update

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
amit-jain: Ok so no update here as these are blob references. Let me know if still think its a problem

thomasmueller (approved):

fabriziofortino (approved):

PR #772, opened 2022-11-30
OAK-10011 : Configure SonarClould for Oak (fixing project key)

nit0906:
Changing Project Key for solr cloud config. Also disabling web-app IT which fails during solr cloud builds (they can be re-enabled once we fix that on infra side.)


anchela (approved):
lgtm

PR #793, opened 2022-12-15
OAK-10032 ensure that m-antrun-p execution comes after m-assembly-p

kwin:
Given them separate phases is the safest approach as according to https://issues.apache.org/jira/browse/MNG-5987: "it was documented and chosen that order of execution is not something you cannot count on: phases are ordered, mojos in a phase simply are not"


kwin:
Better approach suggested in https://github.com/apache/jackrabbit-oak/pull/794.

PR #795, opened 2022-12-16
OAK-10033:Conditions on dates use the wrong range

tihom88:


nit0906 (approved):

fabriziofortino (commented):
Can the test be moved to `PropertyIndexCommonTest` to check Elastic as well?

fabriziofortino (approved):

PR #808, opened 2022-12-20
OAK-9800: Update Maven Plugins and Apache Parent POM to version 29

reschke:
(based on @kwin's work)

### DIFF
--- a/oak-parent/pom.xml
+++ b/oak-parent/pom.xml
@@ -23,7 +23,7 @@
   <parent>
     <groupId>org.apache</groupId>
     <artifactId>apache</artifactId>
-    <version>18</version>
+    <version>27</version>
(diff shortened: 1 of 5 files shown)


oak-parent/pom.xml
```diff
@@ -276,36 +269,32 @@
           <artifactId>gmavenplus-plugin</artifactId>
           <version>1.13.1</version>
         </plugin>
+        <plugin>
```
kwin, line 272: duplicates lines 268ff

kwin (changes requested):

oak-parent/pom.xml
reschke: Indeed.

PR #809, opened 2022-12-21
OAK-10049 - Enable access to the secondary Azure blobstore service endpoint in Oak blobstore

ahanikel:


ahanikel:
@smiroslav @dulceanu Looks like I can't add any reviewers, so I'm using a good old ping.

smiroslav (approved):

reschke:
Please do not forget to track the change in JIRA.

PR #810, opened 2022-12-27
OAK-10051 attach correct artifact

kwin:
Use the one from m-assembly-p instead of the one from m-jar-p


reschke:
Is there a simple way to test this?

kwin:
Just execute `mvn install` on that module and look in the log which artifact is installed in the local repo.

reschke:
Ah, good point.

reschke (approved):
Looks good to me.

PR #812, opened 2023-01-04
OAK-10034 : created an background thread to purge unmerged branch com…

rishabhdaim:
…mits and collision markers for non active clusters

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
@@ -345,6 +345,11 @@ public final class DocumentNodeStore
     private Thread backgroundUpdateThread;
 
     /**
+     * Background thread performing purging of unmerged branch commits & collision markers
+     */
+    private Thread backgroundPurgeThread;
+
+    /**
      * Monitor object to synchronize background writes.
      */
     private final Object backgroundWriteMonitor = new Object();
@@ -729,6 +734,10 @@ public final class DocumentNodeStore
                 new BackgroundReadOperation(this, isDisposed),
                 "DocumentNodeStore background read thread " + threadNamePostfix);
         backgroundReadThread.setDaemon(true);
+        backgroundPurgeThread = new Thread(
+                new BackgroundPurgeOperation(this, isDisposed),
+                "DocumentNodeStore background purge thread " + threadNamePostfix);
+        backgroundPurgeThread.setDaemon(true);
         backgroundUpdateThread = new Thread(
                 new BackgroundUpdateOperation(this, isDisposed),
                 "DocumentNodeStore background update thread " + threadNamePostfix);
@@ -811,6 +820,7 @@ public final class DocumentNodeStore
 
             backgroundUpdateThread.start();
             backgroundSweepThread.start();
+            backgroundPurgeThread.start();
 
             if (sweep2Lock >= 0) {
                 // sweep2 is necessary - so start a sweep2 background task
@@ -857,7 +867,8 @@ public final class DocumentNodeStore
         Utils.joinQuietly(backgroundReadThread,
                 backgroundUpdateThread,
                 backgroundSweepThread,
-                backgroundSweep2Thread);
+                backgroundSweep2Thread,
+                backgroundPurgeThread);
 
         DocumentStoreException ex = null;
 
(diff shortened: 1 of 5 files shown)


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
```diff
@@ -3699,6 +3714,24 @@ private static Supplier<Integer> getDelay(DocumentNodeStore ns) {
...
+        BackgroundPurgeOperation(DocumentNodeStore nodeStore, AtomicBoolean isDisposed) {
+            // run every 60 secs
+            super(nodeStore, isDisposed, Suppliers.ofInstance(60000));
+        }
+
+        @Override
+        protected void execute(@NotNull DocumentNodeStore nodeStore) {
+            ClusterNodeInfoDocument.all(nodeStore.nonLeaseCheckingStore)
+                    .stream().filter(cluster -> !cluster.isActive())
+                    .forEach(nodeStore::purgeUnmergedBranchCommitAndCollisionMarkers);
```
mreutegg: I would like to see an additional safeguard here. It is possible that a clusterId is initially inactive and while `purgeUnmergedBranchCommitAndCollisionMarkers()` is called, the clusterId becomes active. There is a risk that the purge interferes with new commits done by the now active clusterId.
E.g. the purge method could consider a cut-off revision defined by `lastWrittenRootRev` on the `ClusterNodeInfoDocument`. The implementation would only purge revisions that are older than `lastWrittenRootRev`.

mreutegg (changes requested):
Can you also please add a test that exercises the new logic?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
```diff
@@ -2931,6 +2946,22 @@ private int backgroundSweep() throws DocumentStoreException {
...
+    }
+
+    /**
+     * Method to create @{@link java.util.function.Predicate} to filter revisions
+     *
+     * @return @{@link java.util.function.Predicate} to filter revisions older than lastWrittenRootRev
+     */
+    @NotNull
+    @VisibleForTesting
+    static java.util.function.Predicate<Revision> getOlderThanLastWrittenRootRevPredicate(final ClusterNodeInfoDocument cluster) {
```
mreutegg: I would rather move this as a non-static method to `ClusterNodeInfoDocument`. Something like `ClusterNodeInfoDocument.asOlderThanLastWrittenRootRevPredicate()`. 

WDYT?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
```diff
@@ -3704,6 +3735,30 @@ private static Supplier<Integer> getDelay(DocumentNodeStore ns) {
...
+        BackgroundPurgeOperation(DocumentNodeStore nodeStore, AtomicBoolean isDisposed) {
+            // run every 60 secs
+            super(nodeStore, isDisposed, Suppliers.ofInstance(60000));
+        }
+
+        @Override
+        protected void execute(@NotNull DocumentNodeStore nodeStore) {
+            final Clock clock = nodeStore.getClock();
+            long now = clock.getTime();
+            LOG.info("BackgroundPurgeOperation.execute: started purging for non active clusterIds");
```
mreutegg: I would not log info messages for periodic operations like these. Change to debug? `UnmergedBranches.purgeUnmergedBranchCommitAndCollisionMarkers()` will log a message at info level when something was purged.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocument.java
```diff
@@ -569,11 +569,14 @@ public boolean containsRevision(@NotNull Revision revision) {
      *     committed
      * </p>
      *
-     * @param context the revision context.
+     * @param clusterId the clusterId.
      * @param batchSize the batch size to purge uncommitted revisions
+     * @param olderThanLastWrittenRootRevPredicate @{@link java.util.function.Predicate} to filter revisions older than lastWrittenRootRev
```
mreutegg:
```suggestion
     * @param olderThanLastWrittenRootRevPredicate {@link java.util.function.Predicate} to filter revisions older than lastWrittenRootRev
```

mreutegg (approved):

PR #813, opened 2023-01-06
OAK-10054 Improved trace level logging of JCR method calls

thomasmueller:


fabriziofortino (approved):

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/ItemImpl.java
```diff
@@ -17,6 +17,7 @@
 package org.apache.jackrabbit.oak.jcr.session;
 
 import static com.google.common.collect.Lists.newArrayListWithCapacity;
+import static java.lang.String.format;
```
nfsantos, line 20: Nitpick, personal style preference: I was confused reading the code, wondering if the method call format() was part of any 3rd party library or some util class we have made. It took me some time until realizing that it was a static import. My personal preference is to avoid static imports unless it is idiomatic (like Unit tests assert methods), as I think they can make the code a bit less readable.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/SessionDelegate.java
```diff
@@ -78,6 +79,23 @@ public class SessionDelegate {
...
+    // log every 128th call by default
+    private static final long LOG_TRACE_BIT_MASK = Long.getLong(
+            "org.apache.jackrabbit.oak.jcr.operations.bitMask",
+            128L - 1);
+    // log a stack trace every ~1 million calls by default
+    private static final long LOG_TRACE_STACK_BIT_MASK = Long.getLong(
+            "org.apache.jackrabbit.oak.jcr.operations.stack.bitMask",
+            1024L * 1024 - 1);
+    // the counter used for logging
+    private static final AtomicLong LOG_COUNTER = new AtomicLong();
```
nfsantos, line 98: What will happen if the counter overflows? I guess the bit shift operations will continue giving the expected results, but just want to be sure you considered this case. Not sure how long it will take to overflow a long, might be one of those age-of-universe situations, but better be sure.

nfsantos (approved):

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/ItemImpl.java
thomasmueller: Yes, I agree! I did this only because NodeImpl.java already had that before... and I think consistency with that existing code is more important than my reluctance to use static import :-)

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/SessionDelegate.java
thomasmueller: Yes, it will continue to work. Modulo, on the other hand, behaves in a slightly weird way for negative numbers: you get negative reminders. But checking for (x % y) == 0 would still work fine.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/SessionDelegate.java
```diff
@@ -78,6 +79,23 @@ public class SessionDelegate {
...
     static final Logger writeOperationLogger = LoggerFactory.getLogger("org.apache.jackrabbit.oak.jcr.operations.writes");
 
+    // the bitmask used for trace level logging
+    // we use a bitmask instead of a counter to avoid the slow modulo operation:
+    // https://stackoverflow.com/questions/27977834/why-is-modulus-operator-slow
+    // so we use "if ((counter & LOG_TRACE_BIT_MASK) == 0) log(...)"
+    // instead of the slower "if ((counter % LOG_TRACE) == 0) log(...)"
+    // that means the values need to be some power of two, minus one
+    // log every 128th call by default
+    private static final long LOG_TRACE_BIT_MASK = Long.getLong(
```
reschke: please use https://www.javadoc.io/static/org.apache.jackrabbit/oak-commons/1.46.0/org/apache/jackrabbit/oak/commons/properties/SystemPropertySupplier.html

reschke (changes requested):

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/ItemImpl.java
reschke: maybe remove it in NodeImpl then :-)

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/SessionDelegate.java
thomasmueller: Oh, I wasn't aware this is used in the main branch (I new about usage in the test branch). I will use it!

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/ItemImpl.java
thomasmueller: Removing it in NodeImpl would result in a 225 lines diff :-) So probably easier to keep it.

reschke: Oops :-)

reschke (changes requested):

reschke (approved):

stefan-egli (approved):

PR #819, opened 2023-01-11
OAK-10061 : WARN when for an external group a local group with the same name is already present

anchela:
hi @joerghoh , i would appreciate if you could review the suggested improvement wrt log output.


joerghoh (approved):
+1

PR #823, opened 2023-01-13
OAK-10001: Bump up minimal Java version to 11

reschke:


oak-parent/pom.xml
```diff
@@ -111,25 +110,17 @@
       <plugins>
         <plugin>
           <artifactId>maven-compiler-plugin</artifactId>
+          <version>3.8.1</version>
           <configuration>
+            <target>${java.version}</target>
+            <source>${java.version}</source>
+            <release>${java.version}</release>
```
kwin: Rather overwrite `maven.compiler.target` (from ASF parent: https://github.com/apache/maven-apache-parent/blob/5faa5c33e3035e5c83af1ed40b4abf6b131a1149/pom.xml#L95) and set `maven.compiler.release` as well to the same value, as that also affects m-javadoc-p, ...

kwin: Also you should overwrite `java.minimalBuildVersion` (https://github.com/apache/maven-apache-parent/blob/5faa5c33e3035e5c83af1ed40b4abf6b131a1149/pom.xml#L93)

reschke: Please add to the PR...

oak-parent/pom.xml
```diff
@@ -69,8 +68,11 @@
     <testcontainers.version>1.17.5</testcontainers.version>
     <pax-exam.version>4.13.1</pax-exam.version>
     <groovy.version>2.5.20</groovy.version>
-    <java.version>1.8</java.version>
-    <java.version.signature>java18</java.version.signature>
+
+    <java.version>11</java.version>
+    <maven.compiler.release>${java.version}</maven.compiler.release>
+    <maven.compiler.source>${java.version}</maven.compiler.source>
```
kwin: redundant (set in asf parent)

oak-parent/pom.xml
```diff
@@ -111,25 +113,14 @@
       <plugins>
         <plugin>
           <artifactId>maven-compiler-plugin</artifactId>
+          <version>3.8.1</version>
```
kwin: why downgrading the newer version set in ASF parent?

kwin (changes requested):

oak-parent/pom.xml
```diff
@@ -69,8 +68,11 @@
     <testcontainers.version>1.17.5</testcontainers.version>
     <pax-exam.version>4.13.1</pax-exam.version>
     <groovy.version>2.5.20</groovy.version>
-    <java.version>1.8</java.version>
-    <java.version.signature>java18</java.version.signature>
+
+    <java.version>11</java.version>
```
kwin, line 72:
```suggestion
    <java.version>11</java.version>
    <java.minimalBuildVersion>${java.version}</java.minimalBuildVersion>
```

oak-parent/pom.xml
reschke: which part exactly is redundant?

kwin: https://github.com/apache/maven-apache-parent/blob/5faa5c33e3035e5c83af1ed40b4abf6b131a1149/pom.xml#L94

oak-parent/pom.xml
reschke: no idea how that got there; maybe a merge issue

oak-parent/pom.xml
kwin: Just remove line 74

kwin (approved):

anchela (approved):
@reschke , looks good to me at a first glance. but i have to admit that i can't tell what the impact of the removal of the animal-sniffer plugins is.

reschke:
I believe this was needed to check that when targeting Java X, we must ensure not to use APIs from X + 1. AFAIU, this is now taken care by javac itself. @kwin - is that right?

kwin:
Right, look at https://openjdk.org/jeps/247 for details. The release parameter triggers the check.

PR #826, opened 2023-01-18
OAK-10048: DocumentStoreIndexerBase#buildFlatFileStore outputs the wrong path when FlatFileSplitter used

amit-jain:


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -222,26 +235,53 @@ public List<FlatFileStore> buildList(IndexHelper indexHelper, IndexerSupport ind
...
+            List<File> files = getFiles(sortedFilePath);
+            if (files != null) {
+                return files;
+            }
+        } 
+        
+        // Initialize the flat file store again
+        
+        createStoreDir();
+        org.apache.jackrabbit.oak.index.indexer.document.flatfile.SortStrategy strategy = createSortStrategy(flatFileStoreDir);
```
thomasmueller: Couldn't we import the class?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -222,26 +235,53 @@ public List<FlatFileStore> buildList(IndexHelper indexHelper, IndexerSupport ind
...
+        entryCount = strategy.getEntryCount();
+        return Collections.singletonList(result);
+    }
+
+    @Nullable
+    private List<File> getFiles(String sortedFilePath) {
+        File sortedDir = new File(sortedFilePath);
+        if (sortedDir.exists() && sortedDir.canRead() && sortedDir.isDirectory()) {
+            log.info("Reading from provided sorted files directory [{}] (via system property '{}')",
+                sortedDir.getAbsolutePath(), OAK_INDEXER_SORTED_FILE_PATH);
```
thomasmueller: The log message might be a bit misleading... here you write "(via system property x)" but in fact the caller used that system property (createdSortedStoreFiles()), not this class... What about logging in the caller about the system property value, and here just "Reading from provided sorted files directory [{}]"?

This will log twice, but I think that's fine: it is more explicit on what it is doing.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -222,26 +235,53 @@ public List<FlatFileStore> buildList(IndexHelper indexHelper, IndexerSupport ind
...
+     * If the system property OAK_INDEXER_SORTED_FILE_PATH in undefined, or it cannot read relevant files it 
+     * initializes the flat file store.
+     * 
+     * @return list of flat files
+     * @throws IOException 
+     * @throws CompositeException
+     */
+    private List<File> createdSortedStoreFiles() throws IOException, CompositeException {
+        // Check system property defined path
         String sortedFilePath = System.getProperty(OAK_INDEXER_SORTED_FILE_PATH);
```
thomasmueller, line 250: (I have to admit I don't understand why we use a system property, and not a parameter... but this is not the fault of this patch, so let's keep that for now.)

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -31,13 +32,16 @@
 import org.apache.jackrabbit.oak.query.NodeStateNodeTypeInfoProvider;
 import org.apache.jackrabbit.oak.spi.blob.BlobStore;
 import org.apache.jackrabbit.oak.spi.state.NodeStore;
+import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
+import java.io.FilenameFilter;
```
fabriziofortino: not used

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileSplitter.java
```diff
@@ -73,7 +74,9 @@ public class FlatFileSplitter {
     private Set<String> splitNodeTypeNames;
     private boolean useCompression = Boolean.parseBoolean(System.getProperty(OAK_INDEXER_USE_ZIP, "true"));
     private boolean useLZ4 = Boolean.parseBoolean(System.getProperty(OAK_INDEXER_USE_LZ4, "false"));
-
+    
+    static Predicate IS_SPLIT = (Predicate<File>) path -> path.getParent().endsWith(SPLIT_DIR_NAME);
```
fabriziofortino: I would change this with the following
```suggestion
    static Predicate<File> IS_SPLIT = path -> path.getParent().endsWith(SPLIT_DIR_NAME);
```

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilderTest.java
```diff
@@ -60,60 +87,162 @@ public void defaultSortStrategy() throws Exception {
...
+        when(indexerSupport.retrieveNodeStateForCheckpoint()).thenReturn(rootState);
+        
+        List<FlatFileStore> storeList = builder.buildList(indexHelper, indexerSupport, mockIndexDefns());
+        
+        if (split) {
+            assertEquals(new File(dir, "split").getAbsolutePath(), storeList.get(0).getFlatFileStorePath());
+            assertTrue(storeList.size() > 1);
+        } else {
+            assertEquals(dir, storeList.get(0).getFlatFileStorePath());
+            assertTrue(storeList.size() == 1);
```
fabriziofortino: it can be simplified
```suggestion
            assertEquals(1, storeList.size());
```

fabriziofortino (commented):
a few minors

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
amit-jain: The caller (createdSortedStoreFiles()) as well as the definition of the system property is in the same class.

amit-jain: I moved the usage of OAK_INDEXER_SORTED_FILE_PATH to the caller.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
reschke: FWFW, if it needs to be a system property, it should use https://jackrabbit.apache.org/oak/docs/apidocs/org/apache/jackrabbit/oak/commons/properties/SystemPropertySupplier.html

thomasmueller: SystemPropertySupplier is nice because it logs if the value is set (to a non-default value). However in this case I think not using it is fine.

thomasmueller (approved):

fabriziofortino (approved):

PR #827, opened 2023-01-20
OAK-10076 - Bump netty dependency from 4.1.68.Final to 4.1.86.Final

dulceanu:


reschke:
I have reverted the change for now as I'm in the process of cutting a release.

PR #831, opened 2023-01-23
OAK-10081: (tests) create temp folders in .target to avoid leftovers …

fabriziofortino:
…on some platforms


reschke (approved):
Let's give it a try!

amit-jain:
The changes in https://github.com/apache/jackrabbit-oak/pull/832 for this issue. But it's a good idea to create temp files in the TemporaryFolder itself which the test class does in a few tests.

PR #835, opened 2023-01-26
OAK-10087: TypeEditor: constraint exception for invalid child node should contain that node's effective type

reschke:


anchela (approved):
lgtm, thanks a lot.

PR #838, opened 2023-01-31
OAK-10095 | Fix NPE in FieldFactory.dateToLong

nit0906:
Throw a RuntimeException with a proper message in case ISO8601.parse in FieldFactory.dateToLong returns null. This prevents an NPE which was confusing.

We still throw an exception because that is what is expected from the caller - just throwing one with a proper detailed message now.

### DIFF
--- a/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/FieldFactory.java
+++ b/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/FieldFactory.java
@@ -19,6 +19,7 @@ package org.apache.jackrabbit.oak.plugins.index.lucene;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Calendar;
 import java.util.Collection;
 
 import com.google.common.primitives.Ints;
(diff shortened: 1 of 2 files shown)


oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexEditorTest.java
```diff
@@ -233,6 +233,16 @@ public void testLuceneWithNonFullText() throws Exception {
         //Date
         assertEquals("/test", getPath(NumericRangeQuery.newLongRange("creationTime",
                 dateToTime("05/05/2014"), dateToTime("05/07/2014"), true, true)));
+
+        // Call FieldFactory.dateToLong with an unsupported Date format - this should throw a RuntimeException
+        try {
+            getPath(NumericRangeQuery.newLongRange("creationTime",
+                    FieldFactory.dateToLong("05/05/2014"), FieldFactory.dateToLong("05/07/2014"), true, true));
+        } catch (Exception e) {
```
nfsantos: The date parsing method is throwing a `RuntimeException`, while the test is catching the more generic `Exception`. The testing code should match the exact exception.

I was also looking if there is any builtin function to assert that a method throws an exception. JUnit 5 has the `assertThrows`, but I think JUnit 4 doesn't have that method, it only has a parameter for the Test annotation, which I'm not sure if it allows testing for the message inside: `@Test(expected=IndexOutOfBoundsException.class)`

nfsantos (approved):

fabriziofortino (approved):
+1 but I think in the long run we should replace `ISO8601.parse()` with something better (swallowing the exceptions returning `null` is not great).

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/FieldFactory.java
```diff
@@ -186,7 +187,17 @@ public static Long dateToLong(String date){
...
-        return ISO8601.parse(date).getTimeInMillis();
+        Calendar c = ISO8601.parse(date);
+        if (c != null) {
+            return c.getTimeInMillis();
+        } else {
+            // ISO8601.parse returns null in case of multiple exceptions like IllegalFormatException, IndexOutOfBoundsException, NumberFormatException etc
+            // However returning null for us would basically store a null value in the document (which seems wrong).
+            // So throwing an unchecked exception here with a proper description.
+            // Earlier such a situation was leading to an NPE which was confusing to understand.
+            // Refer https://jackrabbit.apache.org/api/2.12/index.html?org/apache/jackrabbit/util/ISO8601.html
```
reschke: please cite a non-obsoleted version of the Jackrabbit API :-)

reschke (commented):
please cite a non-obsoleted version of the Jackrabbit API :-)

reschke:
> +1 but I think in the long run we should replace `ISO8601.parse()` with something better (swallowing the exceptions returning `null` is not great).

Be careful; that method is specifically written to support dates in JCR (AFAIR). Using an entirely different library could introduce subtle bugs. If better diagnostics are needed, we should do that in the same codebase, sharing code with the existing method.

PR #850, opened 2023-02-13
OAK-10113 : lower error to warn

stefan-egli:


stefan-egli:
PS: sonar build failure unrelated to this PR - this is just a log level change, so shouldn't really cause any failiure at all

Joscorbe (approved):

PR #858, opened 2023-02-23
OAK-10114 : created new query API with projections to limit memory us…

rishabhdaim:
…age when fetching huge no. of records

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Document.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Document.java
@@ -152,22 +152,6 @@ public class Document implements CacheValue {
     }
 
     /**
-     * UnSeals this document and turns it into a mutable object.
-     *
-     * @see #seal()
-     */
-    void unSeal() {
-        if (sealed.getAndSet(false)) {
-            data = Maps.newHashMap(data);
-            data.entrySet().stream().filter(entry -> entry.getValue() instanceof Map).forEach(entry -> {
-                @SuppressWarnings("unchecked")
-                Map<Object, Object> map = (Map<Object, Object>) entry.getValue();
-                entry.setValue(transformAndUnSeal(map, 1));
-            });
-        }
-    }
-
-    /**
      * Determines if this document is sealed or not
      * @return true if document is sealed.
      */
@@ -229,34 +213,6 @@ public class Document implements CacheValue {
         } else {
             return Collections.unmodifiableMap(map);
         }
-    }
-
-    /**
-     * Transform and un-seal the data of this document.
-     *
-     * @param map   the map to transform.
-     * @param level the level. Zero for the top level map, one for an entry in
-     *              the top level map, etc.
-     * @return the transformed and un-sealed map.
-     * @see #transformAndSeal(Map, String, int)
-     */
-    @NotNull
-    private Map<?, ?> transformAndUnSeal(@NotNull Map<Object, Object> map, final int level) {
-
-        if (map instanceof NavigableMap) {
-            map = Maps.newTreeMap((NavigableMap<Object, Object>) map);
-        } else {
-            map = Maps.newHashMap(map);
-        }
-
-        map.entrySet().forEach(entry -> {
-            Object value = entry.getValue();
-            if (value instanceof Map) {
-                @SuppressWarnings("unchecked") final Map<Object, Object> childMap = (Map<Object, Object>) value;
-                entry.setValue(transformAndUnSeal(childMap, level + 1));
-            }
-        });
-        return map;
     }
 
     @Override
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Document.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Document.java
@@ -152,6 +152,22 @@ public class Document implements CacheValue {
     }
 
     /**
+     * UnSeals this document and turns it into a mutable object.
+     *
+     * @see #seal()
+     */
+    void unSeal() {
+        if (sealed.getAndSet(false)) {
+            data = Maps.newHashMap(data);
+            data.entrySet().stream().filter(entry -> entry.getValue() instanceof Map).forEach(entry -> {
+                @SuppressWarnings("unchecked")
+                Map<Object, Object> map = (Map<Object, Object>) entry.getValue();
+                entry.setValue(transformAndUnSeal(map, 1));
+            });
+        }
+    }
+
+    /**
      * Determines if this document is sealed or not
      * @return true if document is sealed.
      */
@@ -213,6 +229,34 @@ public class Document implements CacheValue {
         } else {
             return Collections.unmodifiableMap(map);
         }
+    }
+
+    /**
+     * Transform and un-seal the data of this document.
+     *
+     * @param map   the map to transform.
+     * @param level the level. Zero for the top level map, one for an entry in
+     *              the top level map, etc.
+     * @return the transformed and un-sealed map.
+     * @see #transformAndSeal(Map, String, int)
+     */
+    @NotNull
+    private Map<?, ?> transformAndUnSeal(@NotNull Map<Object, Object> map, final int level) {
+
+        if (map instanceof NavigableMap) {
+            map = Maps.newTreeMap((NavigableMap<Object, Object>) map);
+        } else {
+            map = Maps.newHashMap(map);
+        }
+
+        map.entrySet().forEach(entry -> {
+            Object value = entry.getValue();
+            if (value instanceof Map) {
+                @SuppressWarnings("unchecked") final Map<Object, Object> childMap = (Map<Object, Object>) value;
+                entry.setValue(transformAndUnSeal(childMap, level + 1));
+            }
+        });
+        return map;
     }
 
     @Override
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStore.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStore.java
@@ -29,7 +29,6 @@ import org.jetbrains.annotations.Nullable;
 
 import static com.google.common.collect.Sets.newHashSet;
 import static org.apache.jackrabbit.oak.plugins.document.Document.ID;
-import static org.apache.jackrabbit.oak.plugins.document.Document.MOD_COUNT;
 import static org.apache.jackrabbit.oak.plugins.document.Throttler.NO_THROTTLING;
 
 /**
@@ -510,7 +509,7 @@ public interface DocumentStore {
      * @param indexedProperty the name of the indexed property (optional)
      * @param startValue the minimum value of the indexed property
      * @param limit the maximum number of entries to return
-     * @param projections {@link List} of projected keys (optional). Keep this empty to fetch all fields on document.
+     * @param projection {@link List} of projected keys (optional). Keep this empty to fetch all fields on document.
      * @return the list (possibly empty)
      * @throws DocumentStoreException if the operation failed. E.g. because of
      *          an I/O error.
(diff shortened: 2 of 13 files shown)


rishabhdaim:
@mreutegg @stefan-egli could you please review and share the feedback.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStore.java
```diff
@@ -486,4 +490,45 @@ default int getNodeNameLimit() {
...
+     * The returned documents are sorted by key and are immutable.
+     *
+     * @param <T> the document type
+     * @param collection the collection
+     * @param fromKey the start value (excluding)
+     * @param toKey the end value (excluding)
+     * @param indexedProperty the name of the indexed property (optional)
+     * @param startValue the minimum value of the indexed property
+     * @param limit the maximum number of entries to return
+     * @param projections {@link List} of projected keys (optional). Keep this empty to fetch all fields on document.
```
mreutegg: This contract is not intuitive to me. Wouldn't it be better in to return documents with `_id` field only when the `projections` list is empty? One can still use the existing query method if all fields are needed.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStore.java
```diff
@@ -486,4 +490,45 @@ default int getNodeNameLimit() {
...
+                                               final String indexedProperty,
+                                               final long startValue,
+                                               final int limit,
+                                               final List<String> projections) throws DocumentStoreException {
+        final Set<String> projectedSet = newHashSet(projections);
+        projectedSet.add(ID);
+        projectedSet.add(MOD_COUNT);
+
+        final List<T> list = query(collection, fromKey, toKey, indexedProperty, startValue, limit);
+        list.stream().filter(doc -> !doc.isSealed()).forEach(doc -> doc.keySet().retainAll(projectedSet));
```
mreutegg: I would unconditionally create new documents with the projection applied. I'm surprised this works for some DocumentStore implementations. I was expecting all returned documents are sealed. Such documents can be served from cache and it would be dangerous if a caller can modify them.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStore.java
```diff
@@ -486,4 +490,45 @@ default int getNodeNameLimit() {
...
+     *          an I/O error.
+     */
+    @NotNull
+    default <T extends Document> List<T> query(final Collection<T> collection,
+                                               final String fromKey,
+                                               final String toKey,
+                                               final String indexedProperty,
+                                               final long startValue,
+                                               final int limit,
+                                               final List<String> projections) throws DocumentStoreException {
```
mreutegg: I think it should be singular: `projection`

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -722,8 +725,19 @@ public <T extends Document> List<T> query(Collection<T> collection,
                                               String indexedProperty,
                                               long startValue,
                                               int limit) {
-        return queryWithRetry(collection, fromKey, toKey, indexedProperty,
-                startValue, limit, maxQueryTimeMS);
+        return query(collection, fromKey, toKey, indexedProperty, startValue, limit, emptyList());
```
mreutegg, line 727: Ah, now I see why an empty projection must return the entire document. I'm still not sure this is a good idea.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -823,6 +839,14 @@ && canUseModifiedTimeIdx(startValue)) {
                 } else {
                     result = dbCollection.find(query);
                 }
+
+                if (!projections.isEmpty()) {
+                    // it is mandatory to fetch MOD_COUNT because of cacheChangesTracker
+                    projections.add(MOD_COUNT);
+                    Bson projection = include(projections);
+                    result.projection(projection);
```
mreutegg: We need to be careful with the projection in `queryInternal()`. Later in this method documents are put into a cache: `nodeCache.putNonConflictingDocs()`. Documents with a projection are incomplete and later reads for complete documents may be served from cache.
It might be best to simply bypass the cache when there is a projection and not put those documents in the cache.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MongoDocumentStoreTest.java
```diff
@@ -241,6 +242,30 @@ public void queryWithLimit() throws Exception {
...
+        for (int i = 0; i < DocumentMK.MANY_CHILDREN_THRESHOLD * 2; i++) {
+            DocumentNodeState n = new DocumentNodeState(store, Path.fromString("/node-" + i),
+                    new RevisionVector(rev));
+            inserts.add(n.asOperation(rev));
+        }
+        docStore.create(Collection.NODES, inserts);
+        List<NodeDocument> docs = docStore.query(Collection.NODES,
+                Utils.getKeyLowerLimit(Path.ROOT),  Utils.getKeyUpperLimit(Path.ROOT), null, 0,
+                DocumentMK.MANY_CHILDREN_THRESHOLD, Lists.newArrayList(ID));
+        // since both _id & _modCount are mandatory, so data size should be 2
```
mreutegg: This is not consistent with the contract in DocumentStore where it says only `_id` is mandatory.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MongoDocumentStoreTest.java
```diff
@@ -241,6 +242,30 @@ public void queryWithLimit() throws Exception {
...
     }
 
+    @Test
+    public void queryWithProjections() {
+        DocumentStore docStore = openDocumentStore();
+        DocumentNodeStore store = new DocumentMK.Builder()
+                .setDocumentStore(docStore).setAsyncDelay(0).getNodeStore();
+        Revision rev = Revision.newRevision(0);
+        List<UpdateOp> inserts = new ArrayList<UpdateOp>();
+        for (int i = 0; i < DocumentMK.MANY_CHILDREN_THRESHOLD * 2; i++) {
```
mreutegg: Why add so many documents? It seems unnecessary to me for this test.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MongoDocumentStoreTest.java
```diff
@@ -241,6 +242,30 @@ public void queryWithLimit() throws Exception {
...
+            inserts.add(n.asOperation(rev));
+        }
+        docStore.create(Collection.NODES, inserts);
+        List<NodeDocument> docs = docStore.query(Collection.NODES,
+                Utils.getKeyLowerLimit(Path.ROOT),  Utils.getKeyUpperLimit(Path.ROOT), null, 0,
+                DocumentMK.MANY_CHILDREN_THRESHOLD, Lists.newArrayList(ID));
+        // since both _id & _modCount are mandatory, so data size should be 2
+        if (MONGO_DB) {
+            docs.forEach(d -> assertEquals(2 , d.data.size()));
+        }
```
mreutegg: It would be better if behaviour for all implementations was the same, so we don't have to special case here.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStore.java
rishabhdaim: This contract `to fetch all if projections are empty` is in line with how mongo behaves.
If we provide an empty projection array, then mongo returns all the results.
In case we want specific fields, then we need to mention those individually.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MongoDocumentStoreTest.java
rishabhdaim: This won't be required after we make changes regarding putting documents into cache.

reschke:
It would be good if the JIRA issue was more clear about the fact that a new DocumentStore is being introduced; and that API change should actually be described over there.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStore.java
reschke: "This is what MongoDB does" IMHO is not sufficient reason for something that would apply to all DocumentStore implementations. Leaking implementation details...

rishabhdaim: We need a mechanism to fetch all the fields on the document. Since we don't know the field names in advance and they can change at any time (given the schema-less nature of Mongo), I think passing an empty list makes sense to fetch the entire document.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Document.java
```diff
@@ -151,6 +151,22 @@ public void seal() {
         }
     }
 
+    /**
+     * UnSeals this document and turns it into a mutable object.
+     *
+     * @see #seal()
+     */
+    void unSeal() {
```
mreutegg: I don't like adding method. The idea behind sealing a document is to ensure it becomes immutable. A DocumentStore implementation can use this to construct a Document, put it into a cache and then return it to a caller. With this new method a caller can modify a sealed Document, violating a basic principle.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStore.java
```diff
@@ -486,4 +489,55 @@ default int getNodeNameLimit() {
...
+                                               final String toKey,
+                                               final String indexedProperty,
+                                               final long startValue,
+                                               final int limit,
+                                               final List<String> projection) throws DocumentStoreException {
+
+        final List<T> list = query(collection, fromKey, toKey, indexedProperty, startValue, limit);
+
+        if (projection == null || projection.isEmpty()) {
+            list.forEach(Document::seal);
```
mreutegg: This shouldn't be necessary, because implementations should already return sealed documents when they are e.g. served from a cache.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStore.java
```diff
@@ -486,4 +489,55 @@ default int getNodeNameLimit() {
...
+        }
+
+        final Set<String> projectedSet = newHashSet(projection);
+        projectedSet.add(ID);
+
+        list.forEach(t -> {
+            t.unSeal();
+            t.keySet().retainAll(projectedSet);
+            t.seal();
+        });
```
mreutegg: Doesn't this modify Documents potentially coming from a cache? See also earlier comment about un-sealing a Document.

mreutegg: I think this may also be the reason why DocumentNodeStoreBranchesTest.commitHookChangesOnBranchWithInterference is currently failing.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStore.java
rishabhdaim: sure, will revert this change.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStore.java
rishabhdaim: Yes, I agree this would modify the documents coming from the cache.

I will fix this by cloning the document and then returning the cloned document with projected fields only.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MongoDocumentStoreTest.java
```diff
@@ -241,6 +243,80 @@ public void queryWithLimit() throws Exception {
...
+        if (MONGO_DB) {
+            // we have _modCount as additional field in case we use MongoDocumentStore
+            docs.forEach(d -> assertEquals(4 , d.data.size()));
+        } else {
+            docs.forEach(d -> assertEquals(3 , d.data.size()));
+        }
+        assertEquals(10, docs.size());
+        store.dispose();
+    }
+
```
mreutegg: Confusingly MongoDocumentStoreTest does not use MongoDB as a DocumentStore by default.

I would rather move these tests to MongoDocumentStoreIT: https://github.com/rishabhdaim/jackrabbit-oak/pull/2

mreutegg:
Other than moving the tests, this PR now looks good to me.

PR #863, opened 2023-03-01
OAK-10127 - Log warn message when MongoDB document is big

ankitaagar:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -157,6 +157,10 @@ public class MongoDocumentStore implements DocumentStore {
      */
     public static final long DEFAULT_THROTTLING_TIME_MS = Long.getLong("oak.mongo.throttlingTime", 20);
 
+    /**
+     * Document size of 16MB is a limit in Mongo
+     */
+    public static final long SIZE_LIMIT = 16793600;
```
stefan-egli: This is a bit larger than 16MB, what's the reason? I was also wondering whether we shouldn't warn below 16MB already to have some buffer, wdyt?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1350,6 +1358,10 @@ private <T extends Document> BulkUpdateResult sendBulkUpdate(Collection<T> colle
         int i = 0;
         for (UpdateOp updateOp : updateOps) {
             String id = updateOp.getId();
+            int size = updateOp.toString().length();
```
stefan-egli: Is updateOp the right way to measure this though? Isn't that a generic documentMk side object and what mongo sees is the Bson query below?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
ankitaagar: > This is a bit larger than 16MB, what's the reason? I was also wondering whether we shouldn't warn below 16MB already to have some buffer, wdyt?

Actually, Mongo compares with this size only and throws an exception. IMO, it would be good to give a WARN when the size reaches 15MB may be. wdyt?

these are the exceptions, I think keeping 15MB as warning would be good
> org.bson.BsonMaximumSizeExceededException: Payload document size is larger than maximum of 16777216.
>org.bson.BsonMaximumSizeExceededException: Document size of 17826076 is larger than maximum of 16793600.
16MB is 16777216

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1074,6 +1078,10 @@ private <T extends Document> T findAndModify(Collection<T> collection,
                 }
             }
 
+            int size = updateOp.toString().length();
+            if(size > SIZE_LIMIT) {
```
reschke:
```suggestion
            if (size > SIZE_LIMIT) {
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1074,6 +1078,10 @@ private <T extends Document> T findAndModify(Collection<T> collection,
                 }
             }
 
+            int size = updateOp.toString().length();
```
reschke, line 1081: Obtaining the string representation might be expensive, no?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
ankitaagar: updateOp's id/value is passed to the Bson query. So, I thought to give a warn here only.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
ankitaagar: Is there any other way to get the Document size?

ankitaagar: > Obtaining the string representation might be expensive, no?

Is there any other way to get the Document size?

reschke:
I believe the main issue here is that the size of the update has no direct relation to the size of the resulting document.

stefan-egli:
Would it be possible to do this in a try/catch block? That way there would be no performance impact of expensive toStrings and it would only apply to exception cases (ideally there would be a dedicated exeption for this, but I haven't checked)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
rishabhdaim: I believe the size of UpdateOp doesn't reflect the size of the document. UpdateOp only has the delta, not the actual document size.

mreutegg:
> Would it be possible to do this in a try/catch block?

I agree, this might be the better approach.

The PR currently only covers one DocumentStore method in a test. Can you please add tests for each of the write methods?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
stefan-egli: > I think keeping 15MB as warning would be good

+1

stefan-egli: Having said that, as @rishabhdaim  pointed out below, the updateOp might not be the one throwing the document over the 16MB and going via catching the exception might be the only reliable way in the end (in which case we probably don't event need these checks here

ankitaagar: yes adding try/catch only

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1487,8 +1499,23 @@ public <T extends Document> boolean create(Collection<T> collection, List<Update
                 }
                 insertSuccess = true;
                 return true;
-            } catch (MongoException e) {
-                return false;
+            } catch (BsonMaximumSizeExceededException e) {
```
mreutegg, line 1485: The new tests do not seem to trigger this exception.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1487,8 +1499,23 @@ public <T extends Document> boolean create(Collection<T> collection, List<Update
...
+                for (T doc : docs) {
+                    if (doc.toString().length() > SIZE_LIMIT) {
```
mreutegg: IIUC `doc.toString()` simply returns the _id value as a String. I don't think that's the intention here.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1078,23 +1090,25 @@ private <T extends Document> T findAndModify(Collection<T> collection,
...
+            try {
+                oldNode = execute(session -> {
+                    if (session != null) {
+                        return dbCollection.findOneAndUpdate(session, query, update, options);
+                    } else {
+                        return dbCollection.findOneAndUpdate(query, update, options);
+                    }
+                }, collection);
+            } catch (MongoWriteException e) {
+                WriteError error = e.getError();
```
mreutegg: `error` is not used.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1078,23 +1090,25 @@ private <T extends Document> T findAndModify(Collection<T> collection,
...
+            BasicDBObject oldNode = null;
+            try {
+                oldNode = execute(session -> {
+                    if (session != null) {
+                        return dbCollection.findOneAndUpdate(session, query, update, options);
+                    } else {
+                        return dbCollection.findOneAndUpdate(query, update, options);
+                    }
+                }, collection);
+            } catch (MongoWriteException e) {
```
mreutegg: It seems to me the new tests do not exercise this code path with the exception.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1059,13 +1064,20 @@ private <T extends Document> T findAndModify(Collection<T> collection,
...
+                    UpdateResult result = null;
+                    try {
+                        result = execute(session -> {
+                            if (session != null) {
+                                return dbCollection.updateOne(session, query, update);
+                            } else {
+                                return dbCollection.updateOne(query, update);
+                            }
+                        }, collection);
+                    } catch (MongoWriteException e) {
```
mreutegg: It seems to me the new tests do not exercise this code path with the exception.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1059,13 +1064,20 @@ private <T extends Document> T findAndModify(Collection<T> collection,
...
+                        WriteError error = e.getError();
```
mreutegg: `error` is not used.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1031,6 +1035,7 @@ private <T extends Document> T findAndModify(Collection<T> collection,
         }
         final Stopwatch watch = startWatch();
         boolean newEntry = false;
+        Throwable t = null;
```
mreutegg: Throwable t is never used.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1078,23 +1090,25 @@ private <T extends Document> T findAndModify(Collection<T> collection,
...
+                oldNode = execute(session -> {
+                    if (session != null) {
+                        return dbCollection.findOneAndUpdate(session, query, update, options);
+                    } else {
+                        return dbCollection.findOneAndUpdate(query, update, options);
+                    }
+                }, collection);
+            } catch (MongoWriteException e) {
+                WriteError error = e.getError();
+                LOG.warn(e.getMessage(), updateOp.getId(), updateOp.getConditions(), oldNode.size());
```
mreutegg: This looks strange. Does the message format really have three placeholders for id, conditions and size? Why log conditions?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1123,7 +1137,13 @@ private <T extends Document> T findAndModify(Collection<T> collection,
                 }
             }
             return oldDoc;
-        } catch (Exception e) {
+        } catch (MongoCommandException e) {
+            LOG.warn(e.getMessage(), updateOp.getId(), updateOp.getConditions());
```
mreutegg: Same as above. This looks strange to me.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1123,7 +1137,13 @@ private <T extends Document> T findAndModify(Collection<T> collection,
...
+            throw handleException(e, collection, updateOp.getId());
+        } catch (BsonMaximumSizeExceededException e){
+            LOG.warn(e.getMessage(), updateOp.getId());
+            throw handleException(e, collection, updateOp.getId());
```
mreutegg: New tests do not seem to exercise this code/exception. Also, I'm not sure the log message is formatted properly to take the id as argument.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBExceptionTest.java
```diff
@@ -136,32 +138,118 @@ public void idInExceptionMessage() {
...
 
-        // create a 1 MB property
-        char[] chars = new char[1024 * 1024];
+        exceptionMsg = "Document to upsert is larger than 1677721";
+        setExceptionMsg();
+        try {
+           store.createOrUpdate(Collection.NODES, updateOp);
+           fail("DocumentStoreException expected");
+        } catch (DocumentStoreException e) {
+            assertThat(e.getMessage(), containsString(exceptionMsg));
```
mreutegg, line 173: I'd rather be interested to see the exception message contains the id of the document.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBExceptionTest.java
```diff
@@ -136,32 +138,118 @@ public void idInExceptionMessage() {
...
-        Arrays.fill(chars, '0');
-        String content = new String(chars);
+    @Test
+    public void multiCreateOrUpdate16MBDoc() {
+
+        List<UpdateOp> updateOps = new ArrayList<UpdateOp>();
+        UpdateOp updateOp = new UpdateOp("/", true);
+        updateOp = create1MBProp(updateOp);
+        UpdateOp updateOp1 = new UpdateOp("/", true);
+        updateOp1 = create1MBProp(updateOp1);
```
mreutegg: This is a rather unusual combination of update operation. Both target the same document id. Is this intentional?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBExceptionTest.java
```diff
@@ -136,32 +138,118 @@ public void idInExceptionMessage() {
...
-        for (int i = 0; i < 17; i++) {
-            updateOp.set("property"+ Integer.toString(i), content);
+        exceptionMsg = "Resulting document after update is larger than 16777216";
+        setExceptionMsg();
+
+        try {
+            store.createOrUpdate(Collection.NODES, updateOps);
+            fail("DocumentStoreException expected");
+        } catch (DocumentStoreException e) {
+            assertThat(e.getMessage(), containsString(exceptionMsg));
```
mreutegg, line 205: Again, it would be better to check if the exception mentions the document id.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
ankitaagar: It always triggers BsonMaximumSizeExceededException but when the test is executed with DocumentStoreException it throws MongoException

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
ankitaagar: yes, thats wrong. I will use "Entry<Key, Operation> entry : update.getChanges().entrySet()" and get the size of each entryset

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -157,6 +161,10 @@ public class MongoDocumentStore implements DocumentStore {
      */
     public static final long DEFAULT_THROTTLING_TIME_MS = Long.getLong("oak.mongo.throttlingTime", 20);
 
+    /**
+     * Document size of 16MB is a limit in Mongo
+     */
+    public static final long SIZE_LIMIT = 16777216;
```
rishabhdaim: this should be made private.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1055,13 +1064,20 @@ private <T extends Document> T findAndModify(Collection<T> collection,
...
+                        result = execute(session -> {
+                            if (session != null) {
+                                return dbCollection.updateOne(session, query, update);
+                            } else {
+                                return dbCollection.updateOne(query, update);
+                            }
+                        }, collection);
+                    } catch (MongoWriteException e) {
+                        WriteError error = e.getError();
+                        LOG.warn(e.getMessage(), updateOp.getId());
```
rishabhdaim: This would only work with placeholders `{}` not with normal strings.

This would print only the `e.getMessage()` and ignore the `updateOp.getId()` since it won't find any placeholder for this in `e.getMessage()`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
rishabhdaim: Same as the previous comment. It would only print `e.getMessage()` and ignore the other values.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1115,7 +1137,13 @@ private <T extends Document> T findAndModify(Collection<T> collection,
                 }
             }
             return oldDoc;
-        } catch (Exception e) {
+        } catch (MongoCommandException e) {
+            LOG.warn(e.getMessage(), updateOp.getId(), updateOp.getConditions());
+            throw handleException(e, collection, updateOp.getId());
+        } catch (BsonMaximumSizeExceededException e){
+            LOG.warn(e.getMessage(), updateOp.getId());
```
rishabhdaim: I would log some meaningful message, rather than just `e.getMessage()`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
rishabhdaim: You can use `doc.getMemory()`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1470,8 +1499,23 @@ public <T extends Document> boolean create(Collection<T> collection, List<Update
...
+                    if (doc.toString().length() > SIZE_LIMIT) {
+                        LOG.error("The document with Id={} has size={} is greater than the size limit",
+                                doc.getId(), doc.toString().length(), e.getMessage());
+                    }
+                }
+                throw handleException(e, collection, Document.ID);
+            }
+            catch (MongoException e) {
+                for (T doc : docs) {
+                    if (doc.toString().length() > SIZE_LIMIT) {
```
rishabhdaim: Please use `doc.getMemory()`.

rishabhdaim (changes requested):
Changes requested as per review comments.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBExceptionTest.java
```diff
@@ -132,8 +137,129 @@ public void idInExceptionMessage() {
...
+    private UpdateOp create16MBProp(UpdateOp op) {
+        // create a 1 MB property
+        char[] chars = new char[1024 * 1024];
+        Arrays.fill(chars, '0');
+        String content = new String(chars);
+        op.set("property0", content);
+
+        //create 16MB property
+        for (int i = 1; i < 16; i++) {
+            op.set("property"+ Integer.toString(i), content);
```
mreutegg: Integer.toString(i) is unnecessary.
```suggestion
            op.set("property"+ i, content);
```

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBExceptionTest.java
```diff
@@ -30,6 +30,11 @@
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
+import org.junit.Ignore;
```
mreutegg: Unused.
```suggestion
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1055,13 +1063,21 @@ private <T extends Document> T findAndModify(Collection<T> collection,
...
+                                return dbCollection.updateOne(session, query, update);
+                            } else {
+                                return dbCollection.updateOne(query, update);
+                            }
+                        }, collection);
+                    } catch (MongoWriteException e) {
+                        WriteError werr = e.getError();
+                        LOG.error("Failed to update the document with Id={} and size={} with error message '{}'",
+                                updateOp.getId(), updateOp.toString().length(), werr.getMessage());
+                        throw handleException(e, collection, updateOp.getId());
```
mreutegg: I think this exception handling should go to the end of the already existing try/catch block. With this change, the exception is caught, re-thrown, caught and re-thrown again.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1079,14 +1095,22 @@ private <T extends Document> T findAndModify(Collection<T> collection,
...
+                    } else {
+                        return dbCollection.findOneAndUpdate(query, update, options);
+                    }
+                }, collection);
+            } catch (MongoCommandException e) {
+                LOG.error("Failed to update the document with Id={} has size={} " +
+                                "with error message '{}'",
+                                updateOp.getId(), updateOp.toString().length(),
+                                SIZE_LIMIT, e.getMessage());
+                throw handleException(e, collection, updateOp.getId());
```
mreutegg: Same as above.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1055,13 +1063,21 @@ private <T extends Document> T findAndModify(Collection<T> collection,
...
+                            if (session != null) {
+                                return dbCollection.updateOne(session, query, update);
+                            } else {
+                                return dbCollection.updateOne(query, update);
+                            }
+                        }, collection);
+                    } catch (MongoWriteException e) {
+                        WriteError werr = e.getError();
+                        LOG.error("Failed to update the document with Id={} and size={} with error message '{}'",
+                                updateOp.getId(), updateOp.toString().length(), werr.getMessage());
```
mreutegg: Do we really need the size of the UpdateOp? If yes, then I think it would be better to let the UpdateOp calculate it instead of formatting the UpdateOp as a String and getting the number of characters. Converting to String will use quite a bit of memory when the UpdateOp is big.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1079,14 +1095,22 @@ private <T extends Document> T findAndModify(Collection<T> collection,
...
+                        return dbCollection.findOneAndUpdate(session, query, update, options);
+                    } else {
+                        return dbCollection.findOneAndUpdate(query, update, options);
+                    }
+                }, collection);
+            } catch (MongoCommandException e) {
+                LOG.error("Failed to update the document with Id={} has size={} " +
+                                "with error message '{}'",
+                                updateOp.getId(), updateOp.toString().length(),
+                                SIZE_LIMIT, e.getMessage());
```
rishabhdaim: You have provided 4 arguments whereas you only have 3 placeholders.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1470,8 +1495,16 @@ public <T extends Document> boolean create(Collection<T> collection, List<Update
...
                 return true;
-            } catch (MongoException e) {
-                return false;
+            } catch (BsonMaximumSizeExceededException e) {
+                for (T doc : docs) {
+                    // doc.getMemory()/2 - converting from UTF-16 to UTF-8
+                    if (doc.getMemory()/2 > SIZE_LIMIT) {
+                        LOG.error("Failed to create the document with Id={} has size={}" +
+                                        " with error message '{}",
+                                doc.getId(), doc.getMemory()/2, SIZE_LIMIT, e.getMessage());
```
rishabhdaim: Same as above. 4 arguments but only 3 placeholders.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBExceptionTest.java
```diff
@@ -132,8 +137,129 @@ public void idInExceptionMessage() {
...
     private void setExceptionMsg() {
         client.setExceptionBeforeUpdate(exceptionMsg);
         client.setExceptionBeforeQuery(exceptionMsg);
     }
+
+    private UpdateOp create1MBProp(UpdateOp op) {
+        // create a 1 MB property
+        char[] chars = new char[1024 * 1024];
+        Arrays.fill(chars, '0');
+        String content = new String(chars);
```
rishabhdaim: This code can be extracted out to a common method to create a content worth 1MB

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBExceptionTest.java
```diff
@@ -132,8 +137,129 @@ public void idInExceptionMessage() {
...
+        } catch (DocumentStoreException e) {
+            assertThat(e.getMessage(), containsString(exceptionMsg));
+            assertThat(e.getMessage(), containsString(docName));
+        }
+    }
+
+    @Test
+    public void multiCreateOrUpdate16MBDoc() {
+
+        List<UpdateOp> updateOps = new ArrayList<UpdateOp>();
```
rishabhdaim:
```suggestion
        List<UpdateOp> updateOps = new ArrayList<>();
```

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBExceptionTest.java
```diff
@@ -132,8 +137,129 @@ public void idInExceptionMessage() {
...
+            fail("DocumentStoreException expected");
+        } catch (DocumentStoreException e) {
+            assertThat(e.getMessage(), containsString(exceptionMsg));
+        }
+    }
+
+    @Test
+    public void create16MBDoc() {
+
+        List<UpdateOp> updateOps = new ArrayList<UpdateOp>();
```
rishabhdaim:
```suggestion
        List<UpdateOp> updateOps = new ArrayList<>();
```

rishabhdaim (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1470,8 +1495,16 @@ public <T extends Document> boolean create(Collection<T> collection, List<Update
...
+                    }
+                }
+                throw handleException(e, collection, Document.ID);
```
mreutegg: The third parameter is not meaningful. It should be the actual ID values.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1470,8 +1495,16 @@ public <T extends Document> boolean create(Collection<T> collection, List<Update
                 }
                 insertSuccess = true;
                 return true;
-            } catch (MongoException e) {
-                return false;
+            } catch (BsonMaximumSizeExceededException e) {
+                for (T doc : docs) {
+                    // doc.getMemory()/2 - converting from UTF-16 to UTF-8
+                    if (doc.getMemory()/2 > SIZE_LIMIT) {
+                        LOG.error("Failed to create the document with Id={} has size={}" +
```
mreutegg: There are three placeholders, but four arguments.

mreutegg (changes requested):

mreutegg:
Oops, sorry about the duplicate comments. I only saw Rishabh's comments after I submitted mine.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBExceptionTest.java
```diff
@@ -132,8 +136,132 @@ public void idInExceptionMessage() {
...
+
+    private UpdateOp create16MBProp(UpdateOp op) {
+        // create a 1 MB property
+        String content = create1MBContent();
+        op.set("property0", content);
+
+        //create 16MB property
+        for (int i = 1; i < 16; i++) {
+            op.set("property"+ i, content);
+        }
```
rishabhdaim:
```suggestion
        

        //create 16MB property
        for (int i = 0; i < 16; i++) {
            op.set("property"+ i, content);
        }
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1470,8 +1487,18 @@ public <T extends Document> boolean create(Collection<T> collection, List<Update
...
                 insertSuccess = true;
                 return true;
-            } catch (MongoException e) {
-                return false;
+            } catch (BsonMaximumSizeExceededException e) {
+                T doct = null;
+                for (T doc : docs) {
+                    doct = doc;
+                    // doc.getMemory()/2 - converting from UTF-16 to UTF-8
+                    if (doc.getMemory()/2 > SIZE_LIMIT) {
```
stefan-egli: (I haven't followed all discussion) is UTF-16 always the case, or is there something that this could be made dependent upon ?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1115,6 +1123,15 @@ private <T extends Document> T findAndModify(Collection<T> collection,
                 }
             }
             return oldDoc;
+        } catch (MongoWriteException e) {
+            WriteError werr = e.getError();
+            LOG.error("Failed to update the document with Id={} with error message '{}'",
```
```suggestion
            LOG.error("Failed to update the document with Id={} with a MongoWriteException message '{}'",
```

a bit a detail, but perhaps could be useful to explicitly state that it was a MongoWriteException (and a MongoCommandException below)

stefan-egli (commented):
just adding a comment - I haven't followed the entire conversation, so perhaps someone else can also chime in

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
ankitaagar: Mongo returns it in UTF-16 so I have changed it to UTF-8

rishabhdaim (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1470,6 +1487,18 @@ public <T extends Document> boolean create(Collection<T> collection, List<Update
...
+                for (T doc : docs) {
+                    doct = doc;
+                    // doc.getMemory()/2 - converting from UTF-16 to UTF-8
+                    if (doc.getMemory()/2 > SIZE_LIMIT) {
+                        LOG.error("Failed to create the document with Id={} has size={}" +
+                                        " with BsonMaximumSizeExceededException message '{}",
+                                doc.getId(), doc.getMemory()/2, e.getMessage());
+                    }
+                }
+                throw handleException(e, collection, doct.getId());
```
rishabhdaim: AFAICT, this method should return `false` here instead of throwing the exception.
@stefan-egli wdyt?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
reschke: This may be pedantic, but UTF-16 -> UTF-8 does not necessarily imply 50% reduction. I assume the rough number is good enough, but maybe the comment could state more clearly what's going on.

stefan-egli: I think the problem here is that we'd really like to estimate the memory usage within Mongo (which is UTF-8 IIUC), but the method used here is estimating the memory usage within the JVM (which does a multiply by 2, so would estimate for UTF-16). Is that correct?

The problem with the above is that it can potentially be off significantly - as the method used is optimized to estimate JVM side memory use, not Mongo side. And dividing by 2 could skew things..

I wonder if we should indeed lower the threshold for issuing a log.error here (and then phrase the message that the exact size might differ)..

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
stefan-egli: throwing exception looks fine, why would you return `false`?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
reschke: FWIW, MongoDB persists BSON, right? (So a binary representation of JSON).

Also note that that the utility method to estimate JVM memory use in the document store will be incorrect in JDK >= 9.

ankitaagar: @stefan-egli  It was done to mention the document id that caused the exception

stefan-egli: I think it's anyway difficult to convert our JVM estimate to a MongoDB estimate, for that we'd have to know how much space integers, boolean etc are taking up. With the current "divide by 2" we assume they also take half the space, which might or might not be true.

So I would add some margin to this condition, either by already warning at say 14MB - or, more drastically, by dividing by 3 (which would correspond to warning at 10.7MB already). That would be a bit more "loudly", as it would list IDs of more documents than necessary, but it would more likely not miss any. Without any margin there is chance to miss though.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
stefan-egli: +1,
you're right, I was not looking at the whole context. return `false` is correct

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
ankitaagar: I need to add Warn if a doc size reaches 14MB as it won't though any exception. We can have a separate jira to add warn message when a document size reaches 14MB. WDYT?

stefan-egli: The 14MB was just to be on the very safe side, I wasn't suggesting to introduce an additional 14MB check.

Let's say you create a document with 16MB+1byte. Mongo will throw a BsonMaximumSizeExceededException. But for some reason our estimate of that document, divided by 2 is smaller than 16MB (our `getMemory` is only approximate). That case we wouldn't catch atm. But we could catch it as well by eg dividing by 3 or use a lower threshold (whatever is easier). (yes, we would potentially warn about documents that were not the reason mongo issued the exception, but I don't think that's a problem)

reschke (changes requested):
The logic how we compute the size IMHO is flawed (as we actually do not know the size of the BSON representation inside Mongo). If we can't do better, we absolutely should precisely document our assumption, and move the calculation into a single utility method (instead of inlining stuff into the main code flow).

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBExceptionTest.java
```diff
@@ -31,11 +31,13 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
 import static java.util.Collections.singletonList;
 import static org.hamcrest.Matchers.containsString;
-import static org.junit.Assert.assertThat;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
```
mreutegg: Please don't import with a wildcard.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1470,6 +1487,19 @@ public <T extends Document> boolean create(Collection<T> collection, List<Update
                 }
                 insertSuccess = true;
                 return true;
+            } catch (BsonMaximumSizeExceededException e) {
+                T doct = null;
```
mreutegg: `doct` is assigned but never used.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
mreutegg: How about removing this somewhat arbitrary threshold and simply log memory usage for each document? At this point we know some Bson size exceeded the limit. Readers of the log message can then interpret which document was most likely the culprit.

ankitaagar: I understand UTF-16 -> UTF-8 does not necessarily imply 50% reduction.
We decided to show the error message.
Let's say, I change it to 
```
catch (BsonMaximumSizeExceededException e) {
                LOG.error("Failed to create one of the documents from the below list " +
                        " with BsonMaximumSizeExceededException message = '{}'.", e.getMessage());
                for (T doc : docs) {
                    LOG.error("Document ID {} and Size {} ",doc.getId(), doc.getMemory());
                }
                return false;
}
```
The log.error would look like:
```
13:01:03.511 ERROR [main] MongoDocumentStore.java:1494      Failed to create one of the documents from the below list with BsonMaximumSizeExceededException message = 'Payload document size is larger than maximum of 16777216.'.
13:01:03.511 ERROR [main] MongoDocumentStore.java:1494      Document ID /test and Size 2097764 
13:01:03.511 ERROR [main] MongoDocumentStore.java:1494      Document ID /foo and Size 33557724 
```

The exception message from Mongo shows the size limit as 16777216 and we are calculating it with doc.getMemory(). So a document with size 16778862 will show 33557724 with doc.getMemory().

Should I also log that BSON size is calculated in UTF-8 and we are doing it in UTF-16?

reschke: Well, the character encoding is just one source of differences. BSON is a binary serialization. What we use is an *estimate* of memory used in the VM. That is somewhat related, but not directly.

I agree with @mreutegg that just listing the estimated VM size is good enough (maybe: "document id '...' has estimated size xyz in VM"), and maybe sort those in descending order (if it's not too much trouble).

reschke: And it should be "update", not "create", right?

ankitaagar: No, it's create.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -157,6 +161,10 @@ public class MongoDocumentStore implements DocumentStore {
      */
     public static final long DEFAULT_THROTTLING_TIME_MS = Long.getLong("oak.mongo.throttlingTime", 20);
 
+    /**
+     * Document size of 16MB is a limit in Mongo
+     */
+    private static final long SIZE_LIMIT = 16777216;
```
mreutegg: Can you please remove these lines. `SIZE_LIMIT` is not used.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBExceptionTest.java
```diff
@@ -132,8 +137,125 @@ public void idInExceptionMessage() {
...
+        UpdateOp op1 = new UpdateOp("/test", true);
+        op1 = create1MBProp(op1);
+
+        UpdateOp op2 = new UpdateOp("/foo", false);
+        op2 = create1MBProp(op2);
+        op2 = create16MBProp(op2);
+
+        updateOps.add(op1);
+        updateOps.add(op2);
+        assertFalse(store.create(Collection.NODES, updateOps));
```
mreutegg, line 229: Can you please add an assert that error messages are logged? There's an example how to do that in `LoggingDocumentStoreWrapperTest`.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBExceptionTest.java
```diff
@@ -132,8 +137,125 @@ public void idInExceptionMessage() {
...
+        op = create1MBProp(op);
+        store.createOrUpdate(Collection.NODES, op);
+        op = create16MBProp(op);
+        exceptionMsg = "Resulting document after update is larger than 16777216";
+        try {
+            store.findAndUpdate(Collection.NODES, op);
+            fail("DocumentStoreException expected");
+        } catch (DocumentStoreException e) {
+            assertThat(e.getMessage(), containsString(exceptionMsg));
+       }
```
mreutegg, line 248: Please add an assert that the message contains the document id.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBExceptionTest.java
mreutegg: Please add an assert that the message contains the document id.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBExceptionTest.java
```diff
@@ -132,8 +137,125 @@ public void idInExceptionMessage() {
...
+    public void createOrUpdate16MBDoc() {
+
+        UpdateOp updateOp = new UpdateOp("/foo", true);
+        updateOp = create16MBProp(updateOp);
+        exceptionMsg = "Document to upsert is larger than 16777216";
+        try {
+            store.createOrUpdate(Collection.NODES, updateOp);
+            fail("DocumentStoreException expected");
+        } catch (DocumentStoreException e) {
+            assertThat(e.getMessage(), containsString(exceptionMsg));
```
mreutegg, line 153: Please add an assert that the message contains the document id.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
stefan-egli: This one still seems open

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBExceptionTest.java
stefan-egli: IIUC then this one is still open

stefan-egli (commented):
Had another look and IIUC then 2 previous comments weren't resolved yet - I have unresolved them now. Otherwise would look good.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDBExceptionTest.java
ankitaagar: somehow, one of my commits had removed it, added

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
ankitaagar: @stefan-egli with one of my commits, had added this back.

ankitaagar:
> Had another look and IIUC then 2 previous comments weren't resolved yet - I have unresolved them now. Otherwise would look good.

Somehow, mistakenly one was removed. Added it now

stefan-egli (approved):

stefan-egli:
+1, I'll go ahead and merge

PR #865, opened 2023-03-04
OAK-10131 : ExternalGroupPrincipalProvider should return ItemBasedPrincipals if dynamic group option is enabled

anchela:


anchela:
the failures in *-azure modules (failing AzureBlobStoreBackendTest) seem unrelated to the changes made to oak-auth-external

PR #871, opened 2023-03-14
OAK-10145 Rate limit the Missing provider log messages for IndexUpdate

thomasmueller:


tihom88 (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
```diff
@@ -330,9 +334,14 @@ private void collectIndexEditors(NodeBuilder definitions,
                     // then we don't need to handle missing handler
                     if (definition.hasProperty(ASYNC_PROPERTY_NAME) && rootState.async == null) {
                         if (!TYPE_DISABLED.equals(type)) {
-                            log.warn("Missing provider for nrt/sync index: {} (rootState.async: {}). " +
-                                    "Please note, it means that index data should be trusted only after this index " +
-                                    "is processed in an async indexing cycle.", definition, rootState.async);
+                            long now = System.currentTimeMillis();
+                            if (now > lastMissingProviderMessageTime + 60 * 1000) {
```
tihom88: nit-pick: may be we can instantiate delay as variable and use the variable here.
And update "one minute" in log message with delay/1000 seconds.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
```diff
@@ -82,6 +82,10 @@ public class IndexUpdate implements Editor, PathSource {
     // This counter is cyclically incremented till indexJcrTypeInvalidLogLimiter and then reset to 0
     private static volatile long cyclicExecutionCount = INDEX_JCR_TYPE_INVALID_LOG_LIMITER;
 
+    // Warnings about missing index providers are rate limitted, so that the log file is not filled with them.
```
fabriziofortino: typo + minor suggestion
```suggestion
    // Warnings about missing index providers are rate limited, so the log file is not filled with them.
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
```diff
@@ -330,9 +334,14 @@ private void collectIndexEditors(NodeBuilder definitions,
...
-                            log.warn("Missing provider for nrt/sync index: {} (rootState.async: {}). " +
-                                    "Please note, it means that index data should be trusted only after this index " +
-                                    "is processed in an async indexing cycle.", definition, rootState.async);
+                            long now = System.currentTimeMillis();
+                            if (now > lastMissingProviderMessageTime + 60 * 1000) {
+                                lastMissingProviderMessageTime = now;
+                                log.warn("Missing provider for nrt/sync index: {} (rootState.async: {}). " +
+                                        "Please note, it means that index data should be trusted only after this index " +
+                                        "is processed in an async indexing cycle. " +
+                                        "This message is silenced for one minute.", definition, rootState.async);
```
fabriziofortino: `rootState.async` is always null here (see check at line 335)

fabriziofortino: the placeholder in 

```Missing provider for nrt/sync index: {}```

is replaced with the variable `definition` which is an instance of NodeBuilder. Should we instead pass `indexPath` defined at line 312?

**Review by fabriziofortino — CHANGES_REQUESTED** [review]

File: oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java [hunk 4]
```diff
@@ -330,9 +334,14 @@ private void collectIndexEditors(NodeBuilder definitions,
                     // then we don't need to handle missing handler
                     if (definition.hasProperty(ASYNC_PROPERTY_NAME) && rootState.async == null) {
                         if (!TYPE_DISABLED.equals(type)) {
-                            log.warn("Missing provider for nrt/sync index: {} (rootState.async: {}). " +
-                                    "Please note, it means that index data should be trusted only after this index " +
-                                    "is processed in an async indexing cycle.", definition, rootState.async);
+                            long now = System.currentTimeMillis();
+                            if (now > lastMissingProviderMessageTime + 60 * 1000) {
+                                lastMissingProviderMessageTime = now;
```
nfsantos: The read-test-update of `lastMissingProviderMessageTime` is not atomic, so two or more threads may pass the check and print the log more or less at the same time. If this is an intentional loosening of the specification of "log at most every minute" for the sake of performance, please write it in the comments. Otherwise, I think it is more clear and clean if the implementation did not have race conditions that allow it to break the intention that is apparent from the code (log at most once a minute).

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
```diff
@@ -330,9 +334,14 @@ private void collectIndexEditors(NodeBuilder definitions,
                     // then we don't need to handle missing handler
                     if (definition.hasProperty(ASYNC_PROPERTY_NAME) && rootState.async == null) {
                         if (!TYPE_DISABLED.equals(type)) {
-                            log.warn("Missing provider for nrt/sync index: {} (rootState.async: {}). " +
-                                    "Please note, it means that index data should be trusted only after this index " +
-                                    "is processed in an async indexing cycle.", definition, rootState.async);
+                            long now = System.currentTimeMillis();
```
nfsantos: `System.currentTimeMillis()` is not guaranteed to measure correctly the passage of time.

https://www.geeksforgeeks.org/java-system-nanotime-vs-system-currenttimemillis/ 
> It may give wrong results if the user changes the system time, hits a leap second or there are changes in NTP sync.

Even if in this use case it is not important if from time to time log messages are logged more than once a second or less frequently than once a second, I think we should be strictly correct, not to create confusion.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
thomasmueller: It isn't for performance, but for simplicity. In my view, race conditions in this area (logging) are not a problem: worst case, we would log multiple times; that is fine. Sure, if there is a way to do this without adding complexity that would be nice as well! However, AtomicLong can't be used here it seems (it can not do such comparisons). I think using synchronization or locking would add too much unnecessary complexity.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
thomasmueller: I know System.currentTimeMillis() can go backwards, but in this area I don't think this is a problem really. I'll try with nanotime, even thought it has many more zeroes, which might be confusing as well.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
thomasmueller: Done.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
nfsantos: An alternative that relies in `System.nanoTime()` and is somewhat more readable is to use the StopWatch class from Guava. Just showing it here as a reference, not sure if it is worth to use it in this case:

```java
private static volatile Stopwatch lastMissingProviderMessageTime = Stopwatch.createStarted();


long sinceLastLogMillis = lastMissingProviderMessageTime.elapsed(TimeUnit.MILLISECONDS);
if (sinceLastLogMillis > 1000) {
   lastMissingProviderMessageTime = Stopwatch.createStarted();
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
nfsantos: Ok, I also can't think of a way that ensures that we log only once per second using atomic variable. But could you just add a comment to say that the code does not ensure that we log at most once per second but that is ok in this case? So that our future selves or other people will not look at the code and be misled by it.

nfsantos (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
nfsantos: Probably we can do the same using the `Instant` class from the JDK, but internally Instant is more complex, as it tracks a point in real time, while StopWatch just tracks the passing of time without any anchor in real time.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java
nfsantos: I think this would work and is not too complicated. 

```java
private static volatile AtomicLong lastMissingProviderMessageTime = new AtomicLong(0);


  long now = System.currentTimeMillis();
  long previous = lastMissingProviderMessageTime.get();
  if (now > previous + 60 * 1000) {
      if (lastMissingProviderMessageTime.compareAndSet(previous, now)) {
         // log
      }
```
But leaving it here mainly for intelectual curiosity, could be useful in other situations.

fabriziofortino (approved):

PR #874, opened 2023-03-20
OAK-10149: Rebase may be expensive with many siblings

mreutegg:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/AbstractDocumentNodeState.java
```diff
@@ -127,6 +128,48 @@ public boolean compareAgainstBaseState(NodeState base, NodeStateDiff diff) {
                     }
                 }
             }
+        } else if (base instanceof ModifiedNodeState) {
+            ModifiedNodeState mBase = (ModifiedNodeState) base;
+            if (mBase.getBaseState() == this) {
+                // this is the base state of the ModifiedNodeState
+                // do a reverse comparison and report the inverse back to NodeStateDiff
+                return mBase.compareAgainstBaseState(this, new NodeStateDiff() {
```
stefan-egli: (interesting approach. I'm wondering if it could be useful for other cases)
Would it make sense to move this into its own class and have dedicated test case for this reversing?

stefan-egli (approved):
+1, lgtm, only added below comment as to another unit test idea

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/AbstractDocumentNodeState.java
mreutegg: Makes sense. Was thinking about this as well, but too lazy to actually do it :-/

mreutegg:
oak-pojosr failure looks unrelated to this change.

PR #877, opened 2023-03-24
OAK-9989: introduce wrapped-guava and use it in oak-commons

reschke:


oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/FileIOUtils.java
```diff
@@ -395,7 +394,7 @@ public int compare(String s1, String s2) {
      * @deprecated use {@link org.apache.jackrabbit.oak.commons.io.FileLineDifferenceIterator} instead
      */
     @Deprecated(since = "1.20.0", forRemoval = true)
-    public static class FileLineDifferenceIterator extends AbstractIterator<String> implements Closeable {
+    public static class FileLineDifferenceIterator extends com.google.common.collect.AbstractIterator<String> implements Closeable {
```
rishabhdaim, line 397: Is this intentional to use `AbstractIterator` from `com.google.common.*` instead of `org.apache.jackrabbit.oak.guava.common.collect.AbstractIterator`

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/io/LazyInputStream.java
```diff
@@ -48,7 +47,7 @@ public LazyInputStream(Supplier<InputStream> inputStreamSupplier) {
      * @deprecated Use {@link #LazyInputStream(Supplier)} instead
      */
     @Deprecated(since = "1.20.0", forRemoval = true)
-    public LazyInputStream(ByteSource byteSource) {
+    public LazyInputStream(com.google.common.io.ByteSource byteSource) {
```
rishabhdaim, line 50: Shouldn't this be from new jar? If not, then we should revert the changes in this file.

oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/FileIOUtilsTest.java
```diff
@@ -116,7 +116,7 @@ public void writeCustomReadOrgStringsDeprecated() throws Exception {
         Set<String> actual = newHashSet("a", "z", "e", "b");
 
         File f = folder.newFile();
-        int count = writeStrings(added.iterator(), f, false, new Function<String, String>() {
+        int count = writeStrings(added.iterator(), f, false, new com.google.common.base.Function<String, String>() {
```
rishabhdaim, line 119: Shouldn't we use the `Function` from the new jar or is this intentional? I see we are importing `org.apache.jackrabbit.oak.guava.common.base.Function` as well

rishabhdaim (commented):
I still see some instances of the unshaded `guava` jar. If not, intentionally, then we should remove them.

oak-wrapped-guava/README.md
mreutegg, line 1: Needs a license header or must be excluded from rat check

oak-wrapped-guava/dependency-reduced-pom.xml
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mreutegg: Needs a license header.

mreutegg (changes requested):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/FileIOUtils.java
reschke: Yes, that's a case where we need to preserve API compat. This will be removed at a later point.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/io/LazyInputStream.java
reschke: Same as above.

oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/FileIOUtilsTest.java
reschke: Sama as above.

reschke:
@rishabhdaim - thanks for the review. The leftovers in oak-commons are intentional; that is a module that currently has Guava APIs in it's public interfaces. These are deprecated and will be removed in a separate step. (I picked that module for the proof of concept exactly because of the additional complexitiy in this module).

oak-wrapped-guava/dependency-reduced-pom.xml
reschke: That was autogenerated and I believe not needed by us. Reconfigured the plugin not to generate it.

mreutegg (approved):
Looks good to me. I'm undecided on 'wrapped' vs 'shaded' guava.

rishabhdaim (approved):

PR #879, opened 2023-03-28
OAK-10151 : oak-auth-external tests fail with Guava 20

anchela:
@reschke , i would appreciate if you could take a close look. the changes:
- obvious mistake in one auth-external test 
- adding more messages to the assertions in auth-external tests
- simplify AuthorizableIterator construction in AuthorizbleImpl and GroupImpl
- replace usage of guava Iterators.concat in AuthorizableIterator with a custom utility in oak-commons

i hope that this allows us to fully understand why the guava concat fails with guava 20.
once we have that and conclude that the proposed replacement is reasonable, i would replace all usages of guave Iterators.concat in oak security and if used other modules.

### DIFF
--- a/oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/DynamicGroupsTest.java
+++ b/oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/DynamicGroupsTest.java
@@ -302,11 +302,11 @@ public class DynamicGroupsTest extends DynamicSyncContextTest {
         String groupId2 = declaredGroupRefs.get(1).getId();
         Group local = um.createGroup("localGroup");
         local.addMembers(groupId, groupId2);
-        userManager.createGroup(EveryonePrincipal.getInstance());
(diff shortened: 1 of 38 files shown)


reschke:
I'm going to review and test during the day.

reschke:
I can confirm this passes tests with Guava 20.

That said, I'm a little worried about the approach. We have no evidence that "Iterators.concat()" is indeed broken in Guava 20, so this might be exposing a bug in our implementations that we might be just hiding right now.+

Would it be possible to separate the test changes into a distinct ticket/commit and do that first? I'm willing to help with the analysis once I'm done with my current emergency :-).

anchela:
> Would it be possible to separate the test changes into a distinct ticket/commit and do that first?

should be doable.

anchela:
@reschke , wdyt?

reschke (commented):
I don't think we need this. It essentially hides a Guava bug; it's better just not to use that Guava version.

anchela:
@reschke , ok i am closing this now. we can still bring it back if we see a need for it in the future.

PR #880, opened 2023-03-29
OAK-10162: Fix Version copier with preserveOnTarget to ignore diverge…

amit-jain:
…d history

When preserveOnTarget is true
- Ignoring diverged version history or if no change detected
- Not changing version properties if these properties are null in the source which prevents linkage to new successor versions being broken

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/NodeStateCopier.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/NodeStateCopier.java
@@ -16,8 +16,10 @@
  */
 package org.apache.jackrabbit.oak.plugins.migration;
 
+import org.apache.jackrabbit.JcrConstants;
 import org.apache.jackrabbit.oak.api.CommitFailedException;
 import org.apache.jackrabbit.oak.api.PropertyState;
+import org.apache.jackrabbit.oak.api.Type;
 import org.apache.jackrabbit.oak.commons.PathUtils;
 import org.apache.jackrabbit.oak.plugins.memory.EmptyNodeState;
 import org.apache.jackrabbit.oak.spi.commit.CommitHook;
@@ -32,8 +34,11 @@ import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.HashSet;
 import java.util.Set;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.collect.ImmutableSet.copyOf;
@@ -145,28 +150,72 @@ public class NodeStateCopier {
      *
      * @param source The NodeState to copy from.
      * @param target The NodeBuilder to copy to.
+     * @param preserveOnTarget boolean to indicate no changes on target except additions
+     * @param path current path
      * @return Whether changes were made or not.
      */
-    public static boolean copyProperties(NodeState source, NodeBuilder target) {
+    public static boolean copyProperties(NodeState source, NodeBuilder target, boolean preserveOnTarget, String path) {
         boolean hasChanges = false;
 
         // remove removed properties
-        for (final PropertyState property : target.getProperties()) {
-            final String name = property.getName();
-            if (!source.hasProperty(name)) {
-                target.removeProperty(name);
-                hasChanges = true;
+        if (!preserveOnTarget) {
+            for (final PropertyState property : target.getProperties()) {
+                final String name = property.getName();
+                if (!source.hasProperty(name)) {
+                    target.removeProperty(name);
+                    hasChanges = true;
+                }
             }
         }
 
         // add new properties and change changed properties
         for (PropertyState property : source.getProperties()) {
             if (!property.equals(target.getProperty(property.getName()))) {
-                target.setProperty(property);
-                hasChanges = true;
+                if (!isVersionPropertyEmpty(source, property, preserveOnTarget, path)) {
+                    target.setProperty(property);
+                    hasChanges = true;
+                }
             }
         }
         return hasChanges;
+    }
+    
+    private static Set<String> getValues(NodeState nodeState, String prop) {
+        if (nodeState.hasProperty(prop)) {
+            Iterable<String> values = nodeState.getProperty(prop).getValue(Type.STRINGS);
+            return StreamSupport.stream(values.spliterator(), false).filter(s -> !s.isEmpty()).collect(Collectors.toSet());
+        }
+        return new HashSet<>();
+    }
+    
+    private static boolean isVersionPropertyEmpty(NodeState source, PropertyState property, boolean preserveOnTarget,
+        String path) {
+        if (preserveOnTarget) {
+            if (property.getName().equals(JcrConstants.JCR_UUID) || 
+                property.getName().equals(JcrConstants.JCR_SUCCESSORS) || 
+                property.getName().equals(JcrConstants.JCR_PREDECESSORS)) {
+                boolean versionPropertyEmpty = getValues(source, property.getName()).isEmpty();
+                if (versionPropertyEmpty) {
+                    LOG.info("Version Property {} will be skipped for path {}", property.getName(), path);
+                } else {
+                    LOG.info("Version Property {} will be changed for path {}", property.getName(), path);
+                }
+                return versionPropertyEmpty;
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Copies all changed properties from the source NodeState to the target
+     * NodeBuilder instance.
+     *
+     * @param source The NodeState to copy from.
+     * @param target The NodeBuilder to copy to.
+     * @return Whether changes were made or not.
+     */
+    public static boolean copyProperties(NodeState source, NodeBuilder target) {
+        return copyProperties(source, target, false, "");
     }
 
     private boolean copyNodeState(@NotNull final NodeState sourceRoot, @NotNull final NodeBuilder targetRoot) {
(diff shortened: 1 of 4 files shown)


mreutegg (commented):
Looks good. I have some minor suggestions for improvements: https://github.com/apache/jackrabbit-oak/pull/884

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/NodeStateCopier.java
```diff
@@ -145,29 +150,73 @@ public static boolean copyNodeStore(@NotNull final NodeStore source, @NotNull fi
...
         }
         return hasChanges;
     }
+    
+    private static Set<String> getValues(NodeState nodeState, String prop) {
+        if (nodeState.hasProperty(prop)) {
+            Iterable<String> values = nodeState.getProperty(prop).getValue(Type.STRINGS);
+            return StreamSupport.stream(values.spliterator(), false).filter(s -> !s.isEmpty()).collect(Collectors.toSet());
+        }
+        return new HashSet<>();
```
rishabhdaim: Implementation of this method only uses the size of `set` returned by this method. So instead of creating an `set` of non empty properties, we should simply return a boolean indicating whether it is empty or not.
```suggestion
    private static boolean isVersionEmpty(NodeState nodeState, String prop) {
        if (nodeState.hasProperty(prop)) {
            Iterable<String> values = nodeState.getProperty(prop).getValue(Type.STRINGS);
            return StreamSupport.stream(values.spliterator(), false).allMatch(String::isEmpty);
        }
        return true;
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/NodeStateCopier.java
```diff
@@ -145,29 +150,73 @@ public static boolean copyNodeStore(@NotNull final NodeStore source, @NotNull fi
...
+    }
+    
+    private static boolean isVersionPropertyEmpty(NodeState source, PropertyState property, boolean preserveOnTarget,
+        String path) {
+        if (preserveOnTarget) {
+            if (property.getName().equals(JcrConstants.JCR_UUID) || 
+                property.getName().equals(JcrConstants.JCR_SUCCESSORS) || 
+                property.getName().equals(JcrConstants.JCR_PREDECESSORS)) {
+                boolean versionPropertyEmpty = getValues(source, property.getName()).isEmpty();
```
rishabhdaim, line 198:
```suggestion
                boolean versionPropertyEmpty = isVersionEmpty(source, property.getName());
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/NodeStateCopier.java
```diff
@@ -145,29 +150,73 @@ public static boolean copyNodeStore(@NotNull final NodeStore source, @NotNull fi
...
+        if (nodeState.hasProperty(prop)) {
+            Iterable<String> values = nodeState.getProperty(prop).getValue(Type.STRINGS);
+            return StreamSupport.stream(values.spliterator(), false).filter(s -> !s.isEmpty()).collect(Collectors.toSet());
+        }
+        return new HashSet<>();
+    }
+    
+    private static boolean isVersionPropertyEmpty(NodeState source, PropertyState property, boolean preserveOnTarget,
+        String path) {
+        if (preserveOnTarget) {
```
rishabhdaim, line 194: I would simply return if `preserveOnTarget` is false.
```suggestion
        if (!preserveOnTarget) {
        return false;
```
then to go inside if condition.
As a general coding practice, I found:

`if not condition:
  do little stuff
  return
// now...
do lots of stuff.`

increases the code readability and reduce cyclomatic complexity.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/version/VersionCopier.java
```diff
@@ -111,6 +125,72 @@ public boolean copyVersionHistory(String versionableUuid, Calendar minDate, bool
         return false;
     }
 
+    private boolean hasNoConflicts(String versionHistoryPath, String versionableUuid, boolean preserveOnTarget, NodeState sourceVersionHistory) {
+        // if preserveOnTarget is true then check no conflicts which means version history has moved forward only
+        if (preserveOnTarget) {
```
rishabhdaim, line 131: same as above, return if `preserveOnTarget` is false.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/migration/version/VersionCopierTest.java
```diff
@@ -44,18 +50,15 @@
 import org.junit.Test;
 
 import static java.util.Collections.singletonList;
-import static org.apache.jackrabbit.JcrConstants.JCR_FROZENNODE;
-import static org.apache.jackrabbit.JcrConstants.JCR_MIXINTYPES;
-import static org.apache.jackrabbit.JcrConstants.JCR_PRIMARYTYPE;
-import static org.apache.jackrabbit.JcrConstants.JCR_UUID;
-import static org.apache.jackrabbit.JcrConstants.MIX_VERSIONABLE;
-import static org.apache.jackrabbit.JcrConstants.NT_UNSTRUCTURED;
+import static org.apache.jackrabbit.JcrConstants.*;
```
rishabhdaim: Please import them individually.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/version/VersionCopier.java
```diff
@@ -111,6 +125,72 @@ public boolean copyVersionHistory(String versionableUuid, Calendar minDate, bool
...
+            NodeBuilder targetVersionHistory = getVersionHistoryBuilder(targetVersionStorage, versionableUuid);
+            if (targetVersionHistory.exists()) {
+                VersionComparator versionComparator = new VersionComparator();
+
+                // version history id not equal
+                boolean conflictingVersionHistory = !targetVersionHistory.getString(JCR_UUID).equals(sourceVersionHistory.getString(JCR_UUID));
```
rishabhdaim: Since `NodeBuilder.getString()` can return null, I would compare them using `Objects.equal`

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/version/VersionCopier.java
```diff
@@ -111,6 +125,72 @@ public boolean copyVersionHistory(String versionableUuid, Calendar minDate, bool
...
+                // all source versions already exist on target (diverged or no diff)
+                boolean diverged = !targetVersions.contains(sourceVersions.get(0));
+                if (diverged) {
+                    logger.info("Skipping version history for {}: Versions diverged", versionHistoryPath);
+                    return false;
+                }
+
+                // highest source version UUID does not match the corresponding version on target (diverged)
+                boolean conflictingHighestVersion =
+                    !sourceVersionHistory.getChildNode(sourceVersions.get(0)).getString(JCR_UUID).equals(targetVersionHistory.getChildNode(sourceVersions.get(0)).getString(JCR_UUID));
```
rishabhdaim: Please use `Objects.equal`

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/version/VersionCopier.java
```diff
@@ -111,6 +125,72 @@ public boolean copyVersionHistory(String versionableUuid, Calendar minDate, bool
...
+                if (conflictingHighestVersion) {
+                    logger.info("Skipping version history for {}: Old base version id changed", versionHistoryPath);
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
+
+    static class VersionComparator implements Comparator<String> {
```
rishabhdaim, line 183: We are making a reverse comparison here. 
Shouldn't we make a simple comparison and then where we are using we should explicitly call `comparator.reversed()` to highlight that we are using a reverse comparison, wdyt?

rishabhdaim (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/NodeStateCopier.java
amit-jain: I had left it to leave room for furture in case needed but not a strong opinion here.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/NodeStateCopier.java
amit-jain: Here it is not much stuff to do and also adds another if condition but again not a strong opinion.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/version/VersionCopier.java
amit-jain: same reasoning as above, the code is currently concise and change will end up having multiple if blocks

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/version/VersionCopier.java
amit-jain: Maybe adding javadoc will convey the same purpose

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/version/VersionCopier.java
amit-jain: we are in the version history node path which will mandatorily contain jcr:uuid property but Objects.equal comparison is good.

rishabhdaim (approved):

PR #882, opened 2023-03-29
OAK-10163 : add benchmark test cases to measure performance using que…

rishabhdaim:
…ry with projection


rishabhdaim:
@mreutegg @stefan-egli could you please review and share the feedback.

mreutegg (approved):
Looks good to me.

PR #885, opened 2023-03-30
OAK-10170 : simplify usage of authorizableiterator

anchela:
@reschke , as discussed. in a next step i will also adjust the size-calculation such that it returns -1 if duplicates are filtered.


reschke (approved):
LGTM

PR #890, opened 2023-04-04
OAK-10180: switch oak-core-spi to shaded guava

reschke:


oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/gc/DelegatingGCMonitor.java
```diff
@@ -41,15 +40,15 @@ public class DelegatingGCMonitor implements GCMonitor {
      * @param gcMonitors
      */
     public DelegatingGCMonitor(@NotNull Collection<? extends GCMonitor> gcMonitors) {
-        this.gcMonitors = newConcurrentHashSet();
+        this.gcMonitors = Collections.synchronizedSet(new HashSet<>());
```
rishabhdaim: Shouldn't we use `ConcurrentHashMap.newKeySet()` ?

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/gc/DelegatingGCMonitor.java
```diff
@@ -41,15 +40,15 @@ public class DelegatingGCMonitor implements GCMonitor {
...
         this.gcMonitors.addAll(gcMonitors);
     }
 
     /**
      * New instance without any delegate.
      */
     public DelegatingGCMonitor() {
-        this(Sets.<GCMonitor>newConcurrentHashSet());
+        this(Collections.synchronizedSet(new HashSet<>()));
```
rishabhdaim: Same as above

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/lock/LockConstants.java
```diff
@@ -16,14 +16,15 @@
...
 
-import com.google.common.collect.ImmutableSet;
 import org.apache.jackrabbit.JcrConstants;
 
 public interface LockConstants extends JcrConstants {
 
-    Set<String> LOCK_PROPERTY_NAMES = ImmutableSet.of(
-            JCR_LOCKISDEEP,
-            JCR_LOCKOWNER);
+    Set<String> LOCK_PROPERTY_NAMES = Collections
```
rishabhdaim: I believe the goal is to upgrade `guava` not remove it. So we can keep using it, instead of moving this to Java, wdyt ?
cc @mreutegg

reschke: I believe we should avoid use of Guava where we can; maybe at some later point of time we might decide to remove it.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/gc/DelegatingGCMonitor.java
reschke: Thanks.

PR #896, opened 2023-04-12
OAK-10186: switch oak-store-spi to shaded guava

reschke:


oak-it/src/test/java/org/apache/jackrabbit/oak/composite/CompositeNodeStoreTest.java
```diff
@@ -88,6 +86,9 @@
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
 
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Lists;
```
anchela: shouldn't those 2 imports also be replaced by org.apache.jackrabbit.guava.common.* ?

oak-it/src/test/java/org/apache/jackrabbit/oak/composite/CompositeNodeStoreTest.java
```diff
@@ -18,9 +18,9 @@
  */
 package org.apache.jackrabbit.oak.composite;
 
-import static com.google.common.base.Predicates.compose;
-import static com.google.common.collect.Iterables.filter;
 import static com.google.common.collect.Lists.newArrayList;
+import static org.apache.jackrabbit.guava.common.base.Predicates.compose;
+import static org.apache.jackrabbit.guava.common.collect.Iterables.filter;
```
anchela, line 22: for the sake of easy review it would be preferable if the imports would remain on the same line (even if that makes the order a bit off.

oak-it/src/test/java/org/apache/jackrabbit/oak/composite/CompositeNodeStoreTest.java
```diff
@@ -18,9 +18,9 @@
  */
 package org.apache.jackrabbit.oak.composite;
 
-import static com.google.common.base.Predicates.compose;
-import static com.google.common.collect.Iterables.filter;
 import static com.google.common.collect.Lists.newArrayList;
```
anchela: imho that static import of com.google.common.collect.Lists.newArrayList should also be replaced, no?

oak-it/src/test/java/org/apache/jackrabbit/oak/composite/CompositeNodeStoreTest.java
```diff
@@ -46,12 +46,10 @@
...
 
-import com.google.common.base.Predicate;
-import com.google.common.base.Predicates;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Iterables;
-import com.google.common.collect.Lists;
 import org.apache.commons.io.FileUtils;
+import org.apache.jackrabbit.guava.common.base.Predicate;
+import org.apache.jackrabbit.guava.common.base.Predicates;
+import org.apache.jackrabbit.guava.common.collect.Iterables;
```
anchela, line 50: same here... if the imports were on the same lines as before it would be easier to review (and see that some imports didn't get adjusted.

oak-store-spi/src/main/java/org/apache/jackrabbit/oak/plugins/value/jcr/BinaryImpl.java
```diff
@@ -16,7 +16,7 @@
  */
 package org.apache.jackrabbit.oak.plugins.value.jcr;
 
-import static com.google.common.base.Objects.toStringHelper;
+import static org.apache.jackrabbit.guava.common.base.MoreObjects.toStringHelper;
```
anchela, line 19: why does Objects become MoreObjects? that looks like a mistake.

anchela (changes requested):

oak-it/src/test/java/org/apache/jackrabbit/oak/composite/CompositeNodeStoreTest.java
reschke: Yes. Will improve.

oak-it/src/test/java/org/apache/jackrabbit/oak/composite/CompositeNodeStoreTest.java
reschke: I tried, but once we do that consistently, the diffs get really big. Let's take care of these things later on.

oak-store-spi/src/main/java/org/apache/jackrabbit/oak/plugins/value/jcr/BinaryImpl.java
reschke: See https://guava.dev/releases/19.0/api/docs/com/google/common/base/MoreObjects.html

PR #900, opened 2023-04-17
OAK-10192: switch oak-query-spi to shaded guava

reschke:


oak-query-spi/pom.xml
```diff
@@ -39,7 +39,6 @@
           <configuration>
             <instructions>
               <Import-Package>
-                <!-- OAK-7182 -->${guava.osgi.import},
```
mreutegg, line 42: I think the entire Import-Package instruction can be removed. Importing everything that is required is the [default behavior](https://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html#_default_behavior).

reschke: Yep. I'm planning to a cleanup if all POMs once we're done.

mreutegg (approved):

mreutegg (approved):

oak-query-spi/src/main/java/org/apache/jackrabbit/oak/spi/query/CompositeQueryIndexProvider.java
```diff
@@ -23,8 +23,8 @@
 import org.apache.jackrabbit.oak.spi.state.NodeState;
 import org.jetbrains.annotations.NotNull;
 
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Lists;
+import org.apache.jackrabbit.guava.common.collect.ImmutableList;
```
rishabhdaim, line 26: Since we have moved to Java 11, we should use `List.of()` instead of `ImmutableList`.

reschke: Nope. a) I've decided to do only mechanical changes here. b) There may be cases where we want to avoid unnecessary changes that can not be backported to 1.22. So I'll leave this to the actual "owners" of the module.

rishabhdaim (approved):

oak-query-spi/src/main/java/org/apache/jackrabbit/oak/spi/query/CompositeQueryIndexProvider.java
reschke: Nope. a) I've decided to do only mechanical changes here. b) There may be cases where we want to avoid unnecessary changes that can not be backported to 1.22. So I'll leave this to the actual "owners" of the module.

PR #903, opened 2023-04-20
OAK-10196: switch oak-store-document to shaded guava

reschke:


oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/CacheInvalidationIT.java
```diff
@@ -39,7 +39,7 @@
 import org.junit.Before;
 import org.junit.Test;
 
-import static com.google.common.collect.Iterables.*;
+import static org.apache.jackrabbit.guava.common.collect.Iterables.*;
```
rishabhdaim, line 42: could you please import them individually

reschke: No. I'm not going to make any edits that will make the change bigger than necessary.

I agree that wildcard imports are bad, but if we want to change them, that should be done as part of a separate issue.

rishabhdaim (approved):

PR #915, opened 2023-04-28
OAK-10228 : Explain effect of policies for unknown principals and non-existing paths

kwin:


anchela (changes requested):
hi @kwin , i find the mentioning of 'referential integrity' confusing as this is in JCR world associated with referenceable nodes and properties of type REFERENCE. However, neither path nor principal names properties are of type REFERENCE (or WEAK-REFERENCE) and thus the notion of referential integrity does not apply here.

I agree though that adding a paragraph explaining the behavior in case of removal of nodes and principals is indeed helpful. thanks for taking the time to clarify that.

ps: can we also have a jira ticket associated with the change? as a description you could just link to the dev-list conversation.

kwin:
@anchela Any proposal for a snappy headline?

anchela:
@kwin , not sure if am good at 'snappy'..... having said that i would go with

`Unknown principals and non-existing paths`
or
`Effect with unknown principals and non-existing paths`

if you want to elaborate a bit more you could also include (in addition to the mentioning that there is no automatic cleanup):
- restrictions limit the effect of a given access control entry to certain items which may or may not exist (-> link to restriction extension)
- creation of new ACEs for non-existing principals is governed by the 'import-behavior' (-> link to import behavior)

does that make sense?

anchela:
@kwin , i created a ticket for this change and adjusted the summary. hope that's ok

kwin:
@anchela I updated this PR accordingly.

Regarding

> restrictions limit the effect of a given access control entry to certain items which may or may not exist

This is for me a bit outside the scope of unknown principals/non-existing paths so let us rather tackle separately.

kwin:
@anchela Ping, any more feedback? Otherwise I would merge tomorrow.

PR #920, opened 2023-05-02
OAK-10213 : updated DocumentStore API to support bulk modify request

rishabhdaim:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1147,6 +1155,85 @@ private <T extends Document> T findAndModify(Collection<T> collection,
...
+     * @param collection the collection
+     * @param updateOps  the update operation List
+     * @return the list containing old documents or <code>null</code> if the condition is not met
+     * or if the document wasn't found. The order in the result list reflects the order
+     * in the updateOps parameter
+     * @throws DocumentStoreException if the operation failed. E.g. because of
+     *                                an I/O error.
+     * @see #createOrUpdate(Collection, List)
+     */
+    @Override
```
stefan-egli, line 1195: Looks similar to the javadoc of the default method. Do we need it here, ie is there a difference?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1147,6 +1155,85 @@ private <T extends Document> T findAndModify(Collection<T> collection,
...
+    @NotNull
+    public <T extends Document> List<T> findAndUpdate(final @NotNull Collection<T> collection, final @NotNull List<UpdateOp> updateOps) {
```
stefan-egli, line 1197: There are quite some similarities between `findAndUpdate<bulk>` and `createOrUpdate<bulk>`. The singular variants of these `findAndUpdate<single>` and `createOrUpdate<single>` do in fact share the same underlying `findAndModify`. I was wondering if the same could be done for the bulk method (in which case less code would have to be duplicated). Wdyt?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStore.java
```diff
@@ -540,4 +542,33 @@ default <T extends Document> List<T> query(final Collection<T> collection,
...
+     * @param updateOps the update operation List
+     * @return the list containing old documents or <code>null</code> if the condition is not met
+     *         or if the document wasn't found. The order in the result list reflects the order
+     *         in the updateOps parameter
+     * @throws DocumentStoreException if the operation failed. E.g. because of
+     *          an I/O error.
+     */
+    @NotNull
+    default <T extends Document> List<T> findAndUpdate(final @NotNull Collection<T> collection,
+                                                       final @NotNull List<UpdateOp> updateOps) throws DocumentStoreException {
```
stefan-egli, line 571: I agree `@NotNull` would be appropriate, then again all the other methods in this interface do not have it neither, so this method does stand a bit out. I would probably not use them here... 
(if we want to apply these annotations to this interface we could always do it via a separate ticket - but as this is an internal interface to the DocumentNodeStore, I don't see it that important)

stefan-egli (commented):
Generally like having this new interface method. I have added a comment below though to look into whether we can merge the bulk methods createOrUpdate and findAndUpdate (as currently there seems to be quite a bit of duplicate code)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
rishabhdaim: Yes, it can be removed.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
rishabhdaim: Yes, I agree that there are some similarities between `findAndUpdate<bulk>` and `createOrUpdate<bulk>`, especially the code flow, but the similarities end there.

The logic to calculate the resultSet, bulk API to call, updating the cache with updated documents, etc. are handled differently. 

In order to reuse `findAndModify` for both `findAndUpdate<single>` and `createOrUpdate<single>`, we use many boolean flags, and that IMHO only complicates the code.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStore.java
rishabhdaim: I think it's a small step in the right direction (avoiding nulls) and instead of removing `@NotNull`here, we should create a new ticket to update the existing methods if possible.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
stefan-egli: IIUC then in the current state of the branch the javadoc is still there. Did you fix it elsewhere, i.e. is the correct state of your changes pushed already?

rishabhdaim: These are different javadocs as compared to the one earlier which was copy-pasted from the interface. These java docs, explain the functionality of this method.

stefan-egli:
@rishabhdaim , you mentioned offline a fix you did with a different return value, but I can't remember all the details and can't find the diff due to the force pushes. Which change were you referring to?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
stefan-egli: I agree those boolean flags might not be most user friendly. However, the MongoDocumentStore already works that way, i.e. people reading this class will already be somewhat familiar with them - in particular if they are the same for `<single>` and `<bulk>`.

Some additional aspects relevant for this discussion:
* `sendBulkModify` and `sendBulkUpdate` are actually quite similar. I haven't tried it, but it seems like it could be deduplicated with perhaps two boolean flags. If that were the case, i think the gain would be quite large - as we'd only have to maintain 1 variant.
* The other pairs indeed differ a bit more - but I would still see an advantage with deduplication based solely on maintenance (unless the code then becomes entirely unreadable)
* One more related point I think is test coverage. `createOrUpdate<bulk>` has a fair amount of test coverage which includes dedicated classes `BulkCreateOrUpdateTest` and `BulkCreateOrUpdateClusterTest` (besides obviously being used in several higher level use cases). Given the new `findAndUpdate<bulk>` would be new, it would be actually unused at this stage - so it might not immediately require the same test coverage. Having said that, I would argue we should at least look into adding some more test coverage (perhaps using the `createOrUpdate<bulk>` variants as input for ideas or even generalize them (but I think the latter is not necessarily possible).
* If we had full deduplication however, then we would profit from the test coverage for `createOrUpdate<bulk>` to some extend. That could be useful to have as a starting point.

Wdyt?

rishabhdaim: I have updated the code to merge `sendBulkModify` and `sendBulkUpdate` into one method `sendBulkRequest` and used boolean flags to handle upsert & modify requests.

Also, One internal class `BulkUpdateResult` had been modified to `BulkRequestResult` to handle both upsert & modify requests.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1270,6 +1361,86 @@ private Map<String, NodeDocument> getCachedNodes(Set<String> keys) {
...
+                } else {
+                    // some documents might have not been updated, lets fetch them from database
+                    // and found out which had not been updated
+                    updatedDocsMap.forEach((key, value) -> {
+                        T oldDoc = oldDocs.get(key);
+                        if (isNull(oldDoc) || oldDoc == NULL || Objects.equals(oldDoc.getModCount(), value.getModCount())) {
+                            // simply ignore updating the document cache in case
+                            // 1. oldDoc is null
+                            // 2. document didn't get updated i.e. modCount is same after update operation
+                            log("Skipping updating doc cache for {}", key);
```
mreutegg: The log() method does not support markers in the message.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1270,6 +1361,86 @@ private Map<String, NodeDocument> getCachedNodes(Set<String> keys) {
...
+            // document might have been updated, if updated then add oldDoc else add null to result
+            bulkOperations.entrySet().stream().filter(e -> !bulkResult.failedUpdates.contains(e.getKey())).forEach(e -> {
+                T updated = updatedDocsMap.get(e.getKey());
+                T oldDoc = oldDocs.get(e.getKey());
+                if (oldDoc == null || oldDoc == NULL || Objects.equals(oldDoc.getModCount(), updated.getModCount())) {
+                    // add null value in result cause, either this document didn't exist
+                    // at time of modify operation, and we didn't anything for it
+                    // or oldDoc is present and modCount is same,
+                    // so document had not been updated.
+                    log("{} didn't get updated, returning null.", e.getKey());
```
mreutegg: Same as above.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStoreStats.java
```diff
@@ -277,12 +277,22 @@ public void doneCreateOrUpdate(long timeTakenNanos, Collection<? extends Documen
...
     }
 
+    @Override
+    public void doneFindAndModify(long timeTakenNanos, Collection<? extends Document> collection, List<String> ids, boolean newEntry,
+                                  boolean success, int retryCount) {
+
+        modifyMetricUpdater.update(collection, retryCount, timeTakenNanos, success, newEntry, ids, isNodesCollectionUpdated(), getStatsConsumer(),
+                getStatsConsumer(), MeterStats::mark, MeterStats::mark);
+
+        perfLog(perfLog, PERF_LOG_THRESHOLD, timeTakenNanos, "findAndModify [{}]", ids);
```
mreutegg: Remove square brackets? List.toString() usually surrounds elements with square brackets already.
```suggestion
        perfLog(perfLog, PERF_LOG_THRESHOLD, timeTakenNanos, "findAndModify {}", ids);
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStoreStats.java
```diff
@@ -277,12 +277,22 @@ public void doneCreateOrUpdate(long timeTakenNanos, Collection<? extends Documen
...
 
-        modifyMetricUpdater.update(collection, retryCount, timeTakenNanos, success, newEntry, c -> c == NODES, getStatsConsumer(),
+        modifyMetricUpdater.update(collection, retryCount, timeTakenNanos, success, newEntry, of(key), isNodesCollectionUpdated(), getStatsConsumer(),
                 getStatsConsumer(), MeterStats::mark, MeterStats::mark);
 
         perfLog(perfLog, PERF_LOG_THRESHOLD, timeTakenNanos, "findAndModify [{}]", key);
     }
 
+    @Override
+    public void doneFindAndModify(long timeTakenNanos, Collection<? extends Document> collection, List<String> ids, boolean newEntry,
```
mreutegg: Can you please add tests for this method in `DocumentStoreStatsIT` and `DocumentStoreStatsTest`? Similar to method variant with single key/id.

mreutegg (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStoreStatsCollector.java
```diff
@@ -85,6 +85,19 @@ void doneQuery(long timeTakenNanos, Collection<? extends Document> collection, S
     void doneFindAndModify(long timeTakenNanos, Collection<? extends Document> collection, String key,
                            boolean newEntry, boolean success, int retryCount);
 
+    /**
+     * Called when a update operation was completed which could have affected multiple
+     * documents.
+     * @param timeTakenNanos time taken
+     * @param collection the collection
+     * @param ids List of document ids which got updated
+     * @param newEntry true if the document was newly created due to given operation
```
mreutegg: Parameter `newEntry` looks strange to me. This method is about multiple documents. The only call to this method passes `false` and the new 'bulk' method DocumentStore.findAndUpdate() does not allow upsert or insert, right? I think the parameter should be removed.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/ThrottlingStatsCollectorImpl.java
```diff
@@ -126,12 +126,22 @@ public void doneCreateOrUpdate(long throttlingTimeNanos, Collection<? extends Do
...
     }
 
+    @Override
+    public void doneFindAndModify(long throttlingTimeNanos, Collection<? extends Document> collection, List<String> ids, boolean newEntry,
+                                  boolean success, int retryCount) {
+
+        modifyMetricUpdater.update(collection, retryCount, throttlingTimeNanos, success, newEntry, ids, isNodesCollectionUpdated(),
+                getStatsConsumer(), getStatsConsumer(), MeterStats::mark, MeterStats::mark);
+
+        perfLog(perfLog, PERF_LOG_THRESHOLD, throttlingTimeNanos, "findAndModify [{}]", ids);
```
mreutegg: Remove square brackets?
```suggestion
        perfLog(perfLog, PERF_LOG_THRESHOLD, throttlingTimeNanos, "findAndModify {}", ids);
```

reschke (commented):
See comment in ticket: do we absolutely need this in the API? Historically, we have used store-specific code already; I would propose to do that first, and move that into an API iny later on if it is needed in more than a single place.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ThrottlingDocumentStoreWrapper.java
```diff
@@ -190,6 +191,20 @@ public <T extends Document> T findAndUpdate(final Collection<T> collection, fina
...
+    public <T extends Document> List<T> findAndUpdate(@NotNull Collection<T> collection, @NotNull List<UpdateOp> updateOps) {
+        final long throttlingTime = performThrottling(collection);
+        final List<T> results = new ArrayList<>();
+        try {
+            results.addAll(store.findAndUpdate(collection, updateOps));
+        } finally {
+            throttlingStatsCollector.doneFindAndModify(MILLISECONDS.toNanos(throttlingTime), collection,
+                    updateOps.stream().map(UpdateOp::getId).collect(toList()), true, 0);
+        }
+        return results;
```
mreutegg: Adding documents to `results` is unnecessary.
```suggestion
        final long throttlingTime = performThrottling(collection);
        try {
            return store.findAndUpdate(collection, updateOps);
        } finally {
            throttlingStatsCollector.doneFindAndModify(MILLISECONDS.toNanos(throttlingTime), collection,
                    updateOps.stream().map(UpdateOp::getId).collect(toList()), false, true, 0);
        }
```

rishabhdaim:
Thanks, [Julian Reschke](https://github.com/jira/secure/ViewProfile.jspa?name=reschke) for the feedback. However, I see the following issues with putting the new `findAndUpdate` API into MongoVersionGCSupport:



- I don't see it as the right candidate for database-related operations. It's more of a support class for Mongo-related GC ops.

- Having an API in the DocumentStore gives us the following advantages:

1. We are sure of the lease check since it will be done automatically for us. We already have an epic to solve issues which are caused by late-writes i.e. writing after the lease had expired.

1. We can take into account throttling as well. I really don't want to perform heavy garbage collection operations if my system is already under heavy load.

1. I can reuse some of the already existing code.

1. We won't miss the stats for those operations which are done inside DocumentStore.

reschke:
On 26.05.2023 14:07, Rishabh Kumar wrote:
> ...

OK; makes sense.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MongoDocumentStoreIT.java
```diff
@@ -268,4 +277,109 @@ public void queryWithNullProjection() {
...
+                    new RevisionVector(rev), of(new LongPropertyState("prop", 10L)), false, null);
+            inserts.add(n.asOperation(rev));
+        }
+        docStore.create(Collection.NODES, inserts);
+
+        List<UpdateOp> updateOps = new ArrayList<>(10);
+        for (int i = 0; i < 10; i++) {
+            UpdateOp updateOp = new UpdateOp(getIdFromPath(fromString("/node-" + i)), false);
+            updateOp.set("prop", 20L);
+            updateOp.contains("prop", true);
```
mreutegg: This is a rather unusual and a bit confusing. The test setup above create a revision based property 'prop', whereas this update here sets a top level property 'prop' without a revision. These UpdateOps replace the 'prop' sub document with the initial revision with a field of the same name. This is not something the DocumentNodeStore ever does. I think it would be more realistic to update the document with an additional revision in the 'prop' sub document.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MongoDocumentStoreIT.java
```diff
@@ -268,4 +277,109 @@ public void queryWithNullProjection() {
...
+            updateOps.add(updateOp);
+        }
```
mreutegg:
```suggestion
        Revision updateRev = Revision.newRevision(0);
        for (int i = 0; i < 10; i++) {
            UpdateOp updateOp = new UpdateOp(getIdFromPath(fromString("/node-" + i)), false);
            updateOp.setMapEntry("prop", updateRev, "20");
            updateOps.add(updateOp);
        }
```

mreutegg (changes requested):

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MongoDocumentStoreIT.java
```diff
@@ -268,4 +277,109 @@ public void queryWithNullProjection() {
...
+        docStore.create(Collection.NODES, inserts);
+
+        List<UpdateOp> updateOps = new ArrayList<>(10);
+        for (int i = 0; i < 10; i++) {
+            UpdateOp updateOp = new UpdateOp(getIdFromPath(fromString("/node-" + i)), false);
+            updateOp.set("prop", 20L);
+            updateOp.contains("prop", false);
+            updateOps.add(updateOp);
+        }
```
mreutegg: Similar to previous comment. I think it would be better to base the condition on a field that is not the sub document with the revision based prop values. E.g. it could check if the _modified field is absent, which is never the case here.
```suggestion
        Revision updateRev = Revision.newRevision(0);
        for (int i = 0; i < 10; i++) {
            UpdateOp updateOp = new UpdateOp(getIdFromPath(fromString("/node-" + i)), false);
            updateOp.setMapEntry("prop", updateRev, "20");
            updateOp.contains(MODIFIED_IN_SECS, false);
            updateOps.add(updateOp);
        }
```

mreutegg (changes requested):

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MongoDocumentStoreIT.java
```diff
@@ -268,4 +277,109 @@ public void queryWithNullProjection() {
...
+            UpdateOp updateOp = new UpdateOp(getIdFromPath(fromString("/node-" + i)), false);
+            updateOp.set("prop", 20L);
+            updateOp.contains("prop", false);
+            updateOps.add(updateOp);
+
+            updateOp = new UpdateOp(getIdFromPath(fromString("/node-" + (i+100))), false);
+            updateOp.set("prop2", 20L);
+            updateOp.contains("prop", false);
+            updateOps.add(updateOp);
+        }
```
mreutegg: Similar to previous comment. Can you please update accordingly?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1282,6 +1373,86 @@ private Map<String, NodeDocument> getCachedNodes(Set<String> keys) {
...
+        if (collection == NODES) {
+            tracker = nodesCache.registerTracker(bulkOperations.keySet());
+        }
+
+        try {
+            final BulkRequestResult bulkResult = sendBulkRequest(collection, bulkOperations.values(), oldDocs, false);
+            final Set<String> potentiallyUpdatedDocsSet = difference(bulkOperations.keySet(), bulkResult.failedUpdates);
+
+            // fetch all the docs which haven't failed, they might have passed
+            final Map<String, T> updatedDocsMap = findDocuments(collection, potentiallyUpdatedDocsSet);
```
mreutegg: Isn't this rather expensive? Even in the happy case, this will read all modified documents again from MongoDB, right?

rishabhdaim: Yes, you are right, even in normal cases it will fetch all the documents. I will fix this.

mreutegg (approved):
Looks good to me now.

stefan-egli:
+1, 
PS: as mentioned earlier, a merge of the `bulk` methods would have reused quite some existing tests. Since we now don't have that, I would suggest to focus on good test coverage of use cases of this new API instead.

PR #921, opened 2023-05-03
OAK-10221: switch oak-auth-external to shaded guava

reschke:


anchela (approved):
@reschke , lgtm. thanks

PR #924, opened 2023-05-03
OAK-10227: switch oak-auth-ldap to shaded guava

reschke:


anchela (approved):
lgtm.
i noticed however that the changes made to the pom.xml look slightly different than with other modules. why?

reschke:
...edits in the POMs are manual, and I try to minimize them.

PR #936, opened 2023-05-12
OAK-10239 Switch oak-lucene to shaded Guava

jelmini:


oak-lucene/pom.xml
```diff
@@ -220,9 +219,9 @@
 
     <!-- General utility libraries -->
     <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-      <scope>provided</scope>
+      <groupId>org.apache.jackrabbit</groupId>
```
reschke: ...already present above due to the change for oak-search

jelmini: Removed

PR #939, opened 2023-05-15
OAK-10241: NPE while trying to get the BinaryDownload URI

mreutegg:
Add test and implement fix


mattvryan (approved):
LGTM, thank you @mreutegg !

mreutegg:
Build failure in oak-blob-cloud-azure looks unrelated to this change.

bhellema:
@mattvryan @mreutegg could we add some debug logging to spit out the name of the blob and path maybe that is being processed?  We are investigating an issue where the name we have been given doesn't match the name in the container storage causing an error.  Having this debug logging would be very beneficial.

mreutegg:
@bhellema, can you please open a JIRA issue and describe where exactly you would like to have debug messages? Patches are also welcome ;)

PR #947, opened 2023-05-22
OAK-10253: Option to only collect references when calling checkConsis…

amit-jain:
…tency

- Add markOnly option for the check-consistency command


nit0906 (approved):
LGTM - just one thing - could we add the new option with description to the readme of oak-run ?

amit-jain:
> LGTM - just one thing - could we add the new option with description to the readme of [oak-run](https://issues.apache.org/jira/browse/OAK-run) ?

Thanks!! Yes, I missed the doc update

PR #948, opened 2023-05-22
OAK-10252: Distinguish between provider and consumer type interfaces

kwin:


reschke (approved):
LGTM.

Q: in case one of the categorizations would turn out to be incorrect - and we made a release after that - would fixing this afterwards cause any pain (as in version numbers changing in a critical way)?

mreutegg (approved):
Looks good to me.

PR #953, opened 2023-05-24
OAK-10259: oak-core tests running out of memory

mreutegg:
Clear inline mocks at the end of a test

### DIFF
--- a/oak-core/src/test/java/org/apache/jackrabbit/oak/security/authentication/user/LoginModuleImplTest.java
+++ b/oak-core/src/test/java/org/apache/jackrabbit/oak/security/authentication/user/LoginModuleImplTest.java
@@ -86,7 +86,6 @@ import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.anyLong;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.eq;
-import static org.mockito.Mockito.clearInvocations;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
@@ -106,7 +105,6 @@ public class LoginModuleImplTest extends AbstractSecurityTest {
     @Override
     public void after() throws Exception {
         try {
-            clearInvocations(monitor);
             if (user != null) {
                 user.remove();
                 root.commit();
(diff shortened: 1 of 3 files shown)


oak-core/src/test/java/org/apache/jackrabbit/oak/security/authentication/user/LoginModuleImplTest.java
```diff
@@ -106,7 +105,6 @@ public class LoginModuleImplTest extends AbstractSecurityTest {
     @Override
     public void after() throws Exception {
         try {
-            clearInvocations(monitor);
```
anchela: i checked in my local oak and there are more usages of 'Mockito.clearInvocations()' in subclasses of the AbstractSecurityTest both in oak-core and in other security modules.

i am curious to understand why you remove this one and not the others.

anchela (changes requested):
hi @mreutegg , the patch looks good. but i have 1 question regarding the removal of clearInvocations in the LoginModuleImplTest. 
if that is needed we might have to remove the other usages as well?

oak-core/src/test/java/org/apache/jackrabbit/oak/security/authentication/user/LoginModuleImplTest.java
mreutegg: With this line the test failed with:
```
org.mockito.exceptions.misusing.NotAMockException: Argument should be a mock, but is: class org.apache.jackrabbit.oak.spi.security.authentication.LoginModuleMonitor$MockitoMock$AIxpL0Xb
	at org.apache.jackrabbit.oak.security.authentication.user.LoginModuleImplTest.after(LoginModuleImplTest.java:110)
```
It seems the sequence how multiple After methods on the same class are called is determined by method name (lexicographic ordering?). It means the existing `after()` method is called first and then the newly introduced `clearMocks()`.

I will revert the change in `LoginModuleImplTest` and move `Mockito.framework().clearInlineMocks()` to the existing `after()` method. The changes to `LoginContextProviderImplTest` will still be necessary though.

mreutegg:
@anchela WDYT now?

anchela (approved):
lgtm (but i didn't build oak with the modifications nor did i verify that the oom is fixed)... but the patch looks reasonable. hope that helps.

PR #955, opened 2023-05-24
OAK-10262 Document ASCIIFolder and OakAnalyzer

thomasmueller:


oak-doc/src/site/markdown/query/lucene.md
```diff
@@ -762,10 +762,15 @@ defaults to 5
...
+and the `WordDelimiterFilter` with the following options:
+`GENERATE_WORD_PARTS`, `STEM_ENGLISH_POSSESSIVE`, and `GENERATE_NUMBER_PARTS`.
+
 `@since Oak 1.5.5, 1.4.7, 1.2.19`
-Unless custom analyzer is configured (as documented below), in-built analyzer
-can be configured to include original term as well to be indexed. This is
-controlled by setting boolean property `indexOriginalTerm` on analyzers node.
+Unless custom analyzer is explicitly configured (as documented below), the built-in analyzer
+can be configured to include the original term as well (`PRESERVE_ORIGINAL`). This is
+controlled by setting boolean property `indexOriginalTerm` on analyzers node:
```
nfsantos:
```suggestion
controlled by setting boolean property `indexOriginalTerm` on the `analyzers` node:
```

oak-doc/src/site/markdown/query/lucene.md
```diff
@@ -845,7 +850,17 @@ all the other components (e.g. `charFilters`, `Synonym`) are optional.
 
 #### Examples
 
-Adding stemming support
+To convert umlauts using ASCII folding, use:
```
nfsantos: Converting umlauts is just an example of what ASCII folding does, but the sentence above sounds like the example below is just for umlauts. Maybe rephrase to explain that umlaut conversion is just an example and show what it does: Ü -> u. And give another example, for instance: `açaí` -> `acai`. (taken from the Elasticsearch documentation, https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-asciifolding-tokenfilter.html)

nfsantos (approved):

oak-doc/src/site/markdown/query/lucene.md
thomasmueller: I was more coming from the use case... The use case is "umlauts", and one of the solutions for that (probably the easiest) is to use ASCII folding. What you suggest is coming from the ASCII folding, and then list what can you do with it. Sure, we can do that as well; however, to avoid copyright problems, we can't just copy & paste Elasticsearch docs.

nfsantos: I was not suggesting to copy'n'paste, just use as an example the word `açaí`. Otherwise, I can suggest copywrite-free Portuguese words: Conceição, aço, cão, coração... :) But I think most people will know açaí better. But your changes are good, now it's more clear.

PR #956, opened 2023-05-30
OAK-10199 : Added skelton to perform DetailedGC

rishabhdaim:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
```diff
@@ -101,6 +108,22 @@ public VersionGCRecommendations(long maxRevisionAgeMs, Checkpoints checkpoints,
         TimeInterval scope = new TimeInterval(oldestPossible, Long.MAX_VALUE);
         scope = scope.notLaterThan(keep.fromMs);
 
+        detailedGCTimestamp = settings.get(SETTINGS_COLLECTION_DETAILED_GC_TIMESTAMP_PROP);
+        if (detailedGCTimestamp == 0) {
+            if (log.isDebugEnabled()) {
```
stefan-egli: `log.isDebugEnabled()` doesn't seem necessary here (but with `log.info` change below it would entirely become obsolete)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
```diff
@@ -101,6 +108,22 @@ public VersionGCRecommendations(long maxRevisionAgeMs, Checkpoints checkpoints,
...
+                log.debug("No detailedGCTimestamp found, querying for the oldest deletedOnce candidate");
+            }
+            oldestPossibleFullGC = vgc.getOldestModifiedTimestamp(clock) - 1;
+            if (log.isDebugEnabled()) {
+                log.debug("detailedGCTimestamp found: {}", Utils.timestampToString(oldestPossibleFullGC));
+            }
```
stefan-egli: I see that in the existing code above we're also only doing a `log.debug` - but I would actually argue that in both cases it could very well at least be a `log.info` for visibility. wdyt?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCSupport.java
```diff
@@ -160,6 +175,30 @@ public long getOldestDeletedOnceTimestamp(Clock clock, long precisionMs) {
         return ts;
     }
 
+    /**
+     * Retrieve the time of the oldest modified document.
+     *
+     * @return the timestamp of the oldest modified document.
+     */
+    public long getOldestModifiedTimestamp(final Clock clock) {
```
stefan-egli: this method should have a dedicated (few) unit test(s)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
```diff
@@ -193,6 +225,38 @@ public void apply(BasicDBObject document) {
...
     }
 
+    /**
+     * Retrieve the time of the oldest modified document.
+     *
+     * @param clock System Clock to measure time in accuracy of millis
+     * @return the timestamp of the oldest modified document.
+     */
+    @Override
+    public long getOldestModifiedTimestamp(final Clock clock) {
```
stefan-egli: this method should ideally have a (few) unit test(s)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentStoreJDBC.java
```diff
@@ -715,6 +715,7 @@ private PreparedStatement prepareQuery(Connection connection, RDBTableMetaData t
         }
 
         if (sortBy != null) {
+            // FIXME : order should be determined via sortBy field
```
stefan-egli: should this be rather a Jira ticket? (to what extend is it related to OAK-10199, would something fail based on this on RDB?)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBVersionGCSupport.java
```diff
@@ -85,6 +91,29 @@ public Iterable<NodeDocument> getPossiblyDeletedDocs(final long fromModified, fi
...
+     * the granularity of the {@link NodeDocument#MODIFIED_IN_SECS} field and
+     * then perform the comparison.
+     *
+     * @param fromModified the lower bound modified timestamp (inclusive)
+     * @param toModified   the upper bound modified timestamp (exclusive)
+     * @param limit        the limit of documents to return
+     * @return matching documents.
+     */
+    @Override
+    public Iterable<NodeDocument> getModifiedDocs(final long fromModified, final long toModified, final int limit) {
```
stefan-egli: would also consider a unit test here

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBVersionGCSupport.java
```diff
@@ -239,6 +268,33 @@ public long getOldestDeletedOnceTimestamp(Clock clock, long precisionMs) {
...
     }
 
+    /**
+     * Retrieve the time of the oldest modified document.
+     *
+     * @param clock System Clock
+     * @return the timestamp of the oldest modified document.
+     */
+    @Override
+    public long getOldestModifiedTimestamp(Clock clock) {
```
stefan-egli: (same comment about unit test coverage)

stefan-egli (commented):
Besides the line-based comments, what I was wondering more in general was test coverage. I think it would be good if we could have some more test coverage, even with the skeleton (and no actual detail gc yet being done). Thinking of the code that manages the new `detailedGCTimeStamp` etc, and whether tha happens correctly in normal/edge cases with repository restarts etc. I think there is quite a bit of new code in VersionGCSupport and that requires some hardening first still. I guess with the individual cleanups we can add more specific additional tests - but I'd argue that even the skeleton should already be solid. Wdyt?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
```diff
@@ -101,6 +108,22 @@ public VersionGCRecommendations(long maxRevisionAgeMs, Checkpoints checkpoints,
...
+        } else {
+            oldestPossibleFullGC = detailedGCTimestamp - 1;
+        }
+
+        TimeInterval fullGCTimeInternal = new TimeInterval(oldestPossibleFullGC, MAX_VALUE);
```
stefan-egli: `TimeInterval` here expects milliseconds - however in the current version of the code `oldestPossibleFullGC` is seconds - as it comes from `getOldestModifiedTimestamp` which is based on `_modified`.

I would suggest to use milliseconds consistently.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -521,6 +586,87 @@ private VersionGCStats gc(long maxRevisionAgeInMillis) throws IOException {
...
+            try (DetailedGC gc = new DetailedGC(headRevision, monitor, cancel)) {
+                final long fromModified = rec.scopeFullGC.fromMs;
+                final long toModified = rec.scopeFullGC.toMs;
+                if (phases.start(GCPhase.DETAILED_GC)) {
+                    while (foundDoc && oldestModifiedGCed < toModified && docsTraversed <= PROGRESS_BATCH_SIZE) {
+                        // set foundDoc to false to allow exiting the while loop
+                        foundDoc = false;
+                        Iterable<NodeDocument> itr = versionStore.getModifiedDocs(oldestModifiedGCed, toModified, 1000);
+                        try {
+                            for (NodeDocument doc : itr) {
```
stefan-egli: There is an issue with this loop. It uses the `oldestModifiedGCed` as the "from" condition but `getModifiedDocs` uses ">=". So if the last batch returns only documents with all equal "_modified" - or it returns only 1 document - then this will essentially loop excessively, only stopped by `PROGRESS_BATCH_SIZE` eventually.

stefan-egli: There is an additional complication where if there are more than 1000 documents with all equal "_modified", the current code wouldn't get passed these 1000 documents, as it only uses "_modified" as the starting condition but stops after 1000. So if there's more than 1000 it currently can't deliver the 1001st, 1002nd etc.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
rishabhdaim: I agree, we can increase the log level to `info`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
rishabhdaim: `getOldestModifiedTimestamp` is overridden in `MongoVersionGCSupport` which indeed converts the `_modified` value to `millis`.

rishabhdaim: Ah, I see your point now, the `getOldestModifiedTimestamp` inside `VersionGCSupport` return it from `_modified` and it's in seconds.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
rishabhdaim: > There is an additional complication where if there are more than 1000 documents with all equal "_modified", the current code wouldn't get passed these 1000 documents, as it only uses "_modified" as the starting condition but stops after 1000. So if there's more than 1000 it currently can't deliver the 1001st, 1002nd etc.

We are setting `oldestModifiedGCed` to `_modified` field at the end of `for` loop. So, the next time when this loop runs, the `getModifiedDocs` (which uses `>=` clause) will start from 1001st document.

stefan-egli:
PS: Created OAK-10299 to fix VersionGCSupport wrt sorting (at least that aspect was confusing me, so wanted to comment here for context)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
rishabhdaim: > There is an additional complication where if there are more than 1000 documents with all equal "_modified", the current code wouldn't get passed these 1000 documents, as it only uses "_modified" as the starting condition but stops after 1000. So if there's more than 1000 it currently can't deliver the 1001st, 1002nd etc.

I see one more problem here in case if none of the 1000 documents has any garbage, then it will get stuck here infinitely. I would propose using `_id` field in the search criteria of `getModifiedDocs` with `>` conditions and getting results sorted on the basis of `_id`.

We should store the `_id` of the last processed document and save it in the `settings` collection along with `detailedGCTimeStamp`.

PR #957, opened 2023-05-30
OAK-10263 Prevent inconsistent state in TarWriter

jelmini:
* Reduce the probability of transient Azure issues to break TarWriter, by retrying methods AzureSegmentArchiveWriter#doWriteDataFile and AzureSegmentArchiveWriter#afterQueueClosed, which are called on TarWriter#close.
* FileStore#tryFlush, which is called in a background thread, now closes the FileStore in case an unrecoverable exception is caught, to prevent it to continue while in an inconsistent state.
* Extracted & improved existing retry logic in its own class, org.apache.jackrabbit.oak.segment.azure.util.Retries.


dulceanu (approved):
LGTM

PR #958, opened 2023-05-31
OAK-10265 | Refresh the index def after lane revert in oak-run out of…

nit0906:
… the band reindex import


fabriziofortino (approved):

thomasmueller (commented):
I wonder if we could add a test case?

nit0906:
@thomasmueller  - sure. I added assertions to an already existing test to verify this change.

tihom88 (approved):

fabriziofortino (approved):

PR #962, opened 2023-06-01
OAK-10147: Many move operations may consume a lot of memory

mreutegg:
Introduce new implementation that does not reference destination parent MutableTree
Test with feature flag to enable classic move implementation


mreutegg:
There are many build failures for various modules. Looks like an infra issue.

Created https://issues.apache.org/jira/browse/INFRA-24662

reschke (commented):
I fear I do not know enough about oak-core internals to give a useful review here.

mreutegg:
> I fear I do not know enough about [oak-core](https://issues.apache.org/jira/browse/OAK-core) internals to give a useful review here.

That's fine. I just thought you might be interested because you worked on a FileVault memory issue that is very likely caused by this problem.

reschke:
If I only had a test case ore benchmark :-)

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
```diff
@@ -801,7 +802,8 @@ private ContentRepository createNewContentRepository() {
                 queryEngineSettings.unwrap(),
                 indexProvider,
                 securityProvider,
-                new AggregatingDescriptors(t)) {
+                new AggregatingDescriptors(t),
+                newFeature("OAK-10147", whiteboard)) {
```
stefan-egli: elsewhere we used a slightly different naming convention (eg "FT_PREFETCH_OAK-9780"). Is this here the final name?

oak-core/src/main/java/org/apache/jackrabbit/oak/core/MutableRoot.java
```diff
@@ -403,7 +412,31 @@ public PermissionProvider getPermissionProvider() {
...
+         * Apply this and all subsequent moves to the passed tree instance.
+         */
+        Move apply(MutableTree tree);
+    }
+
+    Move createMove(@Nullable Feature classicMove) {
+        // default implementation is memory reduced move, unless classic
+        // implementation is enabled by system property or feature toggle
+        if (CLASSIC_MOVE || (classicMove != null && classicMove.isEnabled())) {
+            return new ClassicMove();
```
stefan-egli, line 433: IIUC then this is unlike previously where we introduced new features `disabled` first. Here it would be `enabled` by default and only `disabled` depending on an active property/toggle. I'm wondering as to why this was was chosen versus the one we did previously?

stefan-egli (commented):
Couple of additional comments besides the inline ones:
* as you mentioned, you still plan to add some more tests. I agree that would be useful to have.
* I was also thinking about benchmarks @reschke  mentioned.
* Would it be possible to reproduce the an OutOfMemoryError by doing an excessive amount of moves?
* As additional testing input, I was wondering if some of the existing tests could be run wiht 2 rules : classicMove and neoMove - that might allow to immediately get some large testing coverage..

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
mreutegg: Agreed. I'll change it to be more consistent with existing feature names.

oak-core/src/main/java/org/apache/jackrabbit/oak/core/MutableRoot.java
mreutegg: I thought about this for quite a while and then decided it is better to enable and use the new implementation by default. It gives us more test coverage by running all tests by default with the change. We already know how Oak behaves with the classic move implementation. If needed, a user can fall back to the current implementation using either the system property or feature mechanism.

mreutegg:
> as you mentioned, you still plan to add some more tests. I agree that would be useful to have.

I added more tests with https://github.com/apache/jackrabbit-oak/pull/962/commits/2676bd6a693ec05bc89d46059f6b921021393b50. Do you have more tests in mind I could add?

> I was also thinking about benchmarks @reschke mentioned.

I'm not too concerned about performance, but I will run some of our existing benchmarks and compare numbers.

> Would it be possible to reproduce the an OutOfMemoryError by doing an excessive amount of moves?

It is certainly possible to create such a test, but I don't see why this would be useful.

> As additional testing input, I was wondering if some of the existing tests could be run wiht 2 rules : classicMove and neoMove - that might allow to immediately get some large testing coverage..

I was considering running all tests with classic and neo move, but then concluded cost outweighs benefits by far. The classic implementation was tested thoroughly so far whenever a test or application code did a move. I therefore decided to enable the new implementation by default and add more test coverage specifically for the classic implementation to get full test coverage even if it is not enabled by default anymore.

stefan-egli:
> I added more tests with https://github.com/apache/jackrabbit-oak/commit/2676bd6a693ec05bc89d46059f6b921021393b50. Do you have more tests in mind I could add?

Mostly I think it would be some more loop tests (a -> b -> c -> a, or a -> b/c -> d/e/f -> a etc), or variations of where the commit is done, but couldn't come up with one that would expose a neo bug right away ..

> I'm not too concerned about performance

I agree, plus neo should anyway be faster than classic for very large moves when including memory effects..

> It is certainly possible to create such a test, but I don't see why this would be useful.

ok, was just thinking of reproducing the original issue - but it might be a bit overkill indeed ...

> I was considering running all tests with classic and neo move, but then concluded cost outweighs benefits by far. The classic implementation was tested thoroughly so far whenever a test or application code did a move. I therefore decided to enable the new implementation by default and add more test coverage specifically for the classic implementation to get full test coverage even if it is not enabled by default anymore.

What I noticed is that there are now two separate tests, ClassicMoveTest and MoveTest - the former on ClassicMove, the latter on NeoMove. What about combining those and running it against classic _and_ neo (eg via a rule)?

I gave it a quick try and interesting enough `MoveTest.moveMoved` does fail on `ClassicMove` (at [this line](https://github.com/mreutegg/jackrabbit-oak/blob/4e6861de0f9e78585f409946728d625688f96f1e/oak-core/src/test/java/org/apache/jackrabbit/oak/core/MoveTest.java#L110)  ) - a potential bug in the old code that NeoMove now fixes as a side-effect?

mreutegg:
> What I noticed is that there are now two separate tests, ClassicMoveTest and MoveTest - the former on ClassicMove, the latter on NeoMove. What about combining those and running it against classic and neo (eg via a rule)?

Good point. I'll combine them and create a parameterized test.

> a potential bug in the old code that NeoMove now fixes as a side-effect?

Yes, I [noticed](https://issues.apache.org/jira/browse/OAK-10147?focusedCommentId=17728395&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17728395) this as well, which is the reason why ClassicMoveTest#moveMoved() is marked as ignored.

mreutegg:
> Mostly I think it would be some more loop tests

I added some more tests with the recent push.

mreutegg:
Created https://issues.apache.org/jira/browse/OAK-10293 for the test failure.

stefan-egli (approved):
+1, lgtm. 
(was curious to see if the known issues with ClassicMove have already been noticed - other than in OAK-10147 - but couldn't find any traces)

protogenes:
Despite this being an old, merged topic, here still seems like the best place for discussion.
I encountered a corner case that works with `ClassicMove` but not with `NeoMove`:
If you indirectly move a tree to a path below its original location `/x -> /x/a`, all `Tree` references to descendants become invalid.

```java
Tree xx = root.getTree("/x/xx");
root.move("/x", "/a");
// workaround: xx.getPath();
root.getTree("/").addChild("x");
root.move("/a", "/x/a");
assertEquals("/x/a/xx", xx.getPath()); // fails as path is "/x/a/a/xx" with NeoMove
```

Looking at `org.apache.jackrabbit.oak.core.MutableTree#applyPendingMoves` the moves are applied multiple times to a path:
1. moves are applied to the parent tree
2. `NeoMove.apply` constructs the most recent path and then
3. applies the same moves again, if the paths overlap as above

The easy solution is to store the path instead of constructing it from the parent which was already rewritten.
The stored path matches the state of pendingMoves at all times and can be rewritten directly.

Alternatively one could restore the path matching the pendingMoves by reverse-rewriting the parent path, requiring a weak back reference in `NeoMove` and effectively doubling the rewrite operations.

PR #969, opened 2023-06-08
OAK-10280: Occasional failure to start docker container

mreutegg:
Consolidate multiple copies of AzuriteDockerRule
Skip test when waiting for port or startup message times out


oak-blob-cloud-azure/src/test/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzuriteDockerRule.java
```diff
@@ -36,23 +37,30 @@
...
     public AzuriteDockerRule() {
         wrappedRule = new DockerRule(ImmutableDockerConfig.builder()
             .image(IMAGE)
             .name("oak-test-azurite-" + CONTAINER_SUFFIX)
             .ports("10000")
             .addStartedListener(container -> {
-                container.waitForPort("10000/tcp");
-                container.waitForLog("Azurite Blob service is successfully listening at http://0.0.0.0:10000");
+                try {
+                    container.waitForPort("10000/tcp");
```
jelmini, line 58: Would it make sense to increase the timeout in `waitForPort()` to prevent tests to be ignored too often, if the origin of the issue is slow Jenkins nodes? The default timeout is 30s and it could be increased to 60s, for example .

jelmini (approved):

oak-blob-cloud-azure/src/test/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzuriteDockerRule.java
mreutegg: I don't know if the problem is a slow Jenkins. 30 seconds already seems generous to me and the actual problem may be something else. Maybe this could be tackled in another JIRA ticket...

PR #977, opened 2023-06-09
OAK-10291: oak-segment-remote: PersistentRedisCacheTest may fail on W…

mbaedke:
…indows due to insufficient pagefile size.

Set maxmemory option for Redis server to 768mb.


mreutegg (approved):
Looks good to me.

reschke (approved):

PR #979, opened 2023-06-12
OAK-10294 - Add to indexing job the Pipelined Strategy to dump Mongo contents

nfsantos:
Add a Pipelined strategy to create the FlatFileStore from MongoDB during the indexing job.


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedStrategy.java
```diff
@@ -0,0 +1,403 @@
...
+
+    private final MongoDocumentStore docStore;
+    private final DocumentNodeStore documentNodeStore;
+    private final RevisionVector rootRevision;
+    private final BlobStore blobStore;
+    private final File storeDir;
+
+    private final Comparator<PipelinedNodeStateHolder> comparator;
+    private final PathElementComparatorStringArray pathComparator;
+    private NotificationEmitter emitter;
```
fabriziofortino: this is never used. Are you planning to use it in the future?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelineMongoDownloadTask.java
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    @Override
+    public Result call() throws Exception {
+        String originalName = Thread.currentThread().getName();
+        Thread.currentThread().setName("mongo-dump");
+        try {
+            LOG.info("Starting to download from MongoDB.");
+            MongoCollection<BasicDBObject> dbCollection = MongoDocumentStoreHelper.getDBCollection(mongoStore, collection);
+            //TODO This may lead to reads being routed to secondary depending on MongoURI
```
thomasmueller: I wonder if this might result in problems if the secondary is far behind

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelineMongoDownloadTask.java
```diff
@@ -0,0 +1,278 @@
...
+            Map<String, Integer> exceptions = new HashMap<>();
+            while (true) {
+                try {
+                    if (lastIdDownloaded != null) {
+                        LOG.info("Recovering from broken connection, finishing downloading documents with _modified={}", nextLastModified);
+                        downloadRange(new TraversingRange(new LastModifiedRange(nextLastModified, nextLastModified + 1), lastIdDownloaded));
+                        // Reset the retries
+                        failureTimestamp = null;
+                        // Continue downloading everything starting from the next _lastmodified value
+                        downloadRange(new TraversingRange(new LastModifiedRange(nextLastModified + 1, upperBoundLastModified), null));
```
thomasmueller: I think there is a small probability that nextLastModified + 1 is larger than upperBoundLastModified. In this case, the constructor of LastModifiedRange might throw an exception:

>         if (lastModifiedTo < lastModifiedFrom) {
>             throw new IllegalArgumentException("Invalid range (" + lastModifiedFrom + ", " + lastModifiedTo + ")");
>         }

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelineMongoDownloadTask.java
```diff
@@ -0,0 +1,278 @@
...
+                            Thread.sleep(retryInterval);
+                        } catch (InterruptedException ignore) {
+                        }
+                        // simple exponential backoff mechanism
+                        retryInterval = Math.min(retryMaxIntervalMillis, retryInterval * 2);
+                    }
+                }
+            }
+        } catch (Throwable t) {
+            LOG.warn("Thread terminating with exception: " + t);
```
thomasmueller: I would log the exception as well: ", t)"

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelineMongoDownloadTask.java
```diff
@@ -0,0 +1,278 @@
...
+    private long getModified(MongoCollection<BasicDBObject> dbCollection, Bson sort) {
+        BsonDocument query = new BsonDocument();
+        query.append(NodeDocument.MODIFIED_IN_SECS, new BsonDocument().append("$ne", new BsonNull()));
+        try (MongoCursor<BasicDBObject> cursor = dbCollection
+                .find(query)
+                .sort(sort)
+                .limit(1)
+                .iterator()) {
+            if (!cursor.hasNext()) {
+                return -1;
```
thomasmueller: It's an edge case... but what about adding a parameter "defaultValue" so that -1 is never used? Let's say in the first call it returns -1, and then a document is added... Very strange edge case indeed.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedNodeStateHolderFactory.java
```diff
@@ -0,0 +1,43 @@
...
+package org.apache.jackrabbit.oak.index.indexer.document.flatfile.pipelined;
+
+import java.util.ArrayList;
+
+import static org.apache.jackrabbit.oak.commons.PathUtils.elements;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.NodeStateEntryWriter.getPath;
+
+/**
+ * Factory of {@link PipelinedNodeStateHolder}, reuse an internal array list to reduce object allocation.
+ * Not thread safe.
```
thomasmueller: As a general comment (not asking to change it):

I wonder if what is more important: (A) thread save + simple code, or (B) add this complexity, risk problems later because it is not thread save, and avoid memory allocation, hopefully speeding it up a bit? 

In this case I would go for simpler and thread-safe code, and only try to optimize if we see it shows up in profiling output. Maybe you have done that already. There are cases where optimizations can't be done later easily, but here it can be. See also https://lemire.me/blog/2023/04/27/hotspot-performance-engineering-fails/

For speed, an alternative is to call PathUtils.getDepth(path) first to calculate the array size, and then construct an array of the right size, and then fill the entries one by one? Sure, we would iterate over the array twice.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PathElementComparatorStringArray.java
```diff
@@ -0,0 +1,70 @@
...
+ * under the License.
+ */
+package org.apache.jackrabbit.oak.index.indexer.document.flatfile.pipelined;
+
+import org.apache.jackrabbit.guava.common.collect.ImmutableSet;
+
+import java.util.Comparator;
+import java.util.Set;
+
+class PathElementComparatorStringArray implements Comparator<String[]> {
```
thomasmueller: I think we are missing a unit test here.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelineMongoDownloadTask.java
```diff
@@ -0,0 +1,278 @@
...
+    public <T extends Document> PipelineMongoDownloadTask(MongoDocumentStore mongoStore,
+                                                          Collection<T> collection,
+                                                          Predicate<String> filter,
+                                                          int batchSize,
+                                                          ArrayBlockingQueue<BasicDBObject[]> queue) {
+        this.mongoStore = mongoStore;
+        this.collection = collection;
+        this.batchSize = batchSize;
+        // TODO
+//        this.filter = filter;
```
thomasmueller: This TODO should be resolved

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelineMongoDownloadTask.java
```diff
@@ -0,0 +1,278 @@
...
+
+        // Default retries for 5 minutes.
+        this.retryDuringSeconds = ConfigHelper.getSystemPropertyAsInt(
+                OAK_INDEXER_PIPELINED_MONGO_CONNECTION_RETRY_SECONDS,
+                DEFAULT_OAK_INDEXER_PIPELINED_MONGO_CONNECTION_RETRY_SECONDS);
+        this.dbCollection = MongoDocumentStoreHelper.getDBCollection(mongoStore, collection);
+
+        //TODO This may lead to reads being routed to secondary depending on MongoURI
+        //So caller must ensure that its safe to read from secondary
+//        this.readPreference = MongoDocumentStoreHelper.getConfiguredReadPreference(mongoStore, collection);
```
thomasmueller: This TODO should be resolved

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedSortBatchTask.java
```diff
@@ -0,0 +1,151 @@
...
+                if (nseBuffer == SENTINEL_NSE_BUFFER) {
+                    LOG.info("Terminating thread, processed {} entries", entriesProcessed);
+                    return new Result(entriesProcessed);
+                }
+                sortAndSaveBatch(nseBuffer);
+                nseBuffer.reset();
+                emptyBuffersQueue.put(nseBuffer);
+            }
+        } catch (Throwable t) {
+            LOG.warn("Thread terminating with exception: " + t);
```
thomasmueller: Log the exception ", t"

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedTransformTask.java
```diff
@@ -0,0 +1,215 @@
...
+                                    sortArray.add(new SortKey(key, bufferPos));
+                                    totalEntryCount++;
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        } catch (Throwable t) {
+            LOG.warn("Thread terminating with exception: " + t);
```
thomasmueller: Log the exception

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/SortKeyPathFactory.java
```diff
@@ -0,0 +1,65 @@
...
+ * <li><a href="https://www.baeldung.com/string/intern">Interns</a> the top level elements of the path.
+ * These are very likely to be repeated in most of the entries, so they are good candidates to be interned.
+ * </ul>
+ *
+ */
+public final class SortKeyPathFactory {
+    private final ArrayList<String> arrayBuilder = new ArrayList<>(16);
+    // Common words that appear in paths.
+    // TODO: confirm that checking for the common words is a worthwhile optimization.
+    private static final Set<String> commonWords = Set.of("content", "dam", "product-assets",
```
thomasmueller: We are missing a unit test

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/SortKeyPathFactory.java
```diff
@@ -0,0 +1,65 @@
...
+            "jcr:content", "jcr:title", "jcr:lastModified", "jcr:created", "jcr:primaryType", "jcr:uuid"
+//            "cq:tags", "cq:lastModified", "cq:lastModifiedBy", "cq:template", "cq:templatePath",
+//            "dc:format", "dc:title", "dc:description", "dc:creator", "dc:modified", "dc:created",
+//            "dam:sha1", "dam:size", "dam:score", "dam:status", "dam:assetState", "dam:imported",
+//            "usages", "predictedTags", "imageFeatures", "contentFragment", "pageTitle", "renditions",
+//            "videoCodec", "audioCodec", "metadata", "original", "profile"
+    );
+
+    public String[] genSortKey(String path) {
```
thomasmueller: generateSortKey?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/SortKeyPathFactory.java
```diff
@@ -0,0 +1,65 @@
...
+        arrayBuilder.clear();
+        int i = 0;
+        for (String part : elements(path)) {
+            // This first levels of the path will very likely be similar for most of the entries (e.g. /content/dam/<company>)
+            // Interning these strings should provide a big reduction in memory usage.
+            // It is not worth to intern all levels because at lower levels the names are more likely to be less diverse,
+            // often even unique, so interning them would fill up the interned string hashtable with useless entries.
+            if (i < 3 || part.length() == 1 || commonWords.contains(part)) {
+                arrayBuilder.add(part.intern());
```
thomasmueller: Sure we can save a bit of memory, but I think it is not worth the complexity.

I think part.length() is not very common.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PathElementComparatorStringArray.java
```diff
@@ -0,0 +1,70 @@
...
+        //Also Rank jcr:content higher i.e. first child
+        while (i1 < p1.length || i2 < p2.length) {
+            if (i1 >= p1.length) {
+                return -1;
+            }
+            if (i2 >= p2.length) {
+                return 1;
+            }
+
+            String pe1 = p1[i1]; i1++;
```
thomasmueller: I would use String pe2 = p2[i2++]; 
as it is less prone to typos

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelineMongoDownloadTask.java
```diff
@@ -0,0 +1,278 @@
...
+    private final Collection<? extends Document> collection;
+    private final Predicate<String> filter;
+    private final int batchSize;
+    private final ArrayBlockingQueue<BasicDBObject[]> mongoDocQueue;
+    private final int retryDuringSeconds;
+    private final Logger traversalLog = LoggerFactory.getLogger(PipelineMongoDownloadTask.class.getName() + ".traversal");
+    private final MongoCollection<BasicDBObject> dbCollection;
+    private final ReadPreference readPreference;
+
+    private long documentsRead = 0;
```
thomasmueller: Initializing to 0 / null is not needed. And it is inconsistent.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedNodeStateHolderFactory.java
nfsantos: This code is on the hot path, so I am willing to sacrifice simplicity for efficiency. However, I have not run any benchmark to validate that creating a new intermediate array for every line read from the intermediate files will cause a measurable performance degradation. Intuitively, I expect so, because this is a tight loop and the files contain millions of lines, but I could be wrong.

I did not knew about the `PathUtils.getDepth(path)`, that seems to be a better solution as it avoids the need for any intermediate object and can be done in a thread-safe way. I will re-rewrite the code to use that method to create an array of the right size, as you suggest, therefore making it thread safe. If I have time, run some microbenchmarks.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelineMongoDownloadTask.java
nfsantos: I think it cannot happen. `upperBoundLastModified` is set to 1 plus the highest _modified value that can be returned by the traversal queries to MongoDB:
```
long upperBoundLastModified = getLatestLastModified(dbCollection) + 1;
```
And `nextLastModified` is only assigned to values returned by the query:
```
this.nextLastModified = next.getLong(NodeDocument.MODIFIED_IN_SECS);
```
Furthermore, the traversing query is exclusive of the upper bound, see `TraversingRange.getFindQuery()`, so it will always return `_modfied` values lower than `upperBoundLastModfied`.

So at most, `nextLastModifed+1` will be equal to `upperBoundLastModified`. Or is there something wrong with my reasoning?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelineMongoDownloadTask.java
nfsantos: I'm trying to avoid polluting the logs with multiple repetitions of the same stack trace. The exception here is propagated and caught by the main thread that is waiting for these tasks to complete (In the PipelinedStrategy class). And then the main thread rethrows the exception wrapped in a RuntimeException. The code that called the `createdSortedStoreFile()` will catch and finally log the exception with the stack trace.

Actually, there is a corner case where the exception will not be logged completely, which is if the main thread already stopped waiting for some of these futures, for instance, if one of the tasks aborted which will cause the main thread to abort with the exception of the first task that failed. In this case, we would only get the name and message of the exceptions thrown by the futures that failed later. Ok, that stack trace may be useful, so I will log the stack trace here as well (except for InterruptedExceptions).

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelineMongoDownloadTask.java
nfsantos: I'm not sure. The MongoDocumentStore implementation has logic to limit the reads from secondaries if they are too far back. Maybe we can implement a similar logic here. But I don't know what could be the problems, maybe it will miss some of the recent documents?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelineMongoDownloadTask.java
nfsantos: This should never happen, it would mean the collection of nodes is empty. What about throwing an exception in this case? In any case, if the collection is empty, there is nothing to index, so we can as well fail the job.

fabriziofortino (approved):
LGTM.
The new strategy is a lot less complicated than the previous ones. The code is a lot easier to understand. It would be great to add some tests.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedStrategy.java
nfsantos: It was copied from the old traverse strategy. It is part of the mechanism to monitor free memory using the management MBeans, which the Pipelined strategy does not use. I removed this field.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeStateHolder.java
```diff
@@ -0,0 +1,54 @@
...
+
+final class NodeStateHolder {
+
+    private final String line;
+    private final String[] pathElements;
+
+    public NodeStateHolder(String line) {
+        this.line = line;
+        String path = getPath(line);
+        int depth = PathUtils.getDepth(path);
```
fabriziofortino, line 34: minor: for consistency you might want to statically import `getDepth`.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMergeSortTask.java
```diff
@@ -0,0 +1,140 @@
...
+
+import static org.apache.jackrabbit.oak.commons.IOUtils.humanReadableByteCountBin;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileStoreUtils.createWriter;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileStoreUtils.getSortedStoreFileName;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileStoreUtils.sizeOf;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.pipelined.PipelinedStrategy.FLATFILESTORE_CHARSET;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.pipelined.PipelinedStrategy.SENTINEL_SORTED_FILES_QUEUE;
+
+/**
+ * Accumulates the intermediate sorted files and, when all files are genereted, merges them into a single sorted file,
```
fabriziofortino: typo
```suggestion
 * Accumulates the intermediate sorted files and, when all files are generated, merges them into a single sorted file,
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMergeSortTask.java
```diff
@@ -0,0 +1,140 @@
...
+            Thread.currentThread().setName(originalName);
+        }
+    }
+
+    private File sortStoreFile(List<File> sortedFilesBatch) throws IOException {
+        Stopwatch w = Stopwatch.createStarted();
+        File sortedFile = new File(storeDir, getSortedStoreFileName(algorithm));
+        try (BufferedWriter writer = createWriter(sortedFile, algorithm)) {
+            Function<String, NodeStateHolder> func1 = (line) -> line == null ? null : new NodeStateHolder(line);
+            Function<NodeStateHolder, String> func2 = holder -> holder == null ? null : holder.getLine();
```
fabriziofortino: can you rename these with more meaningful names?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -0,0 +1,304 @@
...
+            return documentsDownloaded;
+        }
+    }
+
+    private static final Logger LOG = LoggerFactory.getLogger(PipelinedMongoDownloadTask.class);
+
+    /**
+     * Whether to retry on connection errors to MongoDB.
+     * This property affects the query that is used to download the documents from MongoDB. If set to true, the query
+     * will traverse the results by order of the _modfied property (does an index scan), which allows it to resume after
```
fabriziofortino: typo
```suggestion
     * will traverse the results by order of the _modified property (does an index scan), which allows it to resume after
```

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeStateEntryBatchTest.java
```diff
@@ -0,0 +1,97 @@
...
+        batch.addEntry("b", new byte[1]);
+        assertEquals(2, batch.numberOfEntries());
+        assertTrue(batch.isAtMaxEntries());
+        try {
+            batch.addEntry("c", new byte[1]);
+            throw new AssertionError("Expected exception");
+        } catch (IllegalStateException e) {
+            // expected
+        }
+    }
```
fabriziofortino: what about using `expected`? Same for the test below.
```suggestion
    @Test(expected = IllegalStateException.class)
    public void testMaximumNumberOfEntries() {
        NodeStateEntryBatch batch = NodeStateEntryBatch.createNodeStateEntryBatch(1024, 2);
        assertFalse(batch.isAtMaxEntries());
        batch.addEntry("a", new byte[1]);
        assertFalse(batch.isAtMaxEntries());
        batch.addEntry("b", new byte[1]);
        assertEquals(2, batch.numberOfEntries());
        assertTrue(batch.isAtMaxEntries());
        batch.addEntry("c", new byte[1]);
    }```

File: oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedStrategy.java [hunk 22]
```diff
@@ -0,0 +1,424 @@
...
+            int numberOfBuffers = 1 + transformThreads;
+            // The extra memory arena is to account for the memory taken by the SortKey entries that are part of the
+            // entry batches.
+            int memoryArenas = numberOfBuffers + 1;
+            int memoryForEntriesMB = workingMemoryMB / memoryArenas;
+            int maxNumberOfEntries = memoryForEntriesMB * 1024 * 4; // Assuming that 1KB is enough for 4 entries.
+            int maxNumberOfEntriesPerBuffer = maxNumberOfEntries / numberOfBuffers;
+
+            // A ByteBuffer can be at most Integer.MAX_VALUE bytes
+            int bufferSizeBytes;
```
steffenvan: Just for readability purposes, could the assignment here with the local scope be refactored into a function?

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedIT.java
```diff
@@ -0,0 +1,149 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.Set;
+import java.util.function.Predicate;
+
+import static org.junit.Assert.assertEquals;
+
+public class PipelinedIT {
```
fabriziofortino, line 58: can we have a test where the `pathPredicate` does not match any node in the repository? In the past, this has caused issues because the files where empty.

fabriziofortino (approved):
LGT. Just a few improvements.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedStrategy.java
nfsantos: Done

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeStateHolder.java
nfsantos: I have instead replaced all static imports by references of the form `<class>.<method>`. I find this more readable.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMergeSortTask.java
nfsantos: Done, I use the same names as the arguments defined in the function that is called with these variables: `stringToType` and `typeToString`.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeStateEntryBatchTest.java
nfsantos: I replaced by something like:
```assertThrows(BufferOverflowException.class, () -> batch.addEntry("b", new byte[1]));```
which is supported since JUnit 4.13. Oak is using 4.13.1.

File: oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedIT.java (hunk 23 above)
Comment (nfsantos) [reply oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedIT.java:58]: Done

File: oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedStrategy.java [hunk 24]
```diff
@@ -0,0 +1,433 @@
...
+
+            PipelinedMergeSortTask mergeSortTask = new PipelinedMergeSortTask(storeDir, pathComparator, algorithm, sortedFilesQueue);
+            ecs.submit(mergeSortTask);
+
+
+            try {
+                LOG.info("Waiting for tasks to complete.");
+                int tasksFinished = 0;
+                int transformTasksFinished = 0;
+                while (tasksFinished < numberOfThreads) {
```
steffenvan, line 343: Would it be possible to refactor some of the code in this function into more separate functions? The function itself is quite long and with 3 nested `try-catch blocks`. While this might be necessary, I think it would at least improve the readability for the future. I asked ChatGPT to do it for an older version of the code, and the result was quite nice. 

For example: 
```java
private void waitForTasksToComplete(int numberOfThreads, int transformThreads, int entryCount, Stopwatch start,
        ArrayBlockingQueue<BasicDBObject[]> mongoDocQueue,
        ArrayBlockingQueue<NodeStateEntryBatch> emptyBatchesQueue,
        ArrayBlockingQueue<NodeStateEntryBatch> nonEmptyBatchesQueue,
        ArrayBlockingQueue<File> sortedFilesQueue,
        TransformStageStatistics transformStageStatistics) {
        int tasksFinished = 0;
        int transformTasksFinished = 0;
        while (tasksFinished < numberOfThreads) {
            Future<?> completedTask = ecs.take();
         ...
```

And then we could call it:

```java
try {
    waitForTasksToComplete(ecs, numberOfThreads, monitorFuture, mongoDocQueue, emptyBatchesQueue,
        nonEmptyBatchesQueue, sortedFilesQueue, transformStageStatistics, start);
    flatFileStore = mergeSortedFilesTask.getFlatFileStoreFile();
    return flatFileStore;
} finally {
    shutdownExecutors();
}
```

nfsantos: I agree that the method is very long and somewhat hard to read. But the refactoring proposed above is also not very satisfying, as the method has too many arguments. I'll keep this in mind for a future revision of this class, hopefully we can come up with a cleaner and more logical refactor. 

For the final changes to this PR, I'd rather focus only on correctness or on small changes, and avoid doing any refactoring other than trivial ones.

thomasmueller (approved):

PR #986, opened 2023-06-15
OAK-10247: oak-commons: remove Guava from public API - remove remaining uses and add to shaded version to Profiler ignore list

reschke:


mreutegg (commented):
Looks good to me.

PR #987, opened 2023-06-15
OAK-10273: Index-definition json created during oak-run reindex should not serialise index data

tihom88:


tihom88:
on wrong branch

PR #991, opened 2023-06-15
OAK-10273: Index-definition json created during oak-run reindex shoul…

tihom88:
…d not serialise index data

### DIFF
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/IndexerSupport.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/IndexerSupport.java
@@ -121,7 +121,7 @@ public class IndexerSupport {
 
     protected void dumpIndexDefinitions(NodeStore nodeStore) throws IOException {
         IndexDefinitionPrinter printer = new IndexDefinitionPrinter(nodeStore, indexHelper.getIndexPathService());
-        printer.setFilter("{\"properties\":[\"*\", \"-:childOrder\"],\"nodes\":[\"*\", \"-:index-definition\"]}");
+        printer.setFilter("{\"properties\":[\"*\", \"-:childOrder\"],\"nodes\":[\"*\", \"-:index-definition\", \"-:data\", \"-:suggest-data\"]}");
         PrinterDumper dumper = new PrinterDumper(getLocalIndexDir(), IndexDefinitionUpdater.INDEX_DEFINITIONS_JSON,
                 false, Format.JSON, printer);
         dumper.dump();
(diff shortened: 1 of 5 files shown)


oak-run/src/test/java/org/apache/jackrabbit/oak/index/ReindexIT.java
```diff
@@ -19,6 +19,7 @@
 
 package org.apache.jackrabbit.oak.index;
 
+import org.apache.jackrabbit.commons.JcrUtils;
```
fabriziofortino: unused
```suggestion
```

fabriziofortino (approved):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/IndexerSupport.java
```diff
@@ -121,7 +121,7 @@ public void updateIndexDefinitions(NodeBuilder rootBuilder) throws IOException,
 
     protected void dumpIndexDefinitions(NodeStore nodeStore) throws IOException {
         IndexDefinitionPrinter printer = new IndexDefinitionPrinter(nodeStore, indexHelper.getIndexPathService());
-        printer.setFilter("{\"properties\":[\"*\", \"-:childOrder\"],\"nodes\":[\"*\", \"-:index-definition\"]}");
+        printer.setFilter("{\"properties\":[\"*\", \"-:childOrder\"],\"nodes\":[\"*\", \"-:index-definition\", \"-:data\", \"-:suggest-data\"]}");
```
thomasmueller, line 124: I'm not sure if this is actually correct.

Isn't the dump needed for import? If yes, then that would mean if we don't export the ":data" node, import of an index would have to do a full reindex as the index data is empty...

So, first we need to check where this dumpIndexDefinitions method is actually called from... unluckily, there is no documentation it seems.

tihom88: I tested it locally and found out that index definitions don't have :data if --index-paths are provided in command line. But it seems to have :data if --index-paths are not provided.
For :suggest-data also, found out that :suggest-data node only contained   "lastUpdated" property in case --index-paths are provided. So looks like reindexing works without this info.

As :status-data is something which was part of index-definitions, I added a test case just to check if suggest data is preset with this new filter. Let me know if you think we should add few more test cases/ any specific case you think we should incorporate in current test.

tihom88: Hi @thomasmueller , 
Looks like we don't need hidden nodes at all in index-definitions file. While importing index we do cloneVisibleState from index definition.
https://github.com/apache/jackrabbit-oak/blob/16edc2c0b7a3ab20cfbc2dc04b92a5ed245941dd/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/ReindexOperations.java#L75

tihom88:
Testing completed on 1.22.

reschke:
1.22??? This was a merge to "trunk", no?

PR #993, opened 2023-06-19
Oak 10199 : Added Skeleton & Detailed garbage Properties

rishabhdaim:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -521,6 +593,90 @@ private VersionGCStats gc(long maxRevisionAgeInMillis) throws IOException {
...
+            long oldestModifiedDocTimeStamp = rec.scopeDetailedGC.fromMs;
+            String oldestModifiedDocId = rec.detailedGCId;
+            try (DetailedGC gc = new DetailedGC(headRevision, monitor, cancel)) {
+                final long fromModified = rec.scopeDetailedGC.fromMs;
+                final long toModified = rec.scopeDetailedGC.toMs;
+                if (phases.start(GCPhase.DETAILED_GC)) {
+                    while (foundDoc && oldestModifiedDocTimeStamp < toModified && docsTraversed <= PROGRESS_BATCH_SIZE) {
+                        // set foundDoc to false to allow exiting the while loop
+                        foundDoc = false;
+                        Iterable<NodeDocument> itr = versionStore.getModifiedDocs(oldestModifiedDocTimeStamp, toModified, 1000, oldestModifiedDocId);
```
stefan-egli: With the recent change of using `oldestModifiedDocId` here, this iterator now misses that document. Eg if `oldestModifiedDocId="0:/"`, then the underlying call to `DocumentStore.query` has `"0:/"` as the `fromId`, but the javadoc of `DocumentStore.query` states that the [fromId](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentStore.java#L126) is excluded.

stefan-egli:
Looks like `VersionGarbageCollectorIT` now fails in this branch - it works in trunk

rishabhdaim:
> Looks like `VersionGarbageCollectorIT` now fails in this branch - it works in trunk

IT had been fixed with https://github.com/apache/jackrabbit-oak/pull/993/commits/604821d9b9b1230f39e77249ce8e2175cbd78b94 commit

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
rishabhdaim: This would happen only for the very first time when DetailedGC would run.
It had been fixed with https://github.com/apache/jackrabbit-oak/pull/993/commits/dc7092527edbdd565dca86da055fc45fd06ab2f2 commit.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -605,8 +773,147 @@ private void collectDeletedDocuments(GCPhases phases,
...
+                phases.stop(GCPhase.COLLECT_UNMERGED_BC);
+            }
+
+        }
+
+        private void collectDeletedProperties(final NodeDocument doc, final GCPhases phases, final UpdateOp updateOp) {
+
+            // get Map of all properties along with their values
+            if (phases.start(GCPhase.COLLECT_PROPS)) {
+                final Set<String> properties = doc.getPropertyNames();
```
stefan-egli, line 877: There is a problem with using `NodeDocument.getPropertyNames()` here and comparing it directly with `DocumentNodeState.getPropertyNames()` below : the former containes escaped values while the latter does not. The escaping is handled in `Utils.escapePropertyName` / `Utils.unescapePropertyName`

stefan-egli:
Some more concerns I encountered with the detail-gc-skeleton:

##### split classic gc from detail gc
We need a way to separate classic gc from detail gc in the sense that classic gc could be already fully caught up (or eg blocked by a checkpoint), while detail gc would still have to go on and do its repository scan. Currently for example there is a `ignoreDueToCheckPoint` in the `VersionGCRecommendations` which blocks further gc progress if a checkpoint is hit. That applies irrespective of whether the detail gc is actually still way behind. I think the code in `VersionGarbageCollector.GCJob.gc` needs to be split into classic and detail - and to support that, the `VersionGCRecommendations` also needs a `ignore` flag for detail, separated from classic.

##### rounding issue with modified (`-1` vs being 5sec resolution)
There is a rounding issue with `SETTINGS_COLLECTION_DETAILED_GC_TIMESTAMP_PROP`.
* when we read that value, there is currently a [`-1` applied](https://github.com/apache/jackrabbit-oak/blob/ca1c451a84f8eae6dc3f179ffe420cb2cbbc8d02/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java#L132)
* later on in [NodeDocument.getModifiedInSecs](https://github.com/apache/jackrabbit-oak/blob/ca1c451a84f8eae6dc3f179ffe420cb2cbbc8d02/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocument.java#L270) it rounds the value from millis to seconds by dividing - so eg `1657811274999` is rounded to `1657811274`.
* [one line further](https://github.com/apache/jackrabbit-oak/blob/ca1c451a84f8eae6dc3f179ffe420cb2cbbc8d02/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocument.java#L271) it does a modulo 5 - which in the example results in `1657811270`

As a result it goes back a 5 sec lot - which means it will repeat the query and result in duplicate handling of documents

reschke:
By all means, when this is done: please use "squash & merge" :-)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocument.java
```diff
@@ -1669,6 +1670,19 @@ SortedMap<Revision, String> getLocalMap(String key) {
         return map;
     }
 
+    /**
+     * Returns name of all the properties on this document
```
stefan-egli, line 1674: Should we add a note in the javadoc here that these names are escaped?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
```diff
@@ -73,35 +89,57 @@ public class VersionGCRecommendations {
      * @param vgc VersionGC support class
      * @param options options for running the gc
      * @param gcMonitor monitor class for messages
+     * @param detailedGCEnabled whether detailedGC is enabled or not
      */
     public VersionGCRecommendations(long maxRevisionAgeMs, Checkpoints checkpoints, Clock clock, VersionGCSupport vgc,
-            VersionGCOptions options, GCMonitor gcMonitor) {
+                                    VersionGCOptions options, GCMonitor gcMonitor, final boolean detailedGCEnabled) {
```
stefan-egli: One general aspect wrt VersionGCRecommendations came to mind while reviewing this : to some extend we now introduce a second GC - while it is embedded in the classic GC, the second, detail GC is separated in that it has its own from/to timestamps and can be at different positions in the GC process, and parts of the code are therefore separate (like the DetailedGC class). On the other hand, this VersionGCRecommendations contains both "GC lanes" - so is not separated.

From that, I was wondering whether we should separate the recommendation aspect between the two as well, so we'd leave the old class as is and introduce a detailGC variant of it. That way, we'd not risk modifying anything in the classic GC recommendation part - but it also might make code reading/maintenance easier.

Wdyt?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCSupport.java
```diff
@@ -160,6 +203,29 @@ public long getOldestDeletedOnceTimestamp(Clock clock, long precisionMs) {
...
+     * Retrieve the oldest modified document.
+     *
+     * @return the oldest modified document.
+     */
+    public Optional<NodeDocument> getOldestModifiedDoc(final Clock clock) {
+        long ts = 0;
+        long now = clock.getTime();
+        Iterable<NodeDocument> docs = null;
+
+        LOG.info("find oldest modified document");
```
stefan-egli: Noticed in several places in the PR that there are `LOG.info` statements (some replacing `debug`). I do agree we should have a reasonable level of detail logging by default, so `info` is definitely good in some cases. But I'm not sure whether some of these `info` are not leftovers or are indeed here on purpose. This one for example doesn't log any parameters, so at least that should be added. But it also might be somewhat noisy?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
```diff
@@ -120,6 +131,40 @@ public CloseableIterable<NodeDocument> getPossiblyDeletedDocs(final long fromMod
...
+     * within the given range and are greater than given @{@link NodeDocument#ID}.
+     * <p>
+     * The two passed modified timestamps are in milliseconds
+     * since the epoch and the implementation will convert them to seconds at
+     * the granularity of the {@link NodeDocument#MODIFIED_IN_SECS} field and
+     * then perform the comparison.
+     * <p/>
+     *
+     * @param fromModified the lower bound modified timestamp (inclusive)
+     * @param toModified   the upper bound modified timestamp (exclusive)
```
stefan-egli: Noticed that we're not going into details as to whether this is seconds or milliseconds in other places of this class neither. But with new code, I was wondering whether we shouldn't just state here that it is milliseconds. Makes it a bit more explicit and reader friendly.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
```diff
@@ -193,6 +238,31 @@ public void apply(BasicDBObject document) {
...
 
+    /**
+     * Retrieve the time of the oldest modified document.
+     *
+     * @param clock System Clock to measure time in accuracy of millis
+     * @return the timestamp of the oldest modified document.
+     */
+    @Override
+    public Optional<NodeDocument> getOldestModifiedDoc(final Clock clock) {
+        LOG.info("getOldestModifiedDoc() <- start");
```
stefan-egli: (see other comment, left-over `info` statement? maybe `debug` or `trace` instead?)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentStore.java
```diff
@@ -971,8 +971,8 @@ private <T extends Document> void toMapBuilder(ImmutableMap.Builder<String, Stri
     public static String VERSIONPROP = "__version";
 
     // set of supported indexed properties
-    private static final Set<String> INDEXEDPROPERTIES = new HashSet<String>(Arrays.asList(new String[] { MODIFIED,
-            NodeDocument.HAS_BINARY_FLAG, NodeDocument.DELETED_ONCE, NodeDocument.SD_TYPE, NodeDocument.SD_MAX_REV_TIME_IN_SECS, VERSIONPROP }));
+    private static final Set<String> INDEXEDPROPERTIES = new HashSet<>(Arrays.asList(MODIFIED,
```
stefan-egli: Might be worth while if @reschke could comment on RDB changes.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -894,6 +904,17 @@ public static boolean isThrottlingEnabled(final DocumentNodeStoreBuilder<?> buil
...
 
+    /**
+     * Check whether detailed GC is enabled or not for document store.
+     *
+     * @param builder instance for DocumentNodeStoreBuilder
+     * @return true if detailed GC is enabled else false
+     */
+    public static boolean isDetailedGCEnabled(final DocumentNodeStoreBuilder<?> builder) {
+        final Feature docStoreDetailedGCFeature = builder.getDocStoreDetailedGCFeature();
+        return builder.isDetailedGCEnabled() || (docStoreDetailedGCFeature != null && docStoreDetailedGCFeature.isEnabled());
```
stefan-egli, line 915: Noticed that for throttling it was done this way as well. But still, I'm wondering if the behaviour should be AND, nor OR. I think it would be useful if disabling the feature toggle would forcefully disable the detail GC feature - but IIUC that's currently not the case, as it could still be enabled via config?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -356,6 +421,11 @@ private enum GCPhase {
         DELETING,
         SORTING,
         SPLITS_CLEANUP,
+        DETAILED_GC,
+        COLLECT_PROPS,
+        COLLECT_OLD_REVS,
+        COLLECT_UNMERGED_BC,
```
stefan-egli: Was wondering if the three new ones should get a detail gc prefix, to more easily distinguish them from classic?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
```diff
@@ -209,28 +249,75 @@ public void evaluate(VersionGCStats stats) {
...
     }
+
+    private void setLongSetting(final Map<String, Long> propValMap) {
+        UpdateOp updateOp = new UpdateOp(SETTINGS_COLLECTION_ID, true);
+        propValMap.forEach(updateOp::set);
+        vgc.getDocumentStore().createOrUpdate(Collection.SETTINGS, updateOp);
+    }
+
+    @NotNull
+    private static GCResult getResult(VersionGCOptions options, boolean ignoreGC, TimeInterval gcScope, Revision checkpoint) {
```
stefan-egli: Looks like `ignoreGC` is always `false` as parameter. Is it really needed?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -507,6 +581,25 @@ private VersionGCStats gc(long maxRevisionAgeInMillis) throws IOException {
...
+                    } else {
+                        final RevisionVector headRevision = nodeStore.getHeadRevision();
+                        monitor.info("Looking at revisions in {} for detailed GC", rec.scopeDetailedGC);
+                        collectDetailedGarbage(phases, headRevision, rec);
+                    }
+                }
+
+                if (detailedGCEnabled && rec.ignoreDueToCheckPoint && rec.ignoreDetailGCDueToCheckPoint) {
+                    cancel.set(true);
+                } else if (!detailedGCEnabled && rec.ignoreDueToCheckPoint) {
```
stefan-egli: This is a bit complicated to read, I think it could be simplified. Also, in detail gc case, is the first line really correct, is detail gc only cancelled if both "GC lanes" are ignored due to checkpoint?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -521,6 +614,121 @@ private VersionGCStats gc(long maxRevisionAgeInMillis) throws IOException {
...
+            try (DetailedGC gc = new DetailedGC(headRevision, monitor, cancel)) {
+                long fromModified = oldestModifiedMs;
+                String fromId = ofNullable(oldestModifiedDocId).orElse(MIN_ID_VALUE);
+                NodeDocument lastDoc;
+                if (phases.start(GCPhase.DETAILED_GC)) {
+                    while (foundDoc && fromModified < toModified && docsTraversed < PROGRESS_BATCH_SIZE) {
+                        // set foundDoc to false to allow exiting the while loop
+                        foundDoc = false;
+                        lastDoc = null;
+                        Iterable<NodeDocument> itr = versionStore.getModifiedDocs(fromModified, toModified, 1000, fromId);
```
stefan-egli: The `1000` here should probably be replaced with a const, I guess with `PROGRESS_BATCH_SIZE`. Which leads to a second aspect : I think there should be a test case for this that verifies this aspect. Currently `PROGRESS_BATCH_SIZE` is set to `10000`, but here `1000` was used - in which case I think the loop condition would have to be broken in a way. But presumably that was not the case (assuming the tests work).

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -521,6 +614,121 @@ private VersionGCStats gc(long maxRevisionAgeInMillis) throws IOException {
...
+         * @param phases {@link GCPhases}
+         * @param headRevision the current head revision of node store
+         * @param rec {@link VersionGCRecommendations} to recommend GC operation
+         */
+        private void collectDetailedGarbage(final GCPhases phases, final RevisionVector headRevision, final VersionGCRecommendations rec)
+                throws IOException {
+            int docsTraversed = 0;
+            boolean foundDoc = true;
+            final long oldestModifiedMs = rec.scopeDetailedGC.fromMs;
+            final long toModified = rec.scopeDetailedGC.toMs;
```
stefan-egli: Made a similar comment elsewhere about being more explicit whether it is milliseconds or seconds. I think particularly in this method it might not hurt if `toModified` was eg `toModifiedMs`? It is at least otherwise in contrast with `oldestModifiedMs`, which has millis in the name..

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -521,6 +614,121 @@ private VersionGCStats gc(long maxRevisionAgeInMillis) throws IOException {
...
+            long oldModifiedMs = oldestModifiedMs;
+            final String oldestModifiedDocId = rec.detailedGCId;
```
stefan-egli: I think moving all the `final` variables here to the top of the method might make it a bit more readable..

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -521,6 +614,121 @@ private VersionGCStats gc(long maxRevisionAgeInMillis) throws IOException {
...
+                                if (phases.start(GCPhase.COLLECTING)) {
+                                    gc.collectGarbage(doc, phases);
+                                    phases.stop(GCPhase.COLLECTING);
+                                }
+
+                                final Long modified = lastDoc.getModified();
+                                if (modified == null) {
+                                    monitor.warn("collectDetailGarbage : document has no _modified property : {}",
+                                            doc.getId());
+                                } else if (SECONDS.toMillis(modified) < oldestModifiedMs) {
```
stefan-egli: Is this correct? Shouldn't it be `fromModified` here instead of `oldestModifiedMs`?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -605,8 +812,157 @@ private void collectDeletedDocuments(GCPhases phases,
...
+            try {
+                List<NodeDocument> oldDocs = ds.findAndUpdate(NODES, updateOpList);
+                int deletedProps = oldDocs.stream().filter(Objects::nonNull).mapToInt(d -> deletedPropsCountMap.getOrDefault(d.getId(), 0)).sum();
+                updatedDocs = (int) oldDocs.stream().filter(Objects::nonNull).count();
+                stats.updatedDetailedGCDocsCount += updatedDocs;
+                stats.deletedPropsGCCount += deletedProps;
+                log.info("Updated [{}] documents, deleted [{}] properties", updatedDocs, deletedProps);
+                // now reset delete metadata
+                updateOpList.clear();
+                garbageDocsCount = 0;
```
stefan-egli, line 957: `deletedPropsCountMap` should also be cleared here I think

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -521,6 +614,121 @@ private VersionGCStats gc(long maxRevisionAgeInMillis) throws IOException {
...
+                                // continue with GC?
+                                if (cancel.get()) {
+                                    foundDoc = false; // to exit while loop as well
+                                    log.info("Received GC cancel call. Terminating the GC Operation.");
+                                    break;
+                                }
+                                docsTraversed++;
+                                if (docsTraversed % 100 == 0) {
+                                    monitor.info("Iterated through {} documents so far. {} had detail garbage",
+                                            docsTraversed, gc.getGarbageDocsCount());
```
stefan-egli: It looks like the two parameters here are not in sync : the `docsTraversed` is the total number of documents traversed in this entire G run - while `getGarbageDocsCount` is only covering 1 iteration (inner loop).

stefan-egli (changes requested):
Added some detail review comments (see below).

One general thing is the name. Currently there is "detail gc" and "detailed gc" in use. I think we should only use one variant (doesn't much matter which).

PS: I would still like to run some tests against this PR as soon as I find time.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
rishabhdaim: Yes, it was decided to have this flexibility to enable this via either config or feature toggle.
Same approach had been taken for throttling and prefetch as well.
cc @mreutegg

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
rishabhdaim: Good suggestion indeed. Will make the necessary changes.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
stefan-egli: feels non intuitive though...

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
rishabhdaim: I will create a constant for the query limit. yes, we have a test validating that the loop is broken.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
rishabhdaim: yes, it should be `fromModified` here.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
rishabhdaim: Although it only has data per `docId` and wouldn't cause any issues, I agree removing this would be better (at least saves us some space).

stefan-egli:
Discovered below issues:

1. The way it is implemented currently, in VersionGCRecommendations.evaluate it does not set `stats.needRepeat` in case the classic gc is done (or blocked by a checkpoint) but the detail gc is not yet done. I think this is another argument for keeping VersionGCRecommendations as is (i.e not adding detail gc code) and instead add a dedicated detail gc variant of it. Then merge the two results of `evaluate` in a feasible way. (i.e. not sure if `stats.needRepeat` is the only omission).
2. The `collectDeletedProperties` method currently filters out bundled properties entirely. Basically when getting the list of properties via `getNodeAtRevision`, the bundling aspects gets discarded - but the Document's `getProperties` does contain those bundled properties still. So they all get deleted from the document.

rishabhdaim:
@stefan-egli Regarding the `stats.needRepeat`, I don't think this makes sense for our case since we are only deleting a fixed chunk of `detailedGarbage` in every run.

In the case of `revisionGc`, we are deleting on the basis of time only and if we have more documents than a particular threshold we set this to `true` and reduce the time window for the GC cycle.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
stefan-egli: just a nitpick : there's another one few lines below:
```
                long fromModified = oldestModifiedMs;
```

stefan-egli:
+1 for the needRepeat.

Regarding bundled properties : i think it might be useful to add a unit test for this one (assuming it doesn't exist yet)?

Otherwise, there are a few unresolved conversations from the review of [last week](https://github.com/apache/jackrabbit-oak/pull/993#pullrequestreview-1547668975).

(afterwards I'd suggest to merge this to the integration branch - from where we'll anyway have at some later point a PR to trunk, which can get another round of review if we want)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
rishabhdaim: @ stefan-egli I looked into the code and creating a new `Recommendations` class for `DetailedGC` would complicate the code readability.

However, If felt needed, later on, we can separate out the logic. As of now, I would keep them in the same class.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
```diff
@@ -209,28 +249,76 @@ public void evaluate(VersionGCStats stats) {
...
+    private static GCResult getResult(final VersionGCOptions options, final Revision checkpoint, TimeInterval gcScope) {
+        boolean ignoreGC = false;
+        if (checkpoint != null && gcScope.endsAfter(checkpoint.getTimestamp())) {
+            TimeInterval minimalScope = gcScope.startAndDuration(options.precisionMs);
+            if (minimalScope.endsAfter(checkpoint.getTimestamp())) {
+                log.warn("Ignoring GC run because a valid checkpoint [{}] exists inside minimal scope {}.", checkpoint.toReadableString(), minimalScope);
+                ignoreGC = true;
+            } else {
+                gcScope = gcScope.notLaterThan(checkpoint.getTimestamp() - 1);
+                log.info("checkpoint at [{}] found, detailedGCScope now {}", timestampToString(checkpoint.getTimestamp()), gcScope);
```
stefan-egli: I suspect this log.info could become noisy in case of having an old checkpoint. Could it be log.debug?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -605,8 +814,161 @@ private void collectDeletedDocuments(GCPhases phases,
...
+            }
+
+            timer.reset().start();
+            try {
+                List<NodeDocument> oldDocs = ds.findAndUpdate(NODES, updateOpList);
+                int deletedProps = oldDocs.stream().filter(Objects::nonNull).mapToInt(d -> deletedPropsCountMap.getOrDefault(d.getId(), 0)).sum();
+                updatedDocs = (int) oldDocs.stream().filter(Objects::nonNull).count();
+                stats.updatedDetailedGCDocsCount += updatedDocs;
+                stats.deletedPropsGCCount += deletedProps;
+                log.info("Updated [{}] documents, deleted [{}] properties", updatedDocs, deletedProps);
```
stefan-egli: Undecided about this log.info. In classic GC we do `maybeLogStats`, to reduce log frequency. This additional log.info might be noisy, wdyt?

stefan-egli (approved):
+1, lgtm now (only added 2 comments about potentially noisy log.infos). As we're going via an integration branch we'll do more testing and another round of reviewing for the final PR anyway.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreHelper.java
```diff
@@ -72,8 +72,8 @@ public static void garbageReport(DocumentNodeStore dns) {
     }
 
     public static VersionGarbageCollector createVersionGC(
-            DocumentNodeStore nodeStore, VersionGCSupport gcSupport) {
-        return new VersionGarbageCollector(nodeStore, gcSupport);
+            DocumentNodeStore nodeStore, VersionGCSupport gcSupport, final boolean detailedGCEnabled) {
```
reschke:
```suggestion
            DocumentNodeStore nodeStore, VersionGCSupport gcSupport, boolean detailedGCEnabled) {
```

reschke: Unneeded and not consistent with the code base...

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocument.java
```diff
@@ -1669,6 +1670,24 @@ SortedMap<Revision, String> getLocalMap(String key) {
...
+     * <p>
+     *  Note: property names returned are escaped
+     * <p/>
```
reschke:
```suggestion
     *
```

reschke: Not needed.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocument.java
```diff
@@ -1669,6 +1670,24 @@ SortedMap<Revision, String> getLocalMap(String key) {
...
+     * @return Set of all property names (escaped)
```
reschke, line 1678: Question: what kind of escaping is this?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
reschke:
```suggestion
                                    VersionGCOptions options, GCMonitor gcMonitor, boolean detailedGCEnabled) {
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCSupport.java
```diff
@@ -59,28 +69,61 @@ public VersionGCSupport(DocumentStore store) {
...
-            }
+    /**
+     * Returns documents that have a {@link NodeDocument#MODIFIED_IN_SECS} value
+     * within the given range and are greater than given @{@link NodeDocument#ID}.
+     * <p>
+     * The two passed modified timestamps are in milliseconds
+     * since the epoch and the implementation will convert them to seconds at
+     * the granularity of the {@link NodeDocument#MODIFIED_IN_SECS} field and
+     * then perform the comparison.
+     * <p/>
```
reschke:
```suggestion
     *
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -521,6 +614,123 @@ private VersionGCStats gc(long maxRevisionAgeInMillis) throws IOException {
             return stats;
         }
 
+        /**
+         * "Detailed garbage" refers to additional garbage identified as part of OAK-10199
+         * et al: essentially garbage that in earlier versions of Oak were ignored. This
+         * includes: deleted properties, revision information within documents, branch
+         * commit related garbage.
+         * <p/>
```
reschke:
```suggestion
         * <p>
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -521,6 +614,123 @@ private VersionGCStats gc(long maxRevisionAgeInMillis) throws IOException {
...
+         * TODO: limit this to run only on a singleton instance, eg the cluster leader
+         * <p/>
```
reschke:
```suggestion
         * <p>
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentStore.java
```diff
@@ -1853,6 +1853,16 @@ protected <T extends Document> Iterable<T> queryAsIterable(final Collection<T> c
             }
         }
 
+        if (sortBy != null && !sortBy.isEmpty()) {
+            for (String key: sortBy) {
```
reschke:
```suggestion
            for (String key : sortBy) {
```

reschke: (whitespace)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBVersionGCSupport.java
```diff
@@ -85,6 +106,67 @@ public Iterable<NodeDocument> getPossiblyDeletedDocs(final long fromModified, fi
...
 
+    /**
+     * Returns documents that have a {@link NodeDocument#MODIFIED_IN_SECS} value
+     * within the given range and are greater than given @{@link NodeDocument#ID}.
+     * <p>
+     * The two passed modified timestamps are in milliseconds
+     * since the epoch and the implementation will convert them to seconds at
+     * the granularity of the {@link NodeDocument#MODIFIED_IN_SECS} field and
+     * then perform the comparison.
+     * <p/>
```
reschke:
```suggestion
     *
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentStore.java
reschke: I'm not 100% sure why ID wasn't here before. Maybe an oversight, but I woul not bet on it...

reschke (commented):
Various nits re Javadoc (there is no such thing as "\<p/>" in HTML).

Re RDB: I'm a bit reluctant here unless we do actually test this on all databases we support (I bet this has not happened). Maybe it would be better to just state that this mode is not supprted for RDB, and only work on this once a real use case comes up?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
rishabhdaim: this would be called only after every batch of 1000 documents (and that too in case there is any garbage).

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocument.java
rishabhdaim: While saving property names in Mongo, we escape those names that are not supported by Mongo.

e.g. `a.b`. Mongo would store this as a `b` sub-document inside `a`, not as `a.b`.

To avoid such cases we do escape before saving into the database.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
stefan-egli: but also in incremental case once detail gc has caught up and is handling fewer than 1000 documents?

rishabhdaim: `removeGarbage()` won't run in case there is no garbage. So it wouldn't log this message.

And in case of garbage, I think it would be better to log since we are deleting documents.

stefan-egli: We should make sure that a normally running Oak instance doesn't excessively log - and this log.info seems to be going to be a bit noisy. I don't think we need to log every single garbage deletion?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -298,16 +340,26 @@ public String toString() {
...
+                        df.format(deleteDetailedGCDocs.elapsed(MICROSECONDS), MICROSECONDS),
+                        df.format(collectDeletedProps.elapsed(MICROSECONDS), MICROSECONDS),
+                        df.format(collectDeletedOldRevs.elapsed(MICROSECONDS), MICROSECONDS),
+                        df.format(collectUnmergedBC.elapsed(MICROSECONDS), MICROSECONDS),
                         timeDeletingSplitDocs);
             }
 
             return "VersionGCStats{" +
                     "ignoredGCDueToCheckPoint=" + ignoredGCDueToCheckPoint +
+                    "ignoredDetailedGCDueToCheckPoint=" + ignoredDetailedGCDueToCheckPoint +
```
stefan-egli:
```suggestion
                    ", ignoredDetailedGCDueToCheckPoint=" + ignoredDetailedGCDueToCheckPoint +
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentStore.java
rishabhdaim: As discussed over [jira](https://issues.apache.org/jira/browse/OAK-10199?focusedCommentId=17751662&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17751662), will revert RDB changes.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilder.java
```diff
@@ -113,4 +115,27 @@ public MissingLastRevSeeker createMissingLastRevSeeker() {
             return super.createMissingLastRevSeeker();
         }
     }
+
+    @Override
```
reschke, line 119: Would it make sense to remove these, andhave default implementations in the interface?

reschke (approved):
Looks good to me RDB-wise (I previously thought that the new methods are on RDBDocumentStore, not ...Builder)

PR #994, opened 2023-06-19
OAK-10307: oak-shaded-guava leaks original guava as transitive dependency

reschke:


oak-shaded-guava/pom.xml
```diff
@@ -57,13 +57,22 @@
                   <shadedPattern>${pref}</shadedPattern>
                 </relocation>
               </relocations>
-              <createDependencyReducedPom>false</createDependencyReducedPom>
+              <createDependencyReducedPom>true</createDependencyReducedPom>
```
mreutegg, line 60: Default value is true. I would rather remove the line.

PR #995, opened 2023-06-19
OAK-10313: Identify revisions created by late-write scenario

mreutegg:
New tests
Consistency implementation to identify revisions
New consistency option for oak-run documentstore-check

### DIFF
--- a/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/WriteAfterRecoveryTest.java
+++ b/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/WriteAfterRecoveryTest.java
@@ -437,11 +437,23 @@ public class WriteAfterRecoveryTest {
         Set<Revision> inconsistent = checkConsistency(ns2, "/a/b", "/a/b/c");
         assertEquals(lateWriteRevisions, inconsistent);
 
+        String checkpoint = ns2.checkpoint(Integer.MAX_VALUE);
+        assertEquals(stringProperty("p", "u"), getNode(ns1.retrieve(checkpoint), "/a/b").getProperty("p"));
+
         // by making another change on a property with an inconsistency,
         // that inconsistency then becomes unnoticeable going forward
         assertEquals(lateWriteRevisions, checkConsistency(ns2, "/a/b"));
         merge(ns2, setProperty("/a/b", "p", "v2"), true);
         assertEquals(emptySet(), checkConsistency(ns2, "/a/b"));
+
+        // after a change that updates _lastRev on a document with an
+        // inconsistency, the late-write becomes visible to reads from
+        // the checkpoint
+        merge(ns1, createChild("/a/b/e"), true);
+        // invalidation is required here as otherwise it reads a cached node
+        // state that still returns p=u
+        ns1.getNodeCache().invalidateAll();
+        assertEquals(stringProperty("p", "v"), getNode(ns1.retrieve(checkpoint), "/a/b").getProperty("p"));
     }
 
     @Test
--- a/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/WriteAfterRecoveryTest.java
+++ b/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/WriteAfterRecoveryTest.java
@@ -1,0 +1,661 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/WriteAfterRecoveryTest.java
```diff
@@ -0,0 +1,661 @@
...
+        assertTrue(ns2.getLastRevRecoveryAgent().isRecoveryNeeded());
+        assertEquals(0, ns2.getLastRevRecoveryAgent().recover(1));
+
+        // 'late write'
+        store.createOrUpdate(NODES, failed);
+        Set<Revision> lateWriteRevisions = getRevisions(failed);
+
+        // revive clusterId 1
+        ns1 = createNodeStore(1);
+        merge(ns1, createChild("/d"), true);
```
stefan-egli, line 257:
```suggestion
        assertEquals(1, checkConsistency(ns1, "/a/b").size());
        merge(ns1, createChild("/d"), true);
        assertEquals(0, checkConsistency(ns2, "/a/b").size());
```

Might be useful to add for illustration, that `ns1` already sees the inconsistency, while `ns2` requires a backgroundRead.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/WriteAfterRecoveryTest.java
```diff
@@ -0,0 +1,661 @@
...
+
+        ns2.runBackgroundOperations();
+        assertTrue(getNode(ns2.getRoot(), "/d").exists());
+        assertEquals(stringProperty("p", "u"), getNode(ns2.getRoot(), "/a/b").getProperty("p"));
+
+        // current implementation does not fail with conflict when property
+        // is overwritten, but consistency check can still detect it
+        Set<Revision> inconsistent = checkConsistency(ns2, "/a/b", "/a/b/c");
+        assertEquals(lateWriteRevisions, inconsistent);
+    }
```
stefan-egli, line 457:
```suggestion
        // by making another change on a property with an inconsistency,
        // that inconsistency then becomes unnoticeable going forward
        assertEquals(1, checkConsistency(ns2, "/a/b").size());
        merge(ns2, setProperty("/a/b", "p", "v2"), true);
        assertEquals(0, checkConsistency(ns2, "/a/b").size());
    }
```

It might be obvious, but perhaps we can add it in one place, that if a property with an inconsistency is further updated, that inconsistency "disappears".

stefan-egli (approved):
+1, lgtm!
I added just 2 comments with some more assert ideas.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/WriteAfterRecoveryTest.java
stefan-egli: PS: there might be exotic cases that one could construct where a property changes as a result of such an overwrite when reading the node via a checkpoint (it is fine for head though)

stefan-egli: And now I remember another test idea : when making a change on a *different* property on the same node, the original property state also changes - that might be a useful test to add perhaps. 
This could also be further extended with a test where a node was removed via a late write, at which point it would purely be an inconsistency. However, a later modification on a property on that removed node would actually immediately remove the node.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/WriteAfterRecoveryTest.java
mreutegg: Good suggestion. I added a slightly different assert, but functionally what you suggested.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/WriteAfterRecoveryTest.java
mreutegg: Added all those suggestions, except for the exotic case, which I didn't understand. Can you please explain in more detail what you have in mind?

stefan-egli: With exotic I was refering to a case where someone would have a reference on an inconsistent property value - that normally gets filtered out, but only until _lastRev on the document changes because of an unrelated change (which is similar to the test of setting an unrelated property that makes the inconsistency disappear). If someone has a checkpoint to the document between the inconsistency and a new update, then the new update changes the property value for the one reading via the checkpoint. Eg as below:

```
        String checkpoint = ns1.checkpoint(Long.MAX_VALUE);
        assertEquals(stringProperty("p", "u"), getNode(ns1.retrieve(checkpoint), "/a/b").getProperty("p"));
        merge(ns1, createChild("/a/b/e"), true);
        // invalidation is required here as otherwise it reads an old node state that doesn't trigger a failure in below assertEquals
        ns1.getNodeCache().invalidateAll();
        assertEquals(stringProperty("p", "u"), getNode(ns1.retrieve(checkpoint), "/a/b").getProperty("p"));
```

mreutegg: OK, thanks. I added similar asserts to test propertyChanged().

PR #1005, opened 2023-06-26
OAK-10328 fix scope of "jackrabbit-jcr-tests"

kwin:


oak-jcr/pom.xml
```diff
@@ -423,6 +423,7 @@
       <groupId>org.apache.jackrabbit</groupId>
       <artifactId>jackrabbit-jcr-tests</artifactId>
       <version>${jackrabbit.version}</version>
+      <scope>test</scope>
       <optional>true</optional>
```
mreutegg, line 427: And at the same time remove optional flag?

kwin: Although not considered right now, I think the optional flag kind of indicates, that this is not supposed to be part of the Maven test compile classpath (but only necessary for executing some tests). Maven however does not support this kind of granularity (at least yet), but leaving the optional in does not do any harm, because there is some semantics behind that (not evaluated by Maven apart from having an impact as transitive dependency). I am fine with both leaving or removing, up to you to decide...

mreutegg: OK, then let's keep it for now.

mreutegg (approved):

PR #1011, opened 2023-06-30
OAK-10334: Node.addMixin() may overwrite existing mixins

mreutegg:
Test reproducing issue with jcr:mixinTypes


oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/NodeImpl.java
```diff
@@ -1015,6 +1015,9 @@ public void checkPreconditions() throws RepositoryException {
                     throw new VersionException(format(
                             "Cannot add mixin type. Node [%s] is checked in.", getNodePath()));
                 }
+                // OAK-10334: adding mixin requires permission to read existing mixin types
+                PropertyState prop = PropertyStates.createProperty(JCR_MIXINTYPES, singleton(oakTypeName), NAMES);
+                sessionContext.getAccessManager().checkPermissions(dlg.getTree(), prop, Permissions.READ_PROPERTY);
```
anchela: hi @mreutegg , i wonder if it wouldn't be better to also adopt a similar approach as we use in `NodeImpl.getPrimaryType` and `getMixinTypeNames`.... i.e. making sure we don't overwrite the type in the tree-util by potentially reading the value from the read-only tree... in particular since TreeUtil.addMixins is also called internally afaik in the oak code.... in other words: this problem may still occur in case the mixin is added directly in the oak api (and not through JCR api) as this happens in the implementation.

also: if we want to keep the permission check here, we should also adjust Node.canAddMixin

anchela (changes requested):

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/NodeImpl.java
stefan-egli: @anchela,
>  i wonder if it wouldn't be better to also adopt a similar approach as we use in {{NodeImpl.getPrimaryType}} and {{getMixinTypeNames}}.... i.e. making sure we don't overwrite the type in the tree-util by potentially reading the value from the read-only tree

Are you referring to OAK-2441, so basically bypassing the read restriction by force-reading through a `getReadOnlyTree`?

> also: if we want to keep the permission check here, we should also adjust Node.canAddMixin

+1 - except I guess if we went the `getReadOnlyTree` way, then this wouldn't be needed

anchela: @stefan-egli , what i meant was: if the mixin property is not readable to the current session we could obtain the existing values from the read-only-tree. this would prevent the overwrite issue.

> if we went the getReadOnlyTree way, then this wouldn't be needed

not sure i get that.... but what i intended to suggest is to also keep Node.canAddMixin in mind and make sure the 2 methods are consistent. hope that helps.

mreutegg: >  i wonder if it wouldn't be better to also adopt a similar approach as we use in `NodeImpl.getPrimaryType` and `getMixinTypeNames`.... i.e. making sure we don't overwrite the type in the tree-util by potentially reading the value from the read-only tree...

That's a good point and I think would be more consistent. Oak and Jackrabbit allow you to get primary and mixin type information even if the session does not have permission to read properties where those values are stored in. It would be strange if adding a mixin fails in this case.

I'll update the PR accordingly.

mreutegg: Done.

@anchela, please have a look at the changes again. Thanks.

mreutegg: It appears Angela is away. @stefan-egli, can you please review when you get a chance?

reschke (approved):
Looks good to me.

stefan-egli (approved):
+1, lgtm.

What I'm wondering however, is whether this needs to be extended to `removeMixin`. There is code in this PR that fixes this pattern (write-but-no-read-permission) within the `addMixin` part of `updateMixin`, however, I'm unsure if there couldn't be the same issue in the remove part of `updateMixin`. It looks like `NodeDelegate.removeMixin` has some check to see if the to-be-removed mixin is actually part of the current mixin list - and that check is IIUC done with the actual session permission. So in the write-but-no-read-permission case that would probably fail there. But if `addMixin` now becomes more tolerant, then wouldn't `removeMixin` also need to be more tolerant? (But if that'd be the case, we could also handle this in a separate ticket, hence still +1)

mreutegg:
>  It looks like NodeDelegate.removeMixin has some check to see if the to-be-removed mixin is actually part of the current mixin list - and that check is IIUC done with the actual session permission. So in the write-but-no-read-permission case that would probably fail there.

That's correct.

> But if addMixin now becomes more tolerant, then wouldn't removeMixin also need to be more tolerant?

Yes, I agree, but I rather handle this in a separate issue. The impact is a bit different. Removing a mixin would fail and not result in an unintentional change like in the addMixin() case.

mreutegg:
Follow up issue regarding Node.removeMixin(String): [OAK-10425](https://issues.apache.org/jira/browse/OAK-10425)

rishabhdaim (approved):

mreutegg:
All tests are successful, except DocumentStoreIndexerIT. This is a known issue: https://issues.apache.org/jira/browse/OAK-10413

PR #1012, opened 2023-06-30
OAK-10333 Improved logging for queries that that traverse 20'000 nodes

thomasmueller:


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/cursor/TraversingCursor.java
```diff
@@ -159,10 +159,15 @@ private void fetchNext() {
 
                 readCount++;
                 if (readCount % 1000 == 0) {
-                    FilterIterators.checkReadLimit(readCount, settings);
-                    String caller = IndexUtils.getCaller(this.settings.getIgnoredClassNamesInCallTrace());
-                    LOG.warn("Traversed {} nodes with filter {} called by {}; consider creating an index or changing the query" , 
-                            readCount, filter, caller);
+                    if (readCount == 20000) {
+                        LOG.warn("Traversed {} nodes with filter {}; consider creating an index or changing the query",
+                                readCount, filter, new Exception("call stack"));
```
nfsantos, line 164: Very minor suggestion: you could use Throwable instead of Exception, because it is the base class of the exception hierarchy (a matter of being minimalistic).

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndex.java
```diff
@@ -429,7 +429,12 @@ public String next() {
                     readCount++;
                     if (readCount % TRAVERSING_WARNING == 0) {
                         Cursors.checkReadLimit(readCount, settings);
-                        log.warn("Index-Traversed {} nodes with filter {}", readCount, plan.getFilter());
+                        if (readCount == 2 * TRAVERSING_WARNING) {
```
nfsantos, line 432: So this will print the stack trace the second time that the traverse warning message is displayed and then it will not print it again. If the idea is to print it only once, why not print it with the first message, so the first time the user is warned?

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndex.java
```diff
@@ -1180,7 +1180,12 @@ public String next() {
                     readCount++;
                     if (readCount % TRAVERSING_WARNING == 0) {
                         Cursors.checkReadLimit(readCount, settings);
-                        LOG.warn("Index-Traversed {} nodes with filter {}", readCount, filter);
+                        if (readCount == 2 * TRAVERSING_WARNING) {
```
nfsantos, line 1183: Same comment as before, why not print it only on the first warning message.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/cursor/TraversingCursor.java
thomasmueller: Not sure if Throwable also works, according to the docs is needs to be an "exception" object: https://www.slf4j.org/faq.html#paramException

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndex.java
thomasmueller: Yes. There will be a warning (without stack trace) at 10'000 nodes read. And at 20'000, there will be a warning PLUS a stack trace.

> why not print it with the first message

My idea is that 10'000 is kind of OKish... so for those developers that see the message and fix the problem, the more "severe" stack trace is not listed. 

Another reason is that the distribution is right skewed (the median is low), so this will log less stack traces.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/cursor/TraversingCursor.java
nfsantos: They are being imprecise with the use of the term exception, the method takes a Throwable as an argument:
https://www.javadoc.io/doc/org.slf4j/slf4j-api/latest/org/slf4j/Logger.html#warn(java.lang.String,java.lang.Throwable)

nfsantos (approved):

fabriziofortino (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/cursor/TraversingCursor.java
joerghoh, line 164: Why do you log the full stack trace? Isn't logging the caller (as it is done a few lines alter by 
```
String caller = IndexUtils.getCaller(this.settings.getIgnoredClassNamesInCallTrace());
```

good enough? The stack trace creates a lot more noise, and I don't see it to be more actionable in the first place.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/strategy/ContentMirrorStoreStrategy.java
```diff
@@ -457,9 +457,15 @@ private void fetchNextPossiblyDuplicate() {
                         readCount++;
                         if (readCount % TRAVERSING_WARN == 0) {
                             FilterIterators.checkReadLimit(readCount, settings);
-                            String caller = IndexUtils.getCaller(settings.getIgnoredClassNamesInCallTrace());
-                            LOG.warn("Index-Traversed {} nodes ({} index entries) using index {} with filter {}, caller {}", 
-                                    readCount, intermediateNodeReadCount, indexName, filter, caller);
+                            if (readCount == 2 * TRAVERSING_WARN) {
+                                LOG.warn("Index-Traversed {} nodes ({} index entries) using index {} with filter {}",
+                                        readCount, intermediateNodeReadCount, indexName, filter,
+                                        new Exception("call stack"));
```
joerghoh, line 463: same as above.

joerghoh (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/cursor/TraversingCursor.java
thomasmueller: Not it is not good enough.

> The stack trace creates a lot more noise

But it is needed. In the past, we have seem multiple cases where the caller was not sufficient. The stack trace is needed.

joerghoh: Well, but then we have a inconsistent reporting here, and we should harmonize it. And then this magic number ``20000`` makes me wonder what is so special about it (and why this exception is not reported at 10k or 80k)?

What information is needed for analysis? I think that the full stack trace is always noise, but you just need more information than currently provided by ``IndexUtils.getCaller()``. What information is this? Mabye 5 more lines of the stack below that line which is currently reported?

PR #1013, opened 2023-07-02
OAK-10335 TypeEditor misapplies default primary type of child node to its parent when the child primary type is not set

adamcin:


mattvryan (commented):
This change LGTM, but I'm not very familiar with this part of the codebase.

@mreutegg / @mduerig / @thomasmueller perhaps one of you - or someone else more familiar - could also take a look?

anchela (changes requested):
@adamcin , please don't add unrelated changes to this PR like changing variable names. it makes it hard to identify the effective changes you are making.

as far as the effective changes are concerned: i would need to take a closer look at the TypeEditor and it's usages. in general 'childNodeChanged' receives the nodestate of the modified node as 'childNodeAdded' receives the newly added child node. i don't recall exactly what the nature of the 'builder' is.... and if retrieving the child is equivalent to the state that is passed.

anyway: as far as i can see the relevant change is actually only on line 299, where the primary type gets set on the child-builder (which is then already available for the 'validate the child node'). so the PR should only change that and avoid any unrelated changes. i can then take a look again in the ide to get a better sense of then change.

adamcin:
> @adamcin , please don't add unrelated changes to this PR like changing variable names. it makes it hard to identify the effective changes you are making.
> 
> as far as the effective changes are concerned: i would need to take a closer look at the TypeEditor and it's usages. in general 'childNodeChanged' receives the nodestate of the modified node as 'childNodeAdded' receives the newly added child node. i don't recall exactly what the nature of the 'builder' is.... and if retrieving the child is equivalent to the state that is passed.
> 
> anyway: as far as i can see the relevant change is actually only on line 299, where the primary type gets set on the child-builder (which is then already available for the 'validate the child node'). so the PR should only change that and avoid any unrelated changes. i can then take a look again in the ide to get a better sense of then change.

Thanks, @anchela . I'll restore the original variable names.

The `after` NodeState appears to represent the child node state in both the childNodeChanged and childNodeAdded methods, at least according to the `Editor` interface [1], as well the `TypeEditor` implementation of `childNodeAdded` [2], which delegates to `childNodeChanged` by passing the `after` parameter without modifying it.

[1] https://github.com/apache/jackrabbit-oak/blob/03b703e91f9964a2f192ccd34526d5980561839a/oak-store-spi/src/main/java/org/apache/jackrabbit/oak/spi/commit/Editor.java#L93-L120
[2] https://github.com/apache/jackrabbit-oak/blob/03b703e91f9964a2f192ccd34526d5980561839a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/nodetype/TypeEditor.java#L281-L286

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/nodetype/TypeEditorTest.java
```diff
@@ -324,4 +326,34 @@ public void malformedUUID() throws CommitFailedException {
             assertTrue(e.isConstraintViolation());
         }
     }
+
+    @Test
+    public void addChildNodeNotStrictDefaultPrimaryTypeIfUnspecified() throws CommitFailedException {
+        // set strict to 'false' to allow unset jcr:primaryType on new node
```
anchela: maybe saying "to allow for a missing jcr:primaryType on a new node" would be easier to understand.
i am not a native speaker and to me it was not clear what 'unset' means.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/nodetype/TypeEditorTest.java
```diff
@@ -324,4 +326,34 @@ public void malformedUUID() throws CommitFailedException {
...
+
+        NodeBuilder policyBuilder = builder.child("conf")
+                .setProperty(JCR_PRIMARYTYPE, "nt:folder", Type.NAME)
+                .setProperty(JCR_MIXINTYPES, ImmutableList.of(AccessControlConstants.MIX_REP_ACCESS_CONTROLLABLE), Type.NAMES)
+                .child("rep:policy")
+                // create the policy node with the correct primary type in the before state
+                .setProperty(JCR_PRIMARYTYPE, AccessControlConstants.NT_REP_ACL, Type.NAME);
+
+        NodeState before = builder.getNodeState();
+        policyBuilder.child("allow")
```
anchela, line 359: please add a comment here as well: because the crucial piece is that this 'allow' child has no jcr:primaryType property and with 'strict' false it should be set to the default primary type defined by the parent.
so, it's the combination of the parent having a primary type that defines a default-primary-type for child nodes and the fact that this 'allow' child lacks the primary type. and extra comment here would make the test a lot easier to read.... (i had to go back to your bug report to understand it)

anchela (changes requested):
hi @adamcin , i had a close look at your original report and the usage of the TypeEditor. the fix looks sensible to me. 
i would suggest though that you improve the comments on the test case and also add a link to the jira ticket with the test-case. the next one looking at the test would then have an immediate help to understanding what this test-case is trying to verify and why it is there.

PR #1015, opened 2023-07-03
OAK-9837 Update to individual OSGi R7 dependencies

kwin:
Update annotations to R7 as well


reschke (commented):
I don't have sufficient OSGi knowledge to review.

That said, can we please schedule this for post 1.54.0 (which I *plan* to cut sometimes next week)?

kwin:
Sure we can defer to 1.54.0. @reschke Do you know who else of the JR committers has sufficient OSGi know-how?

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/osgi/OsgiWhiteboard.java
```diff
@@ -69,13 +69,13 @@ public <T> Registration register(
         Objects.requireNonNull(properties);
         checkArgument(type.isInstance(service));
 
-        Dictionary<Object, Object> dictionary = new Hashtable<Object, Object>();
+        Dictionary<String, Object> dictionary = new Hashtable<>();
         for (Map.Entry<?, ?> entry : properties.entrySet()) {
-            dictionary.put(entry.getKey(), entry.getValue());
+            dictionary.put((String)entry.getKey(), entry.getValue());
```
reschke: Is a cast safe here? Wouldn't be "toString()" better?

reschke (changes requested):

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/osgi/OsgiWhiteboard.java
kwin: cast is safe, because `Property` always only expose String, String pairs (https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html). Compare with https://stackoverflow.com/questions/873510/why-does-java-util-properties-implement-mapobject-object-and-not-mapstring-st

kwin: Actually the cast is no even necessary here, I am gonna remove it.

kwin: @reschke Is changing this SPI method signature ok to only accept Map<String, ?> because that is the only one OSGi accepts (see https://docs.osgi.org/javadoc/r4v43/core/org/osgi/framework/BundleContext.html#registerService(java.lang.Class,%20S,%20java.util.Dictionary) or do we need to maintain backwards compatibility here as well?

reschke:
I'm seeing:
~~~
ERROR] Errors:
[ERROR]   OSGiIT.bundleStates » ClassNotFound org.apache.jackrabbit.oak.spi.state.NodeSt...
[ERROR]   OSGiIT.listBundles » ClassNotFound org.apache.jackrabbit.oak.spi.state.NodeSto...
[ERROR]   OSGiIT.listServices » ClassNotFound org.apache.jackrabbit.oak.spi.state.NodeSt...
[ERROR]   OSGiIT.testLeaseFailureHandlerIsExported » ClassNotFound org.apache.jackrabbit...
[ERROR]   OSGiIT.testNodeStore » ClassNotFound org.apache.jackrabbit.oak.spi.state.NodeS...
[ERROR]   OSGiIT.testRepository » ClassNotFound org.apache.jackrabbit.oak.spi.state.Node...
[INFO]
[ERROR] Tests run: 11, Failures: 0, Errors: 6, Skipped: 0

~~~

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/osgi/OsgiWhiteboard.java
reschke: if we change it, BND will claim that's an incompatible change, no?

kwin: Not sure, as Generics should be a compile time thing only, but I left the signature as is and clarified the javadoc just a bit. But throwing a ClassCastException is better than silently using the toString() so I left the cast.

kwin:
@reschke I am pretty sure the other IT failures are not related, https://ci-builds.apache.org/blue/organizations/jenkins/Jackrabbit%2Foak-trunk-pr/detail/PR-1015/6/pipeline. WDYT?

reschke:
Testing right now.

mbaedke:
> @reschke I am pretty sure the other IT failures are not related, https://ci-builds.apache.org/blue/organizations/jenkins/Jackrabbit%2Foak-trunk-pr/detail/PR-1015/6/pipeline. WDYT?

Yes, they are not related.

reschke (approved):

PR #1018, opened 2023-07-06
OAK-10338 - Do not use platform-dependent newlines in for files generated when creating a FlatFileStore

nfsantos:
Instead always use \n.


reschke:
That does fix the issue, but I'd really prefer to restrict changes to the test case; changing the actual behavior of oak-commons might not be a good idea given the fact it hasn't changed in ages.

mbaedke:
This doesn't fix the problem if the original file had CRLF (or even CR :)) line endings.

mbaedke:
I think we should either leave the code as is and fix the test or change the code so that the result file has the same line endings as the source file.

nfsantos:
I have reverted the changes in the production code and fixed the test to tolerate different line endings.

fabriziofortino (approved):

PR #1020, opened 2023-07-10
OAK-10341

nfsantos:


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/TreeSortPipelinedTransformTask.java
```diff
@@ -0,0 +1,232 @@
...
+                                    totalEntryCount++;
+
+                                    if (nseBatch.size() == MAX_NSE_BATCH_SIZE) {
+//                                        LOG.info("Buffer full, passing buffer to sort task. Total entries: {}, entries in buffer {}",
+//                                                totalEntryCount, nseBatch.size());
+                                        tryEnqueue(nseBatch);
+                                        nseBatch = new ArrayList<>(MAX_NSE_BATCH_SIZE);
+                                    }
+                                    // Write entry to buffer
+                                    nseBatch.add(new NodeStateEntryJson(nse.getPath(), entryWriter.asJson(nse.getNodeState())));
```
thomasmueller: here we need to add the entry for the child node name... that would be something like this (not compiled or tested):
```
String path = nse.getPath();
if (!path.equals("/")) {
    String nodeName = PathUtils.getName(path);
    String parentPath = PathUtils.getParentPath(path);
    nseBatch.add(new NodeStateEntryJson(parentPath + "\t" + nodeName, ""));
}
```

PR #1021, opened 2023-07-12
OAK-10347 : Adding small util class for exporting a tree as flat file

stefan-egli:


stefan-egli:
@nfsantos, @thomasmueller , wdyt? can you please review?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/SimpleFlatFileUtil.java
```diff
@@ -0,0 +1,109 @@
...
+        totalLines += entryBatch.size();
+        entryBatch.clear();
+    }
+
+    private void addEntryAndTraverseChildren(NodeState ns) throws IOException {
+        addEntry(ns);
+        if (entryBatch.size() > 999) {
+            flush();
+        }
+        for (ChildNodeEntry e : ns.getChildNodeEntries()) {
```
nfsantos: If I understood correctly, this will be querying directly the local MongoDB instance. And this is doing a depth-first traversal of the tree rooted in NodeState given to the createFlatFileFor method. Is return of `getChildNodeEntries` deterministic? That is, can you be sure that the FFS file will always be the same for repositories with equivalent contents? Running a revision GC might change the order by which Mongo answers to some queries, even if the results are equivalent. If the order of results of `getChildNodeEntries` is not deterministic, then you could sort the child nodes before iterating over them.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/SimpleFlatFileUtil.java
```diff
@@ -0,0 +1,109 @@
...
+        DocumentNodeState dns = (DocumentNodeState) ns;
+        NodeStateEntry e = new NodeStateEntry.NodeStateEntryBuilder(dns,
+                dns.getPath().toString()).build();
+        String path = e.getPath();
+        if (!NodeStateUtils.isHiddenPath(path) && pathPredicate.test(path)) {
+            String jsonText = entryWriter.asJson(e.getNodeState());
+            // Here logic differs from NodeStateEntrySorter in sense that
+            // Holder line consist only of json and not 'path|json'
+            StateInBytesHolder h = new StateInBytesHolder(path, jsonText);
+            entryBatch.add(h);
```
nfsantos: Why not writing the entry directly to the file instead of having an intermediate step of storing it in the `entryBatch` array? The code is using a buffered writer, so I don't see what is the advantage of having a second layer of buffering with this entryBatch array.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/SimpleFlatFileUtil.java
```diff
@@ -0,0 +1,109 @@
...
+        entryWriter = new NodeStateEntryWriter(null);
+        fw = new FileWriter(f);
+        bw = new BufferedWriter(fw);
+    }
+
+    public static void createFlatFileFor(NodeState ns, File f) throws IOException {
+        final SimpleFlatFileUtil h = new SimpleFlatFileUtil(f);
+        log.info("createFlatFileFor : writing to {}", f.getCanonicalPath());
+        h.addEntryAndTraverseChildren(ns);
+        h.close();
```
nfsantos: nitpick: would be better to use try with resources to ensure the file is closed even when an error occurs.

nfsantos (approved):
LGTM, my only concern is whether the order of the entries in the FFS is deterministic.

thomasmueller (approved):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/SimpleFlatFileUtil.java
stefan-egli: +1, I'll fix that

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/SimpleFlatFileUtil.java
stefan-egli: Good point, I agree this should be addressed. Actually there's also the issue that `NodeStateEntryWriter` excludes the `:childOrder`.

I'm not sure what the best approach is yet. Might have to do some iteration over this code..

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/SimpleFlatFileUtil.java
stefan-egli: +1, good point, I'll fix that.

stefan-egli:
@nfsantos I tried to address your concerns (except the open ordering question)

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/SimpleFlatFileUtil.java
```diff
@@ -0,0 +1,92 @@
...
+    private SimpleFlatFileUtil(BufferedWriter bw) throws IOException {
+        // blobStore is only used for deserialization - so pass null here:
+        entryWriter = new NodeStateEntryWriter(null);
+        this.bw = bw;
+    }
+
+    public static void createFlatFileFor(NodeState ns, File f) throws IOException {
+        log.info("createFlatFileFor : writing to {}", f.getCanonicalPath());
+        try (FileWriter fw = new FileWriter(f);
+                BufferedWriter bw = new BufferedWriter(fw)) {
```
rishabhdaim:
```suggestion
        try (BufferedWriter bw = Files.newBufferedReader(f.toPath())) {
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/SimpleFlatFileUtil.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private final BufferedWriter bw;
+    private final NodeStateEntryWriter entryWriter;
+    private long totalLines = 0;
+
+    private SimpleFlatFileUtil(BufferedWriter bw) throws IOException {
+        // blobStore is only used for deserialization - so pass null here:
+        entryWriter = new NodeStateEntryWriter(null);
+        this.bw = bw;
+    }
```
rishabhdaim: I would remove this constructor since it is a utility class and would pass the required parameters to methods directly.

rishabhdaim (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/SimpleFlatFileUtil.java
stefan-egli: PS: also noticed that `nodeState.getProperties()` does not always return the same order - so the properties probably need to be sorted explicitly (in the `NodeStateEntryWriter`)

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/SimpleFlatFileUtil.java
stefan-egli: I see. It's currently indeed a bit a mix between a utility class and an (writer) object - so actually the name might a bit off. But for now I don't see it doing other things than `createFlatFileFor`- so I tend to favour leaving it this way (but I agree, it's not the nicest class..)

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/SimpleFlatFileUtil.java
stefan-egli: I have now added sorting of both children and properties (to the `SimpleFlatFileUtil` only, not by default to underlying classes). Also added an option to export the `:childOrder` property - this allows explicit veriication of this property (given the sorting does alphabetic sorting now anyway)

stefan-egli:
@nfsantos , @thomasmueller , @rishabhdaim , I've addressed the review comments plus fixed an unrelated test case. I think the PR is now ready to merge. Can you please +1? Thx!

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
```diff
@@ -77,21 +85,30 @@ public String toString(List<String> pathElements, String nodeStateAsJson) {
     }
 
     public String asJson(NodeState nodeState) {
+        return asJson(StreamSupport.stream(nodeState.getProperties().spliterator(), false));
+    }
+
+    String asSortedJson(NodeState nodeState) {
+        return asJson(StreamSupport.stream(nodeState.getProperties().spliterator(), false)
+                .sorted((o1, o2) -> o1.getName().compareTo(o2.getName())));
```
rishabhdaim:
```suggestion
                .sorted(Comparator.comparing(PropertyState::getName));
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/SimpleFlatFileUtil.java
```diff
@@ -0,0 +1,102 @@
...
+    public static void createFlatFileFor(NodeState ns, Writer writer) throws IOException {
+        SimpleFlatFileUtil h = new SimpleFlatFileUtil(writer);
+        h.addEntryAndTraverseChildren(ns);
+        log.info("createFlatFileFor : done. wrote {} lines in total.", h.totalLines);
+    }
+
+    private void addEntryAndTraverseChildren(NodeState ns) throws IOException {
+        addEntry(ns);
+        StreamSupport.stream(ns.getChildNodeEntries().spliterator(), false)
+                .sorted((o1, o2) -> o1.getName().compareTo(o2.getName())).forEach(e -> {
```
rishabhdaim:
```suggestion
                .sorted(Comparator.comparing(ChildNodeEntry::getName)).forEach(e -> {
```

rishabhdaim (approved):

rishabhdaim:
@stefan-egli made minor changes directly to PR.

stefan-egli:
thx @rishabhdaim for those fixes!

PR #1024, opened 2023-07-13
OAK-10345 - Add debug log message if larger blob is compared byte-wise.

ahanikel:
We want to be able to assess how often it happens that large blobs are compared byte-by-byte.


dulceanu (changes requested):
@ahanikel, I was thinking whether we could issue the log statement only when a boolean flag is set. I don't know if the current approach affects performance, but this way we could enable this only when needed.

ahanikel:
@dulceanu Do you mean something like this?

reschke (changes requested):
For reading system properties, please use oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/properties/SystemPropertySupplier.java

ahanikel:
@reschke Like this?

dulceanu (approved):

ahanikel:
Actually it seems quite hard to set a system property if a feature toggle is enabled. Perhaps it's still easier to directly use a feature toggle instead of a system property?

reschke (approved):

PR #1034, opened 2023-07-21
OAK-10318 : Improve AutoMembershipPrincipals#isInheritedMember

anchela:
@nscendoni, @joerghoh , i would appreciate if you could take a look at this change. benchmark results can be found in the subtasks of https://issues.apache.org/jira/browse/OAK-10318

### DIFF
--- a/oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/AutoMembershipPrincipals.java
+++ b/oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/AutoMembershipPrincipals.java
@@ -16,12 +16,12 @@
  */
 package org.apache.jackrabbit.oak.spi.security.authentication.external.impl.principal;
 
-import org.apache.jackrabbit.guava.common.collect.ImmutableSet;
-import org.apache.jackrabbit.guava.common.collect.Iterators;
-import org.apache.jackrabbit.guava.common.collect.Maps;
 import org.apache.jackrabbit.api.security.user.Authorizable;
 import org.apache.jackrabbit.api.security.user.Group;
 import org.apache.jackrabbit.api.security.user.UserManager;
+import org.apache.jackrabbit.guava.common.collect.ImmutableSet;
+import org.apache.jackrabbit.guava.common.collect.Iterators;
+import org.apache.jackrabbit.guava.common.collect.Maps;
 import org.apache.jackrabbit.oak.spi.security.authentication.external.basic.AutoMembershipConfig;
 import org.apache.jackrabbit.oak.spi.security.principal.GroupPrincipals;
 import org.jetbrains.annotations.NotNull;
@@ -32,6 +32,7 @@ import org.slf4j.LoggerFactory;
 import javax.jcr.RepositoryException;
 import java.security.Principal;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/AutoMembershipPrincipals.java
```diff
@@ -127,23 +128,55 @@ boolean isMember(@NotNull String idpName, @NotNull String groupId, @NotNull Auth
...
+    }
+
+    private static boolean hasInheritedMembership(@NotNull Iterator<Group> declared, @NotNull String groupId, 
+                                                  @NotNull String memberId, @NotNull Set<String> processed) throws RepositoryException {
+        List<Group> groups = new ArrayList<>();
+        while (declared.hasNext()) {
+            Group gr = declared.next();
+            String grId = gr.getID();
+            if (memberId.equals(grId)) {
+                log.error("Cyclic group membership detected for group id {}", memberId);
```
joerghoh: Is it possible to provide more information about the nature of this cycle, which would help to break this cycle?

anchela: hi @joerghoh i am not sure this is possible. the best way to investigate the cycle is using Apache Sling. breaklng it involves regular JCR user management API, but that seems obvious to me.

anchela: but i am happy to rephrase it if you have a suggestion

joerghoh: My idea is just to provide more information which groups are involved in this cycle, but not to break it and/or resolve the underlying root cause. That needs to be done elsewhere and on a different level, I agree.

I could imagine that something like this could already help:

```
private static boolean hasInheritedMembership(..., String auditTrail) {
  [...]
  log.error("Cylic group membership detected for group id {} via {}", memberId, auditTrail);
  [...]
  for (Group group : groups) {
    final String trail = auditTrail + " > " + memberId;
    if (hasInheritedMembership(group.declaredMemberOf(), groupId, memberId, processed, trail)) {
       return true;
      }
    }
  }
}
```
The idea is to keep track of the current stack of resolved groups, and in case of a cycle we know how we ended at this place. 
Of course this can be more sophisticated than just appending to a string, but it illustrates the idea. Also it might not be 100% accurate, as the cycle might not begin with the first invocation of this method. But still it should be possible to detect the cycle just be analysing the trail and finding the group name which appears twice.

anchela: @joerghoh , that's doable :)

anchela: @joerghoh , done. if you can take another look that would be great. i also added more tests and rearranged the code a bit to make easier to read)

joerghoh (approved):
lgtm

anchela:
build failure:
Error:  Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.2:test (default-test) on project oak-store-document: There are test failures, which seem unrelated to my changes.
the ci failed on oak-run which also is unrelated.

i am going to merge it.

PR #1042, opened 2023-07-31
OAK-10358 - Filter by path on MongoDB query when downloading repository in Indexing job

nfsantos:
Use a [MongoDB regex query](https://www.mongodb.com/docs/manual/reference/operator/query/regex/) to download only the Mongo documents that are needed for a given index, based on the `includedPaths`/`excludedPaths` properties. 

This PR only applies filtering in the simple cases, where there is a single `includedPath` and no `excludedPaths`. This should be a common case. In the future, we may expand to filter on more complex combinations of these two properties.

The regex used to match a path is of the form `^[0-9]{1,3}:" + path + ".*$`. Since this regex starts with an wildcard, it cannot use the index on `_id`, so Mongo has to scan all documents, like when downloading without regex filtering. The gain in performance of regex filtering is from only sending to the indexing job the documents for the entries in the path being indexed.

### DIFF
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
@@ -30,7 +30,6 @@ import com.mongodb.client.model.Filters;
 import org.apache.jackrabbit.guava.common.base.Preconditions;
 import org.apache.jackrabbit.guava.common.base.Stopwatch;
 import org.apache.jackrabbit.oak.commons.PathUtils;
-import org.apache.jackrabbit.oak.plugins.document.Document;
 import org.apache.jackrabbit.oak.plugins.document.NodeDocument;
 import org.apache.jackrabbit.oak.spi.filter.PathFilter;
 import org.bson.BsonDocument;
@@ -329,14 +328,20 @@ public class PipelinedMongoDownloadTask implements Callable<PipelinedMongoDownlo
             path = path + "/";
         }
         String quotedPath = path.replace("/", "\\/");
-        return regex("_id", "^[0-9]{1,3}:" + quotedPath + ".*$");
+        // For entries with path sizes above a certain threshold, the _id field contains a hash instead of the path of
+        // the entry. The path is stored instead in the _path field. Therefore, we have to include in the filter also
+        // the documents with matching _path.
+        return Filters.or(
+                regex(NodeDocument.ID, "^[0-9]{1,3}:" + quotedPath + ".*$"),
+                regex(NodeDocument.PATH, quotedPath + ".*$")
+        );
     }
 
     private static Bson ancestorsFilter(String path) {
         ArrayList<Bson> parentFilters = new ArrayList<>();
         int depth = PathUtils.getDepth(path);
         while (true) {
-            parentFilters.add(Filters.eq("_id", depth + ":" + path));
+            parentFilters.add(Filters.eq(NodeDocument.ID, depth + ":" + path));
             if (depth == 0) {
                 break;
             }
@@ -353,7 +358,7 @@ public class PipelinedMongoDownloadTask implements Callable<PipelinedMongoDownlo
             try {
                 while (cursor.hasNext()) {
                     BasicDBObject next = cursor.next();
-                    String id = next.getString(Document.ID);
+                    String id = next.getString(NodeDocument.ID);
                     // If we are retrying on connection errors, we need to keep track of the last _modified value
                     if (retryOnConnectionErrors) {
                         this.nextLastModified = next.getLong(NodeDocument.MODIFIED_IN_SECS);
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
@@ -44,11 +44,13 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import static com.mongodb.client.model.Filters.regex;
 import static com.mongodb.client.model.Sorts.ascending;
@@ -89,7 +91,7 @@ public class PipelinedMongoDownloadTask implements Callable<PipelinedMongoDownlo
     // TODO: Revise this timeout. It is used to prevent the indexer from blocking forever if the queue is full.
     private static final Duration MONGO_QUEUE_OFFER_TIMEOUT = Duration.ofMinutes(30);
     private static final int MIN_INTERVAL_BETWEEN_DELAYED_ENQUEUING_MESSAGES = 10;
-    private final static BsonDocument NATURAL_HINT =  BsonDocument.parse("{ $natural:1}");
+    private final static BsonDocument NATURAL_HINT = BsonDocument.parse("{ $natural:1}");
 
     private final int batchSize;
     private final BlockingQueue<BasicDBObject[]> mongoDocQueue;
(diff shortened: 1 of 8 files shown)


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -225,12 +282,96 @@ private void downloadRange(DownloadRange range) throws InterruptedException, Tim
...
+        if (excludedPaths.isEmpty() && includedPaths.size() == 1) {
+            return includedPaths.stream().iterator().next();
+        } else {
+            return null;
+        }
+    }
+
+    private static Bson childrenFilter(String path) {
+        if (!path.endsWith("/")) {
+            path = path + "/";
```
thomasmueller, line 347: If my understanding is correct, this logic of the code is: if the includedPaths is "/content", then we filter for "everything that starts with /content/". But that seems incorrect: we miss the node "/content".

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedIT.java
```diff
@@ -66,52 +69,95 @@ public class PipelinedIT {
...
     @BeforeClass
-    public static void checkMongoDbAvailable() {
+    public static void setup() throws IOException {
         Assume.assumeTrue(MongoUtils.isAvailable());
-    }
-
-    @Before
-    public void setup() throws IOException {
+        // Generate dynamically the entries expected for the long path tests
+        StringBuilder path = new StringBuilder("/content/dam");
```
thomasmueller, line 84: The test case needs to add /content/dam as well.

thomasmueller (changes requested):

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedIT.java
nfsantos: There are more nodes added in the file below, see in line 266:

```java
    private static final List<String> EXPECTED_FFS = new ArrayList<>(List.of(
            "/|{}",
            "/content|{}",
            "/content/dam|{}",
            "/content/dam/1000|{}",
            "/content/dam/1000/12|{\"p1\":\"v100012\"}",
            "/content/dam/2022|{}",
            "/content/dam/2022/02|{\"p1\":\"v202202\"}",
            "/content/dam/2023|{\"p2\":\"v2023\"}",
            "/content/dam/2023/01|{\"p1\":\"v202301\"}",
            "/content/dam/2023/02|{}",
            "/content/dam/2023/02/28|{\"p1\":\"v20230228\"}"
    ));
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
nfsantos: This is being done, see line 195, there are even comments explaining how this case is handled:

```
        // If regex filtering is enabled, start by downloading the ancestors of the path used for filtering.
        // That is, download "/", "/content", "/content/dam" for a base path of "/content/dam". These nodes will not be
        // matched by the regex used in the Mongo query, which assumes a prefix of "???:/content/dam"
        ...
        if (regexBasePath != null) {
            // Regex path filtering is enabled
            // Download the ancestors in a separate query. No retrials done on this query, as it will take only a few
            // seconds and is done at the start of the job, so if it fails, the job can be retried without losing much work
            LOG.info("Using regex filtering with path {}", regexBasePath);
            Bson ancestorsQuery = ancestorsFilter(regexBasePath);
```

oak-store-spi/src/main/java/org/apache/jackrabbit/oak/spi/filter/PathFilter.java
```diff
@@ -108,8 +108,7 @@ public static PathFilter from(@NotNull NodeBuilder defn) {
      * If both are empty then all paths would be considered to be included
      *
      * @param includes list of paths which should not be included
-     * @param excludes list of p
-     *                 aths which should be included
+     * @param excludes list of paths which should be included
```
fabriziofortino: is this documentation correct? It seems the descriptions have been inverted

fabriziofortino (approved):

oak-store-spi/src/main/java/org/apache/jackrabbit/oak/spi/filter/PathFilter.java
nfsantos: Yes, it looks like a mistake in the documentation. This PR does not change this documentation, I only fixed the wrong line break, but now I fixed the text of the docs as well.

tihom88:
How are we tackling long paths. IIRC long paths are saved as <depth>:h<sha-256>/<nodename>

nfsantos:
@tihom88 This is being addressed, see line 353 of PipelinedMongoDownloadTask, method `childrenFilter()`. I am including below the logic that handles this case:

```java
        // For entries with path sizes above a certain threshold, the _id field contains a hash instead of the path of
        // the entry. The path is stored instead in the _path field. Therefore, we have to include in the filter also
        // the documents with matching _path.
        return Filters.or(
                regex(NodeDocument.ID, Pattern.compile("^[0-9]{1,3}:" + quotedPath + ".*$")),
                regex(NodeDocument.PATH, Pattern.compile(quotedPath + ".*$"))
        );
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -225,12 +282,96 @@ private void downloadRange(DownloadRange range) throws InterruptedException, Tim
...
+        // For entries with path sizes above a certain threshold, the _id field contains a hash instead of the path of
+        // the entry. The path is stored instead in the _path field. Therefore, we have to include in the filter also
+        // the documents with matching _path.
+        return Filters.or(
+                regex(NodeDocument.ID, Pattern.compile("^[0-9]{1,3}:" + quotedPath + ".*$")),
+                regex(NodeDocument.PATH, Pattern.compile(quotedPath + ".*$"))
+        );
+    }
+
+    private static Bson ancestorsFilter(String path) {
```
tihom88, line 359: I think we should add longpath logic here. In case path is long the id will not be \<depth\>:\<path\> but instead it will be '\<depth\>:h\<sha256 of parentpath\>/\<nodename\>'

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -215,8 +190,90 @@ private void reportProgress(String id) {
...
+                    numberOfFailures++;
+                    LOG.warn("Retrying download in {} ms; number of times failed: {}; current series of failures started at: {} ({} seconds ago)",
+                            retryIntervalMs, numberOfFailures, failuresStartTimestamp, secondsSinceStartOfFailures);
+                    exceptions.compute(e.getClass().getSimpleName() + " - " + e.getMessage(),
+                            (key, val) -> val == null ? 1 : val + 1
+                    );
+                    try {
+                        Thread.sleep(retryIntervalMs);
+                    } catch (InterruptedException ignore) {
+                    }
```
steffenvan: Could it be a problem that we are not interrupting the thread while it's sleeping when this exception is caught? 
It seems that if an interruption actually occurs, the code would only catch it and continue what it's doing. That might be the intended behaviour, but I'm not 100 % sure. And if not could we do something like: 

```java
try {
    Thread.sleep(retryIntervalMs);
} catch (InterruptedException ignore) {
    // re-interrupt the current thread to set its interrupted status
    Thread.currentThread().interrupt();
}
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
nfsantos: Good point. I added a test and code to handle the case where the path used for filtering is long enough to be handled as a long path: 2ae36bdea6ae2552d3610c45bcb1e406e703b353

I expect that this will never happen in practice, as it would mean that one of the `includePaths` is extremely long, but we should guard against it.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
nfsantos: Yes, you are right. The method with this code already declares throwing `InterruptedException` and the caller is handling that exception explicitly, so the catch clause is not needed here. I removed it.

tihom88 (approved):

PR #1043, opened 2023-08-01
OAK-9922: Parallel Compaction

dulceanu:


reschke (changes requested):
IMHO (!) this PR is way too big.

- any changes that only affect whitespace and import statements whould be removed
- if ApproximateCounter needs to be moved, that should be done as a separate issue (and the old class should be removed); we also need to discuss whether that API is public or not

dulceanu:
@reschke, I re-worked this PR without the changes to ApproximateCounter, removing unnecessary refactorings, import and whitespace changes. Can you have another look at it, please?

dulceanu:
@reschke, given that the test failures for oak-segment-azure settled now, could you please review this?

reschke (approved):

PR #1045, opened 2023-08-02
OAK-10181: Flaky test: DataStoreBlobStoreStatsTest - ensure that the two created blobs are different

reschke:


mbaedke:
That makes sense. Also I have an old Windows notebook where the test failure is reproducible very well. After applying this patch, the test is currently running in a loop without issues. Great job!

mbaedke (approved):

PR #1051, opened 2023-08-07
OAK-10377: Checked-out state of non-versionable nodes is not handled …

mbaedke:
…correctly

Added versionability check.


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionEditor.java
```diff
@@ -212,7 +212,7 @@ private boolean isLockProperty(PropertyState state) {
     private boolean wasCheckedIn() {
         PropertyState prop = before.getProperty(JCR_ISCHECKEDOUT);
         if (prop != null) {
-            return !prop.getValue(Type.BOOLEAN);
+            return !prop.getValue(Type.BOOLEAN) && isVersionable;
```
mreutegg, line 215: Why not use `isVersionable()` instead? `isVersionable` can be null and needs to be computed first. I see calls to wasCheckedIn() are currently always preceded with isVersionable(), but I'd rather not introduce this implicit dependency.

mreutegg (commented):
Other than that looks good to me.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/version/VersionableTest.java
```diff
@@ -362,6 +363,68 @@ public void testIsCheckedOutOnNonAccessibleParent() throws Exception {
...
+        superuser.save();
+        assertFalse(node.isCheckedOut());
+        node.checkout();
+        superuser.save();
+        assertTrue(node.isCheckedOut());
+        node.removeMixin(mixVersionable);
+        superuser.save();
+        assertTrue(node.isCheckedOut());
+        assertFalse(node.hasProperty(jcrIsCheckedOut));
+    }
```
mreutegg, line 426: I think it would be good to extend this test a bit and add mix:versionable again at the end, then verify version semantics are still correct. WDYT?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionEditor.java
mbaedke: Makes sense.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/version/VersionableTest.java
mbaedke: Also makes sense.

PR #1066, opened 2023-08-14
OAK-10399 Automatically pick a merged index over multiple levels

thomasmueller:


fabriziofortino (approved):

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/IndexName.java
```diff
@@ -231,26 +231,37 @@ public String nextCustomizedName() {
...
+    }
+
+    private static boolean isIndexActive(String indexPath, NodeState rootState, int recursionDepth) {
         // An index is active if it has a hidden child node that starts with ":oak:mount-",
         // OR if it is an active merged index
-        try {
-            NodeState indexNode = rootState;
-            for (String e : PathUtils.elements(indexPath)) {
-                indexNode = indexNode.getChildNode(e);
+        if (recursionDepth > 50) {
```
nfsantos: Maybe the maximum recursion depth should be exposed as an internal configuration property. 50 seems reasonable but maybe in the future we may want to increase it. In the very least, this constant should be defined as a static field in the class, so it has a clear name.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/IndexName.java
```diff
@@ -231,26 +231,37 @@ public String nextCustomizedName() {
...
+            LOG.warn("Fail to check index activeness for {} due to high recursion depth", indexPath);
```
nfsantos: Log the recursion depth. And could this happen because of cycles in the index definition? We could explain in the log messages what are the possible causes and fixes for this error.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/IndexName.java
```diff
@@ -231,26 +231,37 @@ public String nextCustomizedName() {
...
+            return true;
+        }
+        NodeState indexNode = rootState;
+        for (String e : PathUtils.elements(indexPath)) {
+            indexNode = indexNode.getChildNode(e);
+        }
```
nfsantos, line 257: This logic looks strange to me, it took me a while to understand it. It's trying to get the child node of the last element on the indexPath, right? Can you re-write this in a way that it is more clear? Something that reads more like `PathUtils.elements(indexPath).last().getChildNode(e)`?

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/IndexName.java
```diff
@@ -231,26 +231,37 @@ public String nextCustomizedName() {
...
+        for (String c : indexNode.getChildNodeNames()) {
+            if (c.startsWith(":oak:mount-")) {
+                return true;
             }
-            for (String c : indexNode.getChildNodeNames()) {
-                if (c.startsWith(":oak:mount-")) {
-                    return true;
-                }
+        }
+        if (recursionDepth >= 2) {
```
nfsantos, line 263: Shouldn't it >= 1? The recursion starts at 0, so at level 1 we can already have an OOTB index, no? Is there a test for this case? That is, for an index `lucene-7-custom-1` with `merges=["lucene-7", "lucene-6-custom2"]`. And are there tests for other cases? I don't see them in this PR or in the IndexNameTest class.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/IndexName.java
thomasmueller: I'm not quite sure what the added indirection would help in this case: we only use the setting once. But OK, I'll add a constant.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/IndexName.java
thomasmueller: Logging the recursion depths seems only useful if we make it configurable. But OK, it doesn't hurt.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/IndexName.java
thomasmueller: This is a _loop_ over all the entries in the path, to go down to the respective node. This code is not changed.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/IndexName.java
thomasmueller: Level 1 (the direct children) will still need to check. 

> Is there a test for this case?

Yes: CustomizedIndexTest and IndexNameTest are in the patch. The test coverage of the changes is 100% as far as I see.

nfsantos (approved):

PR #1075, opened 2023-08-16
OAK-10370 - Dry-run mode for full GC

ankitaagar:
Created Dry-run mode using flag "gc.infoFullGCEnabled" and command  "infofullgc"

### DIFF
--- a/oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
+++ b/oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
@@ -60,6 +60,7 @@ import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreHelper.createVersionGC;
 import static org.apache.jackrabbit.oak.plugins.document.FormatVersion.versionOf;
 import static org.apache.jackrabbit.oak.plugins.document.util.Utils.getRootDocument;
+import static org.apache.jackrabbit.oak.plugins.document.util.Utils.isDetailedGCEnabled;
 import static org.apache.jackrabbit.oak.plugins.document.util.Utils.timestampToString;
 import static org.apache.jackrabbit.oak.run.Utils.asCloseable;
 import static org.apache.jackrabbit.oak.run.Utils.createDocumentMKBuilder;
@@ -226,7 +227,7 @@ public class RevisionsCommand implements Command {
         useMemoryBlobStore(builder);
         // create a version GC that operates on a read-only DocumentNodeStore
         // and a GC support with a writable DocumentStore
-        VersionGarbageCollector gc = createVersionGC(builder.build(), gcSupport);
+        VersionGarbageCollector gc = createVersionGC(builder.build(), gcSupport, isDetailedGCEnabled(builder));
 
         VersionGCOptions gcOptions = gc.getOptions();
         gcOptions = gcOptions.withDelayFactor(options.getDelay());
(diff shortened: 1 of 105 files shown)


stefan-egli:
It looks like you have accidentally deleted files from `oak-shaded-guava` (which also breaks the build in this branch). Also suspecting some other changes are required to get it compile - which I think might be easier to have done before doing a review.

ankitaagar:
> It looks like you have accidentally deleted files from `oak-shaded-guava` (which also breaks the build in this branch). Also suspecting some other changes are required to get it compile - which I think might be easier to have done before doing a review.

Yes, did a mistake somehow, I have only compiled the `oak-store-document` before `git push`
For me, the branch `DetailedGC/OAK-10199` is also giving compilation errors:
[ERROR] org.apache.jackrabbit.guava.common.annotations: Version increase required; detected 32.1.1, suggested 32.1.2
[ERROR] org.apache.jackrabbit.guava.common.base: Version increase required; detected 32.1.1, suggested 32.1.2

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -685,6 +686,9 @@ private void collectDetailedGarbage(final GCPhases phases, final RevisionVector
                                             modified, fromModifiedMs, toModifiedMs);
                                 }
                             }
+                            if (infoFullGCEnabled) {
+                                return;
```
stefan-egli: there's a problem here : with returning early, the part of the code after the `removeGarbage` bit, is not executed. This would mean the iterating logic would somehow break. I think it would be good to have a test case that proves it would break, and then fix this bit.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -122,6 +122,7 @@ public class VersionGarbageCollector {
      */
     static final String SETTINGS_COLLECTION_DETAILED_GC_DOCUMENT_ID_PROP = "detailedGCId";
 
+    public boolean infoFullGCEnabled = false;
```
rishabhdaim: I would change this name to `dryRun` to indicate the intent of this flag.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -93,6 +94,7 @@ private static class RevisionsOptions extends Utils.NodeStoreOptions {
         static final String CMD_COLLECT = "collect";
         static final String CMD_RESET = "reset";
         static final String CMD_SWEEP = "sweep";
+        static final String CMD_INFOFULLGC = "infofullgc";
```
rishabhdaim: I would rename this to `dryRunDetailedGC`

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
rishabhdaim: Rather than adding the `if-else` check here, I would ignore the deletion part inside `removeGarbage()`.

It would keep the flow simple and inside the `dryRun` mode, we can `log` the `properties/documents` that would have been deleted.

The same should be followed for other deletion operations for `deletingDocs` & `splitCleanUp`.

Following this approach wouldn't skip the iterating logic and thus avoid any unintended consequences.

cc @stefan-egli @mreutegg

rishabhdaim (changes requested):

PR #1079, opened 2023-08-21
OAK-10378 : added metrics to collect detailedGC cycle & deleted properties data

rishabhdaim:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/LoggingDocumentStoreWrapper.java
```diff
@@ -283,6 +283,7 @@ public T call() throws Exception {
     }
 
     @Override
+    @NotNull
```
stefan-egli, line 286: nitpicking: IIUC this is unrelated to this branch (and mostly code beautification). I would usually tend to avoid overloading a PR or a branch with unrelated aspects, as it might later complicate backporting. Maybe not with a one liner like here, hence nitpicking :)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DetailedRevisionGCStatsCollector.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Collector interface for {@link DocumentNodeStore} Detailed revision garbage collection
+ * statistics.
+ */
+public interface DetailedRevisionGCStatsCollector {
+
+    void documentRead();
+
+    void deletedPropertiesRead(long numProps);
```
stefan-egli: Do we need this level of granularity?

I understand that `documentsUpdated` might not contain the full set of documents we wanted to update, as `findAndUpdate` might fail for some. 

But I'm wondering if we rather should have a counter for those failures on the document granularity level, rather than deeper (eg deletedPropertiesRead, but also the same for unmerged branch commits and individual revisions). Wdyt?

stefan-egli (commented):
Generally looks good, I have just one comment (besides a nitpick)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DetailedRevisionGCStatsCollector.java
rishabhdaim: I agree that we should have a counter for documents which had not been updated. I will update the PR.

> Do we need this level of granularity?

Collecting low-level metrics might help us understand the nature of the garbage that we are missing along with the number of documents.

However, In case we find this too noisy, we can stop collecting them.

stefan-egli: We do have `propertiesDeleted` though, which is a low-level metric, ie does provide the granularity of how many properties were deleted. The `deletedPropertiesRead` I'm referring to here is usually identical to `propertiesDeleted` unless `findAndUpdate` fails (which should not fail frequently given we mostly update old document).

I'm just asking whether we need it, as I suspect it is nearly a duplicate of `propertiesDeleted`?

stefan-egli (approved):

PR #1080, opened 2023-08-21
OAK-10410 : fixed potential NPE from AbstractMongoConnectionTest

rishabhdaim:


oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/AbstractMongoConnectionTest.java
```diff
@@ -76,9 +78,14 @@ protected Clock getTestClock() throws InterruptedException {
 
     @After
     public void tearDownConnection() throws Exception {
-        String dbName = mongoConnection.getDBName();
+        String dbName = null;
+        if (nonNull(mongoConnection)) {
```
stefan-egli: Is there an advantage of using `nonNull` here?

Intuitively it makes reading slower, as it is not plain code but a function call that needs to be checked.

stefan-egli (approved):

reschke (approved):

PR #1091, opened 2023-08-31
OAK-10413: DocumentStoreIndexerIT hanging

tihom88:


tihom88:
Reverted back old changes in abstract classes being used in various tests. And added an eventual assertion to check on lock

PR #1094, opened 2023-09-01
OAK-10424 override oak.fastQuerySize system property with OSGi configuration granting same effect, but limited to list of allowed principals

adamcin:
jira: https://issues.apache.org/jira/browse/OAK-10424
- return the index result count estimate instead of -1 for an executing user if principal is listed in an OSGi configurable allow list.

Setting the global QueryEngineSettings.getFastQuerySize() value to true is currently the only way to allow service users to leverage JCR query for collecting accurate repository count metrics in a performant way. However, doing so in a multiuser repository may be inadvisable because the fast result size is returned to the caller without considering the caller's read permissions over the paths returned in the result, which may allow less privileged users to discover the presence of nodes that are not otherwise visible to them.

See https://jackrabbit.apache.org/oak/docs/query/query-engine.html#result-size

To address the security risk, the application should provide a configuration of some kind to restrict the ability to effectively leverage this option to permitted system users, which could be implemented as a JCR repository privilege or as an allowlist property in the QueryEngineSettingsService configuration or similar.

To reduce the impact on security and API surface area, this PR only adds a new OSGi service and interface to oak-core, and simple `QueryCountsSettings` interface  that will encapsulate the migrated fastQueryResultSize preference within the `SessionContext`, in case a small number of additional, but similar user-specific preferences are added in the future.

My approach for testing so far is to ensure that [ResultSizeTest](https://github.com/apache/jackrabbit-oak/blob/30139e3fdbdc3db5fa922c503134fe51cf6f7ec7/oak-lucene/src/test/java/org/apache/jackrabbit/oak/jcr/query/ResultSizeTest.java) continues to pass without modification, and that a new [WhiteboardResultSizeTest](https://github.com/adamcin/jackrabbit-oak/blob/0048ba44c8e3b62e2a019cb2d51e9e1f4cb2a8d1/oak-lucene/src/test/java/org/apache/jackrabbit/oak/jcr/query/WhiteboardResultSizeTest.java) passes for the same queries and expected counts, but with the `SessionQuerySettingsProviderService` registered in the oak whiteboard, and using reflection to reconfigure the service between queries in place of runtime changes to the oak.fastQuerySize system property.


oak-core/src/main/java/org/apache/jackrabbit/oak/query/UnionQueryImpl.java
```diff
@@ -195,7 +195,13 @@ public long getSize(SizePrecision precision, long max) {
         long total = QueryImpl.saturatedAdd(a, b);
         return Math.min(localLimit, total);
     }
-    
+
+    @Override
+    public boolean isInsecureResultSize() {
+        // assumes that the same queryOptions are passed to both left and right queries in setQueryOptions()
+        return right.isInsecureResultSize();
```
thomasmueller: Why not use left.isInsecureResultSize() || right.isInsecureResultSize(): that way the implementation can be changed later.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/xpath/Statement.java
```diff
@@ -382,6 +382,18 @@ private static void appendQueryOptions(StringBuilder buff, QueryOptions queryOpt
                             SQL2Parser::escapeStringLiteral));
             optionValues.add("prefetch (" + list + ")");
         }
+        if (queryOptions.hasInsecureOptions()) {
+            List<String> insecureOpts = new ArrayList<>();
+            insecureOpts.add("insecure");
+            if (queryOptions.insecureResultSize) {
+                insecureOpts.add("resultsize");
```
thomasmueller: I don't like "resultsize" as it's two words concatenated... Options are:
* "resultsize": can be hard to read
* "result_size": looks a bit ugly
* "resultSize": the parser is not case sensitive so it might work
* "result size": this approach is used for "index tag" already

I would prefer "result size" (two words).

thomasmueller (commented):
This is not a full review, just some initial comments.
In general, I think the patch looks good, but it will need a longer review.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryOptions.java
```diff
@@ -106,6 +110,14 @@ public QueryOptions(QueryOptions defaultValues) {
             t.read(']');
             prefetch = list;
         }
+        x = map.get("insecureResultSize");
+        if (x != null) {
```
thomasmueller: I think the following should work:

    insecureResultSize = Boolean.parseBoolean(map.get("insecureResultSize"))

oak-core/src/main/java/org/apache/jackrabbit/oak/query/SQL2Parser.java
```diff
@@ -234,6 +249,28 @@ public Query parse(final String query, final boolean initialise) throws ParseExc
...
+
+    /**
+     * Checks the InsecureQueryOptionsMode for the values ALLOW or DENY. If allowed, the method returns true,
+     * indicating that insecure options should be returned by the parser. If denied, a ParseException is thrown.
+     * For all other modes, the method returns false, indicating that this option should be ignored.
+     *
+     * @return true if the mode is ALLOW, false otherwise
+     * @throws ParseException if the mode is DENY
+     */
+    private boolean allowsInsecureOptions() throws ParseException {
```
thomasmueller: I think checking (and throwing the exception) need to happen at execution time, not while parsing.

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/query/SQL2Parser.java
adamcin: I moved the evaluation to QueryImpl as a distinct `verifyInsecureOptions()` method and call it within the `QueryEngineImpl.executeQuery()` method on each query after it has been parsed/inited but before it has been prepared.

fabriziofortino (approved):
LGTM

oak-lucene/src/test/java/org/apache/jackrabbit/oak/jcr/query/ResultSizeTest.java
```diff
@@ -126,5 +151,90 @@ private void doTestResultSize(boolean union, int expected) throws RepositoryExce
...
+        while (it.hasNext()) {
+            Node n = it.nextNode();
+            buff.append(n.getPath()).append('\n');
+        }
+        String regularResult = buff.toString();
+        if (expected >= 0) {
+            assertEquals(regularResult, fastSizeResult);
+        }
+
+        System.clearProperty("oak.fastQuerySize");
```
thomasmueller: This should be inside a "finally" block, so that if the test fails, other tests are not affected.

oak-doc/src/site/markdown/query/query-engine.md
```diff
@@ -267,6 +269,22 @@ Limitations:
   you need to also set the property `refresh` to `true` (Boolean),
   so that the change is applied. No indexing is required.
 
+#### Query Option Insecure Result Size
+
+`@since Oak 1.60.0 (OAK-10424)`
+
+NOTE: The principal executing the query must have been granted the repository privilege `rep:insecureQueryOptions` (see [Privilege Management / Query Execution](../security/privilege/mappingtoprivileges.md#query-execution)).
+
+Enabling this option activates the same Compatibility behavior for NodeIterator.getSize() as described in [Result Size](#result-size), but only for the query being executed.
```
thomasmueller: Typo(?): lowercase "compatibility"

thomasmueller (approved):

thomasmueller:
I'm not blocking this, but the "finally" block would be nice.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/jcr/query/ResultSizeTest.java
nfsantos: For setting and resetting system properties in tests, have a look at the [RestoreSystemProperties](https://stefanbirkner.github.io/system-rules/apidocs/org/junit/contrib/java/lang/system/RestoreSystemProperties.html) class in JUnit. At the end of a test method, it resets any system property that was set in the method to the original value. It is already a dependency in Oak.

adamcin: I used the try-finally approach since ResultSizeTest is still in pre-junit4 style. Also modified the existing test method in the same way.

thomasmueller (approved):

oak-lucene/src/test/java/org/apache/jackrabbit/oak/jcr/query/ResultSizeTest.java
adamcin: @nfsantos are you able to approve?

nfsantos: I did not look at the PR in detail, I just made a comment in case you were not familiar with the RestoreSystemProperties rule (I was not until recently and discovering it simplified my tests). So I don't feel ready to either approve or to make other comments about the PR. It's fine to use finally to clear the system properties. I'd say to go ahead, you already have two approvals.

The only suggestion I have after a very quick look is to avoid repeating the string literal "oak.fastQuerySize". Better to have a constant defined somewhere, to avoid copy and paste errors.

adamcin:
@thomasmueller I don't have commit rights, so if you are willing to merge it, please do.

thomasmueller:
We discussed this in the team today, and came to the conclusion that we also need somebody from the security component to review the change, before we can deploy it. I'm sorry!

anchela (changes requested):
hi @adamcin , i understand what you attempt to do and i agree that it would be nice to have some authorization level control.

But.... unfortunately i fear your change will have undesired side effects as it alters the built-in permissions. i haven't look at this in a while but if i remember correctly we cannot extend the list of built-in privileges.

what is possible is defining a 'custom' privilege and enforce in an application layer outside of the repositories permission evaluation s.str. and after all this seems to be exactly what you want to do: check if a given privilege is granted in the query code (and not in the permission evaluation code). however, we must make sure that this new privilege does not collide with other custom privileges that might have been registered for a given oak repository (there exist e.g. custom privileges in Adobe AEM). 

i am sorry that in the current form the patch is not possible. let's discuss what options we would have.

adamcin:
@anchela That's unfortunate. I guess the issue is with a new PrivilegeBits constant not being able to reflect the value of `next` in an existing privilege tree that is being upgraded, which probably would mean, for example, that upon upgrade, `rep:insecureQueryOptions` would have the same bit position as `crx:replicate`, and there would either be an error on startup, or there would just be a silent failure to treat them as distinct permissions.

https://github.com/apache/jackrabbit-oak/blob/320df1f83d0580b65aa61e95e81494f973fcdf7d/oak-core/src/main/java/org/apache/jackrabbit/oak/security/privilege/PrivilegeDefinitionWriter.java#L55-L59

Since this approach would likely work if it were defined as a custom privilege, such as `crx:insecureQueryOptions`, I can try reverting the changes under `oak-security-spi` and make the privilege name a configuration option in `QueryEngineSettingsMBean`.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/Result.java
```diff
@@ -74,6 +74,15 @@ public interface Result {
      *         count is higher than max, it returns Long.MAX_VALUE.
      */
     long getSize(SizePrecision precision, long max);
+
+    /**
+     * Indicates whether the query specified the INSECURE RESULT SIZE option.
+     *
+     * @return true if the INSECURE RESULT SIZE query option was set
+     */
+    default boolean isQueryOptionInsecureResultSize() {
```
anchela: from a security point of view, i find it questionable to add public API to make something insecure by design.

i wonder if there is any, remote possiblity to introduce the feature without altering public API contract that we cannot revert or adjust later on.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
```diff
@@ -188,4 +188,21 @@ void setIgnoredClassNamesInCallTrace(
...
     @NotNull
     String[] getIgnoredClassNamesInCallTrace();
+
+    /**
+     * Set the name of a JCR privilege that grants a repository permission to use insecure query options.
+     *
+     * @param privilegeName the insecure query options privilege name
+     * @since 1.60
+     */
+    @Description("Set the name of a JCR privilege that grants a repository permission to use insecure query options.")
```
anchela: the description and the javadoc of this method is not entirely accurate. if a custom privilege is registered (and by doing so relying on application to verify that privilege to be present), we can no longer talk about a repository permission. the repository and it's permission evaluation is agnostic to this custom privilege and will not have the ability to enforce it.... so, i would like the phrasing here to accurately reflect on how the 'insecure-query-option' is being mitigated to some extend

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
```diff
@@ -188,4 +188,21 @@ void setIgnoredClassNamesInCallTrace(
...
+    void setInsecureQueryOptionsPrivilegeName(@NotNull String privilegeName);
+
+    /**
+     * Get the configured name of a JCR privilege that grants a repository permission to use insecure query options.
```
anchela: same as above.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/QueryEngineSettingsMBean.java
```diff
@@ -188,4 +188,21 @@ void setIgnoredClassNamesInCallTrace(
...
+     *
+     * @return the insecure query options privilege name
+     * @since 1.60
+     */
+    String getInsecureQueryOptionsPrivilegeName();
```
anchela: that method name and it's associated 'setter' hurt my security heart. 'insecure' + 'privilegeName' really doesn't got together. i understand your intention but i would strongly recommend to reconsider the naming here.

given that this class is public API and not just an implementation detail, i would like you to put reconsider the naming convention used for this feature. once we have it in the API we cannot get rid of it easily.

oak-query-spi/src/main/java/org/apache/jackrabbit/oak/spi/query/QueryLimits.java
```diff
@@ -43,4 +43,14 @@ default String getStrictPathRestriction() {
...
         return new String[] {};
     }
+
+    /**
+     * Retrieve the configured name of a JCR privilege that grants a repository permission to use insecure query
+     * options.
+     * @return the insecure query options privilege name
+     * @since 1.60
+     */
+    default String getInsecureQueryOptionsPrivilegeName() {
```
anchela: ok.... the same option now is present in

- QueryEngineSettings
- QueryOptions
- QueryLimits

that looks wrong to me. i am mentioning this here as this particular configuration option doesn't sound like a 'limit' to me. 

i am not a query expert but adding the same option to 3 different interfaces in API/Spi range does look like a good design. on top it ends up in the Oak.java class.

oak-query-spi/src/main/java/org/apache/jackrabbit/oak/spi/query/QueryLimits.java
```diff
@@ -43,4 +43,14 @@ default String getStrictPathRestriction() {
     default @NotNull String[] getIgnoredClassNamesInCallTrace() {
         return new String[] {};
     }
+
+    /**
+     * Retrieve the configured name of a JCR privilege that grants a repository permission to use insecure query
```
anchela: see comment above about being accurate with respect to custom privilege vs repository enforced permissions.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -95,6 +95,9 @@
 import org.apache.jackrabbit.oak.spi.query.QueryIndex.OrderEntry;
 import org.apache.jackrabbit.oak.spi.query.QueryIndex.OrderEntry.Order;
 import org.apache.jackrabbit.oak.spi.query.QueryIndexProvider;
+import org.apache.jackrabbit.oak.spi.query.QueryLimits;
```
anchela: this import is unused -> remove. spotted when looking at the code in my IDE

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -95,6 +95,9 @@
...
+import org.apache.jackrabbit.oak.spi.security.authorization.permission.PermissionProvider;
```
anchela: this import is unused -> remove. spotted when looking at the code in my IDE

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -95,6 +95,9 @@
...
+import org.apache.jackrabbit.oak.spi.security.authorization.permission.Permissions;
```
anchela: this import is unused -> remove. spotted when looking at the code in my IDE

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -1421,6 +1425,43 @@ public long getSize(SizePrecision precision, long max) {
...
+                .orElse(false);
+    }
+
+    private boolean hasInsecureQueryOptions() {
+        return this.queryOptions != null && (this.queryOptions.insecureResultSize || this.queryOptions.insecureFacets);
+    }
+
+    @Override
+    public void verifyInsecureOptions() {
+        if (hasInsecureQueryOptions() && !isAllowedInsecureOptions()) {
```
anchela: ouch
that is hard to read

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -1421,6 +1425,43 @@ public long getSize(SizePrecision precision, long max) {
...
+            throw new IllegalArgumentException("Insecure query options are not allowed.");
```
anchela: serious note: this exception may actually bubble up unfiltered like the one recently spotted by a customer and reported by @joerghoh in [OAK-10473](https://issues.apache.org/jira/browse/OAK-10473).

again not a query expert but i would rather avoid that.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -1421,6 +1425,43 @@ public long getSize(SizePrecision precision, long max) {
...
     }
 
+    /**
+     * Check whether the execution context has been granted the privilege represented by
+     * {@link org.apache.jackrabbit.oak.api.jmx.QueryEngineSettingsMBean#getInsecureQueryOptionsPrivilegeName()} as a
+     * repository permission.
+     *
+     * @return true if the context has the insecure query options privilege granted as a repository permission
+     */
+    boolean isAllowedInsecureOptions() {
```
anchela: as far as i can see this method is only used inside QueryImpl itself. is there a reason for not marking it as private?

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -1421,6 +1425,43 @@ public long getSize(SizePrecision precision, long max) {
...
+        return Optional.ofNullable(context.getSettings().getInsecureQueryOptionsPrivilegeName())
+                .filter(name -> !name.isEmpty())
+                .flatMap(privilegeName -> Optional.ofNullable(context)
+                        .map(ExecutionContext::getPermissionProvider)
+                        .map(provider -> provider.hasPrivileges(null, privilegeName)))
```
anchela: this means that the privilege can only be granted as a repository-level privilege. a given subject (as associated with every jcr session) either has the privilege granted or not for the whole repository.

in other words: this applies to all queries and there is no way to turn it on/off for particular query or query results. right?

that would be crucial information on the public facing configuration options and osgi configurations.

anchela: i haven't looked at PermissionProvider.hasPrivieges in a while. so my question would be: what happens if an invalid privilege name has been configured? e.g. one that never got registered. will the method return true or false?
i remember from recent work on OAK-9674 that i had to introduce a way to validate privilege names and i suspect it's not validated for internal calls. the main usage for `PermissionProvider.hasPrivileges` is the JCR API calls AccessControlManager.hasPrivileges which takes a privilege object and not a simple string.

what i suspect (but have not verified): the assertion will return true if an unknown privilege name has been configured.

please check that.

anchela (changes requested):
hi @adamcin , sorry for the delay. today i set aside some time to have a closer look at the feature and your changes addressing my initial first concerns regarding built-in privileges.

those have been addressed by moving to custom privileges. i would still want to point out that there is some risk for problems hidden due to the way privileges are registered: new privileges are just added in the order they are registered... while this works for a given, isolated repository i could envision cases where data are imported outside of regular JCR/Jackrabbit API and due to different order come with different meaning for different privileges. maybe this is a theoretical issue but i want you to verify that it would not cause any regressions in your product before this feature gets merged.

the privilege check in QueryImpl:
i left a comment there regarding the case where an invalid privilege name is configured (i.e.g one that is not associated with a registered privilege name) and how this affects the evaluation.

finally, i had a closer look at the query parts of it and the way this insecure-by-design feature is designed. coming from a security background i obviously feel bad about 'insecure-by-design'. but from a pure software design point of view, i am not super happy with the way this feature is exposed in public API, which makes it impossible to adjust or improve security if let's say (just for the sake of it) it would fail a security audit. regarding that aspect, i left a couple of comments in various places trying to highlight what makes me feel uneasy as an engineer. 

as far as the security aspects are concerned: i am going to discuss it with my team mates to make sure we get additional eyes/brains on this. just looking at the code i feel my review is too focused on the pure change in oak and i want to discuss it with someone that looks at this from a distance and is not closely involved in the impl-details of oak.

i am marking this as 'request change' mainly for the things that i requested a change for in the comments below. but my feeling is that we would do us and everyone else a favor to get this thoroughly reviewed.

hope it helps.

_update_ after having pressed 'submit':
i found your description and in particular the section:

> To address the security risk, the application should also provide a configuration of some kind to restrict the ability to effectively leverage this option to permitted system users, which could be implemented as a JCR repository privilege or as an allowlist property in the QueryEngineSettingsService configuration.

do i understand this right, that this insecure feature should never been used by end users but instead it is intended to be limited services that run oak queries?

if that is the case, i would like to understand what where the pros on cons of the 3 options you outlined in your write up. maybe having the option only available to a very limited set of bundles would be an easier choice and doable without exposing public API (see my concerns in the review). as an example: in AEM we have an OSGi-hook in place that prevents access to the NodeStore service except for a limited set of allowed bundles. did you explore such an option?

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/repository/RepositoryImpl.java
```diff
@@ -564,4 +571,30 @@ public synchronized void cancel() {
...
+        }
+
+        @Override
+        public @NotNull SessionQuerySettings getQuerySettings(@NotNull ContentSession session) {
+            return new SessionQuerySettings() {
+                @Override
+                public boolean useDirectResultCount() {
+                    if (System.getProperty("oak.fastQuerySize") != null) {
+                        return Boolean.getBoolean("oak.fastQuerySize");
+                    }
```
thomasmueller: I would access the system property only once... Something like this:
```suggestion
                    String x = System.getProperty("oak.fastQuerySize");
                    if (x != null) {
                        return Boolean.parseBoolean(x);
                    }
```

I know the risk of concurrent change to the system property is tiny... but still.

thomasmueller (changes requested):
See my comment for "access the system property only once"

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/repository/RepositoryImpl.java
adamcin: @thomasmueller Thanks. I made a change to use an Optional expression for the line.

thomasmueller (approved):

anchela (approved):
hi @adamcin , sorry for the super long delay. wow.... the latest approach looks so much better to me. thanks a ton for your willingness to reconsider the first PR and taking the time to find a solution that doesn't require changing public API in multiple places.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/repository/RepositoryImpl.java
thomasmueller: I personally would have preferred the "if" statement (it's shorter) but OK.

anchela:
thanks a lot @adamcin  for all the energy and effort you put into this!

PR #1098, opened 2023-09-04
OAK-10427: Clear cache on NodeDocumentCache close

mreutegg:


rishabhdaim (approved):

rishabhdaim:
We need to close `changeTrackers` as well.

mreutegg:
> We need to close changeTrackers as well.

That list only contains entries when trackers are in use. This is only the case during some DocumentStore operations, which means the list should generally be small or even empty. I'd rather not make changes, unless we know there is a problem.

PR #1105, opened 2023-09-06
OAK-10371: oak-segment-azure/oak-blob-cloud-azure require provided Guava, embed it instead

reschke:


dulceanu (approved):

mbaedke (approved):
Changed the poms to override <plugin/> with an empty value.

reschke:
running ITs locally; will merge on success

PR #1108, opened 2023-09-08
OAK-10433 | Throttle warn logs for multivalued ordered property

nit0906:


oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -362,10 +371,15 @@ private boolean addTypedOrderedFields(D doc,
...
         // Ignore and warn if property multi-valued as not supported
         if (property.getType().isArray()) {
-            log.warn(
-                    "[{}] Ignoring ordered property {} of type {} for path {} as multivalued ordered property not supported",
-                    getIndexName(), pname,
-                    Type.fromTag(property.getType().tag(), true), path);
+            // Log the warning if the counter goes beyond 1000.
+            // We could miss certain paths being logged here since the DocumentMaker is created for each node state for each index.
+            // But ideally a warning with the property in question should suffice.
+            if (!throttleWarnLogs || (throttleWarnLogs && WARN_LOG_COUNTER_MV_ORDERED_PROPERTY.incrementAndGet() > 1000)) {
```
fabriziofortino: should we make the threshold (1000) configurable?

fabriziofortino: this can be simplified
```suggestion
            if (!throttleWarnLogs || WARN_LOG_COUNTER_MV_ORDERED_PROPERTY.incrementAndGet() > 1000) {
```

fabriziofortino (changes requested):
I don't think the logic is correct: it does not print 1 log every 1k, but it just skips the first 1k.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -362,10 +376,15 @@ private boolean addTypedOrderedFields(D doc,
...
         // Ignore and warn if property multi-valued as not supported
         if (property.getType().isArray()) {
-            log.warn(
-                    "[{}] Ignoring ordered property {} of type {} for path {} as multivalued ordered property not supported",
-                    getIndexName(), pname,
-                    Type.fromTag(property.getType().tag(), true), path);
+            // Log the warning for every 1000 (default to 1000 but configurable) occurrences
+            // We could miss certain paths being logged here since the DocumentMaker is created for each node state for each index.
+            // But ideally a warning with the property in question should suffice.
+            if (!throttleWarnLogs || WARN_LOG_COUNTER_MV_ORDERED_PROPERTY.incrementAndGet() % throttleWarnLogThreshold == 0) {
```
fabriziofortino:
```suggestion
            if (!throttleWarnLogs || WARN_LOG_COUNTER_MV_ORDERED_PROPERTY.incrementAndGet() == 1 || WARN_LOG_COUNTER_MV_ORDERED_PROPERTY.get() % throttleWarnLogThreshold == 0) {
```

fabriziofortino (approved):
LGTM. I have only left an improvement proposal to always log the first occurrence.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -79,6 +90,9 @@ public FulltextDocumentMaker(@Nullable FulltextBinaryTextExtractor textExtractor
         this.path = checkNotNull(path);
         this.logWarnStringSizeThreshold = Integer.getInteger(WARN_LOG_STRING_SIZE_THRESHOLD_KEY,
                 DEFAULT_WARN_LOG_STRING_SIZE_THRESHOLD_VALUE);
+        this.throttleWarnLogs = Boolean.getBoolean(THROTTLE_WARN_LOGS_KEY);
+        this.throttleWarnLogThreshold = Integer.getInteger(THROTTLE_WARN_LOGS_THRESHOLD_KEY,
+                DEFAULT_THROTTLE_WARN_LOGS_THRESHOLD_VALUE);
```
nfsantos: If I understood correctly, a new instance of this class is created for every node that is indexed. Adding two more instance fields here will increase memory and CPU usage during indexing, because this constructor is on the critical path of indexing. Aren't these values constant? Can we make them static variables?

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -362,10 +376,15 @@ private boolean addTypedOrderedFields(D doc,
...
-                    Type.fromTag(property.getType().tag(), true), path);
+            // Log the warning for every 1000 (default to 1000 but configurable) occurrences
+            // We could miss certain paths being logged here since the DocumentMaker is created for each node state for each index.
+            // But ideally a warning with the property in question should suffice.
+            if (!throttleWarnLogs || WARN_LOG_COUNTER_MV_ORDERED_PROPERTY.incrementAndGet() == 1 || WARN_LOG_COUNTER_MV_ORDERED_PROPERTY.get() % throttleWarnLogThreshold == 0) {
+                log.warn(
+                        "[{}] Ignoring ordered property {} of type {} for path {} as multivalued ordered property not supported",
+                        getIndexName(), pname,
+                        Type.fromTag(property.getType().tag(), true), path);
+            }
```
nfsantos: Won't this also risk missing some properties? Imagine there are several properties that fall in this warning. We could end up logging only for a few of the properties and not for others. From what I understood, the important information here is what properties are misconfigured.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
nit0906: good point - I made them static variables.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
nit0906: I have made a change to log the first time any new property is misconfigured - post that the usual 1000 (or defined threshold logic) still persists.
there's a small catch that the counter will not increment for the first time a property occurs which mean the threshold would actually be 1001 instead of 1000 - but  I think that's ok.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -60,6 +63,15 @@ public abstract class FulltextDocumentMaker<D> implements DocumentMaker<D> {
...
     private static final int DEFAULT_WARN_LOG_STRING_SIZE_THRESHOLD_VALUE = 102400;
 
+    private static final String THROTTLE_WARN_LOGS_KEY = "oak.repository.property.throttle.warn.logs";
+    private static final String THROTTLE_WARN_LOGS_THRESHOLD_KEY = "oak.repository.throttle.warn.logs.threshold";
+    private static final int DEFAULT_THROTTLE_WARN_LOGS_THRESHOLD_VALUE = 1000;
+
+    // Counter for multi valued ordered property warnings.
+    // Each path with a multi valued ordered property adds to the counter for every valid index that indexes this property.
+    private static final AtomicInteger WARN_LOG_COUNTER_MV_ORDERED_PROPERTY = new AtomicInteger();
+    private static final Set<String> MV_ORDERED_PROPERTY_SET = new HashSet<>();
```
nfsantos: This set can be accessed from multiple threads, so it should be thread-safe. You can use `ConcurrentHashMap.newKeySet()`.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
nfsantos: That should be ok.

nfsantos (approved):

PR #1090, opened 2023-08-30
OAK-8646 : Clean up changes from orphaned branch commits

stefan-egli:

PR #1111, opened 2023-09-11
OAK-8646 : added metrics for unmergedbranch commits & improved junit coverage

rishabhdaim:


stefan-egli (approved):
+1, let's merge this.

A few after the fact comments:
* I would have preferred to get #1090 done before adding metrics on top. Reason is that the way it's done now breaks the git PR review process and makes it virtually impossible to properly address PR review comments therefore - besides making the review itself unnecessarily much harder.
* Different people might have different preferences when it comes to "stream type processing". As of now, some of the review comments in #1090 have been piggybacked into the #1111 directly, without awaiting PR review comments. So from the current state we'd have to revert it first in this PR were we to revisit that aspect. For this reason I believe it's fine, let's merge this. But there might have been a possibility to discuss pros/cons or just revisit that aspect - which I believe with the #1090 / #1111 mixup was made impossible. That would be another reason for preference of doing PRs sequentially, not combined. 

But overall I think let's merge and then revisit the state in the feature branch if needed.

stefan-egli:
> SonarCloud Quality Gate failed.    Quality Gate failed

maybe we can try to fix that...

rishabhdaim:
> > SonarCloud Quality Gate failed.    Quality Gate failed
> 
> maybe we can try to fix that...

Fixed in https://github.com/apache/jackrabbit-oak/pull/1111/commits/88db3b02bf7ec851db057694663e42bfbd4b1b86

PR #1112, opened 2023-09-11
OAK-6759: Convert oak-blob-cloud-azure to OSGi R6 annotations

reschke:


mbaedke (commented):
LGTM, though I don't understand why the ConfigurationPolicy is REQUIRE. It has been before, so I suppose it's fine.

PR #1121, opened 2023-09-14
OAK-10448 Convert o.a.j.api.security.user.Query back to ConsumerType

kwin:


reschke:
You're the expert; I'll trust you. Do we need to have a second look at the other classifications as well?

reschke (approved):

kwin:
@reschke Sure, a second look always helps, although I assumed that this happened in the context of reviewing https://github.com/apache/jackrabbit-oak/pull/948 :-)

PR #1123, opened 2023-09-16
OAK-10429: Implementation for IncrementalFFS

tihom88:


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -244,73 +255,83 @@ public FlatFileStore build() throws IOException, CompositeException {
...
         File result = strategy.createSortedStoreFile();
+        File metadata = strategy.createMetadataFile();
         entryCount = strategy.getEntryCount();
-        return Collections.singletonList(result);
+        return new ImmutablePair<>(Collections.singletonList(result), Collections.singletonList(metadata));
     }
 
     @Nullable
-    private List<File> getFiles(File sortedDir) {
+    private Pair<List<File>, List<File>> getIndexStoreFiles(File sortedDir) {
```
steffenvan: Reading the return type, it is not immediately obvious what the distinction between the two lists is. Can we either use a custom data class like: 
```java
public class IndexStoreFiles {
    private final List<File> storeFiles;
    private final List<File> metadataFiles;

    public IndexStoreFiles(List<File> storeFiles, List<File> metadataFiles) {
        this.storeFiles = storeFiles;
        this.metadataFiles = metadataFiles;
    }

    public List<File> getStoreFiles() {
        return storeFiles;
    }

    public List<File> getMetadataFiles() {
        return metadataFiles;
    }
}
```

or a `Map<String, List<File>>`?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -319,23 +340,26 @@ private List<File> getFiles(File sortedDir) {
         return null;
```
steffenvan, line 340: With regards to my comment, if we choose to go with the custom data class, can we return an `optional`, or `emptyMap()`? I'd prefer us to not return null wherever possible.

steffenvan, line 350: With regards to my comment, if we choose to go with the custom data class, can we return an `optional`, or `emptyMap()`? I'd prefer us to not return null wherever possible.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
```diff
@@ -118,14 +122,53 @@ public static String getPath(String entryLine) {
         return entryLine.substring(0, getDelimiterPosition(entryLine));
     }
 
+    /*
+        Currently we only have 2 types of store:
```
nit0906: A bit more description would be helpful here.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -207,6 +211,30 @@ private List<FlatFileStore> buildFlatFileStoreList(NodeState checkpointedState,
...
     }
 
+    public IncrementalStore buildIncrementalStore(String initialCheckpoint, String finalCheckpoint) throws IOException, CommitFailedException {
+        IncrementalStoreBuilder builder;
+        IncrementalStore incrementalStore;
+        Set<String> preferredPathElements = new HashSet<>();
+        Set<IndexDefinition> indexDefinitions = getIndexDefinitions();
+        for (IndexDefinition indexDf : indexDefinitions) {
+            preferredPathElements.addAll(indexDf.getRelativeNodeNames());
+        }
```
fabriziofortino: you can use java streams and return an immutable set here
```suggestion
        Set<IndexDefinition> indexDefinitions = getIndexDefinitions();

        Set<String> preferredPathElements = indexDefinitions.stream()
            .flatMap(indexDf -> indexDf.getRelativeNodeNames().stream())
            .collect(Collectors.toUnmodifiableSet());
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/indexstore/IndexStoreMetadata.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Set;
+import java.util.function.Predicate;
+
+public class IndexStoreMetadata {
+
+    private String checkpoint;
+    private String storeType;
+    private String strategy;
+    private Set<String> preferredPaths;
+    private Predicate<String> pathPredicate;
```
fabriziofortino: these should be all `final`s

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/indexstore/IndexStoreSortStrategyBase.java
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Directory where sorted files will be created.
+     */
+    protected File storeDir;
+    protected Compression algorithm;
+    protected Predicate<String> pathPredicate;
+    protected Set<String> preferredPaths;
+    protected String checkpoint;
+
+    private final String DEFAULT_INDEX_STORE_TYPE = "FlatFileStore";
```
fabriziofortino: static?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
```diff
@@ -118,14 +122,53 @@ public static String getPath(String entryLine) {
...
+        }
     }
 
     private static int getDelimiterPosition(String entryLine) {
         int indexOfPipe = entryLine.indexOf(NodeStateEntryWriter.DELIMITER);
         checkState(indexOfPipe > 0, "Invalid path entry [%s]", entryLine);
         return indexOfPipe;
     }
+
+    private static List<Integer> getDelimiterPositions(String entryLine) {
```
steffenvan: I think this function could be refactored a bit - I'd like to avoid the `while(true)` loop. IMO, a more functional approach could look like: 

```java
private static List<Integer> getDelimiterPositions(String entryLine) {
    int firstPos = entryLine.indexOf(NodeStateEntryWriter.DELIMITER);
    if (firstPos == -1) {
        throw new IllegalStateException(String.format("Invalid path entry [%s]", entryLine));
    }

    return IntStream.iterate(firstPos, 
                             pos -> pos >= 0, 
                             pos -> entryLine.indexOf(NodeStateEntryWriter.DELIMITER, pos + 1))
                    .boxed()
                    .collect(Collectors.toList());
}
```

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilderTest.java
```diff
@@ -34,6 +34,7 @@
 import org.apache.jackrabbit.oak.index.IndexHelper;
 import org.apache.jackrabbit.oak.index.IndexerSupport;
 import org.apache.jackrabbit.oak.index.indexer.document.CompositeException;
+import org.apache.jackrabbit.oak.index.indexer.document.indexstore.IndexStoreSortStrategy;
```
fabriziofortino:
```suggestion
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedStrategy.java
```diff
@@ -223,19 +221,37 @@ public PipelinedStrategy(MongoDocumentStore documentStore,
...
         this.documentNodeStore = documentNodeStore;
         this.rootRevision = rootRevision;
         this.blobStore = blobStore;
-        this.storeDir = storeDir;
         this.pathComparator = new PathElementComparator(preferredPathElements);
-        this.pathPredicate = pathPredicate;
-        this.algorithm = algorithm;
         this.pathFilters = pathFilters;
 
         Preconditions.checkState(documentStore.isReadOnly(), "Traverser can only be used with readOnly store");
```
fabriziofortino: this should be replaced with:

```java
this(documentStore, documentNodeStore, rootRevision, preferredPathElements, blobStore, storeDir, algorithm, pathPredicate, pathFilters, null);
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -244,73 +255,83 @@ public FlatFileStore build() throws IOException, CompositeException {
...
+                                         Set<IndexDefinition> indexDefinitions) throws IOException, CompositeException {
         logFlags();
         comparator = new PathElementComparator(preferredPathElements);
         entryWriter = new NodeStateEntryWriter(blobStore);
 
-        List<File> fileList = createdSortedStoreFiles();
-        
+        Pair<List<File>, List<File>> pair = createdSortedStoreFiles();
+        List<File> fileList = pair.getKey();
+        File metadataFile = pair.getValue() == null || pair.getValue().size() == 0 ? null : pair.getValue().get(0);
```
fabriziofortino: use `isEmpty()`
```suggestion
        File metadataFile = pair.getValue() == null || pair.getValue().isEmpty() ? null : pair.getValue().get(0);
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -244,73 +255,83 @@ public FlatFileStore build() throws IOException, CompositeException {
...
-            if (files != null && files.length != 0) {
-                return Arrays.asList(files);
+            File[] storeFiles = sortedDir.listFiles(
+                    (dir, name) -> name.endsWith(FlatFileStoreUtils.getSortedStoreFileName(algorithm)));
+            File[] metadataFiles = sortedDir.listFiles(
+                    (dir, name) -> name.endsWith(FlatFileStoreUtils.getMetadataFileName(algorithm)));
+            // Not throwing error for backward compatibility
+            if (metadataFiles == null || metadataFiles.length == 0) {
+                log.error("Unable to find metadata file in files directory:{}", sortedDir.getAbsolutePath());
+            }
```
fabriziofortino: I see an Exception is not thrown for backward compatibility. But in that case, I think line 334 will throw a runtime Exception because `Arrays.asList` accept a not nullable varargs

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -23,9 +23,11 @@
 import org.apache.jackrabbit.guava.common.collect.Lists;
 import org.apache.jackrabbit.oak.commons.Compression;
 import org.apache.jackrabbit.oak.index.indexer.document.CompositeException;
+import org.apache.jackrabbit.oak.index.indexer.document.indexstore.IndexStoreSortStrategyBase;
 import org.apache.jackrabbit.oak.index.indexer.document.LastModifiedRange;
 import org.apache.jackrabbit.oak.index.indexer.document.NodeStateEntryTraverser;
 import org.apache.jackrabbit.oak.index.indexer.document.NodeStateEntryTraverserFactory;
+import org.apache.jackrabbit.oak.plugins.document.DocumentNodeState;
```
fabriziofortino: unused
```suggestion
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
```diff
@@ -118,14 +122,53 @@ public static String getPath(String entryLine) {
...
+        while (true) {
+            indexOfPipe = entryLine.indexOf(NodeStateEntryWriter.DELIMITER, index);
+            if (indexOfPipe > 0) {
+                indexPositions.add(indexOfPipe);
+                index = indexOfPipe + 1;
+            } else {
+                break;
+            }
+        }
+        checkState(indexPositions.size() > 0, "Invalid path entry [%s]", entryLine);
```
fabriziofortino: use `isEmpty()`
```suggestion
        checkState(!indexPositions.isEmpty(), "Invalid path entry [%s]", entryLine);
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
```diff
@@ -20,6 +20,10 @@
 package org.apache.jackrabbit.oak.index.indexer.document.flatfile;
 
 import org.apache.jackrabbit.guava.common.base.Joiner;
+
+import java.util.LinkedList;
+import java.util.List;
+
```
fabriziofortino: these imports should be together with the other `java.util` imports

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
```diff
@@ -118,14 +122,53 @@ public static String getPath(String entryLine) {
...
+    /*
+        Currently we only have 2 types of store:
+        1. FlatFileStore
+        2. IncrementalFlatFileStore
+     */
     public static String[] getParts(String line) {
-        int pos = getDelimiterPosition(line);
-        return new String[]{line.substring(0, pos), line.substring(pos + 1)};
+        // file is FlatFileStore
+        if (line.endsWith("}")) {
```
fabriziofortino: this looks quite hacky. Is there a cleaner way to understand the type? Or even better, should we have different `NodeStateEntryWriter` implementations to avoid this?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/MergeIncrementalFlatFileStore.java
```diff
@@ -0,0 +1,155 @@
...
+    public MergeIncrementalFlatFileStore(Set<String> preferredPathElements, File baseFFS, File incrementalFFS,
+                                         File merged, Compression algorithm) {
+        this.preferredPathElements = preferredPathElements;
+        this.baseFFS = baseFFS;
+        this.incrementalFFS = incrementalFFS;
+        this.merged = merged;
+        this.algorithm = algorithm;
+    }
+
+    public void doMerge() throws IOException {
```
steffenvan, line 76: I think it's there's a bit too significant amount of logic compacted together in this function. I'd suggest breaking it down for better clarity and maintainability. Here are some specific points:

1. **Variable Naming Clarity:** The distinction between `incLine` and `baseLine` to me isn't immediately clear. I think we should adopt more descriptive names or add comments for clarification. 

2. **Comparator Logic:** The logic for comparison: 

	```compared = comparator.compare(new SimpleNodeStateHolder(baseLine), new SimpleNodeStateHolder(incLine));```

	can be encapsulated within its method. From what I can understand the cases are:
	a. compared < 0 - seems to handle the base line.
	b. compared > 0 - seems to manage the incremental line.
	c. Equal lines - there seems to be special handling for matching lines.

	Each of these cases can be refactored into their dedicated methods for clarity.

3. **File Writing Logic:** There's recurring logic around writing to the file. I think we should refactor it to make it more readable and make it easier to adopt when e.g. we will use the TreeFS. Here are some methods I drafted as an example:

```java
private void writeAndAdvanceLine(BufferedWriter writer, BufferedReader reader) throws IOException {
    writer.write(reader.readLine());
    writer.newLine();
}

private void writeAndAdvanceIncremental(BufferedWriter writer, BufferedReader incrementalReader, String incLine) throws IOException {
    writer.write(removeOperand(incLine));
    writer.newLine();
    incrementalReader.readLine();
}

private void validateAndWriteIncrementalLine(BufferedWriter writer, BufferedReader incrementalReader, String incLine) throws IOException {
    String operand = getOperand(incLine);
    if (MODIFY_OPERAND.equals(operand) || DELETE_OPERAND.equals(operand)) {
        log.error("incremental ffs should not have modify and delete operands: {}", incLine);
        throw new RuntimeException("incremental ffs should not have modify and delete operands" + incLine);
    }
    writer.write(removeOperand(incLine));
    writer.newLine();
    incrementalReader.readLine();
}
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/StoreAndSortStrategy.java
```diff
@@ -22,10 +22,12 @@
 import org.apache.commons.io.FileUtils;
 import org.apache.jackrabbit.guava.common.base.Stopwatch;
 import org.apache.jackrabbit.oak.commons.Compression;
+import org.apache.jackrabbit.oak.index.indexer.document.indexstore.IndexStoreSortStrategyBase;
 import org.apache.jackrabbit.oak.index.indexer.document.LastModifiedRange;
 import org.apache.jackrabbit.oak.index.indexer.document.NodeStateEntry;
 import org.apache.jackrabbit.oak.index.indexer.document.NodeStateEntryTraverser;
 import org.apache.jackrabbit.oak.index.indexer.document.NodeStateEntryTraverserFactory;
+import org.apache.jackrabbit.oak.plugins.document.DocumentNodeState;
```
fabriziofortino: unused
```suggestion
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/StoreAndSortStrategy.java
```diff
@@ -42,31 +45,36 @@
...
     private long entryCount;
     private boolean deleteOriginal = Boolean.parseBoolean(System.getProperty(OAK_INDEXER_DELETE_ORIGINAL, "true"));
     private int maxMemory = Integer.getInteger(OAK_INDEXER_MAX_SORT_MEMORY_IN_GB, OAK_INDEXER_MAX_SORT_MEMORY_IN_GB_DEFAULT);
     private long textSize;
-    private Predicate<String> pathPredicate;
 
+    public StoreAndSortStrategy(NodeStateEntryTraverserFactory nodeStatesFactory, Set<String> preferredPaths,
+                                NodeStateEntryWriter entryWriter, File storeDir, Compression algorithm,
+                                Predicate<String> pathPredicate, String checkpoint) {
+        super(storeDir, algorithm, pathPredicate,preferredPaths, checkpoint);
```
fabriziofortino: missing space
```suggestion
        super(storeDir, algorithm, pathPredicate, preferredPaths, checkpoint);
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/StoreAndSortStrategy.java
```diff
@@ -42,31 +45,36 @@
...
+    @Deprecated
     public StoreAndSortStrategy(NodeStateEntryTraverserFactory nodeStatesFactory, PathElementComparator comparator,
                                 NodeStateEntryWriter entryWriter, File storeDir, Compression algorithm, Predicate<String> pathPredicate) {
+        super(storeDir, algorithm, pathPredicate, null, null);
         this.nodeStatesFactory = nodeStatesFactory;
         this.comparator = comparator;
         this.entryWriter = entryWriter;
-        this.storeDir = storeDir;
-        this.algorithm = algorithm;
-        this.pathPredicate = pathPredicate;
```
fabriziofortino, line 69: this should be replaced with

```java
    this(nodeStatesFactory, null, entryWriter, storeDir, algorithm, pathPredicate, null);
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseWithSortStrategy.java
```diff
@@ -23,10 +23,12 @@
 import org.apache.jackrabbit.guava.common.base.Stopwatch;
 import org.apache.jackrabbit.oak.commons.Compression;
 import org.apache.jackrabbit.oak.commons.sort.ExternalSort;
+import org.apache.jackrabbit.oak.index.indexer.document.indexstore.IndexStoreSortStrategyBase;
 import org.apache.jackrabbit.oak.index.indexer.document.LastModifiedRange;
 import org.apache.jackrabbit.oak.index.indexer.document.NodeStateEntry;
 import org.apache.jackrabbit.oak.index.indexer.document.NodeStateEntryTraverser;
 import org.apache.jackrabbit.oak.index.indexer.document.NodeStateEntryTraverserFactory;
+import org.apache.jackrabbit.oak.plugins.document.DocumentNodeState;
```
fabriziofortino: unused
```suggestion
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseWithSortStrategy.java
```diff
@@ -92,17 +93,22 @@ class TraverseWithSortStrategy implements SortStrategy {
     private File sortWorkDir;
     private List<File> sortedFiles = new ArrayList<>();
     private ArrayList<NodeStateHolder> entryBatch = new ArrayList<>();
-    private Predicate<String> pathPredicate;
-
+    TraverseWithSortStrategy(NodeStateEntryTraverserFactory nodeStatesFactory, Set<String> preferredPaths,
```
fabriziofortino, line 94: add a space
```suggestion

    TraverseWithSortStrategy(NodeStateEntryTraverserFactory nodeStatesFactory, Set<String> preferredPaths,
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseWithSortStrategy.java
```diff
@@ -92,17 +93,22 @@ class TraverseWithSortStrategy implements SortStrategy {
...
+    @Deprecated
     TraverseWithSortStrategy(NodeStateEntryTraverserFactory nodeStatesFactory, PathElementComparator pathComparator,
                              NodeStateEntryWriter entryWriter, File storeDir, Compression algorithm, Predicate<String> pathPredicate) {
+        super(storeDir, algorithm, pathPredicate,null, null);
         this.nodeStatesFactory = nodeStatesFactory;
         this.entryWriter = entryWriter;
-        this.storeDir = storeDir;
         this.comparator = (e1, e2) -> pathComparator.compare(e1.getPathElements(), e2.getPathElements());
-        this.pathPredicate = pathPredicate;
-        this.algorithm = algorithm;
```
fabriziofortino, line 105: this should be replaced with
```java
this(nodeStatesFactory, null, entryWriter, storeDir, algorithm, pathPredicate, null);
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/IncrementalFlatFileStoreStrategy.java
```diff
@@ -0,0 +1,181 @@
...
+import static org.apache.jackrabbit.oak.commons.IOUtils.humanReadableByteCount;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileNodeStoreBuilder.OAK_INDEXER_MAX_SORT_MEMORY_IN_GB;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileNodeStoreBuilder.OAK_INDEXER_MAX_SORT_MEMORY_IN_GB_DEFAULT;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileStoreUtils.getMetadataFileName;
+import static org.apache.jackrabbit.oak.index.indexer.document.flatfile.FlatFileStoreUtils.getSortedStoreFileName;
+
+public class IncrementalFlatFileStoreStrategy implements IncrementalIndexStoreSortStrategy {
+
+    private final Logger log = LoggerFactory.getLogger(getClass());
+    private final String STORE_TYPE = "IncrementalFlatFileStore";
```
fabriziofortino: static?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/IncrementalFlatFileStoreStrategy.java
```diff
@@ -0,0 +1,181 @@
...
+    @Override
+    public File createSortedStoreFile() throws IOException {
+        Stopwatch sw = Stopwatch.createStarted();
+        File file = new File(storeDir, getSortedStoreFileName(algorithm));
+        try (BufferedWriter w = FlatFileStoreUtils.createWriter(file, algorithm)) {
+            NodeState before = Objects.requireNonNull(nodeStore.retrieve(beforeCheckpoint));
+            NodeState after = Objects.requireNonNull(nodeStore.retrieve(afterCheckpoint));
+            EditorDiff.process(VisibleEditor.wrap(new IncrementalFlatFileStoreEditor(w, entryWriter, pathPredicate, this)), before, after);
+        }
+        String sizeStr = algorithm == null || algorithm.equals(Compression.NONE) ? "" : String.format("compressed/%s actual size", humanReadableByteCount(textSize));
```
fabriziofortino: I don't think `algorithm` can be null here
```suggestion
        String sizeStr = algorithm.equals(Compression.NONE) ? "" : String.format("compressed/%s actual size", humanReadableByteCount(textSize));```

File: oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/IncrementalFlatFileStoreStrategy.java [hunk 25]
```diff
@@ -0,0 +1,181 @@
...
+        }
+        log.info("Created metadataFile:{} with strategy:{} ", metadataFile.getPath(), indexStoreMetadata.getStoreType());
+        return metadataFile;
+    }
+
+
+    private void writeMetadataToFile(BufferedWriter w, IncrementalIndexStoreMetadata indexStoreMetadata) throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
+        mapper.writeValue(w, indexStoreMetadata);
+    }
```
fabriziofortino: these methods do pretty much the same thing as the ones in `IndexStoreSortStrategyBase`. I think we could improve this code by introducing generics and avoid to duplicate most of the logic

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/IncrementalIndexStoreMetadata.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class IncrementalIndexStoreMetadata {
+
+    private String beforeCheckpoint;
+    private String afterCheckpoint;
+    private String storeType;
+    private String strategy;
+
+    private List<String> preferredPaths;
+    private Predicate<String> pathPredicate;
```
fabriziofortino: finals

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/IncrementalStoreBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Predicate;
+
+import static java.util.Collections.unmodifiableSet;
+
+public class IncrementalStoreBuilder {
+    private final Logger log = LoggerFactory.getLogger(getClass());
+
+    private final String INCREMENTAL_STORE_DIR_NAME_PREFIX = "inc-store";
```
fabriziofortino: static?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/IncrementalStoreBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+        return this;
+    }
+
+
+    public IncrementalStore build() throws IOException, CompositeException {
+        logFlags();
+        File dir = createStoreDir();
+
+        switch (sortStrategyType) {
+            case INCREMENTAL_FFS_STORE:
```
fabriziofortino: this switch has only one case, I would replace it with a if for better readability
```suggestion
if (Objects.requireNonNull(sortStrategyType) == IncrementalSortStrategyType.INCREMENTAL_FFS_STORE) {
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/MergeIncrementalFlatFileStore.java
```diff
@@ -0,0 +1,155 @@
...
+
+    private static String getOperand(String incLine) {
+        String[] parts = NodeStateEntryWriter.getParts(incLine);
+        String operand = parts[2];
+        return operand;
+    }
+
+    private String removeOperand(String line) {
+        String[] parts = NodeStateEntryWriter.getParts(line);
+        return new StringBuilder().append(parts[0]).append("|").append(parts[1]).toString();
```
fabriziofortino: I would avoid to use `StringBuilder` here
```suggestion
        return parts[0] + "|" + parts[1];
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/MergeIncrementalFlatFileStore.java
```diff
@@ -0,0 +1,155 @@
...
+                    throw new RuntimeException(e);
+                }
+            });
+        }
+    }
+
+    private static String getOperand(String incLine) {
+        String[] parts = NodeStateEntryWriter.getParts(incLine);
+        String operand = parts[2];
+        return operand;
```
fabriziofortino: redundant
```suggestion
        return parts[2];
```

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MergeIncrementalFFSTest.java
```diff
@@ -0,0 +1,103 @@
...
+        merge.doMerge();
+
+        try(BufferedReader br = FlatFileStoreUtils.createReader(merged, algorithm)) {
+            for (String line : expectedList) {
+                String actual = br.readLine();
+                System.out.println(actual);
+                Assert.assertEquals(line, actual);
+
+            }
+            Assert.assertEquals(null, br.readLine());
```
fabriziofortino: use `assertNull()`
```suggestion
            Assert.assertNull(br.readLine());
```

fabriziofortino (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
thomasmueller: I'm also not quite sure... 

If we have one metafile per store file (?) then we could also use List<Pair<File, File>>.

But a separate class might be even better.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
thomasmueller: (For me, returning null is fine... My experience is that changing existing code is not worth the trouble.)

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
```diff
@@ -118,14 +122,53 @@ public static String getPath(String entryLine) {
...
+        } else {
+            // file is IncrementalFlatFileStore
+            List<Integer> positions = getDelimiterPositions(line);
+            checkState(positions.size() >= 2, "Invalid path entry [%s]", line);
+            // there are 4 parts in incrementalFFS and default delimiter is |
+            // path|nodeData|checkpoint|operand
+            // Node's data can itself have many | so we split based on first and last 2 |
+            String[] parts = new String[4];
+            parts[0] = line.substring(0, positions.get(0));
+            parts[1] = line.substring(positions.get(0) + 1, positions.get(positions.size() - 2));
```
thomasmueller, line 144: Why not use the simpler

    line.substring(positions.get(0) + 1, positions.get(1));

? I think it should be equivalent.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
```diff
@@ -118,14 +122,53 @@ public static String getPath(String entryLine) {
...
+            parts[2] = line.substring(positions.get(positions.size() - 2) + 1, positions.get(positions.size() - 1));
+            parts[3] = line.substring(positions.get(positions.size() - 1) + 1);
```
thomasmueller, line 146: same here

    parts[3] = line.substring(positions.get(2) + 1);


(I think)

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
thomasmueller: I don't actually like the functional approach here. For me, the loop is easier to debug.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
```diff
@@ -118,14 +122,53 @@ public static String getPath(String entryLine) {
...
         return indexOfPipe;
     }
+
+    private static List<Integer> getDelimiterPositions(String entryLine) {
+        List<Integer> indexPositions = new LinkedList<>();
+        int index = 0;
+        int indexOfPipe;
+        while (true) {
+            indexOfPipe = entryLine.indexOf(NodeStateEntryWriter.DELIMITER, index);
+            if (indexOfPipe > 0) {
```
thomasmueller: Actually, it should be ">= 0"

thomasmueller (changes requested):
(Partial review)

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
```diff
@@ -234,18 +230,42 @@ public int getValue() {
...
+        taskQueue = new LinkedBlockingQueue<>();
+        phaser = new Phaser() {
+            @Override
+            protected boolean onAdvance(int phase, int registeredParties) {
+                //terminate phaser only if it is advancing from last phase and there are no registered parties
+                return phase == Phases.WAITING_FOR_RESULTS.value && registeredParties == 0;
+            }
+        };
+        // arrives on final file sorted
+        mergePhaser = new Phaser(1);
```
thomasmueller, line 253: (It's a bit sad that we have to copy&paste code here... if it can be avoided easily, that would be great. If not, then OK: anyway we want to retire + remove MultithreadedTraverseWithSortStrategy soon.)

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
steffenvan: That's fair. Then we could do something like this: 

```java
private static List<Integer> getDelimiterPositions(String entryLine) {
    List<Integer> indexPositions = new ArrayList<>();

    int index = 0;
    while ((index = entryLine.indexOf(NodeStateEntryWriter.DELIMITER, index)) != -1) {
        indexPositions.add(index);
        index++;
    }

    if (indexPositions.isEmpty()) {
        throw new IllegalStateException(String.format("Invalid path entry [%s]", entryLine));
    }

    return indexPositions;
}
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
thomasmueller: I think the JSON can contain the delimiter (NodeStateEntryWriter.DELIMITER). No?

The path can not. But the JSON, I think it can.

So we would have to get the first delimiter from the left. And the remaining 2 delimiter from the right (from end to beginning).

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
thomasmueller: I think it should be a public method, so we can add a few unit tests.

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
thomasmueller: This would be good, except for:

    // we expect 3 delimiters
    List<Integer> indexPositions = new ArrayList<>(3);

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
```diff
@@ -118,14 +122,53 @@ public static String getPath(String entryLine) {
...
+        if (line.endsWith("}")) {
+            int pos = getDelimiterPosition(line);
+            return new String[]{line.substring(0, pos), line.substring(pos + 1)};
+        } else {
+            // file is IncrementalFlatFileStore
+            List<Integer> positions = getDelimiterPositions(line);
+            checkState(positions.size() >= 2, "Invalid path entry [%s]", line);
+            // there are 4 parts in incrementalFFS and default delimiter is |
+            // path|nodeData|checkpoint|operand
+            // Node's data can itself have many | so we split based on first and last 2 |
```
thomasmueller: I didn't see this comment :-)

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
```diff
@@ -118,14 +122,53 @@ public static String getPath(String entryLine) {
...
-        int pos = getDelimiterPosition(line);
-        return new String[]{line.substring(0, pos), line.substring(pos + 1)};
+        // file is FlatFileStore
+        if (line.endsWith("}")) {
+            int pos = getDelimiterPosition(line);
+            return new String[]{line.substring(0, pos), line.substring(pos + 1)};
+        } else {
+            // file is IncrementalFlatFileStore
+            List<Integer> positions = getDelimiterPositions(line);
+            checkState(positions.size() >= 2, "Invalid path entry [%s]", line);
```
thomasmueller: The following would pass: "/content/dam/abc|checkpoint|operand"

I think it should be:

    checkState(positions.size() >= 3, "Invalid path entry [%s]", line);

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
thomasmueller: I wonder if we really need to have a List<File> metadataFiles. Isn't the metadata file always the same as the storedFile, except for an added suffix?

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/IncrementalStore.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.Iterator;
+
+public interface IncrementalStore extends Closeable {
+
+    Iterator<NodeStateEntry> iterator();
+
+    String getFlatFileStorePath();
```
nit0906: change this to getIncrementalStorePath()

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStore.java
```diff
@@ -38,12 +38,13 @@ public class FlatFileStore implements Iterable<NodeStateEntry>, Closeable {
...
     private final BlobStore blobStore;
     private final File storeFile;
+    private final File metadataFile;
     private final NodeStateEntryReader entryReader;
     private final Set<String> preferredPathElements;
     private final Compression algorithm;
     private long entryCount = -1;
 
-    public FlatFileStore(BlobStore blobStore, File storeFile, NodeStateEntryReader entryReader, Set<String> preferredPathElements, Compression algorithm) {
+    public FlatFileStore(BlobStore blobStore, File storeFile, File metadataFile, NodeStateEntryReader entryReader, Set<String> preferredPathElements, Compression algorithm) {
```
nit0906, line 49: Should we make an interface IndexingStore (like IncrementalStore) and make this class implement that. Later on TreeStore can also implement the same interface.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/TraverseWithSortStrategy.java
tihom88: We can do it but in that case I would have to change access modifiers for few variables.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/IncrementalStoreBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+        throw new IllegalStateException("Not a valid sort strategy value " + sortStrategyType);
+    }
+
+    private File createStoreDir() throws IOException {
+        return Files.createTempDirectory(workDir.toPath(), getDirNamePrefix()).toFile();
+    }
+
+    private String getDirNamePrefix() {
+        return INCREMENTAL_STORE_DIR_NAME_PREFIX + sortStrategyTypeString;
```
nit0906: sortStrategyTypeString - this can be null if system property is not set , we should probably use the string rep of sortStrategyType instead.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/StoreAndSortStrategy.java
tihom88: I didn't wanted to add both comparator and preferredPaths in constructor as comparator can be derived from  preferredPaths.
This will also lead to double initialisation of final comparator, as comparator is also initialised in other constructor.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/IncrementalFlatFileStoreStrategy.java
tihom88: Added @NotNull on constructor parameter and removed this condition.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/IncrementalStoreBuilder.java
tihom88: done

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
tihom88: created an inner class.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
tihom88: created new class.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
tihom88: Now creating an object instead of pair and only returning 1 metadata file.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
tihom88: Created a new class extending NodeStateEntryWriter.java

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/IncrementalFlatFileStoreStrategy.java
tihom88: Lets discuss this.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -207,6 +212,29 @@ private List<FlatFileStore> buildFlatFileStoreList(NodeState checkpointedState,
         return storeList;
     }
 
+    public IndexStore buildIncrementalStore(String initialCheckpoint, String finalCheckpoint) throws IOException, CommitFailedException {
```
amit-jain: Can we just have a `buildStore` which returns IndexStore and taken in `Options` which represent what type if indexing to be done

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
```diff
@@ -118,14 +122,53 @@ public static String getPath(String entryLine) {
         return entryLine.substring(0, getDelimiterPosition(entryLine));
     }
 
+    /*
+        Currently we only have 2 types of store:
+        1. FlatFileStore
+        2. IncrementalFlatFileStore
+     */
     public static String[] getParts(String line) {
```
amit-jain: Can we create a json encompassing the fields  4 fields listed `path|nodeData|checkpoint|operand` converting to [jsonl](https://jsonlines.org/) format?  This will make the code more readable as the whole parsing and writing is a little hacky and will have to change again to parse postions if we need to add more fields.

- It will increase the size of the field and file a little but storage requirements can be managed by compressing it (we should get good compression).
- Parsing we can manage by just reading json at 1 level and not parsing the node data.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/MergeIncrementalFlatFileStore.java
tihom88: Made some changes in code structure let me know if it makes sense now.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
tihom88: We can remove this duplication but to do that access modifiers of some variables need to be changed. Let me know if you feel that is the right approach. We can discuss this.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStore.java
tihom88: IndexStore is now used for both base and incrementalStore. As currently both stores have similar implementations. We can later add IncrementalStore in case implementation changes.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/MultithreadedTraverseWithSortStrategy.java
thomasmueller: I don't have a strong opinion; if you think it is better that way then let's keep it.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/StoreAndSortStrategy.java
```diff
@@ -42,39 +44,43 @@
...
     }
 
     @Override
     public File createSortedStoreFile() throws IOException {
         try (NodeStateEntryTraverser nodeStates = nodeStatesFactory.create(
                 new TraversingRange(new LastModifiedRange(0, Long.MAX_VALUE), null))
         ) {
-            File storeFile = writeToStore(nodeStates, storeDir, getSortedStoreFileName(algorithm));
+            File storeFile = writeToStore(nodeStates, this.getStoreDir(), getSortedStoreFileName(this.getAlgorithm()));
```
thomasmueller: this. shouldn't be needed here

thomasmueller (commented):
I wonder what is the test coverage of the new classes... do you have any numbers?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
tihom88: I think changing method name will make it a little ambious. This method is to create incrementalStore as it takes 2 checkpoints as inputs.
 Though I am a little confused on should we have 2 interfaces IndexStore and IncrementalIndexStore. If we do there won't be any change in both interfaces for current implementation.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
tihom88: Yes, it will be a better fit with structure like json. I think it would make sense to discuss this and do it in a separate ticket.

tihom88:
> I wonder what is the test coverage of the new classes... do you have any numbers?

Currently we have unit test around Merge and an IT IncrementalStoreIT. I will be adding few more test cases mainly around the the NodeStateWrite getParts.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStore.java
```diff
@@ -54,8 +57,18 @@ public FlatFileStore(BlobStore blobStore, File storeFile, NodeStateEntryReader e
...
+
+    public FlatFileStore(BlobStore blobStore, File storeFile, NodeStateEntryReader entryReader, Set<String> preferredPathElements, Compression algorithm) {
+        this(blobStore, storeFile, null, entryReader, preferredPathElements, algorithm);
+    }
+
+    public String getStorePath() {
+        return storeFile.getParentFile().getAbsolutePath();
     }
 
+    @Deprecated
```
nit0906, line 75: for documentation purposes of deprecation - 

`
It is strongly recommended that the reason for deprecating a program element be explained in the documentation, using the @deprecated javadoc tag. The documentation should also suggest and link to a recommended replacement API, if applicable. A replacement API often has subtly different semantics, so such issues should be discussed as well.
It is recommended that a since value be provided with all newly annotated program elements. Note that since cannot be mandatory, as there are many existing annotations that lack this element value.
There is no defined order among annotation elements. As a matter of style, the since element should be placed first.
The @Deprecated annotation should always be present if the @deprecated javadoc tag is present, and vice-versa.`

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
amit-jain: On the surface both need to get checkpoints either between 0 -> c1 or c1 -> c2. So, just having it goverened by options should make it simpler to understand. 
We probably don't need to have 2 different interfaces. But need to identify type (FFS, TreeFS), not sure if have to identify it as iincremental.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/IncrementalFlatFileStoreStrategy.java
tihom88: Added generics to avoid code duplication. Though the generics type is open and there is no condition on what datatype it should be. Though we can add another interface if this doesn't feel right.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
tihom88: This method is now removed added unit test got its parent methos getParts()

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
tihom88: Created overloaded method buildStore

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStore.java
tihom88: done

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/MergeIncrementalFlatFileStore.java
```diff
@@ -0,0 +1,220 @@
...
+        if (baseFFSMetadataFile.exists() && incrementalMetadataFile.exists()) {
+            IndexStoreMetadata indexStoreMetadata = new IndexStoreMetadataOperatorImpl<IndexStoreMetadata>()
+                    .getIndexStoreMetadata(baseFFSMetadataFile, algorithm, new TypeReference<IndexStoreMetadata>() {
+                    });
+            IncrementalIndexStoreMetadata incrementalIndexStoreMetadata = new IndexStoreMetadataOperatorImpl<IncrementalIndexStoreMetadata>()
+                    .getIndexStoreMetadata(incrementalMetadataFile, algorithm, new TypeReference<IncrementalIndexStoreMetadata>() {
+                    });
+            return mergeIndexStores(indexStoreMetadata, incrementalIndexStoreMetadata);
+        } else {
+            throw new RuntimeException("either one or both metadataFiles donot exist at path: " +
```
fabriziofortino: typo
```suggestion
            throw new RuntimeException("either one or both metadataFiles do not exist at path: " +
```

nit0906 (approved):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/MergeIncrementalFlatFileStore.java
```diff
@@ -0,0 +1,220 @@
...
+        }
+    }
+
+    private IndexStoreMetadata getIndexStoreMetadataForMergedFile() throws IOException {
+        File baseFFSMetadataFile = IndexStoreUtils.getMetadataFile(baseFFS, algorithm);
+        File incrementalMetadataFile = IndexStoreUtils.getMetadataFile(incrementalFFS, algorithm);
+
+        if (baseFFSMetadataFile.exists() && incrementalMetadataFile.exists()) {
+            IndexStoreMetadata indexStoreMetadata = new IndexStoreMetadataOperatorImpl<IndexStoreMetadata>()
+                    .getIndexStoreMetadata(baseFFSMetadataFile, algorithm, new TypeReference<IndexStoreMetadata>() {
```
fabriziofortino: the explicit type can be removed
```suggestion
                    .getIndexStoreMetadata(baseFFSMetadataFile, algorithm, new TypeReference<>() {
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/MergeIncrementalFlatFileStore.java
```diff
@@ -0,0 +1,220 @@
...
+                    });
+            IncrementalIndexStoreMetadata incrementalIndexStoreMetadata = new IndexStoreMetadataOperatorImpl<IncrementalIndexStoreMetadata>()
+                    .getIndexStoreMetadata(incrementalMetadataFile, algorithm, new TypeReference<IncrementalIndexStoreMetadata>() {
```
fabriziofortino: the explicit type can be removed
```suggestion
                    .getIndexStoreMetadata(incrementalMetadataFile, algorithm, new TypeReference<>() {
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/MergeIncrementalFlatFileStore.java
```diff
@@ -0,0 +1,220 @@
...
+        try (BufferedWriter writer = IndexStoreUtils.createWriter(IndexStoreUtils.getMetadataFile(merged, algorithm), algorithm)) {
+            ObjectMapper mapper = new ObjectMapper();
+            mapper.writeValue(writer, getIndexStoreMetadataForMergedFile());
+        }
+    }
+
+    /**
+     * We only merge indexStore and incrementalStore if:
+     * 1. IndexStore's checkpoint equals incrementalStore's before checkpoint.
+     * 2. IndexStore's preferredPaths equals incrementalStore's prefferedPaths.
```
fabriziofortino: typo
```suggestion
     * 2. IndexStore's preferredPaths equals incrementalStore's preferredPaths.
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/indexstore/IndexStoreMetadataOperatorImpl.java
```diff
@@ -0,0 +1,70 @@
...
+        return metadataFile;
+    }
+
+    private void writeMetadataToFile(BufferedWriter bufferedWriter, M m) throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
+        mapper.writeValue(bufferedWriter, m);
+    }
+
+    /*
+        deserialization with generic type doesnot happen automatically. So we explicitly added this 3rd argument to
```
fabriziofortino: typo
```suggestion
        deserialization with generic type does not happen automatically. So we explicitly added this 3rd argument to
```

fabriziofortino (approved):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/indexstore/IndexStoreMetadataOperatorImpl.java
```diff
@@ -0,0 +1,70 @@
...
+        provide JavaType info to jackson.
+     */
+    @Override
+    public M getIndexStoreMetadata(File metadataFile, Compression algorithm, TypeReference<M> clazz) throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
```
nfsantos: ObjectMapper is an expensive object to build. Since it is thread-safe, it is preferable to reuse it. There are a few more similar situations in this PR.

PR #1135, opened 2023-09-29
OAK-10463: Retrieve flatFileStore for indexing from baseFlatFileStore

tihom88:


oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/sort/ExternalSort.java
```diff
@@ -460,11 +602,21 @@ public static <T> File sortAndSave(List<T> tmplist,
...
      */
     public static <T> File sortAndSave(List<T> tmplist,
                                        Comparator<T> cmp, Charset cs, File tmpdirectory,
-                                       boolean distinct, Compression algorithm, Function<T, String> typeToString) throws IOException {
-        Collections.sort(tmplist, cmp);
+                                       boolean distinct, Compression algorithm,
+                                       Function<T, String> typeToString,
+                                       @Nullable Predicate<T> filterPredicate
+    ) throws IOException {
+        if ( filterPredicate == null){
```
amit-jain: Isn't this at odds with how the predicate is intended to be used? All other places predicate t -> true is added to retain all elements (Line 423) but here it is removing. 

Should be changed as below (lambda in removeIf is unneccasary) 

`if (filterPredicate != null) {
    tmplist.removeIf(filterPredicate.negate());
}
Collections.sort(tmplist, cmp);`

amit-jain (changes requested):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/sort/ExternalSort.java
```diff
@@ -410,11 +536,19 @@ public static File sortAndSave(List<String> tmplist,
...
+     * @param filterPredicate
+     *            predicate to filter out data which need to be sorted
      */
+    public static File sortAndSave(List<String> tmplist,
+                                       Comparator<String> cmp, Charset cs, File tmpdirectory,
+                                       boolean distinct, boolean usegzip, Predicate<String> filterPredicate) throws IOException {
+        return sortAndSave(tmplist, cmp, cs, tmpdirectory, distinct, usegzip, identity(), filterPredicate);
+    }
+
     public static File sortAndSave(List<String> tmplist,
```
amit-jain, line 547: Not sure why this method uses List<String> and a couple others but all other methods use List<T>. These methods may have been already available. I didn't see any usage of these withink Oak. Maybe mark them deprecated?

tihom88: As this is from external code. So I didn't removed this. But yes there are methods we don't use.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/sort/ExternalSort.java
tihom88: I will add another comment here but adding it here can lead to creation of many empty files.

> 
                       We are setting filterPredicate as null as we already filtered the results based
                       on filterPredicate and there is no need to again filter in sortAndSave method.
                        Another reason to do it here is to avoid creating empty files in sortAndSave
                        method as filtering remove all entries in tmplist.
                     
This is the reason I added this null check. Refer https://github.com/apache/jackrabbit-oak/blob/f3c434e4206db5a35335d809eb6e8f51757f67cd/oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/sort/ExternalSort.java#L436

amit-jain: null check is not a problem but the usage of the filter predicate is ` tmplist.removeIf(t -> filterPredicate.test(t));`. Also a change for simplification and clarity

tihom88: Updated code. Also updated comment above, please have a look at it.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/sort/ExternalSort.java
```diff
@@ -460,11 +603,20 @@ public static <T> File sortAndSave(List<T> tmplist,
      * @param typeToString
      *        function to map string to custom type. User for coverting line to custom type for the
      *        purpose of sorting
+     * @param filterPredicate
```
amit-jain, line 605: Can we clarify this to say that (filter out comes out as removing to me but here we negate the keep predicate)
`predicate to keep relevant data which need to be sorted`

amit-jain (approved):
lgtm! suggested a javadoc improvement

PR #1136, opened 2023-10-02
OAK-10458 - Set LZ4 as the default compression algorithm for the indexing job.

nfsantos:
Additionally, since the configuration of the compression algorithm is read from multiple classes, centralize this logic in IndexStoreUtils.


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -130,7 +131,7 @@ public class FlatFileNodeStoreBuilder {
     private Predicate<String> pathPredicate = path -> true;
 
     private final boolean compressionEnabled = Boolean.parseBoolean(System.getProperty(OAK_INDEXER_USE_ZIP, "true"));
-    private final boolean useLZ4 = Boolean.parseBoolean(System.getProperty(OAK_INDEXER_USE_LZ4, "false"));
+    private final boolean useLZ4 = Boolean.parseBoolean(System.getProperty(OAK_INDEXER_USE_LZ4, DEFAULT_OAK_INDEXER_USE_LZ4));
     private final Compression algorithm = compressionEnabled ? (useLZ4 ? new LZ4Compression() : Compression.GZIP) :
```
nit0906: Should we move the logic to read the system prop for compression to some util class - maybe IndexStoreUtils - assuming this would also be needed for the TreeFS builder later on ?

reschke: Note that we have https://www.javadoc.io/doc/org.apache.jackrabbit/oak-commons/latest/org/apache/jackrabbit/oak/commons/properties/SystemPropertySupplier.html

nfsantos: > Should we move the logic to read the system prop for compression to some util class - maybe IndexStoreUtils - assuming this would also be needed for the TreeFS builder later on ?

I created static methods in IndexStoreUtils to read the compression configuration, the logic is all there.

nfsantos: > Note that we have https://www.javadoc.io/doc/org.apache.jackrabbit/oak-commons/latest/org/apache/jackrabbit/oak/commons/properties/SystemPropertySupplier.html

I was not aware of this class, it's a useful class. I will use it in future but I'd rather not introduce more changes in this PR.

nit0906 (approved):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/indexstore/IndexStoreUtils.java
```diff
@@ -39,6 +40,23 @@
...
+
+    public static boolean compressionEnabled() {
+        return Boolean.parseBoolean(System.getProperty(OAK_INDEXER_USE_ZIP, "true"));
+    }
+
+    public static boolean useLZ4() {
+        return Boolean.parseBoolean(System.getProperty(OAK_INDEXER_USE_LZ4, "true"));
+    }
+
+    public static Compression compressionAlgorithm() {
```
steffenvan, line 54: Small thing, but I think this looks a bit simpler:
```java
public static Compression compressionAlgorithm() {
    if (!compressionEnabled()) {
        return Compression.NONE;
    }
    return useLZ4() ? LZ4_INSTANCE : Compression.GZIP;
}
```

Perhaps we could also add a comment like: `LZ4 takes precedence over GZIP if both are enabled.`

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/indexstore/IndexStoreUtils.java
```diff
@@ -39,6 +40,23 @@
 public class IndexStoreUtils {
     public static final String METADATA_SUFFIX = ".metadata";
 
+    public static final String OAK_INDEXER_USE_ZIP = "oak.indexer.useZip";
+    public static final String OAK_INDEXER_USE_LZ4 = "oak.indexer.useLZ4";
+
+    public static boolean compressionEnabled() {
```
steffenvan, line 46: I think this is slightly ambiguous. It seems like USE_ZIP has two meanings: whether to enable compression **and** use GZIP to compress? But LZ4 only indicates whether we use LZ4. So if we disable ZIP but enable LZ4, it wouldn't be used. On the other hand, if we enable ZIP, we will use compression *and* GZIP.

steffenvan, line 46: I think this is slightly ambiguous. It seems like USE_ZIP has two meanings: whether to enable compression **and** use GZIP to compress? But LZ4 only indicates whether we use LZ4. So if we disable ZIP but enable LZ4, it wouldn't be used. On the other hand, if we enable ZIP, we will use compression *and* GZIP.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/indexstore/IndexStoreUtils.java
nfsantos: Done.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/indexstore/IndexStoreUtils.java
nfsantos: Yes, I agree, it's not very clear. I would prefer to have a single property, like: `oak.indexer.flatfilestore_compression = [none | gzip | lz4]`. But that is a bigger change, as it would change the public API of this class, breaking any clients that are relying on the current variable names. Probably it's worth to do it to reduce technical debt sometime in the future, but we should not piggy back it in an unrelated PR.

dalbani: > But that is a bigger change, as we would have to update helm charts and bootstrap rules.

Hum, interesting, it's the first that see a _Helm chart_ being mentioned in relation with Oak.  
I'm curious: is there a such a chart which is publicly available?

nfsantos: Not that I'm aware. My comment should not have mentioned helm charts, which is something we use internally, it should instead have been about breaking public APIs. I rephrased the comment. Thanks for the remark.

PR #1139, opened 2023-10-06
OAK-10472 - Move ApproximateCounter from oak-core to a place where it…

dulceanu:
… can be used by oak-segment-tar

### DIFF
--- a/oak-core/pom.xml
+++ b/oak-core/pom.xml
@@ -55,7 +55,6 @@
               org.apache.jackrabbit.oak.plugins.identifier,
               org.apache.jackrabbit.oak.plugins.index,
               org.apache.jackrabbit.oak.plugins.index.aggregate,
-              org.apache.jackrabbit.oak.plugins.index.counter,
               org.apache.jackrabbit.oak.plugins.index.cursor,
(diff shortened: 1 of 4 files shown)


reschke (changes requested):
That split o.a.j.oak.plugins.index between two projects; technically this may not be a problem, but it might create confusion.

Proposal:

1. Short term: just copy ApproximateCounter into oak-segment-tar (under a different package name), so we can release 1.58.0 without public API changes.
2. Mid-term: decide what do do with org.apache.jackrabbit.oak.plugins.index - should it all go into oak-store-spi?

cc @mreutegg and @anchela

dulceanu:
@reschke, I made the changes proposed. Can you please take another look?

reschke:
Thanks a lot. I will update the issue description; open a new ticket regarding future refactoring and add a note to the copied class pointing to that new issue.

Thanks for the qucik feedback!

reschke:
Overtaken by events; see https://github.com/apache/jackrabbit-oak/commit/c747535b9274403f4d56a15ebf25aadd109700d0 - thanks @dulceanu for the quick feedback.

PR #1141, opened 2023-10-09
OAK-10462: o.a.j.o.plugins.version.VersionEditor#propertyAdded() may …

mbaedke:
…mistakenly assume an ongoing restore operation.

Improved heuristics to identify the sentinel node of a restore operation; added unit test.


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionEditor.java
```diff
@@ -108,9 +112,15 @@ public void propertyAdded(PropertyState after)
                 && this.after.hasProperty(JcrConstants.JCR_VERSIONHISTORY)
                 && !this.after.hasProperty(JCR_ISCHECKEDOUT)
                 && !this.before.exists()) {
-            // sentinel node for restore
-            vMgr.restore(node, after.getValue(Type.REFERENCE), null);
-            return;
+            Tree tree = new TreeProviderService().createReadOnlyTree(this.node.getNodeState());
+            if (vMgr.getNodeTypeManager().isNodeType(
+                    TreeUtil.getPrimaryTypeName(tree), TreeUtil.getMixinTypeNames(tree), MIX_VERSIONABLE)) {
```
rishabhdaim, line 117: Shouldn't this be replaced by `isVersionable()`. 
Both are using the same underlying apis and `isVersionable()` caches the results in `Boolean` as well.

mbaedke: @rishabhdaim No, that doesn't work, because isVersionable() also checks for the presence of the jcr:isCheckedOut property (indirectly, it uses another version of the NodeTypeManager.isNodeType() method, which uses this shortcut when checking for the presence of mix:versionable. That's ok for "real" nodes, because the property is autocreated and mandatory, but does not work for the sentinel node of the restore operation, which is just used internally and breaks this contract). So if you use isVersionable() here, you will break restore().

rishabhdaim (approved):

reschke (approved):

PR #1142, opened 2023-10-09
OAK-10475 - Expose the Mongo client connection in MongoDocumentNodeStoreBuilderBase

nfsantos:
This is a follow up of OAK-10453, which changed the indexing job to use custom codecs when downloading from Mongo. For this, the indexing logic needs access to the Mongo Connection to register the codec. If a library client uses MongoDocumentNodeStoreBuilderBase to build a MongoDocumentStore instance, then it may also need access to the Mongo connection to pass it to the indexing logic.

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentNodeStoreBuilderBase.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentNodeStoreBuilderBase.java
@@ -47,6 +47,7 @@ public abstract class MongoDocumentNodeStoreBuilderBase<T extends MongoDocumentN
     private String uri;
     private String name;
     private String collectionCompressionType;
+    private MongoClient mongoClient;
 
     /**
      * Uses the given information to connect to to MongoDB as backend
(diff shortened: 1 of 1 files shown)


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentNodeStoreBuilderBase.java
```diff
@@ -212,6 +213,15 @@ MongoStatus getMongoStatus() {
         return mongoStatus;
     }
 
+    /**
+     * Returns the MongoDB client configured in the {@link #setMongoDB(String, String, int)} method.
+     *
+     * @return the client or null if the {@link #setMongoDB(String, String, int)} method hasn't been called.
+     */
+    public MongoClient getMongoClient() {
+        return mongoClient;
```
thomasmueller, line 222: This would make the mongoClient field obsolete:
```suggestion
        return mongoDBConnection.getClient();
```

nfsantos: The class is not keeping a reference to `mongoDBConnection` in a field. We could change it to keep `mongoDBConnection` in a field instead of the references to `mongoClient` for `mongoDBConnection.getClient()` and `mongoStatus` for `mongoDBConnection.getStatus()`, but this would create unnecessary coupling between `MongoDocumentNodeStoreBuilderBase` and `MongoDBConnection`. MongoDBConnection contains fields that are not needed by `MongoDocumentNodeStoreBuilderBase`. I would prefer to leave it as is.

thomasmueller (approved):

PR #1144, opened 2023-10-10
OAK-6760: Convert oak-blob-cloud to OSGi R6 annotations

reschke:


dulceanu (approved):

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3DataStoreService.java
```diff
@@ -19,10 +19,23 @@
...
+    @Reference
+    private StatisticsProvider statisticsProvider;
+
     public static final String NAME = "org.apache.jackrabbit.oak.plugins.blob.datastore.S3DataStore";
+
+    protected StatisticsProvider getStatisticsProvider(){
+        return statisticsProvider;
+    }
+
+    protected void setStatisticsProvider(StatisticsProvider statisticsProvider) {
```
kwin, line 38: I would switch to Constructor injection here and still store the field in the common super class. Then you can also make it final. Constructor injection also allows to easily set it to some mock value for tests.

kwin (changes requested):

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3DataStoreService.java
reschke: This change just replicates what we have in oak-blob-cloud-azure. If there's a better way to do it, we should do it in "-azure" first, as this apparently has better test coverage.

PR #1146, opened 2023-10-11
OAK-9459: ConstraintViolationException in VersionManagerImplRestore w…

mbaedke:
…hen target node has a property definition unknown by the frozen node

Fixed by retrieving the property definitions before setting the node types.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionableState.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionableState.java
@@ -295,18 +295,19 @@ class VersionableState {
             restoreFrozenTypeAndUUID(frozen, dest);
             frozenTypeRestored = true;
         }
-        HashMap<PropertyState, Integer> opvs = new HashMap<>();
+        Map<PropertyState, Integer> opvs = new HashMap<>();
         for (PropertyState p : dest.getProperties()) {
             String propName = p.getName();
-            if (BASIC_PROPERTIES.contains(propName)) {
-                continue;
-            }
-            if (frozen.hasProperty(propName)) {
-                continue;
-            }
-            int action = getOPV(dest, p);
-            if (action == COPY || action == VERSION || action == ABORT || action == INITIALIZE || action == COMPUTE) {
-                opvs.put(p, action);
+            if (!BASIC_PROPERTIES.contains(propName) && !frozen.hasProperty(propName)) {
+                int action = getOPV(dest, p);
+                switch (action) {
+                    case COPY:
+                    case VERSION:
+                    case ABORT:
+                    case INITIALIZE:
+                    case COMPUTE:
+                        opvs.put(p, action);
+                }
             }
         }
         if (!frozenTypeRestored) {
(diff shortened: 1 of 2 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionableState.java
```diff
@@ -285,8 +287,32 @@ private void restoreFrozen(@NotNull NodeBuilder frozen,
...
-        restoreFrozenTypeAndUUID(frozen, dest);
+        //OAK-9459: if the NodeState is not empty, retrieve OPVs before restoring types to avoid constraint violations
+        boolean frozenTypeRestored = false;
+        if (!dest.hasProperty(JCR_PRIMARYTYPE)) {
+            // empty NodeState
+            // 15.7.2 Restoring Type and Identifier
+            restoreFrozenTypeAndUUID(frozen, dest);
+            frozenTypeRestored = true;
+        }
+        HashMap<PropertyState, Integer> opvs = new HashMap<>();
```
reschke:
```suggestion
        Map<PropertyState, Integer> opvs = new HashMap<>();
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionableState.java
```diff
@@ -285,8 +287,32 @@ private void restoreFrozen(@NotNull NodeBuilder frozen,
...
+        for (PropertyState p : dest.getProperties()) {
+            String propName = p.getName();
+            if (BASIC_PROPERTIES.contains(propName)) {
+                continue;
```
reschke: Can we change the logic not to use "continue"?

reschke (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionableState.java
mbaedke: It's completely irrelevant here, but as you wish.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionableState.java
mbaedke: Certainly, I did not touch the logic here. What is wrong with continue? This is exactly the kind of "precondition-check" continue that makes the code more readable.

reschke: I find code that uses "continue" usually harder to read. In this case, it seems putting the remainder of the block into an "if" wrapper works as well.

But yes, in the meantime I realized it's old code, just was moved around.

mbaedke: I don't care, so I'll remove the continue and and I think we should also turn the if-else if sequence into a switch.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionableState.java
```diff
@@ -285,43 +287,68 @@ private void restoreFrozen(@NotNull NodeBuilder frozen,
...
         for (PropertyState p : dest.getProperties()) {
             String propName = p.getName();
-            if (BASIC_PROPERTIES.contains(propName)) {
-                continue;
-            }
-            if (frozen.hasProperty(propName)) {
-                continue;
+            if (!BASIC_PROPERTIES.contains(propName) && !frozen.hasProperty(propName)) {
+                int action = getOPV(dest, p);
+                switch (action) {
```
reschke, line 303: can we just change that to 

   action != IGNORE
   
?

Or, actually, wouldn't it be clearer just to store all OPV states here?

reschke (changes requested):
I believe that we can simplify the copying of OPV values, thus making the code change smaller.

reschke (approved):
Works for me, but see https://github.com/apache/jackrabbit-oak/tree/issue/OAK-9459b (https://github.com/apache/jackrabbit-oak/commit/ba022742eacab5552df29c5b20687942c75179c9) for a proposal to make it (IMHO) more readable.

mbaedke:
Replaced by https://github.com/apache/jackrabbit-oak/pull/1183

PR #1147, opened 2023-10-11
OAK-10486 : Resolution of inherited groups may terminate pre-maturely for external users

anchela:
hi @reschke , i would very much appreciate if you had time to review this.

### DIFF
--- a/oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/InheritedMembershipIterator.java
+++ b/oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/InheritedMembershipIterator.java
@@ -54,6 +54,7 @@ class InheritedMembershipIterator extends AbstractLazyIterator<Group> {
             try {
                 // call 'memberof' to cover nested inheritance
                 Iterator<Group> it = gr.memberOf();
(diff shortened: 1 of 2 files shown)


reschke (commented):
So if I understand correctly, the problem was the early return in line 73? That would imply that swapping line 72..74 with 75..77 would fix it, too?

mbaedke:
> So if I understand correctly, the problem was the early return in line 73? That would imply that swapping line 72..74 with 75..77 would fix it, too?

That is also my understanding. The two new testcases agree, too.

reschke (approved):

mbaedke (approved):

anchela:
@reschke , to answer your questions: yes and yes. do you agree that the approach chosen in the PR is equivalent? i find it easier to understand as the decision on what iterator to read from, when to move on to the next and when all iterators have been processed is kept in a single place.

mbaedke:
It is definitely equivalent.

> i find it easier to understand as the decision on what iterator to read from, when to move on to the next and when all iterators have been processed is kept in a single place.

And that is a good point.

reschke:
bq. @reschke , to answer your questions: yes and yes. do you agree that the approach chosen in the PR is equivalent? i find it easier to understand as the decision on what iterator to read from, when to move on to the next and when all iterators have been processed is kept in a single place.

Yes, they look equivalent. I just wanted to double-check that I had understood the problem.

PR #1148, opened 2023-10-12
OAK-10490 Suggest queries return duplicate entries if prefetch is enabled

thomasmueller:


oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndex.java
```diff
@@ -465,6 +465,9 @@ public void remove() {
         @Override
         public IndexRow next() {
             final IndexRow pathRow = pathCursor.next();
+            // we need to copy the reference to the current row here,
+            // otherwise all the returned IndexRows might reference the same row
+            final FulltextResultRow currentRow = currentRowInPathIterator;
             return new IndexRow() {
 
                 @Override
```
nfsantos: To avoid unintentional capture of state from the parent class, you could create a static subclass of IndexRow and declare an explicit constructor. This requires the caller here to explicitly provide all the arguments that are needed and to copy them into the new object of IndexRow, thus avoiding any accidental use of mutable state from the parent class.

thomasmueller: Yes, that would probably be cleaner... My plan was to create the most simple fix, but this seems better.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndex.java
```diff
@@ -465,67 +470,13 @@ public void remove() {
...
-                            LOG.warn(e.getMessage());
-                            LOG.debug(e.getMessage(), e);
-                            throw new RuntimeException(e);
-                        }
-                    }
-                    return pathRow.getValue(columnName);
-                }
+            // we need to copy the reference to the current row here,
+            // otherwise all the returned IndexRows might reference the same row
+            // (and then we use _this_ row in the IndexRow class below)
```
nfsantos: This comment may not be needed anymore. Or else rephrase it because the copy of the reference is now implicit, it's just the standard call-by-value Java behaviour so maybe there is no need for comments.

nfsantos (approved):

fabriziofortino (approved):

PR #1150, opened 2023-10-13
OAK-10311 - Optimize SegmentBlob#equals for segment blobs that origin…

dulceanu:
…ate from the same blob store

### DIFF
--- a/oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/SegmentBlob.java
+++ b/oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/SegmentBlob.java
@@ -199,8 +199,8 @@ public class SegmentBlob extends Record implements Blob {
 
             if (FAST_EQUALS_SAME_BLOBSTORE) {
                 if (blobStore != null && this.blobStore.equals(that.blobStore) && this.isExternal() && that.isExternal()) {
-                    if (this.getContentIdentity() != null && that.getContentIdentity() != null) {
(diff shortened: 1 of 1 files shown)


oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/SegmentBlob.java
```diff
@@ -192,6 +197,14 @@ public boolean equals(Object object) {
                 }
             }
 
+            if (FAST_EQUALS_SAME_BLOBSTORE) {
+                if (blobStore != null && this.blobStore.equals(that.blobStore) && this.isExternal() && that.isExternal()) {
+                    if (this.getContentIdentity() != null && that.getContentIdentity() != null) {
```
smiroslav: Here, I believe we should use SegmentBlob#getBlobId. It returns not null only for external blobs. Probably, in that case ```isExternal``` check is not needed.
Besides that, ```getBlobId``` is not inherited from  ```org.apache.jackrabbit.oak.api.Blob``` like ```getContentIdentity``` is (and we should not use the latter one in a way that contradicts its contract).

dulceanu: @smiroslav, based on our recent discussion I switched as advised to using directly the blob id. However, I kept the checks for figuring out if the blob is external or not for clarity and easier comprehension. Thanks for reviewing!

PR #1152, opened 2023-10-13
OAK-10473 do not throw an IAE on invalid names

joerghoh:


oak-core/src/test/java/org/apache/jackrabbit/oak/security/user/UserImporterMembershipIgnoreTest.java
```diff
@@ -65,7 +65,7 @@ public void before() throws Exception {
         assertTrue(importer.start(memberRefList));
     }
 
-    @Test(expected = IllegalArgumentException.class)
```
anchela, line 68: if the IdentifierManager no longer verifies the validity of an given uuid (as they are used for storing group membership, we might consider doing so in the UserImporter.

reason: there exist 3 different import modes
- ABORT: must fail with ConstraintViolationException if the member uuid cannot be resolved to a valid user
- IGNORE: ignore the uuids that cannot be resolved
- BESTEFFORT: assume the member does just not yet exist and add the value.

that latter is the one that bothers me here: if the validity of the uuid format is no longer verified, we will end up with garbage in the rep:members properties that might later on fail the resolution (i haven't checked). the goal of BESTEFFORT is to not mandate that members exist already.... but obviously storing completely illegal values should be avoided.

i am not sure yet what the best option is.
throwing RepositoryException in the IdentifierManager instead of illegalargumentexception would have the same problem btw as the UserImporter is catching them (and then proceeding as described above).

anchela (changes requested):
hi @joerghoh as discussed i left a comment describing the finding with the userimporter. i am not entirely sure yet what the best solution would be.... will let that simmer a bit.... :)

oak-core/src/test/java/org/apache/jackrabbit/oak/security/user/UserImporterMembershipIgnoreTest.java
joerghoh: I think that the best way would be to subclass the the InvalidArgumentException and throw this specialized exception in this case; and in some codepaths we can catch this specialized exception and handle it differently, while to fallback to the existing handling IAE handling for the others.

PR #1156, opened 2023-10-16
OAK-10460 - PIPELINED strategy: support eager merging of intermediate sorted files

nfsantos:
At the end of the download phase, the intermediate sorted files are merged into a single FFS file. For the same total data to merge, the time for an n-way merge increases with n. As an example, these are the times to merge n files of total size 7GB:

| n | time |
|--------|--------|
| 15 | 00:02:32 |
| 50 | 00:03:06 | 
| 72 | 00:03:20 |
|143 | 00:03:40 | 

Additionally, having no bounds on the number of intermediate files can lead to edge cases with hundreds or even thousands of intermediate files, which can cause OOME or an unacceptable slowdown of the merge phase.

This task adds support for eager merging. During the download phase, the thread responsible for merge sorting the files will evaluate the number of intermediate files and their sizes, and possibility run a merge operation with the smallest files available. Taking the smallest files is the best way of reducing the number of total files and additionally it should completely quickly. 

This is done in a loop by the merge-sort thread. Whenever a new intermediate file is produced, the thread evaluates if the conditions for eager merging are met. If so, it takes the smallest files and merges them. Since this is done on a separate thread, it should not affect the download speed (as long as there are enough CPUs resources available). It is also important to be careful so that at the end of the download, there is no eager merging running, which would delay the final merge phase.

The conditions for eager merging are the following:
- The number of intermediate files must be greater than a given value.
- There must be enough small candidate files for merging. Merging just 2 or 3 files will provide no significant speedups in the final merge phase.
- The sum of the size of the files that are to be merged is not bigger than a certain limit. Eager merging should complete quickly, so that it does not delay the final merge. 

New configuration properties:
- `oak.indexer.pipelined.eagerMergeTriggerThreshold` (64) - Minimum number of intermediate files that must exist before trying to do an eager merge.
- `oak.indexer.pipelined.eagerMergeMaxFilesToMerge` (32) - Maximum number of files to eagerly merge. This is to keep the eager merges efficient, as efficiency decreases with the number of files in a merge
- `oak.indexer.pipelined.eagerMergeMinFilesToMerge` (4) - Minimum number of files to eagerly merge at a time. The minimum number of files to merge can be seen as the reduction factor for the final merge, if every time we merge 4 files, then there can be up to 4x fewer files for the final merge.
- `oak.indexer.pipelined.eagerMergeMaxSizeToMergeMB` (2048) - Maximum size of intermediate files that can be eagerly merged.

## Optimize merge operation
This PR also introduces an optimized merge sorting variant. The existing implementation was loading each line of the intermediate files as a Java String, which requires a conversion from UTF-8 to Java String when reading and another conversion from Java String to UTF-8 when writing. But for comparing the values in the n-way merge, we only use the first part of the line, that represents the path of the node. And in many cases, the lines are orders of magnitude bigger than the path section. This is wasteful, as it is time-consuming and generates many large String objects.

This PR adds a variant of merge sorting that reads the lines as byte arrays, converting only the first section of the line to a String, so it can be used for comparisons. It then writes back the byte array to the final file, without ever converting it to Java Strings.

In tests done locally on my machine, I measured an speed-up of around 20% (a n-way merge went from ~55 to 43 seconds).

### DIFF
--- a/oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/sort/ExternalSortByteArray.java
+++ b/oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/sort/ExternalSortByteArray.java
@@ -1,0 +1,202 @@
+package org.apache.jackrabbit.oak.commons.sort;
+
+import org.apache.jackrabbit.oak.commons.Compression;
+
+import java.io.BufferedOutputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.EOFException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.PriorityQueue;
+import java.util.function.Function;
+
+/**
+ * Variation of ExternalSort that stores the lines read from intermediate files as byte arrays to avoid the conversion
+ * from byte[] to String and then back.
+ */
+public class ExternalSortByteArray {
+    public static <T> int mergeSortedFilesBinary(List<Path> files, BufferedOutputStream fbw, final Comparator<T> cmp,
+                                                 boolean distinct, Compression algorithm,
+                                                 Function<T, byte[]> typeToByteArray, Function<byte[], T> byteArrayToType) throws IOException {
+        ArrayList<BinaryFileBufferBinary<T>> bfbs = new ArrayList<>();
+        try {
+            for (Path f : files) {
+                InputStream in = algorithm.getInputStream(Files.newInputStream(f));
+                BinaryFileBufferBinary<T> bfb = new BinaryFileBufferBinary<>(in, byteArrayToType);
+                bfbs.add(bfb);
+            }
+            return mergeBinary(fbw, cmp, distinct, bfbs, typeToByteArray);
+        } finally {
+            for (BinaryFileBufferBinary<T> buffer : bfbs) {
+                try {
+                    buffer.close();
+                } catch (Exception ignored) {
+                }
+            }
+            for (Path f : files) {
+                Files.deleteIfExists(f);
+            }
+        }
+    }
+
+    public static <T> int mergeBinary(BufferedOutputStream fbw, final Comparator<T> cmp, boolean distinct,
+                                      List<BinaryFileBufferBinary<T>> buffers, Function<T, byte[]> typeToByteArray)
+            throws IOException {
+        PriorityQueue<BinaryFileBufferBinary<T>> pq = new PriorityQueue<>(
+                11,
+                (i, j) -> cmp.compare(i.peek(), j.peek())
+        );
+        for (BinaryFileBufferBinary<T> bfb : buffers) {
+            if (!bfb.empty()) {
+                pq.add(bfb);
+            }
+        }
+        int rowcounter = 0;
+        T lastLine = null;
+        try (fbw) {
+            while (!pq.isEmpty()) {
+                BinaryFileBufferBinary<T> bfb = pq.poll();
+                T r = bfb.pop();
+                // Skip duplicate lines
+                if (!distinct || lastLine == null || cmp.compare(r, lastLine) != 0) {
+                    fbw.write(typeToByteArray.apply(r));
+                    fbw.write('\n');
+                    lastLine = r;
+                }
+                ++rowcounter;
+                if (bfb.empty()) {
+                    bfb.fbr.close();
+                } else {
+                    pq.add(bfb); // add it back
+                }
+            }
+        } finally {
+            for (BinaryFileBufferBinary<T> bfb : buffers) {
+                bfb.close();
+            }
+        }
+        return rowcounter;
+    }
+
+    /**
+     * WARNING: Uses '\n' as a line separator, it will not work with other line separators.
+     */
+    public static class BinaryFileBufferBinary<T> {
+        private final static int BUFFER_SIZE = 64 * 1024;
+        public final InputStream fbr;
+        private final Function<byte[], T> byteArrayToType;
+        private T cache;
+        private boolean empty;
+
+        // Used to reassemble the lines read from the source input stream.
+        private final ByteArrayOutputStream bais = new ByteArrayOutputStream();
+        private final byte[] buffer = new byte[BUFFER_SIZE];
+        private int bufferPos = 0;
+        private int bufferLimit = 0;
+
+        public BinaryFileBufferBinary(InputStream r, Function<byte[], T> byteArrayToType)
+                throws IOException {
+            this.fbr = r;
+            this.byteArrayToType = byteArrayToType;
+            reload();
+        }
+
+        public boolean empty() {
+            return this.empty;
+        }
+
+        private void reload() throws IOException {
+            try {
+                byte[] line = readLine();
+                this.cache = byteArrayToType.apply(line);
+                this.empty = this.cache == null;
+            } catch (EOFException oef) {
+                this.empty = true;
+                this.cache = null;
+            }
+        }
+
+        private boolean bufferIsEmpty() {
+            return bufferPos >= bufferLimit;
+        }
+
+        /*
+         * Read a line from the source input as a byte array. This is adapted from the implementation of
+         * BufferedReader#readLine() but without converting the line to a String.
+         */
+        private byte[] readLine() throws IOException {
+            bais.reset();
+
+            for (; ; ) {
+                if (bufferIsEmpty()) {
+                    bufferLimit = fbr.read(buffer);
+                    bufferPos = 0;
+                }
+                if (bufferIsEmpty()) { /* EOF */
+                    // Buffer is still empty even after trying to read from input stream. We have reached the EOF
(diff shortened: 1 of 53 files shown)


thomasmueller (approved):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/sort/ExternalSortByteArray.java
```diff
@@ -0,0 +1,219 @@
...
+                } catch (Exception ignored) {
+                }
+            }
+            for (Path f : files) {
+                Files.deleteIfExists(f);
+            }
+        }
+    }
+
+    public static <T> int mergeBinary(BufferedOutputStream fbw, final Comparator<T> cmp, boolean distinct,
```
fabriziofortino: should this be private? It's only used by `mergeSortedFilesBinary` and accepts `BinaryFileBuffer` as a parameter which has private visibility.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/sort/ExternalSortByteArray.java
```diff
@@ -0,0 +1,219 @@
...
+                (i, j) -> cmp.compare(i.peek(), j.peek())
+        );
+        for (BinaryFileBuffer<T> bfb : buffers) {
+            if (!bfb.empty()) {
+                pq.add(bfb);
+            }
+        }
+        int rowcounter = 0;
+        T lastLine = null;
+        try (fbw) {
```
fabriziofortino: the BufferedOutputStream `fwb` is closed twice: here and where it gets created (PipelineMergeSortTask.sortStoreFile#line 314). I think we can remove it from here.

```suggestion
        try {
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreUtils.java
```diff
@@ -75,6 +77,11 @@ public static BufferedOutputStream createOutputStream(File file, Compression alg
         return new BufferedOutputStream(algorithm.getOutputStream(out));
     }
 
+    public static BufferedOutputStream createOutputStream(Path file, Compression algorithm) throws IOException {
+        OutputStream out = Files.newOutputStream(file);
+        return new BufferedOutputStream(algorithm.getOutputStream(out));
+    }
+
```
fabriziofortino: can we remove this? It is never used and the class is deprecated.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeStateHolderFactory.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.function.Function;
+
+final class NodeStateHolderFactory implements Function<byte[], NodeStateHolder> {
+    private final static byte PIPE = (byte) '|';
+    private final static byte PATH_SEPARATOR = (byte) '/';
+
+    // In UTF-8, the ASCII characters have similar encoding as in ASCII, so we can search for then without decoding the
```
fabriziofortino: typo
```suggestion
    // In UTF-8, the ASCII characters have similar encoding as in ASCII, so we can search for them without decoding the
```

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedIT.java
```diff
@@ -41,12 +41,15 @@
...
 import org.junit.Assume;
+import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.contrib.java.lang.system.RestoreSystemProperties;
 import org.junit.rules.TemporaryFolder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
fabriziofortino: unused
```suggestion
```

fabriziofortino (approved):
LGTM. Just a few minors.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/sort/ExternalSortByteArray.java
nfsantos: Yes, changed to private.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/sort/ExternalSortByteArray.java
nfsantos: Done.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreUtils.java
nfsantos: Done.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/sort/ExternalSortByteArray.java
```diff
@@ -0,0 +1,212 @@
...
+            }
+        }
+        return rowcounter;
+    }
+
+    /**
+     * WARNING: Uses '\n' as a line separator, it will not work with other line separators.
+     */
+    private static class BinaryFileBuffer<T> {
+        private final static int BUFFER_SIZE = 64 * 1024;
```
thomasmueller: I think this should be configurable with a system property. The reason is, if there are many files (say 10'000), then having them open at the same time can result in out-of-memory.

I wonder if 64 KB is the right size; my experience is that 8 KB doesn't slow down merging significantly.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMergeSortTaskTest.java
```diff
@@ -19,103 +19,134 @@
...
+
+        // Generate FFS
+        List<String> ffs = generateFFS(LINES_IN_FFS);
+        // Shuffle entries to simulate retrieving from MongoDB by an arbitrary order
+        Collections.shuffle(ffs);
+
+        // Generate the expected results by sorting using the node state entries comparator,
+        List<NodeStateHolder> nodesOrdered = sortAsNodeStateEntries(ffs);
+        // Convert back to a list of Strings
+        String[] expectedFFS = nodesOrdered.stream().map(f -> new String(f.getLine())).toArray(String[]::new);
```
thomasmueller: This uses the platforms default charset; should it use FLATFILESTORE_CHARSET, always UTF-8, or a configurable charset?

thomasmueller (commented):
I would not do the optimization with the byte array. This reasons are:
* Adds complexity to the code. Copying code from the JDK sounds wrong. But more importantly, maintenance cost is increased: having to maintain 2 versions.
* Adds implementation details (where does the path end and the payload begin?)
* Problems with charsets (UTF-8 vs default charset).
* Changes the memory usage (larger buffer?).
* In my view, the performance improvement does not compensate for the disadvantages.

I do think using byte arrays instead of Strings does make sense, to avoid the JSON de-serialization. However, having the "hybrid" String + byte array seems wrong to me.

thomasmueller (commented):
I would not do the optimization with the byte array.

thomasmueller (commented):
.

thomasmueller (commented):
general feedback without explicit approval.

thomasmueller (changes requested):
For some reason, I can not just "comment" on the PR.

thomasmueller (commented):
.

thomasmueller (commented):
.

thomasmueller (approved):

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMergeSortTaskTest.java
nfsantos: Done.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/sort/ExternalSortByteArray.java
nfsantos: Made it configurable with a default of 16KB.
In my tests, I saw some improvements from 8 to 16KB. And at 16KB, if we have 1000 files to merge, that's 16MB of memory, I don't having 16 instead of 8KB should be an issue at all, as anyway, we should always be much below 1000 files to merge.

PR #1158, opened 2023-10-17
Oak-10497 | Properties order in FFS can be different across runs

nit0906:


nit0906:
imo, we could actually hold off on this change and simply handle in the client verification script instead.
Changing the behaviour of the public asJson method doesn’t seem correct to me and the other alternative would be make it configurable for all sort strategies (which would probably require another system property).

PR #1159, opened 2023-10-17
OAK-10497 Properties order in FFS can be different across runs: option to sort

thomasmueller:
I think it's easier to sort.
Performance-wise, it doesn't have a big impact, according to a micro-branchmark I wrote (a few percent slower per serialized node). Anyway writing is not on the critical path if the flat file store is generated in a separate job.

If we want to speed up things, we should not use JSON.

### DIFF
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
@@ -86,10 +86,6 @@ public class NodeStateEntryWriter {
     }
 
     public String asJson(NodeState nodeState) {
-        return asJson(StreamSupport.stream(nodeState.getProperties().spliterator(), false));
-    }
-
-    String asSortedJson(NodeState nodeState) {
         return asJson(StreamSupport.stream(nodeState.getProperties().spliterator(), false)
                 .sorted(Comparator.comparing(PropertyState::getName)));
     }
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
@@ -38,7 +38,7 @@ import java.util.stream.StreamSupport;
 import static org.apache.jackrabbit.guava.common.base.Preconditions.checkState;
 
 public class NodeStateEntryWriter {
-    private static final boolean UNSORTED_PROPERTIES = Boolean.getBoolean("oak.NodeStateEntryWriter.unsortedProps");
+    private static final boolean SORTED_PROPERTIES = Boolean.getBoolean("oak.NodeStateEntryWriter.sort");
     private static final String OAK_CHILD_ORDER = ":childOrder";
     public static final String DELIMITER = "|";
     private final JsopBuilder jw = new JsopBuilder();
@@ -87,11 +87,11 @@ public class NodeStateEntryWriter {
     }
 
     public String asJson(NodeState nodeState) {
-        if (UNSORTED_PROPERTIES) {
-            return asJson(StreamSupport.stream(nodeState.getProperties().spliterator(), false));
+        if (SORTED_PROPERTIES) {
+            return asJson(StreamSupport.stream(nodeState.getProperties().spliterator(), false)
+                    .sorted(Comparator.comparing(PropertyState::getName)));
         }
-        return asJson(StreamSupport.stream(nodeState.getProperties().spliterator(), false)
-                .sorted(Comparator.comparing(PropertyState::getName)));
+        return asJson(StreamSupport.stream(nodeState.getProperties().spliterator(), false));
     }
 
     private String asJson(Stream<? extends PropertyState> stream) {
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
@@ -38,6 +38,7 @@ import java.util.stream.StreamSupport;
 import static org.apache.jackrabbit.guava.common.base.Preconditions.checkState;
 
 public class NodeStateEntryWriter {
+    private static final boolean UNSORTED_PROPERTIES = Boolean.getBoolean("oak.NodeStateEntryWriter.unsortedProps");
     private static final String OAK_CHILD_ORDER = ":childOrder";
     public static final String DELIMITER = "|";
     private final JsopBuilder jw = new JsopBuilder();
@@ -86,6 +87,9 @@ public class NodeStateEntryWriter {
     }
 
     public String asJson(NodeState nodeState) {
+        if (UNSORTED_PROPERTIES) {
+            return asJson(StreamSupport.stream(nodeState.getProperties().spliterator(), false));
+        }
         return asJson(StreamSupport.stream(nodeState.getProperties().spliterator(), false)
                 .sorted(Comparator.comparing(PropertyState::getName)));
     }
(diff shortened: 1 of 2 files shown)


thomasmueller:
Micro-benchmark and results:

```
package org.apache.jackrabbit.oak.index.indexer.document.flatfile;

import java.util.ArrayList;

import org.apache.jackrabbit.oak.commons.Profiler;
import org.apache.jackrabbit.oak.plugins.memory.EmptyNodeState;
import org.apache.jackrabbit.oak.plugins.memory.MemoryNodeBuilder;
import org.apache.jackrabbit.oak.spi.blob.BlobStore;
import org.apache.jackrabbit.oak.spi.blob.MemoryBlobStore;
import org.apache.jackrabbit.oak.spi.state.NodeBuilder;
import org.junit.Test;

public class MicroBench {
    @Test
    public void test() {
        BlobStore blobStore = new MemoryBlobStore();
        NodeStateEntryWriter entryWriter = new NodeStateEntryWriter(blobStore);
        ArrayList<NodeState> list = new ArrayList<>();
        for (int j = 0; j < 1000000; j++) {
            NodeBuilder b = new MemoryNodeBuilder(EmptyNodeState.EMPTY_NODE);
            for (int i = 0; i < 5; i++) {
                b.setProperty("p" + i, "Hello, World");
            }
            NodeState ns = b.getNodeState();
            list.add(ns);
        }
        // Profiler prof = new Profiler().startCollecting();
        for(int test = 0; test < 10; test++) {
            long start = System.currentTimeMillis();
            int len = 0;
            for (NodeState ns : list) {
                len += entryWriter.asUnsortedJson(ns).length();
            }
            long time = System.currentTimeMillis() - start;
            System.out.println(time + " ms; string length " + len + " unsorted");
    
            start = System.currentTimeMillis();
            len = 0;
            for (NodeState ns : list) {
                len += entryWriter.asJson(ns).length();
            }
            time = System.currentTimeMillis() - start;
            System.out.println(time + " ms; string length " + len + " sorted");
            System.out.println();
        }
        // System.out.println(prof.getTop(10));
    }

}

1069 ms; string length 101000000 unsorted
610 ms; string length 101000000 sorted

401 ms; string length 101000000 unsorted
425 ms; string length 101000000 sorted

439 ms; string length 101000000 unsorted
612 ms; string length 101000000 sorted

411 ms; string length 101000000 unsorted
514 ms; string length 101000000 sorted

356 ms; string length 101000000 unsorted
428 ms; string length 101000000 sorted

362 ms; string length 101000000 unsorted
371 ms; string length 101000000 sorted

417 ms; string length 101000000 unsorted
380 ms; string length 101000000 sorted

387 ms; string length 101000000 unsorted
381 ms; string length 101000000 sorted

355 ms; string length 101000000 unsorted
542 ms; string length 101000000 sorted

611 ms; string length 101000000 unsorted
621 ms; string length 101000000 sorted
```

It is initially slower due to JVM warmup.
Afterwards, sorting is a bit slower, but not much.
For 100 MB of data, sorting overhead is about 0.1 seconds at most, which is insignificant in my view. (If we want to speed up things, we should not use JSON.)

nfsantos:
I rerun the tests with 20 properties per node and with larger values in the property, which I think is a more realistic scenario. The sorted version is slower, around 15% slower. 

I think the impact in performance will be measurable but maybe not very significant. In the case of the Pipelined strategy, the overhead from sorting will happen in the transform threads, which can be easily scaled up. The Mongo download thread is the main bottleneck, as this stage is single threaded and there is no good way to parallelize it, so I would resist adding overhead in the work done by this thread, but I don't have objections to slightly increasing the work of the transform threads.

I would anyway suggest having a configuration setting to enable/disable sorting of the properties when writing to the FFS.

```java
package org.apache.jackrabbit.oak.index.indexer.document.flatfile;

import java.util.ArrayList;

import org.apache.commons.lang3.RandomStringUtils;
import org.apache.jackrabbit.oak.plugins.memory.EmptyNodeState;
import org.apache.jackrabbit.oak.plugins.memory.MemoryNodeBuilder;
import org.apache.jackrabbit.oak.spi.blob.BlobStore;
import org.apache.jackrabbit.oak.spi.blob.MemoryBlobStore;
import org.apache.jackrabbit.oak.spi.state.NodeBuilder;
import org.junit.Test;

public class MicroBenchmark {
    public void test() {
        BlobStore blobStore = new MemoryBlobStore();
        NodeStateEntryWriter entryWriter = new NodeStateEntryWriter(blobStore);
        ArrayList<NodeState> list = new ArrayList<>();
        for (int j = 0; j < 1000000; j++) {
            NodeBuilder b = new MemoryNodeBuilder(EmptyNodeState.EMPTY_NODE);
            for (int i = 0; i < 20; i++) {
                b.setProperty("p" + i, RandomStringUtils.random(40, true, true));
            }
            NodeState ns = b.getNodeState();
            list.add(ns);
        }
        // Profiler prof = new Profiler().startCollecting();
        for(int test = 0; test < 10; test++) {
            long start = System.currentTimeMillis();
            int len = 0;
            for (NodeState ns : list) {
                len += entryWriter.asJson(ns).length();
            }
            long time = System.currentTimeMillis() - start;
            System.out.println(time + " ms; string length " + len + " unsorted");

            start = System.currentTimeMillis();
            len = 0;
            for (NodeState ns : list) {
                len += entryWriter.asSortedJson(ns).length();
            }
            time = System.currentTimeMillis() - start;
            System.out.println(time + " ms; string length " + len + " sorted");
            System.out.println();
        }
        // System.out.println(prof.getTop(10));
    }

    public static void main(String[] args) {
        new MicroBenchmark().test();
    }
}
```

```
4418 ms; string length 971000000 unsorted
2399 ms; string length 971000000 sorted

1886 ms; string length 971000000 unsorted
2156 ms; string length 971000000 sorted

1760 ms; string length 971000000 unsorted
2039 ms; string length 971000000 sorted

1667 ms; string length 971000000 unsorted
2000 ms; string length 971000000 sorted

1665 ms; string length 971000000 unsorted
2000 ms; string length 971000000 sorted

1665 ms; string length 971000000 unsorted
1999 ms; string length 971000000 sorted

1667 ms; string length 971000000 unsorted
2000 ms; string length 971000000 sorted

1664 ms; string length 971000000 unsorted
2001 ms; string length 971000000 sorted

1663 ms; string length 971000000 unsorted
1999 ms; string length 971000000 sorted

1667 ms; string length 971000000 unsorted
1999 ms; string length 971000000 sorted
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
```diff
@@ -86,10 +86,6 @@ public String toString(List<String> pathElements, String nodeStateAsJson) {
     }
 
     public String asJson(NodeState nodeState) {
```
nfsantos, line 89: I don't think we should change the behavior of this method, as this is breaking the public contract. The API already had a method for returning node with sorted properties and another unsorted, so there is no need to change the behavior of the methods. Instead, it should be the API users that switch to using the method that returns the properties sorted by name.

The Pipelined strategy is using this method and if you change the behavior, this will affect its performance characteristics and may have other undesirable effects.

nfsantos (changes requested):
Please do not change the behavior of the existing public methods. See comment.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
thomasmueller: > this is breaking the public contract

"not sorting" is not the public contract... The public contract is just that this is a JSON.

> this will affect its performance characteristics 

According to the benchmark, it doesn't have a big impact: ~0.1 second for 100 MB

amit-jain (approved):

nfsantos (changes requested):
I don't think we should do this change.

The NodeStateEntryWriter is an utility class, and as such it should not be opinionated. It should offer sorted and unsorted variants, and let clients decide what to use. If we expose a single `asJson` method and make its behavior depend on a global system property, then clients have no choice on what to use. Or else they have to stop using this class and have their own conversion logic where they can control the ordering of the properties. This is unnecessarily restrictive. 

If you want to have a default, then we could have a new method that returns the default as defined by the system property, in addition to methods for sorted and unsorted.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/NodeStateEntryWriter.java
thomasmueller: Now, it doesn't sort by default.

nfsantos (approved):

PR #1165, opened 2023-10-19
OAK-10504- Add log messages with start and end of the indexing job.

nfsantos:
Additional minor changes:
- Remove deprecated constructor.
- Improve logging of size units.

### DIFF
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
@@ -257,6 +257,8 @@ public abstract class DocumentStoreIndexerBase implements Closeable {
     }
 
     public void reindex() throws CommitFailedException, IOException {
+        log.info("[TASK:INDEXING_JOB:START] Starting indexing job");
+        Stopwatch indexJobWatch = Stopwatch.createStarted();
         IndexingProgressReporter progressReporter =
                 new IndexingProgressReporter(IndexUpdateCallback.NOOP, NodeTraversalCallback.NOOP);
         configureEstimators(progressReporter);
@@ -311,6 +313,10 @@ public abstract class DocumentStoreIndexerBase implements Closeable {
                 .build());
 
         indexerSupport.postIndexWork(copyOnWriteStore);
+        log.info("[TASK:INDEXING_JOB:END] Metrics {}", MetricsFormatter.newBuilder()
+                .add("duration", FormattingUtils.formatToSeconds(indexJobWatch))
+                .add("durationSeconds", indexJobWatch.elapsed(TimeUnit.SECONDS))
+                .build());
     }
 
     private void indexParallel(List<FlatFileStore> storeList, CompositeIndexer indexer, IndexingProgressReporter progressReporter)
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
@@ -313,7 +313,7 @@ public abstract class DocumentStoreIndexerBase implements Closeable {
                 .build());
 
         indexerSupport.postIndexWork(copyOnWriteStore);
-        log.info("[TASK:FULL_INDEX_CREATION:START] Metrics {}", MetricsFormatter.newBuilder()
+        log.info("[TASK:FULL_INDEX_CREATION:END] Metrics {}", MetricsFormatter.newBuilder()
                 .add("duration", FormattingUtils.formatToSeconds(indexJobWatch))
                 .add("durationSeconds", indexJobWatch.elapsed(TimeUnit.SECONDS))
                 .build());
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
@@ -257,7 +257,7 @@ public abstract class DocumentStoreIndexerBase implements Closeable {
     }
 
     public void reindex() throws CommitFailedException, IOException {
-        log.info("[TASK:INDEXING_JOB:START] Starting indexing job");
+        log.info("[TASK:FULL_INDEX_CREATION:START] Starting indexing job");
         Stopwatch indexJobWatch = Stopwatch.createStarted();
         IndexingProgressReporter progressReporter =
                 new IndexingProgressReporter(IndexUpdateCallback.NOOP, NodeTraversalCallback.NOOP);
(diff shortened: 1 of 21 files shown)


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -257,6 +257,8 @@ public FlatFileStore buildFlatFileStore() throws IOException, CommitFailedExcept
     }
 
     public void reindex() throws CommitFailedException, IOException {
+        log.info("[TASK:INDEXING_JOB:START] Starting indexing job");
```
chibulcuteanu: I'm a bit confused here. Isn't this reindexing? Why are we calling it INDEXING JOB here? also the variable

nfsantos: I was unsure about the name. Here we want to denote the full time of the job, so including building the FFS, indexing, upload files, catchup. But one of the phases of the job is the actual indexing. So if we use REINDEX for the total duration and INDEXING for duration of indexing phase, this is also confusing. I though that INDEXING_JOB would be a better name to denote the total time, but I'm open to other suggestions.

fabriziofortino: Here is a proposal: the entire process is named `Indexing`. We can rename the current `indexing` phase with `ingestion`. Not sure if this would require too many changes though.

nfsantos: The first idea that comes to my mind for `ingestion` is the upload phase. And so far we have been using indexing to refer to the actual reindex operation, so it would be a change of terminology. But I'm ok with changing the name and introducing a new term, if no one has a better proposal.

The changes to replace the current INDEXING references by INGESTION are small on the Oak side. But users of Oak that are already parsing the logs for the term INGESTION would have to be modified, but that is probably not much work. Since this is a recent feature, it should not be an issue.

chibulcuteanu: I got confused by the method it was in: `reindex`. I think something more generic should be used for the name (for the entire process) and not `INDEXING_JOB`. 
Some proposals:
- `Full Index Creation` - for the whole process
- `Indexing` - for the actual indexing process - we should also probably not use `reindexing` anymore
- `Ingestion` - for the upload part

nit0906:
Please note that there could also be a case where the FFS has been built before hand and then the build FFS part here would simply read the system property and consume the built FFS from it's location - in which this total time could be misleading.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
nfsantos: I'm ok with this proposal @fabriziofortino ? And others?

nfsantos: I have change the name used for the whole process to `FULL_INDEX_CREATION`. I did not change the name of the upload phase because this is derived automatically from existing code:

https://github.com/apache/jackrabbit-oak/blob/abcfb39909afbaf58b436a9ad6e34998e864a833/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/IndexImporter.java#L131

Here is an example of how the metric messages look like currently for the phases after building the FFS:
```
15:42:35.440 [main] INFO  o.a.j.o.i.i.d.DocumentStoreIndexer - [TASK:INDEXING:START] Starting indexing
16:22:26.236 [main] INFO  o.a.j.o.i.i.d.DocumentStoreIndexer - [TASK:INDEXING:END] Metrics: {"duration":"00:39:50","durationSeconds":2390}
16:22:26.236 [main] INFO  o.a.j.o.i.i.d.DocumentStoreIndexer - [TASK:MERGE_NODE_STORE:START] Starting merge node store
16:22:26.237 [main] INFO  o.a.j.o.i.i.d.DocumentStoreIndexer - [TASK:MERGE_NODE_STORE:END] Metrics: {"duration":"00:00:00","durationSeconds":0}
16:23:46.397 [main] INFO  o.a.j.o.p.i.importer.IndexImporter - [TASK:SWITCH_LANE:START]
16:23:46.451 [main] INFO  o.a.j.o.p.i.importer.IndexImporter - [TASK:SWITCH_LANE:END] Metrics: {"duration":"00:00:00","durationSeconds":0}
16:23:46.452 [main] INFO  o.a.j.o.p.i.importer.IndexImporter - [TASK:IMPORT_INDEX_DATA:START]
16:29:48.386 [main] INFO  o.a.j.o.p.i.importer.IndexImporter - [TASK:IMPORT_INDEX_DATA:END] Metrics: {"duration":"00:06:01","durationSeconds":361}
16:29:48.387 [main] INFO  o.a.j.o.p.i.importer.IndexImporter - [TASK:BRING_INDEX_UPTODATE:START]
16:31:23.963 [main] INFO  o.a.j.o.p.i.importer.IndexImporter - [TASK:BRING_INDEX_UPTODATE:END] Metrics: {"duration":"00:01:35","durationSeconds":95}
16:31:23.964 [main] INFO  o.a.j.o.p.i.importer.IndexImporter - [TASK:RELEASE_CHECKPOINT:START]
16:31:23.969 [main] INFO  o.a.j.o.p.i.importer.IndexImporter - [TASK:RELEASE_CHECKPOINT:END] Metrics: {"duration":"00:00:00","durationSeconds":0}
```

chibulcuteanu (approved):

nfsantos:
> Please note that there could also be a case where the FFS has been built before hand and then the build FFS part here would simply read the system property and consume the built FFS from it's location - in which this total time could be misleading.

If the FFS is provided as a system property, `IndexStoreSortStrategy.createSortedStoreFile()` will not be called. All the logging metrics for the Pipelined strategy is inside this method, so these phases will not be printed.

https://github.com/apache/jackrabbit-oak/blob/abcfb39909afbaf58b436a9ad6e34998e864a833/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java#L297-L319

But it's important for users to be aware that the metrics for download might not be printed if the FFS was already present.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -311,6 +313,10 @@ public void reindex() throws CommitFailedException, IOException {
                 .build());
 
         indexerSupport.postIndexWork(copyOnWriteStore);
+        log.info("[TASK:FULL_INDEX_CREATION:START] Metrics {}", MetricsFormatter.newBuilder()
```
tmvlad: Here it should be `[TASK:FULL_INDEX_CREATION:END]`, right?
Besides this, the naming convention is fine by me.

nfsantos: Yes, good catch. Fixed.

PR #1171, opened 2023-10-19
OAK-10511 - Get blobId without network roundtrip

ahanikel:


ahanikel:
@smiroslav `BlobStoreBlob` instances receive their `blobId` in the constructor. That constructor is called in two places: `SegmentNodeStore.getBlob(reference)` where the blobstore is called to get the blobId, and `SegmentBlob.getBlob(blobId)`, which is used by `SegmentBlob.getNewStream()` and `SegmentBlob.getLength()`, and in both cases, the blobId was already written to the segment (because the SegmentBlob is a Record and the blobId read from there). So I think we can safely assume that that blob exists in the blobstore if we're dealing with a `BlobStoreBlob` instance. WDYT?

ahanikel:
@smiroslav @dulceanu Could you have a look when you have time? Thanks!

ahanikel:
@dulceanu I've changed the lazy init as described in https://stackoverflow.com/a/3580658 - thanks for the hint!
I ran the tests like this and they all passed:
```mvn -DskipTests -Dbaseline.skip clean install && (cd oak-segment-tar && mvn integration-test) && (cd oak-it && mvn integration-test)```

dulceanu (approved):

ahanikel:
@dulceanu Thanks for the approval! Could you also merge the PR? I don't have the rights to do that.

PR #1175, opened 2023-10-19
OAK-10497 Properties order in FFS can be different across runs: sort

thomasmueller:


thomasmueller:
Fixes the bug in my previous PR

PR #1180, opened 2023-10-25
OAK-10518 IndexInfo should have a isActive() method

thomasmueller:
This was surprisingly complex.

I moved (well, copied) the IndexName class to oak-core, and deprecated the old class.


fabriziofortino (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexInfoServiceImpl.java
```diff
@@ -62,11 +63,21 @@ public IndexInfoServiceImpl(NodeStore nodeStore, IndexPathService indexPathServi
 
     @Override
     public Iterable<IndexInfo> getAllIndexInfo() {
+        HashSet<String> allIndexes = new HashSet<>();
+        indexPathService.getIndexPaths().forEach(allIndexes::add);
+        HashSet<String> activeIndexes = new HashSet<>();
+        if (indexPathService.getMountInfoProvider().hasNonDefaultMounts()) {
+            IndexName.filterReplacedIndexes(allIndexes, nodeStore.getRoot(), true).forEach(activeIndexes::add);
```
nfsantos:
```suggestion
            activeIndexes.addAll(IndexName.filterReplacedIndexes(allIndexes, nodeStore.getRoot(), true));
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexInfoServiceImpl.java
```diff
@@ -62,11 +63,21 @@ public IndexInfoServiceImpl(NodeStore nodeStore, IndexPathService indexPathServi
...
+            activeIndexes.addAll(allIndexes);
+        }
         return Iterables.filter(Iterables.transform(indexPathService.getIndexPaths(), new Function<String, IndexInfo>() {
             @Override
             public IndexInfo apply(String indexPath) {
                 try {
-                    return getInfo(indexPath);
+                    IndexInfo info = getInfo(indexPath);
+                    info.setActive(activeIndexes.contains(indexPath));
+                    return info;
```
nfsantos: Is it possible for getInfo to return null? the method can return null in some conditions, could it happen in this particular context? If so, guard against that case when calling `setActive`.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/CustomizedIndexTest.java
```diff
@@ -100,6 +112,8 @@ private void compositeLibs2() throws Exception {
                 "/jcr:root//*[@foo] order by @jcr:path",
                 "test-2",
                 "[/content/test, /libs/test2]");
+        assertEquals("[/oak:index/lucene, /oak:index/test-2]",
```
nfsantos, line 122: There is no test case with no active indexes. Although this probably should not happen in practice, it may be good to have that test case. And also a test case with a single active index.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/CustomizedIndexTest.java
```diff
@@ -211,4 +233,25 @@ private void createFolders() throws IOException {
...
+                continue;
+            }
+            if (!info.isActive()) {
+                continue;
+            }
+            list.add(info.getIndexPath());
+        }
+        Collections.sort(list);
+        return list.toString();
+    }
```
nfsantos: I find the name of this method confusing. From the name, without looking at the return type, I expected that it returns a list of the active indexes. But in fact, this converts to String description, so it should have a name like toString().

Why are you doing the test assertions on a string representation of the list of the indexes instead of using Java lists or Java sets? I think using lists or sets is less error prone and more clear, because conceptually that is what we are comparing, not strings.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/IndexName.java
```diff
@@ -33,16 +33,8 @@
...
- *
- * The format of an index node name is:
- * - The name of the index,
- * - optionally a dash ('-') and the product version number,
- * - optionally "-custom-" and the customer version number.
- *
- * If the node name doesn't contain version numbers / dashes, then version 0 is
- * assumed (for both the product version number and customer version number).
+ * @deprecated
+ * Use oak-core org.apache.jackrabbit.oak.plugins.index.IndexName instead.
```
nfsantos, line 37: Is it enough to have `@deprecated` in the comments, without having the `@Deprecated` annotation on the class?

nfsantos (approved):
Just some minor questions.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexInfoServiceImpl.java
thomasmueller: Right (I tried using addAll first but it didn't compile, so I switched to this... but I must have made some mistake, because now it does compile)

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexInfoServiceImpl.java
thomasmueller: I don't think so, but let's be careful...

oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/CustomizedIndexTest.java
thomasmueller: By using string comparison (assertEquals with Strings) It is very easy to fix the test if the returned value doesn't match the expected value.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/CustomizedIndexTest.java
thomasmueller: Sure, I can add such a test. I don't see how this would improve code coverage or so, but it's easy, by just removing the "lucene" index.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/CustomizedIndexTest.java
thomasmueller: Why would it be less error prone?

thomasmueller: Whether lists of strings are compared or string representations of lists of strings... I don't see what difference it would make... But OK, I'll change it.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/IndexName.java
thomasmueller: I think it doesn't matter: the compiler and IDE does see the deprecation. That's probably due to the history of Java: the annotation was added later. But I can add the annotation as well, no issue.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/composite/blueGreen/CustomizedIndexTest.java
nfsantos: If we use strings, we are comparing two things:
- that the index names are the same (the goal of the test)
- that the String format used to convert a list of names to a single string is the same in the expected and actual values

We care only about the index names being the same, the string format is an additional complexity that we are introducing only for the test. For instance, we need to get the commas and the spacing right, while if we compare based on `Set<String>` we avoid these details. 

Why would it be easier to fix the tests with Strings as compared with having something like ` assertEquals(Set.of("/oak:index/lucene", "/oak:index/test-2")`, ...`

Anyway, it's not a big issue.

thomasmueller: With a single String, my IDE allows to copy & paste... I'll try if that works with sets as well.

PR #1182, opened 2023-10-27
OAK-10519 - Export metrics related to reindexing 

nfsantos:
Uses [StatisticsProvider](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-core-spi/src/main/java/org/apache/jackrabbit/oak/stats/StatisticsProvider.java) to collect metrics in the following modules:
- [IndexImporter](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/IndexImporter.java) - Duration of each phase of import: SWITCH_LANE, IMPORT_INDEX_DATA, BRING_INDEX_UPTODATE, RELEASE_CHECKPOINT. Metric name: `oak_indexer_import_${phaseName}_duration_seconds`.
- DocumentStoreIndexer:
`oak_indexer_indexing_duration_seconds`
`oak_indexer_indexing_merge_node_store_duration_seconds`
`oak_indexer_indexing_full_index_creation_duration_seconds`
- Pipelined strategy:
`oak_indexer_pipelined_documentsDownloaded`
`oak_indexer_pipelined_documentsTraversed`
`oak_indexer_pipelined_documentsRejectedSplit`
`oak_indexer_pipelined_documentsAccepted`
`oak_indexer_pipelined_documentsRejected`
`oak_indexer_pipelined_documentsRejectedEmptyNodeState`
`oak_indexer_pipelined_entriesTraversed`
`oak_indexer_pipelined_entriesAccepted`
`oak_indexer_pipelined_entriesRejected`
`oak_indexer_pipelined_entriesRejectedHiddenPaths`
`oak_indexer_pipelined_entriesRejectedPathFiltered`
`oak_indexer_pipelined_extractedEntriesTotalSize`
`oak_indexer_pipelined_mongoDownloadEnqueueDelayPercentage`
`oak_indexer_pipelined_mergeSortIntermediateFilesCount`
`oak_indexer_pipelined_mergeSortEagerMergesRuns`
`oak_indexer_pipelined_mergeSortFinalMergeFilesCount`
`oak_indexer_pipelined_mergeSortFinalMergeTime`

The metrics are exported as counters and set only at the end of the job. This PR is intended only to provide an overview of how the job went after it finishes, not for real time monitoring of the job.


oak-run/src/test/java/org/apache/jackrabbit/oak/index/DocumentStoreIndexerIT.java
```diff
@@ -380,6 +387,95 @@ protected CompositeIndexer prepareIndexers(NodeStore nodeStore, NodeBuilder buil
 
     }
 
+    @Test
+    public void metrics() throws Exception {
```
steffenvan, line 391: From what I can see, this test has a very similar setup as the bundling() test. Could you refactor that and create a setup function? Then it would be easier to know what the setup is and what is being tested.

oak-run/src/test/java/org/apache/jackrabbit/oak/index/DocumentStoreIndexerIT.java
```diff
@@ -380,6 +387,95 @@ protected CompositeIndexer prepareIndexers(NodeStore nodeStore, NodeBuilder buil
...
+                "jcr:content/renditions/original",
+                "jcr:content/renditions/original/jcr:content"
+        );
+        builder.child("test").setChildNode("book.jpg", appNB.getNodeState());
+        store.merge(builder, EmptyHook.INSTANCE, CommitInfo.EMPTY);
+
+        String checkpoint = store.checkpoint(100000);
+
+        //Shut down this store and restart in readOnly mode
+        store.dispose();
```
steffenvan: Why is it necessary to call `dispose()` twice? Perhaps the test could be divided into 2?

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMergeSortTaskTest.java
```diff
@@ -89,27 +94,41 @@ private PipelinedMergeSortTask.Result runTest(Compression algorithm, Path... fil
...
         // +1 for the Sentinel.
         ArrayBlockingQueue<Path> sortedFilesQueue = new ArrayBlockingQueue<>(files.length + 1);
-        PipelinedMergeSortTask mergeSortTask = new PipelinedMergeSortTask(sortRoot, pathComparator, algorithm, sortedFilesQueue);
-        // Enqueue all the files that are to be merged
-        for (Path file : files) {
-            // The intermediate files are deleted after being merged, so we should copy them to the temporary sort root folder
-            Path workDirCopy = Files.copy(file, sortRoot.resolve(file.getFileName()));
-            sortedFilesQueue.put(workDirCopy);
+        ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
+        try (MetricStatisticsProvider metricStatisticsProvider = new MetricStatisticsProvider(null, executor)) {
```
steffenvan: Since `null` is passed into the `MetricStatisticsProvider`, I think it would be nice to know why that is, instead of `ManagementFactory.getPlatformMBeanServer()`?

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMergeSortTaskTest.java
```diff
@@ -89,27 +94,41 @@ private PipelinedMergeSortTask.Result runTest(Compression algorithm, Path... fil
         Path sortRoot = sortFolder.getRoot().toPath();
```
steffenvan, line 110: Would it make sense to have a setup/teardown function for instantiating these objects? Like the root, `ScheduledExecutorService` and `MetricStatisticsProvider`? Then it would clearly signal that it is being instantiated for each test case. And we can ensure to deregister the statistics provider etc after each test case.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/IndexImporter.java
```diff
@@ -465,19 +475,31 @@ interface IndexImporterStepExecutor {
...
                         MetricsFormatter.newBuilder()
                                 .add("duration", FormattingUtils.formatToSeconds(start))
-                                .add("durationSeconds", start.elapsed(TimeUnit.SECONDS))
+                                .add("durationSeconds", durationSeconds)
                                 .build()
                 );
+
+                String name = "oak_indexer_import_" + indexImportPhaseName.toLowerCase() + "_duration_seconds";
+                CounterStats metric = statisticsProvider.getCounterStats(name, StatsOptions.METRICS_ONLY);
+                LOG.debug("Adding metric: {} {}", name, durationSeconds);
```
steffenvan: In `MetricUtils.java` I see the same logic of resetting the counter if it's not 0. Could be refactored to one function.

nfsantos: The MetricsUtils class was not in a package accessible from from IndexImporter. I moved MetricsUtils to a common package so it can be used more widely.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMergeSortTaskTest.java
nfsantos: I moved the executor to the top level, to be created before all tests and shutdown after the tests. There was a potential resource leak because I was not calling the shutdown method. For the `MetricStatisticsProvider`. I prefer to create it close to the code that does the testing. Almost all tests call this `runTest` method and the tests that don't do not need needs this object. So in this case, allocation of this helper object is closer to where they are used, which makes the logic simpler.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMergeSortTaskTest.java
nfsantos: This constructor is called in many tests with null, I don't want to spread the same comment in so many places. The constructor should have a comment explaining that null is accepted and what that means (it does not, but that's a separate issue). And from the comments that should be in the constructor, it would be clear why we are passing null (i.e., the test does not need an MBean server).

oak-run/src/test/java/org/apache/jackrabbit/oak/index/DocumentStoreIndexerIT.java
nfsantos: This variable is used to reference two different objects. And I agree that the test could be improved, but I just copy'n'pasted the existing test and do not want to spend much more time in refactoring it.

oak-run/src/test/java/org/apache/jackrabbit/oak/index/DocumentStoreIndexerIT.java
nfsantos: It's not so simple to share the common code because it is creating several objects that we need during the test. In this case, I think it is not worth the effort to improve the code, some duplication is not a big issue here as this is testing code and the duplication is contained in the same file.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMergeSortTaskTest.java
steffenvan: Is that because you are not interested in testing whether the values of the metrics? To me it's not 100 % clear why the tests don't need an MBean server - in other tests, I see they are there. E.g. I remember it was necessary in the tests for the `NodeCounter` a while ago.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -191,15 +196,19 @@ public Result call() throws Exception {
             } else {
                 downloadWithNaturalOrdering();
             }
-            String enqueueingDelayPercentage = String.format("%1.2f", (100.0 * totalEnqueueWaitTimeMillis) / downloadStartWatch.elapsed(TimeUnit.MILLISECONDS));
+            double enqueueingDelayPercentage = (100.0 * totalEnqueueWaitTimeMillis) / downloadStartWatch.elapsed(TimeUnit.MILLISECONDS);
```
thomasmueller, line 199: Possible division by zero

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/MetricsUtils.java
```diff
@@ -0,0 +1,40 @@
...
+import org.slf4j.LoggerFactory;
+
+public class MetricsUtils {
+    private final static Logger LOG = LoggerFactory.getLogger(MetricsUtils.class);
+
+    public static void setCounter(StatisticsProvider statisticsProvider, String name, long value) {
+        CounterStats metric = statisticsProvider.getCounterStats(name, StatsOptions.METRICS_ONLY);
+        LOG.debug("Adding metric: {} {}", name, value);
+        if (metric.getCount() != 0) {
+            LOG.warn("Counter was not 0: {} {}", name, metric.getCount());
```
thomasmueller, line 48: I understand that metrics are used for logging in this case, so each metric is only set once, and then it is not supposed to be called. But that's not the normal usage of metrics. In my view, it is using metrics for something that metrics are not designed for: to log. The main issue, of course, is that delivering the metrics update may not work. It is fine if metrics are not sent from time to time. I understand you do this on purpose, kind of a "best effort".

But someone else that sees this class doesn't know that. He will think "MetricsUtils.setCounter... nice, I'll use that". And then we fill the log with warnings...

What about adding Javadocs to the class, then renaming the class to "SetOnceMetricsUtils" and the method to "setCounterOnce"? That way, it is more clear about the intended usage, and avoid confusion (and so unnecessarily filling the log file).

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
nfsantos: This is fixed in https://github.com/apache/jackrabbit-oak/pull/1202

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/MetricsUtils.java
nfsantos: I added documentation explaining the intended use of this method and renamed it to `setCounterOnce`. I did not rename the class as that seems excessive. And in the future we may want to have other util methods in this class that do not have the restriction of this method. 

I'm not totally satisfied with this design, maybe this can be revised in a future PR. One option would be to update the metrics directly when the events occur instead of having internal counters. But for the time being, I think this will be enough.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMergeSortTaskTest.java
nfsantos: If we pass null to the constructor, the `MetricStatisticsProvider` will use the `ManagementFactory.getPlatformMBeanServer()`. I think the problem here is lack of documentation on the construtor of `MetricStatisticsProvider`. I don't think this PR should correct the missing documentation.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MetricsUtils.java
```diff
@@ -0,0 +1,41 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.jackrabbit.oak.index.indexer.document.flatfile.pipelined;
+
+interface PipelinedMetrics {
```
fabriziofortino: this is considered by many an antipattern (see Constant interface anti-pattern). In general, I tend to have the constant in the classes where they are used most. Arguments are mainly stylistic, so feel free to find alternatives.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedStrategy.java
```diff
@@ -222,9 +226,10 @@ private static void prettyPrintTransformStatisticsHistograms(TransformStageStati
 
 
     /**
-     * @param pathPredicate Used by the transform stage to test if a node should be kept or discarded.
-     * @param pathFilters   If non-empty, the download stage will use these filters to try to create a query that downloads
-     *                      only the matching MongoDB documents.
+     * @param pathPredicate      Used by the transform stage to test if a node should be kept or discarded.
+     * @param pathFilters        If non-empty, the download stage will use these filters to try to create a query that downloads
+     *                           only the matching MongoDB documents.
+     * @param statisticsProvider
```
fabriziofortino: add a description or remove the `@param`

fabriziofortino (approved):
LGTM

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MetricsUtils.java
nfsantos: I changed it to a class, following the advice here: https://www.baeldung.com/java-constants-good-practices
I prefer to keep all the metrics in a single place instead of spreading them over all the classes in the pipelined package, this makes it easier to get an overview of all the metrics. And in a way, which class publishes which metric is an implementation detail, I prefer to see it as the pipelined strategy publishing metrics.

tihom88 (approved):

PR #1185, opened 2023-11-02
OAK-10526 : test to reproduce split doc GC with referenced revisions

mreutegg:
Suggested changes to test

### DIFF
--- a/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollectorIT.java
+++ b/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollectorIT.java
@@ -299,7 +299,7 @@ public class VersionGarbageCollectorIT {
     public void gcSplitDocsWithReferencedRevisions() throws Exception {
         final String exp;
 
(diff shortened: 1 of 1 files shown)


stefan-egli (commented):
+1 to the t(x) notation, makes it easier to read indeed.

Question regarding the move of the checkpoint creation back 1 week : what is the idea / reasoning of that?

(The bug is triggered in both cases (recent or 1 week old checkpoint) - so we might actually create two flavours of the test as well. My original idea was to illustrate that even using a recent checkpoint triggers the bug. Actually the bug would also be triggered without checkpoint but just using a head revision of root taken before the split/gc cycle.)

mreutegg:
> Question regarding the move of the checkpoint creation back 1 week : what is the idea / reasoning of that?

It is easier for me to understand what the test is doing and as far as I can see is functionally equivalent. Instead of waiting one week and then create a checkpoint that reflects the state a week ago (mentally going forward and backward in time), creating the checkpoint and then wait a week does the same. In both cases there are no changes on the NodeStore for a week and both checkpoints refer to the same state. The time when a checkpoint is taken is not relevant, only the state of the repository at that moment.

stefan-egli (approved):

PR #1186, opened 2023-11-02
OAK-10522: avoid use of SecurityManager in indexing test

reschke:


mbaedke (changes requested):
Great solution. For consistency, I'd replace all the System.exit() calls in IndexCommand.

mbaedke (approved):

PR #1187, opened 2023-11-03
OAK-10527 Improve readability of the explain query output

thomasmueller:
I had a to change a lot of tests, because they depended (too much) on implementation details.

Even thought this would be a good idea, I did not switch to Json output so far, because:
* some upstream UI tools depend on the query plans to have a certain form, which would break,
* it would make the patch even larger,
* the Lucene query string might contain double quotes, which would then be escaped, and so the output hard to read

So basically, first the upstream UI tools would 

The Elasticsearch query is Json, which wouldn't need to be escaped if we add it as a child object.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/nodetype/NodeTypeIndex.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/nodetype/NodeTypeIndex.java
@@ -102,7 +102,10 @@ class NodeTypeIndex implements QueryIndex, JcrConstants {
     
     @Override
     public String getPlan(Filter filter, NodeState root) {
-        return "nodeType " + filter.toString();
+        return "nodeType\n" +
+                "    path: " + filter.getPath() + "\n" +
+                "    primaryTypes: " + filter.getPrimaryTypes().toString() + "\n" +
+                "    mixinTypes: " + filter.getMixinTypes().toString() + "\n";
     }
 
     @Override
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/nodetype/NodeTypeIndex.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/nodetype/NodeTypeIndex.java
@@ -104,8 +104,8 @@ class NodeTypeIndex implements QueryIndex, JcrConstants {
     public String getPlan(Filter filter, NodeState root) {
         return "nodeType\n" +
                 "    path: " + filter.getPath() + "\n" +
-                "    primaryTypes: " + filter.getPrimaryTypes().toString() + "\n" +
-                "    mixinTypes: " + filter.getMixinTypes().toString() + "\n";
+                "    primaryTypes: " + filter.getPrimaryTypes() + "\n" +
+                "    mixinTypes: " + filter.getMixinTypes() + "\n";
     }
 
     @Override
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
@@ -23,15 +23,17 @@ import java.util.Locale;
 /**
  * Formatter for JCR queries in order to make them easier to read. Formatting is
  * done on a best-effort basis.
- * 
- * Warning: formatting is also done within e.g. string literals. So there is no
- * guarantee that the formatted query is semantically equal to the original one!
+ *
+ * Warning: Care was taken to not add newlines inside string literals and so on,
+ * but there is still no guarantee that the formatted query is semantically
+ * equal to the original one. It is recommended to run queries that are returned
+ * by these methods.
  */
 public class QueryFormatter {
 
     /**
      * Detect whether the query is an XPath query.
-     * 
+     *
      * @param query the query
      * @param language the language, if known, or null
      * @return true if xpath
@@ -46,7 +48,7 @@ public class QueryFormatter {
             query = query.substring("explain".length()).trim();
             if (query.startsWith("measure")) {
                 query = query.substring("explain".length()).trim();
-            }            
+            }
         }
         // union queries
         while (query.startsWith("(")) {
@@ -61,7 +63,7 @@ public class QueryFormatter {
     /**
      * Format the query into a more human-readable way, by adding newlines.
      * Warning: newlines are also added inside e.g. string literals.
-     * 
+     *
      * @param query the query (may not be null)
      * @param language the query language, or null if unknown
      * @return the formatted query
@@ -76,28 +78,61 @@ public class QueryFormatter {
     }
 
     private static String formatXPath(String query) {
-        query = query.replaceAll("\\[", "\\[\n  ");
-        for (String term : new String[] {
-                "and ", "or ", "order by ", "option\\("
-        }) {
-            // xpath is case sensitive
-            query = query.replaceAll(" (" + term + ")", "\n  $1");
+        StringBuilder buff = new StringBuilder(query);
+        for (int i = 0; i < buff.length(); i++) {
+            char c = buff.charAt(i);
+            if (c == '\'' || c == '"') {
+                while (++i < buff.length() && buff.charAt(i) != c) {
+                    // skip
+                }
+            } else if (c =='[') {
+                buff.insert(i + 1, "\n  ");
+                i += 3;
+            } else if (c <= ' ') {
+                String sub = buff.substring(i, Math.min(i + 10, buff.length()));
+                if (sub.startsWith(" and ")
+                        || sub.startsWith(" or ")
+                        || sub.startsWith(" order by ")
+                        || sub.startsWith(" option(")) {
+                    buff.setCharAt(i, '\n');
+                    buff.insert(i + 1, "  ");
+                    // just skip over the whitespace - but that's OK
+                    i += 2;
+                }
+            }
         }
-        // remove duplicate newlines
-        query = query.replaceAll("\n+", "\n");
-        return query;
+        return buff.toString();
     }
-    
+
     private static String formatSQL(String query) {
-        for (String term : new String[] {
-                "union ", "from ", "where ", "and ", "or ", "order by ", "option\\("
-        }) {
-            // SQL is case insensitive, so we use (?i)
-            query = query.replaceAll("(?i) (" + term + ")", "\n  $1");
+        StringBuilder buff = new StringBuilder(query);
+        for (int i = 0; i < buff.length(); i++) {
+            char c = buff.charAt(i);
+            if (c == '\'' || c == '"') {
+                while (++i < buff.length() && buff.charAt(i) != c) {
+                    // skip
+                }
+            } else if (c <= ' ') {
+                String sub = buff.substring(i, Math.min(i + 10, buff.length()));
+                if (startsWithIgnoreCase(sub, " and ")
+                        || startsWithIgnoreCase(sub, " or ")
+                        || startsWithIgnoreCase(sub, " union ")
+                        || startsWithIgnoreCase(sub, " from ")
+                        || startsWithIgnoreCase(sub, " where ")
+                        || startsWithIgnoreCase(sub, " order by ")
+                        || startsWithIgnoreCase(sub, " option(")) {
+                    buff.setCharAt(i, '\n');
+                    buff.insert(i + 1, "  ");
+                    // just skip over the whitespace - but that's OK
+                    i += 2;
+                }
+            }
         }
-        // remove duplicate newlines
-        query = query.replaceAll("\n+", "\n");
-        return query;
+        return buff.toString();
     }
-    
+
+    private static boolean startsWithIgnoreCase(String s, String prefix) {
+        return s.regionMatches(true, 0, prefix, 0, prefix.length());
+    }
+
 }
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
@@ -47,7 +47,7 @@ public class QueryFormatter {
         if (query.startsWith("explain")) {
             query = query.substring("explain".length()).trim();
             if (query.startsWith("measure")) {
-                query = query.substring("explain".length()).trim();
+                query = query.substring("measure".length()).trim();
             }
         }
         // union queries
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
@@ -1,0 +1,103 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.jackrabbit.oak.query;
+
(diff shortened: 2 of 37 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
```diff
@@ -0,0 +1,103 @@
...
+        for (String term : new String[] {
+                "union ", "from ", "where ", "and ", "or ", "order by ", "option\\("
+        }) {
+            // SQL is case insensitive, so we use (?i)
+            query = query.replaceAll("(?i) (" + term + ")", "\n  $1");
+        }
+        // remove duplicate newlines
+        query = query.replaceAll("\n+", "\n");
+        return query;
+    }
```
nfsantos: Some of these keywords are common enough that they will likely appear in the text of some queries, which will then result in line breaks being added in the wrong places. Would it be very complicated to change the logic to detect if they words are used as sql keywords or not? Maybe something with checking if the word appears inside quotes or not? Or maybe it's ok for the output to sometimes be wrongly formatted, as this is mostly a visual aid.

nfsantos (commented):
I did not look at the tests, just have a minor comment.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
thomasmueller: Yes this is a concern, if the output is then run as a query... (By the human or automatically).
I was thinking on how to solve it, but I don't think there is a simple solution that would cover all cases. (Basically, don't use regular expressions.) Maybe we could add a comment in the output, but that's ugly as well...

The output is only used in one place: LOG.debug, so I think the risk is not all that great now.
My recommendation is to just accept that risk.
But not use the code anywhere else, that's why I added the warning in the Javadoc.

thomasmueller: Well, actually, I implemented the slightly more complex solution now (including randomized test, which is anyway good to have).

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private static String formatSQL(String query) {
+        StringBuilder buff = new StringBuilder(query);
+        for (int i = 0; i < buff.length(); i++) {
+            char c = buff.charAt(i);
+            if (c == '\'' || c == '"') {
+                while (++i < buff.length() && buff.charAt(i) != c) {
+                    // skip
+                }
```
nfsantos, line 126: Won't this fail if there is some string in the query that contains " inside it? For instance, a string like `"he said \"Hello\""`.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/nodetype/NodeTypeIndex.java
```diff
@@ -102,7 +102,10 @@ public Cursor query(Filter filter, NodeState root) {
     
     @Override
     public String getPlan(Filter filter, NodeState root) {
-        return "nodeType " + filter.toString();
+        return "nodeType\n" +
+                "    path: " + filter.getPath() + "\n" +
+                "    primaryTypes: " + filter.getPrimaryTypes().toString() + "\n" +
+                "    mixinTypes: " + filter.getMixinTypes().toString() + "\n";
```
fabriziofortino: `toString` calls are redundant:

```suggestion
                "    primaryTypes: " + filter.getPrimaryTypes() + "\n" +
                "    mixinTypes: " + filter.getMixinTypes() + "\n";
```

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
```diff
@@ -0,0 +1,150 @@
...
+    public static boolean isXPath(String query, String language) {
+        if (language != null) {
+            return "xpath".equals(language);
+        }
+        query = query.trim().toLowerCase(Locale.ENGLISH);
+        // explain queries
+        if (query.startsWith("explain")) {
+            query = query.substring("explain".length()).trim();
+            if (query.startsWith("measure")) {
+                query = query.substring("explain".length()).trim();
```
fabriziofortino: should it be like this? can we add a test for this case?
```suggestion
                query = query.substring("measure".length()).trim();
```

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
```diff
@@ -0,0 +1,150 @@
...
+            }
+        }
+        // union queries
+        while (query.startsWith("(")) {
+            query = query.substring("(".length()).trim();
+        }
+        if (query.startsWith("select")) {
+            return false;
+        }
+        return true;
```
fabriziofortino, line 64: this could be simplified 

```suggestion
        return !query.startsWith("select");
```

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * Format the query into a more human-readable way, by adding newlines.
+     * Warning: newlines are also added inside e.g. string literals.
+     *
+     * @param query the query (may not be null)
+     * @param language the query language, or null if unknown
+     * @return the formatted query
+     */
+    public static String format(String query, String language) {
```
fabriziofortino: since the `query` param cannot be `null`, what about annotating it?
```suggestion
    public static String format(@NotNull String query, String language) {
```

oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/SelectorImpl.java
```diff
@@ -427,10 +426,14 @@ public String getPlan(NodeState rootState) {
                 buff.append(index.getPlan(createFilter(true), rootState));
             }
         } else {
-            buff.append("no-index");
+            buff.append("no-index\n");
         }
         if (!selectorConstraints.isEmpty()) {
```
fabriziofortino, line 431: this `if` block is empty. Should we remove it?

oak-core/src/test/java/org/apache/jackrabbit/oak/query/QueryFormatterTest.java
```diff
@@ -0,0 +1,178 @@
...
+ */
+package org.apache.jackrabbit.oak.query;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.util.Random;
+
+import org.apache.jackrabbit.oak.query.QueryFormatter;
```
fabriziofortino: not needed
```suggestion
```

fabriziofortino (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
thomasmueller: Backslash is not an escape character. The escape character for ' is ', so that it would be:

    "he said ""Hello"""

and

    'he said ''Hello'''

Because of this, we just need to search for the next ' / ". The "" and '' are correctly recognised. There's also a test for that.

thomasmueller: See the docs at https://jackrabbit.apache.org/oak/docs/query/grammar-sql2.html#static_operand and https://jackrabbit.apache.org/oak/docs/query/grammar-xpath.html#static_operand

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
thomasmueller: Actually... I think the current code is clearer in this case: The "not" can be confusing.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/SelectorImpl.java
thomasmueller: Well, it is not empty: I would like to keep the comment and the commented code...

fabriziofortino (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
```diff
@@ -0,0 +1,154 @@
...
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Formatter for JCR queries in order to make them easier to read. Formatting is
+ * done on a best-effort basis.
+ *
+ * Warning: Care was taken to not add newlines inside string literals and so on,
+ * but there is still no guarantee that the formatted query is semantically
+ * equal to the original one. It is recommended to run queries that are returned
+ * by these methods.
```
nfsantos: Don't you mean that it is recommended **not** to run queries...?

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Detect whether the query is an XPath query.
+     *
+     * @param query the query
+     * @param language the language, if known, or null
+     * @return true if xpath
+     */
+    public static boolean isXPath(String query, String language) {
+        if (language != null) {
+            return "xpath".equals(language);
```
nfsantos, line 46: Being case sensitive is intentional? The query string is treated as case insensitive. If the language is supposed to be case sensitive, maybe add a note about it in the comments.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+        query = query.trim().toLowerCase(Locale.ENGLISH);
```
nfsantos, line 49: If we do not want any locale related interpretation, it's better to use [Locale.ROOT](https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#ROOT).

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
```diff
@@ -0,0 +1,154 @@
...
+        // explain queries
+        if (query.startsWith("explain")) {
+            query = query.substring("explain".length()).trim();
+            if (query.startsWith("measure")) {
+                query = query.substring("measure".length()).trim();
+            }
+        }
+        // union queries
+        while (query.startsWith("(")) {
+            query = query.substring("(".length()).trim();
```
nfsantos, line 59: Maybe replace with `substring(1)` since we are just removing the first character? I understand the reason for this convention when it is a word, prevents mistakes, but when it's only a single character, maybe it is simpler to use 1.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+        if (query.startsWith("select")) {
+            return false;
+        }
+        return true;
+    }
+
+    /**
+     * Format the query into a more human-readable way, by adding newlines.
+     * Warning: newlines are also added inside e.g. string literals.
```
nfsantos:
```suggestion
     * Warning: newlines are also added inside string literals.
```
In what conditions? Do you mean that they may also be added in some rare cases?

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
```diff
@@ -0,0 +1,154 @@
...
+            query = query.substring("(".length()).trim();
+        }
+        if (query.startsWith("select")) {
+            return false;
+        }
+        return true;
+    }
+
+    /**
+     * Format the query into a more human-readable way, by adding newlines.
```
nfsantos:
```suggestion
     * Format the query into a more human-readable way by adding newlines.
```
Adding newlines where? This sentence could be a bit more clear.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
```diff
@@ -0,0 +1,154 @@
...
+     * @param query the query (may not be null)
+     * @param language the query language, or null if unknown
+     * @return the formatted query
+     */
+    public static String format(@NotNull String query, String language) {
+        boolean xpath = isXPath(query, language);
+        if (xpath) {
+            return formatXPath(query);
+        } else {
+            return formatSQL(query);
```
nfsantos, line 80: We are not sure here if the language is sql. If language is specified, add a check that it is either sql or xpath. If we autodetect, then can we check if it is sql by looking at if it starts with `select`? That way, we could have a nice error message saying that we do not recognize the language instead of failing with a syntax error or returning a string with newlines added in random places.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+
+    private static String formatXPath(String query) {
+        StringBuilder buff = new StringBuilder(query);
+        for (int i = 0; i < buff.length(); i++) {
+            char c = buff.charAt(i);
+            if (c == '\'' || c == '"') {
```
nfsantos, line 88: This could be transformed into a switch.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/property/MultiPropertyOrTest.java
```diff
@@ -121,22 +121,17 @@ public void unionSortResultCount() throws Exception {
         root.commit();
 
         List<Integer> nodes = Lists.newArrayList();
-        Random r = new Random();
-        int seed = -2;
+        Random r = new Random(1);
```
nfsantos, line 124: Why? You want the test to be deterministic?

nfsantos (commented):
Just a few more comments.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
thomasmueller: Of course 😵‍💫

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
thomasmueller: The _language_ is case sensitive. The query, for SQL, is not (for XPath it is).

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
thomasmueller: Locale.ENGLISH is safe according to https://stackoverflow.com/questions/10336730/which-locale-should-i-specify-when-i-call-stringtolowercase . We use that in other parts too, so I would keep this.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
thomasmueller: I used this for consistency, and to prevent errors in maintenance.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
thomasmueller: > we could have a nice error message saying that we do not recognize the language

I don't see a need for this currently. The current use case is debug log, where we actually always know which language it is.

I do want to re-use this in the near future for http://oakutils.appspot.com/generate/index where the language detection is (unfortunately) incorrect.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryFormatter.java
thomasmueller: Yes. But that wouldn't necessarily be better. I prefer the "if else" in this case.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/property/MultiPropertyOrTest.java
thomasmueller: Yes.

nfsantos (approved):

amit-jain (approved):
LGTM. Haven't looked at the tests carefully though.

PR #1190, opened 2023-11-06
OAK-10526 : set split doc maxRev to 'now' at split time, to avoid it …

stefan-egli:
…being GCed too early


mreutegg (approved):
Looks good to me.

PR #1195, opened 2023-11-08
OAK-10542 : test reproducing ignorance of split deleted properties

stefan-egli:


stefan-egli:
Additional test case created which combines OAK-10542 with OAK-10526 - i.e. that test requires both bugs to be fixed to succeed

stefan-egli:
merged without review to move ahead with sibling https://github.com/apache/jackrabbit-oak/pull/1198

PR #1196, opened 2023-11-08
OAK-10545: oak-jcr: add test for observing effect of remapping a namespace prefix to a different namespace name

reschke:


oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/NamePathTest.java
```diff
@@ -212,6 +218,47 @@ public void testSpaceInNames() throws RepositoryException {
         session.getNode(n.getPath());
     }
 
+    @Test
+    @Ignore("OAK-10544")
```
kwin, line 222: Shouldn't we rather add it to the https://svn.apache.org/repos/asf/jackrabbit/trunk/jackrabbit-jcr-tests/? It doesn't use any Oak API IIUC and JR2 should work just fine with that.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/NamePathTest.java
```diff
@@ -212,6 +218,47 @@ public void testSpaceInNames() throws RepositoryException {
...
+    public void testPrefixRemapping() throws NamespaceException, RepositoryException {
+        Random r = new Random();
+        int i1 = r.nextInt();
+        int i2 = r.nextInt();
+        String prefix = "nstest" + i1;
```
kwin, line 227: can we make the prefix really hardcoded here? Adding the random number here just complicates the error output.

anchela (approved):
lgtm

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/NamePathTest.java
reschke: That's a good point in the mid term: those reside in Jackrabbit though, so it would take some time to have a release that contains them and can be used in Oak. But yes.

Also, in the meantime I learned that there maybe a more specific test class I could use; I'll update the PT soonish.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/NamePathTest.java
reschke: In Oak we probably can assume that we're working on an empty repo. But that's something we wouldn't be able to rely on in TCK tests.

reschke:
@kwin 's TCK proposal reminds me that we may actually have that test, but that might be marked as ignored for Oak. Will check tomorrow.

reschke:
So, in Jackrabbit, there is https://github.com/apache/jackrabbit/blob/90b60f4561eae41256c2110df3a72b854ead962d/jackrabbit-jcr-tests/src/main/java/org/apache/jackrabbit/test/api/NamespaceRemappingTest.java#L113 - but apparently that does not trigger the failure in Oak.

PR #1198, opened 2023-11-08
OAK-10542 : getLiveRevision must inspect split docs if local deleted …

stefan-egli:
…does not contain most recent committed revision.

This is currently waiting for https://github.com/apache/jackrabbit-oak/pull/1195 to be merged first - after which the tests will be unignored.


reschke (commented):
Not my area of expertise, but then, and under the assumptions that we have good test coverage and the new tests are correct...:

The actual change looks good to me. IIUC, it disables an optimization that avoids looking at split docs; so in the worst case, we'd see a performance degradation, right?

stefan-egli:
> it disables an optimization that avoids looking at split docs; so in the worst case, we'd see a performance degradation, right?

right. 
* in most of the cases there wouldn't be any split documents anyway (in which cases this change is a virtual no-op)
* in those cases where there are split document it would mostly not consider them neither as `isMostRecentCommitted` would return true (in which case this change should not have any noticeable performance impact)
* but in heavily updated nodes it would have an impact, yes. but that's also where the fix is needed.

btw: it does follow the same logic as in `getNodeAtRevision` ([here](https://github.com/apache/jackrabbit-oak/blob/a5300d8088eacf7d5662ef5b8a7a64abad1ce7bf/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocument.java#L969-L993))

reschke:
Oh, good point.

Maybe open a TODO ticket to refactor this later into a single utility method?

stefan-egli:
yes I was considering a refactoring actually. there's a few subtleties and non-straight-forwards, so I opted out for now. Will add a TODO

stefan-egli:
change of plan : refactored the code duplication. /cc @reschke

reschke (approved):

rishabhdaim (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocument.java
```diff
@@ -2181,6 +2171,34 @@ private String getCommitValue(Revision revision) {
...
+    /**
+     * Check if there may be more recent values in a previous document and return
+     * null if the latter is the case
+     *
+     * @param localValue value as resolved from local value map
+     * @param local      local value map
+     * @param context    the revision context
+     * @return localValue if it is most recent, null otherwise
+     */
+    private Value checkIsMostRecentCommitted(@Nullable Value localValue,
```
Joscorbe: Just an opinion but, the way this function works and how it is used above, it may be more intuitive and clean if it just returns a boolean for example.

Joscorbe (changes requested):

Joscorbe (approved):
I approve it, my comment is just a suggestion, I think it improves understandability of the code.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocument.java
stefan-egli: +1, renamed as suggested. wdyt?

Joscorbe: +1

PR #1202, opened 2023-11-10
OAK-10541 - Pipelined strategy: improve memory management of transform stage

nfsantos:
This PR makes several improvements to memory management of the buffers between the transform and batch-sort threads, with the goal of reducing the working set and memory requirements. 

Just to recap, the data with the entries is serialized into a ByteBuffer by the transform thread. But to sort, it is necessary to use an array of Java objects with the keys and pointers to locations in the ByteBuffer. Previously, the array with the sort keys was being created by the transform threads at the same time as the entries were written to the ByteBuffer. However, the array with sort keys is needed only later, by the sort thread. Until then, this array is taking up space in memory. This is an issue because there are n transform threads to 1 sort thread, so we were keeping n sort arrays when only one will ever be needed at a time.

This PR changes this, so that the transform threads do not build the sort array. Instead, they only write the entries to the ByteBuffer. The sort-batch thread will then create the sort array from the contents of the ByteBuffer just before sorting. This simplifies and reduces the work done by the transform threads, adding a little extra work to the sort thread. But in the tests that I performed, the creation of the sort array takes less time than sorting the array and writing the file to disk. Additionally, these tasks are faster than building a batch by the transform threads, so a single sort thread can still easily keep up with several at least up to 3 transform threads. 

The size of the sort array is taken from the working memory. This memory is now divided between a region for the sort array and n regions for the bytebuffers, so: `WorkingMemory = SortArrayMaxSize + n*ByteBufferSize`. To estimate SortArrayMaxSize is estimated as a fraction of the ByteBufferSize, that is `ByteBufferSize*FractionOfKeys/100`, where FractionOfKeys is a value between 1 and 100, that represents approximately how much space the paths of the entries take in proportion of the full FFS line (so path + json). The default is 25, so assuming that a path is on average 25% of a line in the FFS. This can be set with this configuration property:

- `oak.indexer.pipelined.sortBufferMemoryPercentage`

Additional changes:
- Release the references to the buffers used between the different stages of the pipelined strategy as soon as those stages are complete, to allow them to be GCed. In particular, the buffers between transform and sort-batch threads are very large and they were being kept alive during the final merge-sort phase, long after the transform and sort-batch threads had terminated. This was creating memory pressure during sort-merge phase, which could lead to OOME during this phase.
- When writing to compressed sorted file, do not put a buffered stream on top of the compression streams. The compression streams already do their own buffering. Instead, put a buffered stream between the compression stream and the file stream.


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeStateEntryBatch.java
```diff
@@ -51,49 +49,47 @@ public static NodeStateEntryBatch createNodeStateEntryBatch(int bufferSizeBytes,
     }
 
     private final ByteBuffer buffer;
-    private final ArrayList<SortKey> sortBuffer;
     private final int maxEntries;
+    private int numberOfEntries;
+    private int sizeOfEntries;
```
thomasmueller: sizeOfEntries sounds like bytes. At some point, we might want to go beyond 2 GB. Could we switch to "long"? Maybe better for all of then entries: maxEntries, numberOfEntries, but most importantly, sizeOfEntries.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedSortBatchTask.java
```diff
@@ -116,48 +136,90 @@ public Result call() throws Exception {
...
-        LOG.info("Stored batch of size {} (uncompressed {}) with {} entries in {}",
-                humanReadableByteCountBin(Files.size(newtmpfile)),
+        timeWritingMillis += saveClock.elapsed().toMillis();
+        long compressedSize = Files.size(newtmpfile);
+        LOG.info("Wrote batch of size {} (uncompressed {}) with {} entries in {} at {} MB/s",
+                humanReadableByteCountBin(compressedSize),
                 humanReadableByteCountBin(textSize),
-                sortBuffer.size(), saveClock);
+                sortBuffer.size(), saveClock,
+                String.format("%2.2f", ((1.0 * compressedSize / saveClock.elapsed().toMillis()) * 1000) / FileUtils.ONE_MB)
```
thomasmueller: There's a risk of division by 0...

What about adding another utility method for this division?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedStrategy.java
```diff
@@ -276,21 +276,19 @@ public PipelinedStrategy(MongoDocumentStore documentStore,
         );
         this.mongoDocQueueSize = mongoDocQueueReservedMemoryMB / mongoDocBatchMaxSizeMB;
 
-        // Transform threads <-> merge-sort
+        // Derived values for transform <-> sort-save
+        int nseWorkingMemoryMB = readNSEBuffersReservedMemory();
         this.nseBuffersCount = 1 + numberOfTransformThreads;
-
-        long nseBuffersReservedMemoryBytes = nseBuffersReservedMemoryMB * FileUtils.ONE_MB;
+        long nseWorkingMemoryBytes = nseWorkingMemoryMB * FileUtils.ONE_MB;
```
thomasmueller: FileUtils.ONE_MB is a "long", so this is fine...

An option would be to cast nseWorkingMemoryMB to long here, so this is very clear.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedStrategy.java
```diff
@@ -276,21 +276,19 @@ public PipelinedStrategy(MongoDocumentStore documentStore,
...
+        long memoryReservedForSortKeysArray = estimateMaxSizeOfSortKeyArray(nseWorkingMemoryBytes, nseBuffersCount, sortBufferMemoryPercentage);
+        long memoryReservedForBuffers = nseWorkingMemoryBytes - memoryReservedForSortKeysArray;
 
-        // Assuming 1 instance of SortKey takes around 256 bytes. We have #transformThreads + 1 regions of nseBufferSizeBytes.
-        // The extra region is for the SortKey instances. Below we compute the total number of SortKey instances that
-        // fit in the memory region reserved for them, assuming that each SortKey instance takes 256 bytes. Then we
-        // distribute equally these available entries among the nse buffers
-        this.nseBufferMaxEntriesPerBuffer = (this.nseBuffersSizeBytes / 256) / this.nseBuffersCount;
+        // A ByteBuffer can be at most Integer.MAX_VALUE bytes long
+        this.nseBuffersSizeBytes = limitToIntegerRange(memoryReservedForBuffers / nseBuffersCount);
```
thomasmueller, line 298: We had "/ (nseBuffersCount + 1)" before, which prevents division by 0.

Maybe another utility method?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedStrategy.java
```diff
@@ -304,11 +302,21 @@ public PipelinedStrategy(MongoDocumentStore documentStore,
...
+                IOUtils.humanReadableByteCountBin(memoryReservedForSortKeysArray)
+        );
+    }
+
+    static long estimateMaxSizeOfSortKeyArray(long nseWorkingMemoryBytes, long nseBuffersCount, int sortBufferMemoryPercentage) {
+        // We reserve a percentage of the size of a buffer for the sort keys array. That is, we are assuming that for every line
+        // in the sort buffer, the memory needed to store the SortKey of the path section of the line will not be more
+        // than sortBufferMemoryPercentage of the total size of the line in average
+        // Estimate memory needed by the sort keys array. We assume each entry requires 256 bytes.
+        long approxNseBufferSize = limitToIntegerRange(nseWorkingMemoryBytes / nseBuffersCount);
```
thomasmueller, line 325: Possibly division by 0

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedSortBatchTask.java
```diff
@@ -116,48 +136,90 @@ public Result call() throws Exception {
...
+                LOG.debug("Resizing copy buffer from {} to {}", copyBuffer.length, pathLength);
+                copyBuffer = new byte[pathLength];
+            }
+            buffer.get(copyBuffer, 0, pathLength);
+            // Skip the json
+            int entryLength = buffer.getInt();
+            buffer.position(buffer.position() + entryLength);
+            // Create the sort key
+            String path = new String(copyBuffer, 0, pathLength, StandardCharsets.UTF_8);
+            String[] pathSegments = SortKey.genSortKeyPathElements(path);
```
thomasmueller, line 153: Just an idea: if we don't need "path" and only the array, then maybe we could store the number of entries and then the entries separately?

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeStateEntryBatch.java
nfsantos: We cannot go beyond 2GB because that's the size limit of a `ByteBuffer`. I have removed the sizeOfEntries field because this is the same as `buffer.position()` and renamed the method to get this value to `sizeOfEntriesBytes()`.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedSortBatchTask.java
nfsantos: Could you elaborate? I do not understand the suggestion.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedSortBatchTask.java
nfsantos: Done.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedStrategy.java
nfsantos: `nseBuffersCount` will always be bigger than 2. In the constructor, it is defined as 
```
this.nseBuffersCount = 1 + numberOfTransformThreads;
```
And `numberOfTransformThreads` is checked to make sure that it is 1 or more.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedStrategy.java
nfsantos: Not possible as explained before.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedStrategy.java
nfsantos: ok, I added the explicit cast.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedSortBatchTask.java
thomasmueller: Something like that:

    int pathSegmentCount = buffer.getInt();
    String[] pathSegments = new String[pathSegments];
    for (int i = 0; i < pathSegmentCount; i++) {
        pathSegments[i] = readString(buffer);
    }

this is with a readString utility method, which maybe would anyway make sense.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedUtils.java
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public static String formatAsPercentage(long numerator, long denominator) {
+        if (denominator == 0) {
+            return "N/A";
+        } else {
+            return String.format("%1.2f", (100.0 * numerator) / denominator);
+        }
+    }
+
+    public static String formatAsTransferSpeedMBs(long numberOfBytes, long timeMiilis) {
```
thomasmueller: Typo 
```suggestion
    public static String formatAsTransferSpeedMBs(long numberOfBytes, long timeMillis) {
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedSortBatchTask.java
nfsantos: I don't think this would be faster. I tried a similar idea when optimizing the merge phase, when reading lines from the intermediate sorted files. It was about the same speed to read the segments of the path directly one by one than to read the full path as a string and then break it into segments, even though in the second option we are allocating an extra String. I was very surprised. I think the reason is that with the path as a String, we can use the `String.indexOf` method, which is an intrinsic and likely heavily optimized. If we read the path segments from the byte buffer, we have to iterate byte by byte in the buffer until finding the `/` characters, then build a string. This will very likely be slower than just reading the bytes with the path in a single operation from the byte buffer and then breaking the path string into segments. Furthermore, the String with the path is very short lived, so the JIT is likely applying escape analysis to allocate it on the stack.

thomasmueller: > if we read the path segments from the byte buffer, we have to iterate byte by byte in the buffer until finding the / characters

No. The above code first writes the number of path segments, and then each path segment. There is no search for '/'.

thomasmueller: Maybe I should also add readString:

    private static String readString(ByteBuffer buff) {
        int len = buff.getShort();
        if (len == -2) {
            return null;
        } else if (len == -1) {
            len = buff.getInt();
            int pos = buff.position();
            buff.position(buff.position() + len);
            return new String(buff.array(), pos, len, StandardCharsets.UTF_8);
        } else {
            int pos = buff.position();
            buff.position(buff.position() + len);
            return new String(buff.array(), pos, len, StandardCharsets.UTF_8);
        }
    }

nfsantos: I have changed the code to read directly from the ByteBuffer instead of copying to an intermediate buffer, that is, to do this:
```
new String(buff.array(), pos, len, StandardCharsets.UTF_8);
```
However, the proposed solution of writing to the byte buffer the path segments separately just moves the work of splitting the string with the path into segments from the sort-batch thread into the transform thread, and it would complicate somewhat reading and writing to the ByteBuffer, which instead of having two strings for each entry (path and json) would now have n strings for the path and 1 for the json. Therefore, I will leave it as it is.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeStateEntryBatch.java
```diff
@@ -51,49 +49,46 @@ public static NodeStateEntryBatch createNodeStateEntryBatch(int bufferSizeBytes,
     }
 
     private final ByteBuffer buffer;
-    private final ArrayList<SortKey> sortBuffer;
     private final int maxEntries;
+    private int numberOfEntries = 0;
```
thomasmueller, line 53: 0 is the default
```suggestion
    private int numberOfEntries;
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeStateEntryBatch.java
```diff
@@ -51,49 +49,46 @@ public static NodeStateEntryBatch createNodeStateEntryBatch(int bufferSizeBytes,
...
+    private int sizeOfEntriesBytes = 0;
```
thomasmueller, line 54:
```suggestion
    private int sizeOfEntriesBytes;
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedSortBatchTask.java
```diff
@@ -66,9 +68,12 @@ public long getTotalEntries() {
     private final BlockingQueue<NodeStateEntryBatch> nonEmptyBuffersQueue;
     private final BlockingQueue<Path> sortedFilesQueue;
     private final Path sortWorkDir;
-    private final byte[] copyBuffer = new byte[4096];
+    private final ArrayList<SortKey> sortBuffer = new ArrayList<>(32 * 1024);
     private long entriesProcessed = 0;
     private long batchesProcessed = 0;
+    private long timeCreatingSortArrayMillis = 0;
+    private long timeSortingMillis = 0;
+    private long timeWritingMillis = 0;
```
thomasmueller, line 76:
```suggestion
    private long entriesProcessed;
    private long batchesProcessed;
    private long timeCreatingSortArrayMillis;
    private long timeSortingMillis;
    private long timeWritingMillis;
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedTransformTask.java
```diff
@@ -88,6 +89,7 @@ public int getThreadId() {
     private final TransformStageStatistics statistics;
     private final int threadId = threadIdGenerator.getAndIncrement();
     private long totalEnqueueDelayMillis = 0;
+    private long totalEmptyBatchQueueWaitTimeMillis = 0;
```
thomasmueller, line 92:
```suggestion
    private long totalEnqueueDelayMillis;
    private long totalEmptyBatchQueueWaitTimeMillis;
```

thomasmueller (approved):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedStrategy.java
```diff
@@ -270,10 +271,9 @@ public PipelinedStrategy(MongoDocumentStore documentStore,
         Preconditions.checkArgument(numberOfTransformThreads > 0,
                 "Invalid value for property " + OAK_INDEXER_PIPELINED_TRANSFORM_THREADS + ": " + numberOfTransformThreads + ". Must be > 0");
 
-        // Derived values for transform <-> sort-save
-        int nseBuffersReservedMemoryMB = readNSEBuffersReservedMemory();
-
-        // Calculate values derived from the configuration settings
+        int sortBufferMemoryPercentage = ConfigHelper.getSystemPropertyAsInt(OAK_INDEXER_PIPELINED_SORT_BUFFER_MEMORY_PERCENTAGE, DEFAULT_OAK_INDEXER_PIPELINED_SORT_BUFFER_MEMORY_PERCENTAGE);
+        Preconditions.checkArgument(sortBufferMemoryPercentage > 0 && sortBufferMemoryPercentage <= 100,
+                "Invalid value for property " + OAK_INDEXER_PIPELINED_TRANSFORM_THREADS + ": " + numberOfTransformThreads + ". Must be > 0");
```
fabriziofortino: minor: this should say "must be between 0 and 100"

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeStateEntryBatch.java
nfsantos: As discussed privately, I prefer to initialize the values explicitly, to convey the intention that 0 is the initial value for the field, as opposed to a variable that is not assigned in its definition because it will be initialized in the construtor.

fabriziofortino (approved):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedStrategy.java
nfsantos: Good point. In fact, it should say between 1 and 100, 0 is not a valid value. It was also printing the wrong name of the property. I fixed the message.

PR #1203, opened 2023-11-12
OAK-10547 - Fix: Indexing job fails at the end of reindexing if it took more than 24h to run

nfsantos:
Fix support formatting durations longer than 24h.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/FormattingUtils.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/FormattingUtils.java
@@ -18,18 +18,28 @@ package org.apache.jackrabbit.oak.plugins.index;
 
 import org.apache.jackrabbit.guava.common.base.Stopwatch;
 
-import java.time.LocalTime;
-import java.time.format.DateTimeFormatter;
 import java.util.concurrent.TimeUnit;
 
 public class FormattingUtils {
     public static String formatToSeconds(Stopwatch stopwatch) {
-        LocalTime seconds = LocalTime.ofSecondOfDay(stopwatch.elapsed(TimeUnit.SECONDS));
-        return DateTimeFormatter.ISO_TIME.format(seconds);
+        long seconds = stopwatch.elapsed(TimeUnit.SECONDS);
+        long absSeconds = Math.abs(seconds);
+        return String.format(
+                "%02d:%02d:%02d",
+                absSeconds / 3600,
+                (absSeconds % 3600) / 60,
+                absSeconds % 60);
     }
 
     public static String formatToMillis(Stopwatch stopwatch) {
-        LocalTime nanoSeconds = LocalTime.ofNanoOfDay(stopwatch.elapsed(TimeUnit.MILLISECONDS)*1000000);
-        return DateTimeFormatter.ISO_TIME.format(nanoSeconds);
+        long millis = stopwatch.elapsed(TimeUnit.MILLISECONDS);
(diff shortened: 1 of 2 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/FormattingUtils.java
```diff
@@ -18,18 +18,28 @@
...
-import java.time.LocalTime;
-import java.time.format.DateTimeFormatter;
 import java.util.concurrent.TimeUnit;
 
 public class FormattingUtils {
     public static String formatToSeconds(Stopwatch stopwatch) {
-        LocalTime seconds = LocalTime.ofSecondOfDay(stopwatch.elapsed(TimeUnit.SECONDS));
-        return DateTimeFormatter.ISO_TIME.format(seconds);
+        long seconds = stopwatch.elapsed(TimeUnit.SECONDS);
+        long absSeconds = Math.abs(seconds);
```
thomasmueller, line 26: That way we don't detect if the time went backwards, which may be the case in some edge case.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/FormattingUtils.java
```diff
@@ -18,18 +18,28 @@
...
+        return String.format(
+                "%02d:%02d:%02d",
+                absSeconds / 3600,
```
thomasmueller: The formulas look correct, but it should be possible to use Java utilities...

```
long millis = -123456789;
long hours = Math.abs(TimeUnit.MILLISECONDS.toHours(millis));
long minutes = Math.abs(TimeUnit.MILLISECONDS.toMinutes(millis)) % 60;
long seconds = Math.abs(TimeUnit.MILLISECONDS.toSeconds(millis)) % 60;
long milliseconds = Math.abs(millis) % 1000;

String formattedTime = String.format("%s%02d:%02d:%02d.%03d", millis < 0 ? "-" : "", hours, minutes, seconds, milliseconds);
System.out.println(formattedTime);
```

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/FormattingUtils.java
nfsantos: Changed to print negative time spans: prefix with `-`.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/FormattingUtils.java
nfsantos: Yes, it becomes easier to read. I have refactored the code.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/FormattingUtilsTest.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+package org.apache.jackrabbit.oak.plugins.index;
+
+import org.apache.jackrabbit.guava.common.base.Stopwatch;
+import org.apache.jackrabbit.guava.common.base.Ticker;
+import org.junit.Test;
+
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.*;
```
fabriziofortino: avoid `*` imports

fabriziofortino (approved):

thomasmueller (approved):

PR #1213, opened 2023-11-15
OAK-10551: o.a.j.o.jcr.osgi.RepositoryManager declares OSGi propertie…

mbaedke:
…s that are not used at all.

Removed misleading property name declarations.


reschke (commented):
Maybe the intent was to use "label", not "name"?

mbaedke:
@reschke that is very likely, but in view of https://github.com/apache/jackrabbit-oak/pull/1214 it doesn't really matter. There is, afaik, no replacement for the label parameter.

reschke:
I agree these are wrong, but they *are* used as names in the generated OSGi metadata. We need to fix them, butI'm not sure we can simply remove (thus rename) them.

reschke (approved):

PR #1214, opened 2023-11-15
OAK-6765: Convert oak-jcr to OSGi R7 annotations

mbaedke:
org.apache.felix.scr.annotations -> org.osgi.service.component.annotations


oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/osgi/RepositoryManagerTest.java
```diff
@@ -85,6 +86,7 @@ public void repositoryShutdown() throws Exception{
 
 
     private void registerRequiredServices() {
+        context.registerService(StatisticsProvider.class, StatisticsProvider.NOOP);
```
reschke: Why was this needed? (Did the test make bad assumptions or do we actually have a detectable behavioral change?)

mbaedke: Technically it was a behavioral change, because previously the RepositoryManager would start without the reference being fulfilled. In practice it probably doesn't make a difference, because e.g. org.apache.jackrabbit.oak.plugins.blob.datastore.AbstractDataStoreService requires a StatisticsProvider service. Anyway I removed the whole thing, because it wasn't being used at all.

reschke (approved):

PR #1215, opened 2023-11-15
OAK-10549 Improve performance of facet count at scale (Lucene)

thomasmueller:
Local tests pass now.

### DIFF
--- a/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
+++ b/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
@@ -209,6 +209,9 @@ public class LucenePropertyIndex extends FulltextIndex {
     public final static String CACHE_FACET_RESULTS_NAME = "oak.lucene.cacheFacetResults";
     private final boolean CACHE_FACET_RESULTS =
             Boolean.parseBoolean(System.getProperty(CACHE_FACET_RESULTS_NAME, "true"));
+    public final static String EAGER_FACET_CACHE_FILL_NAME = "oak.lucene.cacheFacetEagerFill";
+    private final static boolean EAGER_FACET_CACHE_FILL =
+            Boolean.parseBoolean(System.getProperty(EAGER_FACET_CACHE_FILL_NAME, "true"));
 
     private static boolean FLAG_CACHE_FACET_RESULTS_CHANGE = true;
 
@@ -1644,8 +1647,36 @@ public class LucenePropertyIndex extends FulltextIndex {
                 return cachedResults.get(cacheKey);
             }
             LOG.trace("columnName = {} facet Data not present in cache...", columnName);
+            if (EAGER_FACET_CACHE_FILL) {
+                fillFacetCache(numberOfFacets);
+                if (cachedResults.containsKey(cacheKey)) {
+                    LOG.trace("columnName = {} now found");
+                    return cachedResults.get(cacheKey);
+                }
+                LOG.warn("Facet data for {} not found: read using query", cacheKey);
+            }
             List<Facet> result = getFacetsUncached(numberOfFacets, columnName);
             cachedResults.put(cacheKey, result);
+            return result;
+        }
+
+        private List<Facet> fillFacetCache(int numberOfFacets) throws IOException {
+            List<Facet> result = null;
+            LuceneIndexNode indexNode = index.acquireIndexNode(plan);
+            try {
+                IndexSearcher searcher = indexNode.getSearcher();
+                Facets facets = FacetHelper.getFacets(searcher, query, plan, config);
+                if (facets != null) {
+                    List<String> allColumnNames = FacetHelper.getFacetColumnNamesFromPlan(plan);
+                    for (String column : allColumnNames) {
+                        result = getFacetsUncached(facets, numberOfFacets, column);
+                        String cc = column + "/" + numberOfFacets;
+                        cachedResults.put(cc, result);
+                    }
+                }
+            } finally {
+                indexNode.release();
+            }
             return result;
         }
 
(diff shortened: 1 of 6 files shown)


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
```diff
@@ -1644,11 +1647,39 @@ public List<Facet> getFacets(int numberOfFacets, String columnName) throws IOExc
                 return cachedResults.get(cacheKey);
             }
             LOG.trace("columnName = {} facet Data not present in cache...", columnName);
+            if (EAGER_FACET_CACHE_FILL) {
+                fillFacetCache(numberOfFacets);
+                if (cachedResults.containsKey(cacheKey)) {
+                    LOG.trace("columnName = {} now found");
```
fabriziofortino: `cacheKey` is missing
```suggestion
                    LOG.trace("columnName = {} now found", cacheKey);
```

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
```diff
@@ -1677,6 +1708,28 @@ private List<Facet> getFacetsUncached(int numberOfFacets, String columnName) thr
...
+                }
+                for (LabelAndValue lav : topChildren.labelValues) {
+                    res.add(new Facet(
+                            lav.label, lav.value.intValue()
+                    ));
+                }
+                return res.build();
+            } catch (IllegalArgumentException iae) {
+                LOG.debug(iae.getMessage(), iae);
+                LOG.warn("facets for {} not yet indexed: " + iae, facetFieldName);
```
fabriziofortino: this message mixes placeholders + interpolation. I would prefer to use just placeholders:
```suggestion
                LOG.warn("facets for {} not yet indexed: {}", facetFieldName, iae);
```

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
```diff
@@ -1677,6 +1708,28 @@ private List<Facet> getFacetsUncached(int numberOfFacets, String columnName) thr
                 indexNode.release();
             }
         }
+
+        private List<Facet> getFacetsUncached(Facets facets, int numberOfFacets, String columnName) throws IOException {
```
fabriziofortino, line 1712: `getFacetsUncached(int numberOfFacets, String columnName)` has the exact same logic at lines 1690-1705. What about reducing the code in this class (that is already huge) and call the new `getFacetsUncached` in there?

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/hybrid/ManyFacetsTest.java
```diff
@@ -0,0 +1,280 @@
...
+            assertTrue(it.hasNext());
+            while (it.hasNext()) {
+                Row r = it.nextRow();
+                for (int i = 0; i < qr.getColumnNames().length; i++) {
+                    String columnName = qr.getColumnNames()[i];
+                    String v = r.getValue(columnName).getString();
+                    JsonObject json = JsonObject.fromJson(v, true);
+                    for (int j = 0; j < NUM_LABELS; j++) {
+                        String n = json.getProperties().get("foo" + i + "x" + j);
+                        assertTrue(n != null);
```
fabriziofortino: `assertNotNull` is a better option here
```suggestion
                        assertNotNull(n);
```

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/hybrid/ManyFacetsTest.java
```diff
@@ -0,0 +1,280 @@
...
+        root.refresh();
+    }
+    
+    @Test
+    public void facet() throws Exception {
+        // Explicitly setting following configs to run DelayedLuceneFacetProvider and a thread sleep of 50 ms in refresh readers. Refer: OAK-8898
+        System.setProperty(LucenePropertyIndex.OLD_FACET_PROVIDER_CONFIG_NAME, "false");
+        // The variable is static final so once set it remains same for all tests and which will lead to slow execution
+        // of other tests as this add a sleep of specified milliseconds in refresh reader method in LuceneIndexNodeManager.
+       // System.setProperty(LuceneIndexNodeManager.OLD_FACET_PROVIDER_TEST_FAILURE_SLEEP_INSTRUMENT_NAME, "40");
```
fabriziofortino: nitpick: fix indentation
```suggestion
        // System.setProperty(LuceneIndexNodeManager.OLD_FACET_PROVIDER_TEST_FAILURE_SLEEP_INSTRUMENT_NAME, "40");
```

fabriziofortino (changes requested):

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
thomasmueller: I did this on purpose actually: to keep the old code as is as much as possible, in case we need to switch back to it. I agree it would be better, but hopefully we don't need the old code in the future - in which case we should remove it. I'm fine to create a jira issue about removing the old code in 6 month, so we don't forget. OK?

fabriziofortino (approved):

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
```diff
@@ -1644,11 +1647,39 @@ public List<Facet> getFacets(int numberOfFacets, String columnName) throws IOExc
                 return cachedResults.get(cacheKey);
             }
             LOG.trace("columnName = {} facet Data not present in cache...", columnName);
+            if (EAGER_FACET_CACHE_FILL) {
```
tihom88, line 1650: As we are filling cache with all facets. Would it make sense to initialise it in constructor itself using future. Like @fabriziofortino did in https://github.com/apache/jackrabbit-oak/pull/1217

thomasmueller: I prefer to avoid concurrency. it is filled when the first one is accessed, so I think it "should be fine". (I know, it sounds like "famous last words")

PR #1221, opened 2023-11-20
OAK-9796: oak-segment-remote Redis tests fail with ARM processor (App…

mreutegg:
Skip test if redis cannot be started


mreutegg:
Build timeout in unrelated oak-jcr:ut module.

PR #1223, opened 2023-11-20
OAK-10382 : added flatfile command to the project

stefan-egli:


stefan-egli:
The sonar warning is about 
```
Failed condition 11.0% 11.0% Duplicated Lines (%) on New Code (is greater than 3%)
```
which I would argue are fine in this oak-run context. To avoid this duplication would mean to restructure oak-run commands and introduce base classes and I'm not sure that is helping much in this case.

Joscorbe (approved):
Looks good to me

PR #1238, opened 2023-11-28
OAK-10570 : add detailedGC capability to oak-run revisions collect/re…

stefan-egli:
…set/info


stefan-egli:
moved from OAK-10382 to OAK-10570

Joscorbe (approved):
Looks good to me

PR #1240, opened 2023-11-29
OAK-10370 : added dryRun mode for detailedGC

rishabhdaim:


stefan-egli (commented):
I guess dry-run could be used together with `revisions reset` and thus be executed multiple times. And even with a reset after a dry-run to have the same situations as before dry-run. So the current state of the PR already has its use cases.

However, one could argue that it currently is not really dry - as it indeed does modify the settings collection.

I'm wondering if we could get it to be really dry. As it would allow to run without any interference whatsoever on the database or even a running Oak system. (we could even argue if the DocumentNodeStore should then not be opened in actual read-only mode as well)

wdyt?

rishabhdaim:
I agree that the `dry-run` shouldn't affect the document node store, even updating the `settings` collections.

IIRC, we discussed this point earlier and agreed that we would try to implement this if possible.

stefan-egli:
As an alternative to making those VersionGarbageCollector constants public, what about moving RevisionsCommandTest to `org.apache.jackrabbit.oak.plugins.document` ? oak-run/test already has a couple other similar cases, and it seems there is some level of coupling between RevisionsCommandTest and VersionGarbageCollector anyway, so having them in the same package (even if in different bundles) would fit, wdyt? Otherwise lgtm.

rishabhdaim:
>what about moving RevisionsCommandTest to org.apache.jackrabbit.oak.plugins.document ?

done in https://github.com/apache/jackrabbit-oak/pull/1240/commits/a8550ca24920df643a181339cfe432712d9c03e5

stefan-egli (approved):
I see that we have some overlap with detailedGC flag in the revisions command (coming from https://github.com/apache/jackrabbit-oak/pull/1238). But I'm fine with merging this PR and consolidating it in the DetailedGC branch afterwards, so +1

PR #1244, opened 2023-11-30
Issue/oak 10573 azure write timeout

smiroslav:
https://issues.apache.org/jira/browse/OAK-10573

### DIFF
--- a/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
+++ b/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
@@ -139,6 +139,7 @@ public class AzureArchiveManager implements SegmentArchiveManager {
             getBlobs(archiveName)
                     .forEach(cloudBlob -> {
                         try {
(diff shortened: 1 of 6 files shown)


smiroslav:
Now I see that AzureSegmentArchiveWriter, also has retry parameters. Need to align it with parameters introduced in this PR.
https://github.com/apache/jackrabbit-oak/blob/trunk/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureSegmentArchiveWriter.java#L45

smiroslav:
> Now I see that AzureSegmentArchiveWriter, also has retry parameters. Need to align it with parameters introduced in this PR. https://github.com/apache/jackrabbit-oak/blob/trunk/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureSegmentArchiveWriter.java#L45

The code in this PR does not need to be changed, since already existing parameters, ```azure.segment.archive.writer.retries.max```  and ```azure.segment.archive.writer.retries.intervalMs``` are configuring ```Retrier```, and in execute method there is always ```writeAccessController.checkWritingAllowed();``` check.

dulceanu (approved):

PR #1247, opened 2023-12-05
OAK-10577 Advanced repository statistics

thomasmueller:
Ready for review


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/StatsBuilder.java
```diff
@@ -0,0 +1,129 @@
...
+     *
+     * @param args the file name
+     */
+    public static void main(String... args) throws Exception {
+
+        String fileName = args[0];
+        String filter = null;
+        if (args.length > 1) {
+            filter = args[1];
+        }
```
nfsantos: Are you planning on creating a command-line tool somewhere to use this? Or integrate in oak-run? It would be nice to have an easy-to-use entry point, with documentation and embedded help (`--help`).

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/StatsBuilder.java
```diff
@@ -0,0 +1,129 @@
...
+    private static final boolean ONLY_READ = false;
+
+    /**
+     * Read a flat file store and build statistics.
+     *
+     * @param args the file name
+     */
+    public static void main(String... args) throws Exception {
+
+        String fileName = args[0];
```
nfsantos: Add check for number of arguments and print a message if no args are provided.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/StatsBuilder.java
```diff
@@ -0,0 +1,129 @@
...
+        collectors.add(new TopLargestBinaries(10));
+        if (filter != null) {
+            collectors.add(new NodeNameFilter(filter, new BinarySize(100_000_000, 1)));
+            collectors.add(new NodeNameFilter(filter, new BinarySizeEmbedded(100_000, 1)));
+            collectors.add(new NodeNameFilter(filter, new BinarySizeHistogram(1)));
+            collectors.add(new NodeNameFilter(filter, new TopLargestBinaries(10)));
+        }
+
+        Profiler prof = new Profiler().startCollecting();
+
```
nfsantos: Print message saying which file is going to be processed.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/StatsBuilder.java
```diff
@@ -0,0 +1,129 @@
...
+        NodeLineReader reader = NodeLineReader.open(fileName);
```
nfsantos: As lines are processed individually, maybe this can be parallelized? I think the bottleneck will be CPU, not I/O, so 2 threads should boost speed significantly. Or we make the logic so efficient that it become I/O bounded even with a single thread.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/stream/NodeLineReader.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    public NodeData readNode() throws IOException {
+        String line = reader.readLine();
+        if (line == null) {
+            close();
+            return null;
+        }
+        if (++lineCount % 1000000 == 0) {
+            System.out.println(lineCount + " lines");
```
nfsantos: Could this also print a progress report on the number of bytes read and/or the percentage of the file processed so far?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/modules/BinarySize.java
```diff
@@ -0,0 +1,111 @@
...
+        long size = 0;
+        for(NodeProperty p : node.getProperties()) {
+            if (p.getType() == ValueType.BINARY) {
+                for (String v : p.getValues()) {
+                    if (!v.startsWith(":blobId:")) {
+                        continue;
+                    }
+                    v = v.substring(":blobId:".length());
+                    if (v.startsWith("0x")) {
+                        // embedded
```
nfsantos, line 66: I have seen in some FFS that the embedded binaries also have the size at the end. But I don't know if this is always the case.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/modules/BinarySize.java
```diff
@@ -0,0 +1,111 @@
...
+                    } else {
+                        // reference
+                        int hashIndex = v.indexOf('#');
```
nfsantos: `lastIndexOf()` should be a bit faster.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/modules/BinarySizeEmbedded.java
```diff
@@ -0,0 +1,108 @@
...
+        for(NodeProperty p : node.getProperties()) {
+            if (p.getType() == ValueType.BINARY) {
+                for (String v : p.getValues()) {
+                    if (!v.startsWith(":blobId:")) {
+                        continue;
+                    }
+                    v = v.substring(":blobId:".length());
+                    if (v.startsWith("0x")) {
+                        // embedded
+                        size = (v.length() - 2) / 2;
```
nfsantos: I found a blob with the following:
```blobId:0x89504e470d0a1....e426082#8658```
So it's embedded and it has the size at the end. The formula above will reach the wrong result because it is not subtracting the size suffix.

**Review by nfsantos — COMMENTED** [review]
Great idea! I see this can become quite useful in the future. I do have some comments for improvements:

- Would be nice to have a more structured output. I see two main uses for this tool: for diagnostics/debugging and to produce machine readable data for further processing. Even for human consumption I find the output should be more structured, broken up into sections with clear description of what they are and with the possibility of requesting only a few of those sections. Since this tool may produce a lot of data, it would also be nice if it could write all output to files in disk, for instance, in a directory provided by the user as a list of separate files. 


Units of display
```
NodeCount (millions)
/: 17
/content: 17
/content/dam: 17
...
/content/dam/sesame: 1
```
This is confusing and does not have enough detail for small folders and environments. Instead I would suggest something like this:

```
NodeCount
/: 17.3M
/content: 17.3M
/content/dam: 10.5M
...
/content/dam/sesame: 650.4K
...
/content/dam/sesame/a/b/c: 33
```
This for human friendly output. For machine parseable output, then include the values in bytes. Similar comment for other types of output, the histograms are also hard to read:

```
...
refs / 33 (2147483649..4294967296): 21
refs / 34 (4294967297..8589934592): 1
refs / 35 (8589934593..17179869184): 1
...
```
This would be better:
```
...
refs / 33 (2G..4G): 21
refs / 34 (4G..8G): 1
refs / 35 (8G..17G): 1
...
```
And please don't make me try to parse this in my brain:
```
refs total size: 35212917115714
```

Sort entries at the same level of the tree by size. Instead of:
```
/content/dam/acme: 60
/content/dam/acme/bbbbb-exchange: 1
/content/dam/acme/aaaaaa: 1
/content/dam/acme/bbbbbb-ccccc: 2
/content/dam/acme/acme: 13
/content/dam/acme/acme-outlet: 2
/content/dam/acme/acme/2013: 1
```


```
/content/dam/acme: 60
/content/dam/acme/acme: 13
/content/dam/acme/foo-bar: 2
/content/dam/acme/acme-outlet: 2
/content/dam/acme/dddd-exchange: 1
/content/dam/acme/aaaa: 1
/content/dam/acme/acme/2013: 1
```

The tree of the node count is truncated at 5 levels, but the tree with BinarySize seems to include all levels. Why? Shouldn't they show the same levels? And once again, it is easier to read if the tree is shown in order of bigger children. 

On the other hand, it would also be nice to have a view where all the statistics of a path were shown after that path, like a table of entries and size. 
```
/content/dam/acme/ddddd-exchange:    236  34MB
```
So not sure what to recommend.

**Comment by thomasmueller** [comment]
> * Would be nice to have a more structured output.

Right. However I would prefer to first get some real-world data before we spend a lot of time tweaking the output. Right now we don't know what is most useful and what is not. We can guess, but likely, a lot of the data turns out to be not useful (or only useful initially), and so we will remove it later. Most of the time I sent on making sure things are "correct" so far (tests). Documentation is sparse on purpose. We can spend a lot of time tweaking, but then it turns out that the data is not useful, or that some other data is a lot more useful. I would prefer if we don't try to do this kind of "optimization" right now, except for the obvious cases.

Having human readable text values: I would prefer machine-readable output with a useful fixed magnitude, e.g. GiB for binaries, and for counts: the number of millions.

Node count in millions: I think that "million" is a useful metric. Small environments will have 0 million, but that's fine. I don't remember I had a use case to get more accurate data.

Histogram: I agree it's not readable. I changed it to eg.: 3 (2B..4B), 13 (2KiB..4KiB), 23 (2MiB..4MiB), 33 (2GiB..4GiB), 43 (2TiB..4TiB). 

For "size" and "count" data: I'll also add "size GiB", I guess that's the most interesting number; for "count", I'll add "count million".

**Comment by thomasmueller** [comment]
> it is easier to read if the tree is shown in order of bigger children.

Interesting idea! I tried that for the NodeCount, and now I'm not sure if it's really better. In a way, it looks confusing, because it's not sorted alphabetically:

```
/: 24
/content: 17
/content/dam: 17
/content/dam/braba: 10
/etc: 4
/etc/commerce: 4
/content/dam/acme: 3
/content/dam/braba/part-002: 2
/content/dam/braba/part-001: 2
/var: 2
/etc/commerce/products: 2
/etc/commerce/products/acme: 2
/content/dam/braba/part-004: 1
/var/audit: 1
/var/audit/com.day.cq.replication: 1
/content/dam/naru: 1
/var/audit/com.day.cq.replication/content: 1
/var/audit/com.day.cq.replication/content/dam: 1
/var/audit/com.day.cq.replication/content/dam/acme: 1
/content/dam/sesame: 1
/content/dam/braba/part-001/department-C: 1
/content/dam/braba/studios: 1
```

**Comment by thomasmueller** [comment]
> The tree of the node count is truncated at 5 levels, but the tree with BinarySize seems to include all levels. Why?

It is not truncated at a fixed number of levels. It's just that entries with value 0 are not shown.

File: oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/StatsBuilder.java (hunk 1 above)
Comment (thomasmueller) [reply oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/StatsBuilder.java]: At some point, yes it should be integrated into oak-run. We should probably also support segment store. And then we need to add command line documentation. But it's still in an early phase. I would prefer to not spend the time on that currently.

File: oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/StatsBuilder.java (hunk 4 above)
Comment (thomasmueller) [reply oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/StatsBuilder.java]: I would like to get real-world data before we try to optimize it.

File: oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/stream/NodeLineReader.java (hunk 5 above)
Comment (thomasmueller) [reply oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/stream/NodeLineReader.java]: I don't think that's easy to add: we don't know how many lines. We could estimate the number of bytes we read so far.

But again, it seems unnecessary currently.

Comment (nfsantos) [reply oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/stream/NodeLineReader.java]: We can easily know the number of bytes read so far with a counting input stream and we know the size of the file, so we can print a percentage of bytes processed. Not of lines, that as you say is harder.

File: oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/modules/BinarySize.java (hunk 6 above)
Comment (thomasmueller) [reply oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/modules/BinarySize.java:66]: You are right! I didn't know. (It seems unnecessary, because the length of such binaries can be easily calculated, and so adding the length at the end wastes space.)

File: oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/stream/NodeLineReader.java (hunk 5 above)
Comment (thomasmueller) [reply oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/stream/NodeLineReader.java]: Using a CountingInputStream seems like a simple solution, I'll try that.

File: oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/analysis/modules/BinarySize.java [hunk 9]
```diff
@@ -0,0 +1,134 @@
...
+        return result;
+    }
+
+    public String toString() {
+        StringBuilder buff = new StringBuilder();
+        buff.append("BinarySize");
+        buff.append(embedded ? " embedded" : " references");
+        buff.append(" in " + unit);
+        buff.append(" (resolution: " + resolution + ")\n");
+        buff.append(getRecords().stream().map(s -> s + "\n").collect(Collectors.joining()));
```
nfsantos, line 129: This method does not seem to be performance critical, but just for coherence, since a StringBuilder is used with the intent to reduce allocations, it would be better in this line to append directly the individual records to the String builder instead of using Collectors.joining() to create a separate String.

BTW, in recent versions of Java (as in Java 11 or later), String concatenation with + (when not done in loops) is often faster than using StringBuilder because of [JEP 280: Indify String Concatenation](https://openjdk.org/jeps/280). Probably does not apply here because there is an implicit loop.

nfsantos (approved):

PR #1249, opened 2023-12-13
OAK-10580 - Support multiple include paths in regex path filtering

nfsantos:
Generalize support for regex path filtering when downloading from Mongo, by also supporting filters when the indexes to be downloaded have multiple values in the `includedPaths` property. Regex path filtering will not be applied only if there are `excludedPaths`.


fabriziofortino (approved):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -353,63 +354,101 @@ private void downloadWithNaturalOrdering() throws InterruptedException, TimeoutE
...
-            return null;
+            return Set.of();
+        }
+        Set<String> includedPaths = new HashSet<>();
+        Set<String> excludedPaths = new HashSet<>();
+        for (PathFilter pathFilter : pathFilters) {
+            includedPaths.addAll(pathFilter.getIncludedPaths());
+            excludedPaths.addAll(pathFilter.getExcludedPaths());
+        }
+        // Sort by length to make it easier to compute the common ancestors of include paths
```
tihom88: Comment - > "Sort by natural order..."

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -353,63 +354,101 @@ private void downloadWithNaturalOrdering() throws InterruptedException, TimeoutE
...
         if (!path.endsWith("/")) {
             path = path + "/";
         }
         String quotedPath = Pattern.quote(path);
         // For entries with path sizes above a certain threshold, the _id field contains a hash instead of the path of
         // the entry. The path is stored instead in the _path field. Therefore, we have to include in the filter also
         // the documents with matching _path.
-        return Filters.or(
+        return Stream.of(
                 regex(NodeDocument.ID, Pattern.compile("^[0-9]{1,3}:" + quotedPath + ".*$")),
```
tihom88, line 426: Is this condition correct.  The getIdFromPath has a different format.
https://github.com/apache/jackrabbit-oak/blob/0b8f4ab2e736c6561ae745a5fe6040a59911eeb3/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java#L291C61-L291C61

nfsantos: Can you elaborate? 

From what I understand, the document id will be one of these:
- depth + ":" + path; (last line of the method that you refer)
- an hash - as returned by `createHashedId(depth, hash, path.getName());`

This is the full line that you quoted:

```
        return Stream.of(
                regex(NodeDocument.ID, Pattern.compile("^[0-9]{1,3}:" + quotedPath + ".*$")),
                regex(NodeDocument.PATH, Pattern.compile(quotedPath + ".*$")
                ));
```

If the id is in the first format, the regex filter on the `_id` will pick it up. If the id is an hash, then the second regex on `_path` will match the document.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
nfsantos: Corrected.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
tihom88: Yeah this condition looks fine. 
Another query:
As we won't be able to filter all documents based on condition1, mongo would have to anyways traverse through all documents, would keeping condition1 make any difference?

nfsantos: We need both conditions. The `_path` field is only present on documents with a "long path", that is, where the `_id` field was replaced by an hash. For instance:

https://github.com/apache/jackrabbit-oak/blob/0b8f4ab2e736c6561ae745a5fe6040a59911eeb3/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeState.java#L446-L459

tihom88 (approved):

PR #1251, opened 2023-12-14
OAK-10584 : catch exception in getOldestRevisionToKeep when in read-o…

stefan-egli:
…nly mode

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Checkpoints.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Checkpoints.java
@@ -151,7 +151,7 @@ class Checkpoints {
                 LOG.debug("Purged {} expired checkpoints", op.getChanges().size());
             } catch (UnsupportedOperationException uoe) {
                 LOG.info("getOldestRevisionToKeep : could not clean up expired checkpoints"
-                        + " due to excepttion : " + uoe, uoe);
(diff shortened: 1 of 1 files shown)


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Checkpoints.java
```diff
@@ -146,8 +146,13 @@ public Revision getOldestRevisionToKeep() {
         }
 
         if (op.hasChanges()) {
-            store.findAndUpdate(Collection.SETTINGS, op);
-            LOG.debug("Purged {} expired checkpoints", op.getChanges().size());
+            try {
+                store.findAndUpdate(Collection.SETTINGS, op);
+                LOG.debug("Purged {} expired checkpoints", op.getChanges().size());
+            } catch (UnsupportedOperationException uoe) {
+                LOG.info("getOldestRevisionToKeep : could not clean up expired checkpoints"
```
rishabhdaim, line 153: It should be `LOG.warn` at least.

stefan-egli: Reason for info was that using this in read-only (eg via oak-run revisions info), this exception is likely met. In which case a warn would be confusing. I think we'd need to expose whether or not a DocumentStore is read-only somehow - without that we're left with catching this exception (which I agree is not nice)

rishabhdaim (approved):

PR #1254, opened 2023-12-19
OAK-10590 - If includedPaths and excludedPaths are specified as a String instead of array of String, interpret them as a one-element array of Strings

nfsantos:
The includedPaths property of an index definition should be an array of strings. But a common mistake made by users is to define it as a String when it has a single element. That is, instead of:
```
        "includedPaths": [ "/a/b"] , 
```
it is defined as:
```
        "includedPaths": "/a/b", 
```

If `includedPaths` is defined as a String, the indexing job would ignore its value and instead default to use the root as the `includedPaths`, which results in downloading the full node store and creating an FFS containing everything (except hidden paths). This will slow down significantly the indexing job, as it will negate any benefits from using regex filtering. And even if regex filtering is not enabled or cannot be used, using `/` as `includedPaths` will also result in the FFS containing more nodes than it should, which will once again slow down the indexing job.

The same is true for `excludedPaths`, but in this case, the default value is an empty list of `excludedPaths`, so it will ignore the value of this property and will not exclude anything. This may result in parts of the node store being indexed that should not be indexed.

The handling of `includedPaths` and `excludedPaths` is also inconsistent with the handling of `queryPaths`, which is being interpreted as a one-element array if it is defined as a String.

This PR makes the logic that reads the `includedPaths` and `excludedPaths` properties more lenient, by treating Strings as one-element arrays and issuing a warning with a suggested fix.

Additionally, the PR makes some minor cleanups in the files that had to be modified (remove usages of Guava and fix some compilation warnings).

### DIFF
--- a/oak-store-spi/src/main/java/org/apache/jackrabbit/oak/spi/filter/PathFilter.java
+++ b/oak-store-spi/src/main/java/org/apache/jackrabbit/oak/spi/filter/PathFilter.java
@@ -21,7 +21,6 @@
 package org.apache.jackrabbit.oak.spi.filter;
 
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.List;
 import java.util.Set;
 
@@ -29,7 +28,6 @@ import org.apache.jackrabbit.oak.api.PropertyState;
 import org.apache.jackrabbit.oak.api.Type;
 import org.apache.jackrabbit.oak.commons.PathUtils;
 import org.apache.jackrabbit.oak.spi.state.NodeBuilder;
-import org.apache.jackrabbit.oak.spi.state.NodeState;
 import org.jetbrains.annotations.NotNull;
 
 import static org.apache.jackrabbit.guava.common.base.Preconditions.checkState;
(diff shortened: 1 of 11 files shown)


oak-store-spi/src/main/java/org/apache/jackrabbit/oak/spi/filter/PathFilter.java
```diff
@@ -164,26 +167,36 @@ public String toString() {
...
+     *
+     * @return the value of the property if the property is set and is of type Strings or String, otherwise the default value
+     */
+    private static Iterable<String> getStringsLenient(NodeBuilder builder, String propertyName, Collection<String> defaultVal) {
         PropertyState property = builder.getProperty(propertyName);
         if (property != null && property.getType() == Type.STRINGS) {
             return property.getValue(Type.STRINGS);
+        } else if (property != null && property.getType() == Type.STRING) {
+            String value = property.getValue(Type.STRING);
+            LOG.warn("Property \"{}\"=\"{}\" has type String but it should be array of String. Proceeding by treating it as a " +
```
thomasmueller: I would probably not log a warning. Instead, I would document that this is also supported.

Reason: we already have many many entries that are defined as string (mostly "/dummy") instead of array.

thomasmueller: I would make it public and add a unit test case for it.

fabriziofortino (approved):

PR #1258, opened 2023-12-26
OAK-10589 - Improve regex path filtering to also handle cases where excludedPaths are defined

nfsantos:
Generalize regex path filtering to also be applied when some indexes also have `excludedPaths` defined. The idea is to use a filter condition like the following:
```
(_id $in [ <list of regex for included paths> ]) AND (_id $nin [<list of regex for excluded paths>])
```
This is just a simplified example, the real condition is more complex because it has to also consider the case of long paths, so it needs to filter both on `_id` and on `_path`.

This filter is also a simplification of the previous version of the filter. Instead of using `$or` on individual regex filters (one for each path included/excluded) instead use the `$in` operator, e.g.: `$in [ <regex p1>, <regex p2>, ... ]`. This format will likely not be faster than the first option but it is shorter and easier to understand.


oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTaskTest.java
```diff
@@ -142,42 +144,101 @@ private List<PathFilter> createIncludedPathFilters(String... paths) {
...
 
         List<PathFilter> multipleIncludeFiltersDifferent2 = createIncludedPathFilters("/a/a1/a2", "/a/a1", "/b", "/c", "/cc");
-        assertEquals(Set.of("/a/a1", "/b", "/c", "/cc"), PipelinedMongoDownloadTask.extractIncludedPaths(multipleIncludeFiltersDifferent2));
+        assertEquals(new MongoFilterPaths(List.of("/a/a1", "/b", "/c", "/cc"), List.of()), PipelinedMongoDownloadTask.buildMongoFilter(multipleIncludeFiltersDifferent2));
+    }
 
-        List<PathFilter> withExcludeFilter = List.of(
-                new PathFilter(List.of("/"), List.of("/var"))
+    @Test
+    public void regexFilters_included_and_excluded_paths() {
```
thomasmueller: This is mixing camel case and snake case... Here we use excluded_paths, in other places it's includedPaths.

I would try to be a bit more consistent. But I understand purely camel case is not nice either, regexFiltersIncludedAndExcludedPaths... but I think still better.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -349,95 +353,132 @@ private void downloadWithNaturalOrdering() throws InterruptedException, TimeoutE
...
             download(childrenIterable);
         }
     }
 
-    private Set<String> getPathsForRegexFiltering() {
+    private MongoFilterPaths getPathsForRegexFiltering() {
         if (!regexPathFiltering) {
             LOG.info("Regex path filtering disabled.");
-            return Set.of();
+            return null;
```
thomasmueller: I wonder if we should have a "empty MongoFilterPaths" instance (e.g. MongoFilterPaths.ALL), instead of null comparisons... You can then still have comparisons:

    if (filter == MongoFilterPaths.ALL) {
        ...
    }

instead of

    if (filter == null) {
        ...
    }

it might be a bit clearer... not sure.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -349,95 +353,132 @@ private void downloadWithNaturalOrdering() throws InterruptedException, TimeoutE
...
-
-        LOG.info("Paths considered for regex filtering. IncludedPaths: {}, excludedPaths: {}", sortedIncludedPaths, excludedPaths);
-        if (sortedIncludedPaths.contains("/")) {
-            LOG.info("Disabling regex path filtering because root path is in the includedPaths: {}", sortedIncludedPaths);
-            return Set.of();
+        // Keep only unique included paths. That is, if paths "/a/b" and "/a/b/c" are both in the list, keep only "/a/b"
+        ArrayList<String> finalIncludedPathsRoots = new ArrayList<>();
+        LOG.debug("Candidate included paths: {}, candidate excluded paths: {}", sortedCandidateIncludedPaths, candidateExcludedPaths);
+        // The quadratic algorithm below could easily be replaced by linear time algorithm, but it's not worth the added
+        // complexity because the list of paths should never be more than a few tens
```
thomasmueller: Features are often used in surprising ways... What if we don't filter if the list is larger than a 20? With 20 times 100, it would be 10'000, which should still be OK. But above that, it might not.

Also, the MongoDB query might become long even if just one path is too very long (not sure if there is a limit). So maybe we should restrict on the length as well.

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
nfsantos: Done.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTaskTest.java
nfsantos: Changed all the methods to be in camel case. It's the convention for Java.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
nfsantos: I added a configuration setting to limit on the number of paths that can be used to build regex path filters. By default, the value is 20. This is limit is applied after merging and de-duplicating the paths in all the path filters, so the algorithm to de-duplicate the paths might still be exposed to the quadratic complexity when using many paths. I tested in my M1 mac, with 10k unique paths it takes two seconds to de-duplicate the list of paths. I think it's ok, it would be very hard to have a index definition with more than 10k paths, probably other parts of the system would break before getting to this point.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoRegexPathFilterFactory.java
```diff
@@ -0,0 +1,149 @@
...
+
+    private final static Logger LOG = LoggerFactory.getLogger(MongoRegexPathFilterFactory.class);
+
+    private final int maxNumberOfPaths;
+
+    public MongoRegexPathFilterFactory(int maxNumberOfPaths) {
+        this.maxNumberOfPaths = maxNumberOfPaths;
+    }
+
+    public MongoFilterPaths buildMongoFilter(List<PathFilter> pathFilters) {
```
amit-jain, line 78: Question: From a quick look it seems that the computation done is largely duplicate of what is done in PathFilter(). Can't we just initialize a new PathFilter by concatenating the includes and excludes from the filters list?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoRegexPathFilterFactory.java
```diff
@@ -0,0 +1,149 @@
...
+        if (pathFilters == null || pathFilters.isEmpty()) {
```
steffenvan, line 79: Do we need to include the root path here, i.e the equivalent of `pathFilters == ["/"]`?

thomasmueller: I see we have a special handling for "includedPaths includes /" below. So that's good.

Just FYI, I only know about one case currently that has this:

    "includedPaths": ["/"],

With the change we did for the index definition generator https://oakutils.appspot.com/generate/index - to always add "includedPaths" - we might get more of those in the future.

So yes, it might make sense to support "/". But it's already supported, so that's good.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoRegexPathFilterFactory.java
```diff
@@ -0,0 +1,149 @@
...
+        // included paths but too many excluded paths. The excluded paths are always children of the included paths, so
+        // if we do not filter on excludedPaths, we simply download everything inside the included paths. This might be
+        // much less than the full repository.
+        if (finalExcludedPaths.size() > maxNumberOfPaths) {
+            LOG.info("Number of excludedPaths ({}) exceed the maximum allowed ({}), disabling filtering of excludedPaths",
+                    finalExcludedPaths.size(), maxNumberOfPaths);
+            finalExcludedPaths = List.of();
+        }
+
+        if (finalExcludedPaths.isEmpty()) {
```
steffenvan, line 137: I know this PR is not supposed to address this, but we do have the recommendation to not have either only `excludedPaths` or `includedPaths`. Not sure if it's helpful, but would logging a warning make sense here?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -101,12 +100,22 @@ public long getDocumentsDownloaded() {
...
     public static final String OAK_INDEXER_PIPELINED_MONGO_REGEX_PATH_FILTERING = "oak.indexer.pipelined.mongoRegexPathFiltering";
     public static final boolean DEFAULT_OAK_INDEXER_PIPELINED_MONGO_REGEX_PATH_FILTERING = false;
+
+    /**
+     * Maximum number of elements in the included/excluded paths list used for regex path filtering. If after
+     * merging and de-deduplication of the paths of all the path filters the number of included or excluded paths exceeds
+     * this value, then disable path filtering to avoid creating Mongo queries with large number of filters
+     */
+    public static final String OAK_INDEXER_PIPELINED_MONGO_REGEX_PATH_FILTERING_MAX_PATHS = "oak.indexer.pipelined.mongoRegexPathFilteringMaxPaths";
+    public static final int DEFAULT_OAK_INDEXER_PIPELINED_MONGO_REGEX_PATH_FILTERING_MAX_PATHS = 20;
```
thomasmueller, line 116: It would be good to verify what is the maximum number of entries that MongoDB supports... I could imagine it's less than 20 includedPath + 20 excludedPaths...

And I wonder if there is a limit on the length of the paths, not just the number of paths.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedIT.java
```diff
@@ -418,14 +515,23 @@ private PipelinedStrategy createStrategy(Backend backend, Predicate<String> path
...
+        contentDamBuilder.child("2022").child("02").child("04").setProperty("p1", "v20220204");
+        contentDamBuilder.child("2022").child("03").setProperty("p1", "v202203");
+        contentDamBuilder.child("2022").child("04").setProperty("p1", "v202204");
         contentDamBuilder.child("2023").setProperty("p2", "v2023");
+        contentDamBuilder.child("2023").child("01").setProperty("p1", "v202301");
+        contentDamBuilder.child("2023").child("01").setProperty("p1", "v202301");
         contentDamBuilder.child("2023").child("02").child("28").setProperty("p1", "v20230228");
 
         // Node with very long name
+//        @NotNull NodeBuilder node = contentDamBuilder.child("long_path_test");
```
thomasmueller:
```suggestion
```

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoRegexPathFilterFactory.java
nfsantos: I added the special handling of '/' recently, mainly to simplify the query in this case. But even before adding this special case, the logic handling the download of subpaths was doing the right thing when the includedPaths was '/', but it would do unnecessary queries to download the ancestors of '/' and would use a filters to match '/' on the query for the children, which is unnecessary. 
But yes, now it is handled and if '/' is one of the includedPath, it will use a simple query without any filters that downloads everything.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
nfsantos: The maximum size of a query is limited by the maximum size of a bson document, which is 16MB. I did not make a test myself, but I found several posts on StackOverflow that say that it is possible to have hundreds of fields in the arrays passed to `$in` and `$nin`. But we should anyway have a much lower limit in the tens because otherwise the performance would degrade.

I'm wondering if we should have hard limits to the number of included/excluded paths. Or just log a warning if there are more than, let's say, 100. I would go with a warning, because there may be valid reasons to have many included/excluded paths and Mongo might actually handle it correctly, so we do not want to restrict our freedom in the future. But a warning would already be useful, as if something would fail, the warning would point to the potential problem. What do you think?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoRegexPathFilterFactory.java
nfsantos: It is similar but not quite the same. For instance, if there is an index with 
```
IndexA
includedPaths=/a
excludedPaths=/a/a1

IndexB
includedPaths=/a, /b
excludedPaths=[]
```
Then `/a/a1` should not be excluded. But if we concatenate the included and excluded paths we get:
```
IndexPathsMerged
includedPaths=/a, /b
excludedPaths=/a/a1
```
which will exclude `/a/a1`.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoRegexPathFilterFactory.java
thomasmueller: We have documentation for having includedPaths (and queryPaths... but that's when running queries). We don't need documentation for excludedPaths in my view.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
thomasmueller: > The maximum size of a query is limited by the maximum size of a bson document, which is 16MB.

Great! 

A warning might make sense, but I don't think it's important to have a warning here. I think it's fine.

thomasmueller (approved):

PR #1262, opened 2024-01-10
OAK-10597 : introduce embedded verification to detailedGC

stefan-egli:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1222,6 +1229,29 @@ public void removeGarbage(final VersionGCStats stats) {
...
             timer.reset().start();
             try {
+                if (embeddedVerification) {
+                    // embedded verification is done completely independent of DocumentStore
+                    // also, it's done irrespective of dry-run or not
+                    final Iterator<UpdateOp> it = updateOpList.iterator();
+                    while(it.hasNext()) {
+                        final UpdateOp update = it.next();
+                        NodeDocument oldDoc = ds.find(Collection.NODES, update.getId());
+                        if (oldDoc == null) {
```
rishabhdaim, line 1239: In case oldDoc is absent `findAndUpdate` won't do anything. So we don't need to return from here, just ignore this.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1222,6 +1229,29 @@ public void removeGarbage(final VersionGCStats stats) {
 
             timer.reset().start();
             try {
+                if (embeddedVerification) {
+                    // embedded verification is done completely independent of DocumentStore
+                    // also, it's done irrespective of dry-run or not
+                    final Iterator<UpdateOp> it = updateOpList.iterator();
+                    while(it.hasNext()) {
+                        final UpdateOp update = it.next();
+                        NodeDocument oldDoc = ds.find(Collection.NODES, update.getId());
```
rishabhdaim, line 1238: I think we should fetch them all in one go. That would be much faster.
Since no such API exists in `DocumentStore`, I suggest we use Mongo APIs directly here.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -924,6 +930,7 @@ public DetailedGC(@NotNull RevisionVector headRevision, long toModifiedMs, @NotN
             this.timer = createUnstarted();
             // clusterId is not used
             this.revisionForModified = Revision.newRevision(0);
+            this.ownHeadRevision = headRevision.getRevision(nodeStore.getClusterId());
```
rishabhdaim, line 933: Why do we need different headRevisions here? We already have a headRevision for the whole root/tree.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1252,6 +1282,37 @@ public void removeGarbage(final VersionGCStats stats) {
                 delayOnModifications(timer.stop().elapsed(MILLISECONDS), cancel);
             }
         }
+
+        private boolean verify(NodeDocument oldDoc, NodeDocument newDoc, UpdateOp update) {
```
rishabhdaim, line 1286: This comparison is doing more than required. In case the oldDoc is null, we should ignore it and move forward cause that is already taken care of by Mongo.

Mongo only updates if it finds the document else, it ignores the update operation.

rishabhdaim (changes requested):
I think we can even reuse the output of ` versionStore.getModifiedDocs` in `collectDetailedGarbage()` and pass the documents to `collectGarbage()` rather than querying the database again.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
stefan-egli: agree that a bulk `find` would be handy. In this case it would mostly be reading from the `nodesCache` though, as detailedGC just needed the oldDoc as well. So it's mostly not going to read from mongo. 
What about splitting bulk API into a separate ticket if we see it is necessary?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
stefan-egli: One is a RevisionVector, the other is just the Revision of the own clusterId of the same headRevision. It is needed for `getNodeAtRevision`

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
stefan-egli: good find, fixed in https://github.com/apache/jackrabbit-oak/pull/1262/commits/a83aee1a702b2dade42a18a37a683f0c72705fdf

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
stefan-egli: the comparison is at the `DocumentNodeState` level though, not `NodeDocument`. oldDoc and newDoc are both not null here

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1252,6 +1282,37 @@ public void removeGarbage(final VersionGCStats stats) {
...
+            if (oldNS == null && newNS == null) {
+                // both don't exist - fine, that's considered equal
+                return true;
+            } else if ((oldNS == null && newNS != null)
+                    || (oldNS != null && newNS == null)) {
+                // failure : one is deleted/missing, the other not
+                log.error("removeGarbage.verify : failure in DetailedGC"
+                        + " with id : {}, oldNS exists : {}, newNS exists: {}, update: {}",
+                        oldDoc.getId(), oldNS == null, newNS == null, update);
+                return false;
```
rishabhdaim, line 1303: In this logic if `oldNs` is null then we should ignore and return true.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
rishabhdaim: What I meant is that https://github.com/apache/jackrabbit-oak/pull/1262/files#diff-e47c4905644873320ba8f5ed72096b1c6b2ad3fdec6d51241b27a413c371562aR1297-R1303

In the above code, if `oldNs` is null then we should ignore and return true.

stefan-egli:
>  think we can even reuse the output of  versionStore.getModifiedDocs in collectDetailedGarbage() and pass the documents to collectGarbage() rather than querying the database again.

would be possible, agreed. would require to replace the iterator approach with keeping the underlying, full list. possible. not sure if we'd gain a lot though, since MongoDocumentStore has the nodesCache, so we're not fetching again from mongo, most of the time. I think the current iterator approach works somewhat better with larger lists / documents where eg there isn't really much garbage.

rishabhdaim (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
stefan-egli: can you elaborate why?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
stefan-egli: (see my comment/question in other thread at https://github.com/apache/jackrabbit-oak/pull/1262#discussion_r1448755387 )

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
rishabhdaim: I thought _because null would mean that the document has been deleted and in that case, it doesn't make sense to compare them cause it has already been gone._

Now I think you are right, deleted document should be null for both old & new.

stefan-egli: Yes, I think we should compare for all variations there. The _node_ can be deleted in the old or new state - but it has to be the same, otherwise `removeGarbage` cleaned up some wrong garbage

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
stefan-egli: (I'd probably avoid creating a jira ticket already now though, only once we find out it is a performance problem)

rishabhdaim (approved):

PR #1265, opened 2024-01-11
[SITES-18348] Add conflict handler for cq:lastRolledout property

alexandru-stancioiu:


anchela (changes requested):
hi @alexandru-stancioiu , as the ticket reference suggests this is an Adobe AEM specific extension that does not make too much sense in the context of Oak.

i would suggest to create a patch against the AEM component that is in charge of creating the Oak repository with AEM specific configuration and setup values.
hope that helps

alexandru-stancioiu:
@anchela makes sense, closing PR.

PR #1267, opened 2024-01-15
OAK-10592 - Add support to specify a custom regex to exclude documents from being downloaded from Mongo during indexing

nfsantos:
Adds the following configuration property:

- `oak.indexer.pipelined.mongoCustomExcludeEntriesRegex` - Documents whose paths match this regex will not be downloaded from Mongo by the indexing job.
 
For instance, the regex: `"/metadata.xml$|/renditions/.*.jpg$"` would exclude any nodes whose names end in `/metadata.xml` or `/renditions/.*.jpg`

This can be used to exclude certain nodes that are not indexed even though they are in the `includedPaths` of the index definition.


fabriziofortino (approved):

thomasmueller (commented):
I would probably make the "private static" methods (those that return Bson) public and add unit test for them. That would help with code coverage, and also help developers that want to understand what the code is doing.

nfsantos:
> I would probably make the "private static" methods (those that return Bson) public and add unit test for them. That would help with code coverage, and also help developers that want to understand what the code is doing.

I have added several unit tests for these methods.

thomasmueller (approved):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -129,6 +133,117 @@ public long getDocumentsDownloaded() {
 
     private static final String THREAD_NAME = "mongo-dump";
 
+    /**
+     * Creates the filter to be used in the Mongo query
+     *
+     * @param mongoFilterPaths         The paths to be included/excluded in the filter. These define subtrees to be included or excluded.
+     *                                 (see {@link MongoFilterPaths} for details)
+     * @param customExcludedPathsRegex A custom regex pattern to exclude paths from the download.
```
thomasmueller: I would add:

    This is typically a _suffix_, for example "/metadata.xml$|/renditions/.*.jpg$".
    To exclude subtrees such as /content/abc, use mongoFilterPaths instead.

PR #1271, opened 2024-01-15
OAK-10607 Rename Maven property "java.version"

kwin:
This prevents hiding the same named Java system property


reschke (approved):
Not tested; but looks good to me.

PR #1273, opened 2024-01-18
OAK-10608 - [Indexing job] Improve regex expression used to download from Mongo to make better used of Mongo indexes

nfsantos:
The current regex expression used to filter from Mongo the included/excluded paths has conditions on both the fields `_id` and `_path`. In most cases, the `_id` field contains the path of the node, but when the path is too long, the `_id` is replaced by an hash of the path and the full path is added to the document as an additional `_path` field. For these cases, the regex expression must also check the `_path` field.

When running an ordered traversal, we use a Mongo index on `(_modified, _id)`. So checks on `_id` can be done with just the data retrieved from the index. But for the check on `_path`, Mongo needs to read the full document from the column store, which slows down significantly the traversal.

Currently, if `_id` does not match, the regex expression will always check `_path`, forcing a retrieval of the document. But we only need to check `_path` if the `_id` is of the form of a long path id, that is, of the pattern `4:h<hash>...`, otherwise, if the `_id` is not a long path, then if it does not match the regex, we can be sure that the document is not needed. The check that `_id` is an hash can be done without retrieving the full document from the column store, so it will be fast. And in the common case, the document is not a long path, so this simple check will avoid retrieving the document from the column store.

This optimization will have a big impact when the regex expression matches a small fraction of the repository. In the current implementation, Mongo has to traverse both the index and the column store for all possible regex filters. But with the additional check for long paths, Mongo has still to traverse the full index but it will only retrieve from the column store the documents that match the filter or the long path documents. And since the index is much smaller than the column store and can more easily be cached, this will significantly improve performance.

In a test environment, with an index job that downloads 12M out of 230M mongo documents, I got the following results:
- regex filtering, without long path optimization: 1h45
- regex filtering with long path optimization: 0h45m

### DIFF
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
@@ -143,15 +143,15 @@ public class PipelinedMongoDownloadTask implements Callable<PipelinedMongoDownlo
     /**
      * Creates the filter to be used in the Mongo query
      *
-     * @param mongoFilterPaths         The paths to be included/excluded in the filter. These define subtrees to be included or excluded.
-     *                                 (see {@link MongoFilterPaths} for details)
+     * @param mongoFilterPaths          The paths to be included/excluded in the filter. These define subtrees to be included or excluded.
+     *                                  (see {@link MongoFilterPaths} for details)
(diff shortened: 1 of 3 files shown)


steffenvan:
Good description and was easy to get an overview of the problem. Looks good to me 👍

fabriziofortino (approved):

thomasmueller:
We are adding more and more document store logic here. I think it's best of the document store team could review the PR.

Joscorbe (approved):
+1 It looks good to me.

Thanks for the detailed explanation, it helped a lot to fully understand the objective. The performance gains seem promising, it will depend in the amount of long paths in the repository, but in general we should see an improvement in almost any case.

PR #1274, opened 2024-01-18
OAK-10613: Set Java version in Jenkinsfile to 17 (for SonarCloud)

reschke:


Jenkinsfile
```diff
@@ -56,7 +56,7 @@ def buildModule(moduleSpec) {
             '''
             timeout(70) {
                 checkout scm
-                withEnv(["Path+JDK=$JAVA_JDK_11/bin","Path+MAVEN=$MAVEN_3_LATEST/bin","JAVA_HOME=$JAVA_JDK_11","MAVEN_OPTS=-Xmx1536M"]) {
+                withEnv(["Path+JDK=$JAVA_JDK_17/bin","Path+MAVEN=$MAVEN_3_LATEST/bin","JAVA_HOME=$JAVA_JDK_17","MAVEN_OPTS=-Xmx1536M"]) {
```
kwin, line 59: Usually you want to separate checking with SonarCloud from the actual build. Look at https://github.com/apache/sling-tooling-jenkins/blob/fa79d93ce00c6f52fafaad348a04d2c0757460ed/vars/slingOsgiBundleBuild.groovy#L234C15-L234C15 or https://github.com/apache/jackrabbit-filevault-jenkins-lib/blob/7b2e2f0fce955141c1bc62ebb4a1ba36f6ba7c62/vars/vaultStageBuild.groovy#L69. That way different JDKs can be used for those.
I think building and testing with just JDK17 is probably not enough (if we state that we support building with Java11+, https://github.com/apache/jackrabbit-oak?tab=readme-ov-file#getting-started)

reschke: ack, thanks

so what can we do to get the SonarCloud check working again?

kwin: Actually SonarQube is executed by GitHub Actions not ASF Jenkins.

PR #1277, opened 2024-01-22
OAK-10616: Make error messages from o.a.j.o.namepath.JcrNameParser/Jc…

mbaedke:
…rPathParser consistent and less misleading

Unified error messages from JcrPathParser and JcrNameParser.


reschke (changes requested):
I find it problematic to make the error message prefix part of the API.

I understand that this so that tests and actual code do not close coupling, but IMHO it would be completely acceptable to just hardwire the copy of the prefix in the string class.

reschke (approved):

PR #1278, opened 2024-01-22
OAK-10622 : oak.documentMK.reuseDelayAfterRecoverySecs and recoveryTime…

stefan-egli:
… added

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/ClusterNodeInfo.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/ClusterNodeInfo.java
@@ -88,9 +88,9 @@ public class ClusterNodeInfo {
     public static final String LEASE_END_KEY = "leaseEnd";
 
     /**
-     * The time a recover was done, if the last shutdown required a recover - not set otherwise.
+     * The time a recovery was done, if the last shutdown required a recover - not set otherwise.
      */
-    public static final String RECOVER_TIME_KEY = "recoverTime";
+    public static final String RECOVERY_TIME_KEY = "recoveryTime";
 
     /**
      * The start time.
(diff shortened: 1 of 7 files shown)


Joscorbe (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/ClusterNodeInfo.java
```diff
@@ -615,6 +650,18 @@ private static Map.Entry<ClusterNodeInfo, Long> createInstance(DocumentStore sto
             // if we get here the cluster node entry is inactive. if recovery
             // was needed, then it was successful
 
+            if (reuseAfterRecoveryMillis > 0) {
+                Long lastRecoveryTime = (Long) doc.get(RECOVERY_TIME_KEY);
```
rishabhdaim, line 660: We are setting `RECOVERY_TIME_KEY` as null in many places and this could lead to an NPE in case it is null here.

I would suggest using default value in place of null when we are setting `RECOVERY_TIME_KEY` or doing a null check here before casting to `Long`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/ClusterNodeInfo.java
```diff
@@ -229,6 +234,12 @@ static RecoverLockState fromString(String state) {
     /** OAK-3399 : max number of times we're doing a 1sec retry loop just before declaring lease failure **/
     private static final int MAX_RETRY_SLEEPS_BEFORE_LEASE_FAILURE = 5;
 
+    /** OAK-10622 : seconds to prevent reuse of clusterId if it crashed and thus required a recover */
+    private static final int DEFAULT_REUSE_AFTER_RECOVERY_SECS = SystemPropertySupplier.create("oak.documentMK.reuseDelayAfterRecoverySecs", -1)
+            .loggingTo(LOG).validateWith(value -> value >= -1)
+            .formatSetMessage((name, value) -> String.format("Reuse delay after recovery set to (secs): %ss (using system property %s)", name, value)).get();
+    public static final int DEFAULT_REUSE_AFTER_RECOVERY_MILLIS = 1000 * DEFAULT_REUSE_AFTER_RECOVERY_SECS;
```
rishabhdaim: Scope could be reduced to `default` since it is only used in `ClusterNodeInfoTest`.

rishabhdaim (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/ClusterNodeInfo.java
stefan-egli: If the document doesn't contain a `"recoveryTime"` key or the value of that key is `null`, then this will result in `lastRecoveryTime` here being null. No NPE is thrown. 

The intent was:
* `"recoveryTime"` is null (non existent) in legacy cases : in this case no delay should apply (hence it skips the block below)
* `"recoveryTime"` is null (non existent) after a graceful shutdown : in these cases also no delay should apply
* `"recoveryTime"` is set to an actual value at the end of a recovery : only in this case should the block below be handled

All of these cases should be handled by the null check in the next line, or am I missing something?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/ClusterNodeInfo.java
stefan-egli: +1, will reduce it

stefan-egli: done at https://github.com/apache/jackrabbit-oak/pull/1278/commits/697668209d86295004cde06ccaf3bd9c8585fdd3

rishabhdaim:
On second thought, IIRC we discussed that we would try to use Osgi Configs instead of System properties.
Using Osgi Config here instead of the System property would give us more flexibility in changing the value and rolling out changes. wdyt ?

stefan-egli:
Good point about osgi config vs system properties. And I think we might have some inconsistent usages of both of them in various places throughout oak. Given DocumentNodeStore is not supporting restarts or dynamic reconfigurations I don't think it makes one or the other more flexible. But at least osgi configs are cleaner in that they don't make the command line longer.

stefan-egli:
@rishabhdaim done at https://github.com/apache/jackrabbit-oak/pull/1278/commits/20e5481e6872518f7c89e18d263afe56ee8e15de

rishabhdaim (approved):

PR #1280, opened 2024-01-23
OAK-10615 - Azure Service Principal Support in oak-run segment-copy, compact, console

dulceanu:
First un-tested implementation

### DIFF
--- a/oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/tool/ToolUtilsTest.java
+++ b/oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/tool/ToolUtilsTest.java
@@ -21,10 +21,14 @@ package org.apache.jackrabbit.oak.segment.azure.tool;
 import com.microsoft.azure.storage.StorageCredentials;
 import com.microsoft.azure.storage.StorageCredentialsAccountAndKey;
 import com.microsoft.azure.storage.StorageCredentialsSharedAccessSignature;
+
+import java.net.URISyntaxException;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
+import com.microsoft.azure.storage.StorageException;
+import com.microsoft.azure.storage.blob.CloudBlobDirectory;
 import org.apache.jackrabbit.oak.blob.cloud.azure.blobstorage.AzuriteDockerRule;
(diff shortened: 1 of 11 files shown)


oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/tool/ToolUtilsTest.java
```diff
@@ -18,59 +18,95 @@
...
         StorageCredentialsAccountAndKey credentials = expectCredentials(
             StorageCredentialsAccountAndKey.class, 
-            () -> ToolUtils.createCloudBlobDirectory(SEGMENT_STORE_PATH, environment)
+            () -> ToolUtils.createCloudBlobDirectory(DEFAULT_SEGMENT_STORE_PATH, environment),
+            DEFAULT_CONTAINER_URL
         );
-        
-        assertEquals("myaccount", credentials.getAccountName());
+
+        assertTrue(checkLogContainsMessage(AZURE_SECRET_KEY_WARNING, logAppender.list.stream().map(ILoggingEvent::getFormattedMessage).collect(Collectors.toList())));
```
smiroslav, line 89: I am not in favor of this type of check. Instead, I would prefer to examine the concrete class returned (subtype of StorageCredentials) or use another method for verification.

dulceanu: Ack. I tried now to tweak it via concrete class returned, as you mentioned, but that would be irrelevant IMO, since we already know that we expect `StorageCredentialsAccountAndKey`. To me, checking the logs for the warning is still valuable, as this log could witness later deprecated authentication method used.
We use the same mechanism here: [DefaultSegmentWriterTest#testHugeMapRecordErrorSize](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-segment-tar/src/test/java/org/apache/jackrabbit/oak/segment/DefaultSegmentWriterTest.java#L326-L327).

smiroslav: Thanks for the additional info. 
In another task, we can perhaps evaluate other options to perform verification of such nature.

smiroslav (approved):

PR #1283, opened 2024-01-23
OAK-10544: oak-jcr: remapping a namespace prefix leaves namespace resolver in broken state

reschke:
...based on @mreutegg 's suggestion to fix this on oak-jcr.

I don't believe we'd need a feature toggle for this.

Feedback appreciated.

### DIFF
--- a/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionNamespaces.java
+++ b/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionNamespaces.java
@@ -90,6 +90,8 @@ public class SessionNamespaces extends LocalNameMapper {
         local.put(prefix, uri);
 
         // make sure the previously mapped URI has a prefix
+        // (getNamespacePrefix has the side effect of generating
+        // a new prefix if none was found, and adding that to the local mapping)
         if (previouslyMappedUri != null) {
             getNamespacePrefix(previouslyMappedUri);
         }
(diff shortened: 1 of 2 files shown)


oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionNamespaces.java
```diff
@@ -86,6 +88,11 @@ synchronized void setNamespacePrefix(String prefix, String uri)
 
         // add the new mapping
         local.put(prefix, uri);
+
+        // make sure the previously mapped URI has a prefix
+        if (previouslyMappedUri != null) {
+            getNamespacePrefix(previouslyMappedUri);
```
thomasmueller, line 96: So this works because getNamespacePrefix will put put a new prefix in the local mapping, if not found.

The behavior of getNamespacePrefix is not changed (always was like that). And I assume setNamespacePrefix was not called a lot. So I think this change has a low risk.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionNamespaces.java
```diff
@@ -86,6 +88,11 @@ synchronized void setNamespacePrefix(String prefix, String uri)
 
         // add the new mapping
         local.put(prefix, uri);
+
+        // make sure the previously mapped URI has a prefix
```
thomasmueller, line 92:
```suggestion
        // make sure the previously mapped URI has a prefix
        // (getNamespacePrefix has the side effect of generating
        // a new prefix if none was found, and adding that to the local mapping)
```

thomasmueller (changes requested):
I find it very weird that adding two "get" methods will make the test pass, so I think it makes sense to add a comment to explain what is going on.

PR #1287, opened 2024-01-25
OAK-10632 :  make embedded verification configurable in dryRun mode

rishabhdaim:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
```diff
@@ -642,7 +642,7 @@ public DocumentNodeStore(DocumentNodeStoreBuilder<?> builder) {
         this.branches = new UnmergedBranches();
         this.asyncDelay = builder.getAsyncDelay();
         this.versionGarbageCollector = new VersionGarbageCollector(
-                this, builder.createVersionGCSupport(), isDetailedGCEnabled(builder), false);
+                this, builder.createVersionGCSupport(), isDetailedGCEnabled(builder), false, false);
```
stefan-egli, line 645: Is there a particular reason for switching the default to `false`? I would have preferred to keep it at `true`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -153,13 +153,14 @@ public class VersionGarbageCollector {
...
-                            final boolean isDetailedGCDryRun) {
+                            final boolean isDetailedGCDryRun,
+                            final boolean embeddedVerification) {
         this.nodeStore = nodeStore;
         this.versionStore = gcSupport;
         this.ds = gcSupport.getDocumentStore();
         this.detailedGCEnabled = detailedGCEnabled;
         this.isDetailedGCDryRun = isDetailedGCDryRun;
-        this.embeddedVerification = true; //TODO: make this "configurable"
+        this.embeddedVerification = embeddedVerification; //TODO: make this "configurable" for actual detailedGC
```
stefan-egli, line 163: detail: this TODO could now also move to DocumentNodeStore (assuming it will have to be adjusted)

stefan-egli (changes requested):
I would keep the default to `true`, otherwise lgtm.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
rishabhdaim: The only reason, I didn't move this to `DocumentNodeStore` is because of the https://issues.apache.org/jira/browse/OAK-10633 ticket. This ticket would eventually remove this TODO once finished.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
rishabhdaim: I agree that `true` makes more sense here and would use that value while making code changes for https://issues.apache.org/jira/browse/OAK-10633 ticket.

stefan-egli (approved):

PR #1294, opened 2024-02-05
OAK-10633 :  make embedded verification configurable in detailedGC

rishabhdaim:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilder.java
```diff
@@ -138,4 +138,26 @@ public RDBDocumentNodeStoreBuilder setDocStoreDetailedGCFeature(@Nullable Featur
     public Feature getDocStoreDetailedGCFeature() {
         return null;
     }
+
+    @Override
+    public boolean isEmbeddedVerificationEnabled() {
+        return false;
```
stefan-egli, line 144: what was again the reason why detailedGC is not supported for RDB?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
```diff
@@ -185,7 +186,12 @@ public class DocumentNodeStoreService {
     /**
      * Feature toggle name to enable detailed GC for Mongo Document Store
      */
-    private static final String FT_NAME_DEATILED_GC = "FT_DETAILED_GC_OAK-10199";
+    private static final String FT_NAME_DETAILED_GC = "FT_DETAILED_GC_OAK-10199";
+
+    /**
+     * Feature toggle name to enable embedded verification for detailed GC mode for Mongo Document Store
+     */
+    private static final String FT_NAME_EMBEDDED_VERIFICATION = "FT_EMBEDDED_VERIFICATION_OAK-10633";
```
stefan-egli, line 194: Starting to think that we might have to revisit our use of feature toggle for DocumentNodeStore (besides the fact that we still have legacy system properties) : it looks like we more and more opt to not only introduce new config options, but we also make them increasingly available as feature toggle. Which makes our list of feature toggles at some point fairly large.

Not saying that we shouldn't introduce this one - just noting that we make excessive use of it lately.

stefan-egli (approved):
looks good to me in general, just added a question and a comment for discussion (but that shouldn't block the PR).

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilder.java
rishabhdaim: Please find the reason here : https://issues.apache.org/jira/browse/OAK-10199?focusedCommentId=17751662&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17751662

stefan-egli: thx - (I think we also introduced a new method to DocumentStore that DetailedGC requires)

PR #1296, opened 2024-02-06
Issue/oak 10636

smiroslav:
https://issues.apache.org/jira/browse/OAK-10636


dulceanu:
@smiroslav, I will take a look at the changes asap.

ahanikel:
@smiroslav If I understand correctly, while the segment containing the blob reference is recovered, the file store tries to load that very segment uncached and fails. Instead of creating a separate Map, can't we just put the recovered segment in the segment cache, like this?
```modified   oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/AbstractFileStore.java
@@ -233,6 +233,7 @@ public abstract class AbstractFileStore implements SegmentStore, Closeable {
         w.recoverEntry(msb, lsb, data, 0, data.length, generation);
         if (SegmentId.isDataSegmentId(lsb)) {
             Segment segment = new Segment(tracker, segmentReader, tracker.newSegmentId(msb, lsb), buffer);
+            segmentCache.putSegment(segment);
             populateTarGraph(segment, w);
             populateTarBinaryReferences(segment, w);
         }
```

smiroslav:
@ahanikel, that actually might work as well and requires minimal code change.

smiroslav:
>  Instead of creating a separate Map, can't we just put the recovered segment in the segment cache, like this?


Hi @ahanikel 

The problem could be if the FileStore is started with a segment cache that is not big enough to contain all recovered segments. 

Another issue could be if the segment is not in the tar archive that is being recovered. To find the segment in that case, we need TarReader to check other archives since we will get a cache miss.

ahanikel:
> Another issue could be if the segment is not in the tar archive that is being recovered. To find the segment in that case, we need TarReader to check other archives since we will get a cache miss.

@smiroslav I think that shouldn't happen because the blob reference should be either in the same segment or in a segment written before, which should also be in the cache.

But I agree that the cache size is a problem. It would need to keep not only the segments of the current tar file but of all the tar files. But then, that is also true for a separate Map --- wouldn't that Map then grow exceptionally big for large repositories?

Can we perhaps find a way to initialise `FileStore#tarFiles` without the recovery and run the recovery process after that?

smiroslav:
>  But then, that is also true for a separate Map --- wouldn't that Map then grow exceptionally big for large repositories?

No, that map contains only segments from the archive which  is being recovered. If segment is in another tar archive, ```tarFiles.readSegment``` is used to find it. 

https://github.com/apache/jackrabbit-oak/blob/82021c78d62dc031aaf7c72345eaa55885108026/oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/AbstractFileStore.java#L294-L294

ahanikel:
But isn't `tarFiles` still "null" at that point, causing the NPE?

smiroslav:
> But isn't tarFiles still "null" at that point, causing the NPE?

Not after the changes I made in this PR. More specifically, initializing tar readers in the ```init``` method, instead of doing it in the constructor
https://github.com/apache/jackrabbit-oak/blob/issue/OAK-10636/oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/tar/TarFiles.java#L406

ahanikel:
Ah, I see, thanks! LGTM :-)

ahanikel (approved):

PR #1301, opened 2024-02-08
OAK-10281 : draft of handling recoveryDelayMillis directly as paramet…

stefan-egli:
…er in isRecoveryNeeded


stefan-egli:
@rishabhdaim , this (which is actually part of OAK-10651) was created as part of an [earlier discussion](https://github.com/apache/jackrabbit-oak/pull/1288#pullrequestreview-1849297188) - question now is : how to move forward with this. Not sure we need to keep this PR open - we could as well close it but not delete the branch and have still have the code around for a while (until we'd prune branches). Wdyt?

stefan-egli:
Closing this to indicate that this was a draft and is currently not being worked upon. It can always be picked up again - the underlying branch won't be deleted.

PR #1303, opened 2024-02-09
OAK-10642: Add tests for operations on very large ordered collections

mbaedke:
Testing repeated reordering in a long running transaction.


mbaedke:
3 new tests added, 2 of them are expected to fail when run in oak-jcr with 
` mvn -DargLine="-Xmx4g" -Dupdate.limit=50 -Dnsfixtures="DOCUMENT_NS" clean install -Dtest=ManyChildrenIT -Djacoco.skip
`
The move test should succeed, but will take very long.

PR #1313, opened 2024-02-20
OAK-10656: MongoDocumentStore: keep metrics about document size related exceptions

reschke:
...we can later expand this to cover different exceptions.

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
@@ -1,4 +1,4 @@
-/*
+/*c
  * Licensed to the Apache Software Foundation (ASF) under one or more
  * contributor license agreements.  See the NOTICE file distributed with
  * this work for additional information regarding copyright ownership.
@@ -34,6 +34,7 @@ import java.util.Objects;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.Lock;
 
 import org.apache.jackrabbit.guava.common.base.Optional;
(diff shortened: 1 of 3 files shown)


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -2317,10 +2320,19 @@ private <T extends Document> DocumentStoreException handleException(Throwable ex
                 invalidateCache(collection, id);
             }
         }
+
+        if (ex instanceof MongoWriteException) {
```
stefan-egli, line 2324: Wondering if `MongoWriteException` is indeed the one we like to count.

Looking at the [MongoServerException](https://mongodb.github.io/mongo-java-driver/3.12/javadoc/com/mongodb/MongoServerException.html) it looks like there are bulk siblings that we might also be interested : `BulkWriteException`, `MongoBulkWriteException` ?

If that's the case, then I'm wondering if we should count `MongoServerException` instead - but arguably that's less specific than the current `MongoWriteException` ...

reschke: I believe I checked, but will double-check again.

At the end of the day, we may want to count multiple types of exceptions, but I wanted to keep this one focused.

stefan-egli (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
stefan-egli: +1 to go ahead with this one and then consider the bulk ones later

mbaedke (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
reschke: I verified that this is the right one; see https://issues.apache.org/jira/browse/OAK-10642?focusedCommentId=17819206&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17819206

reschke: So proposal is to go ahead with this and expand this later on when we have more time.

PR #1314, opened 2024-02-21
OAK-10657: shrink in-DB documents after updates fail due to 16MB limit

reschke:
Just a proof-of-concept.

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
@@ -279,8 +279,13 @@ public class Utils {
 
     /**
      * Produce an {@link UpdateOp} suitable for shrinking branch revision entries for given property in {@link Document}, {@code null} otherwise.
+     * 
+     * @param doc document to inspect for repeated branch commits
+     * @param propertName property to check for
+     * @param revisionChecker filter for revisions (for instance, to check for cluster id)
+     * @return {@link UpdateOp} suitable for shrinking document, {@code null} otherwise
      */
-    public static @Nullable UpdateOp getShrinkOp(Document doc, String propertyName) {
+    public static @Nullable UpdateOp getShrinkOp(Document doc, String propertyName, Predicate<Revision> revisionChecker) {
         Object t_bc = doc.get("_bc");
         Object t_property = doc.get(propertyName);
         if (t_bc instanceof Map && t_property instanceof Map) {
@@ -291,9 +296,11 @@ public class Utils {
             List<Revision> revs = new ArrayList<>();
             for (Map.Entry<Revision, String> en : pMap.entrySet()) {
                 Revision r = en.getKey();
-                String bcv = _bc.get(r);
-                if ("true".equals(bcv)) {
-                    revs.add(r);
+                if (revisionChecker.apply(r)) {
+                    String bcv = _bc.get(r);
+                    if ("true".equals(bcv)) {
+                        revs.add(r);
+                    }
                 }
             }
             // sort by age
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
@@ -280,7 +280,7 @@ public class Utils {
     /**
      * Produce an {@link UpdateOp} suitable for shrinking branch revision entries for given property in {@link Document}, {@code null} otherwise.
      */
-    public static @Nullable UpdateOp getShrinkOp(Document doc, String propertyName, String replacementValue) {
+    public static @Nullable UpdateOp getShrinkOp(Document doc, String propertyName) {
         Object t_bc = doc.get("_bc");
         Object t_property = doc.get(propertyName);
         if (t_bc instanceof Map && t_property instanceof Map) {
@@ -314,7 +314,7 @@ public class Utils {
             for (Revision r : revs) {
                 if (last != null) {
                     if (last.getClusterId() == r.getClusterId()) {
-                        clean.setMapEntry(propertyName, last, replacementValue);
+                        clean.removeMapEntry(propertyName, last);
                     }
                 }
                 last = r;
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
@@ -26,6 +26,7 @@ import java.sql.Timestamp;
 import java.time.Instant;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.Iterator;
@@ -46,6 +47,7 @@ import org.apache.jackrabbit.oak.commons.StringUtils;
 import org.apache.jackrabbit.oak.plugins.document.ClusterNodeInfo;
 import org.apache.jackrabbit.oak.plugins.document.ClusterNodeInfoDocument;
 import org.apache.jackrabbit.oak.plugins.document.Collection;
+import org.apache.jackrabbit.oak.plugins.document.Document;
 import org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreBuilder;
 import org.apache.jackrabbit.oak.plugins.document.DocumentStore;
 import org.apache.jackrabbit.oak.plugins.document.DocumentStoreException;
@@ -54,6 +56,7 @@ import org.apache.jackrabbit.oak.plugins.document.Path;
 import org.apache.jackrabbit.oak.plugins.document.Revision;
 import org.apache.jackrabbit.oak.plugins.document.RevisionVector;
 import org.apache.jackrabbit.oak.plugins.document.StableRevisionComparator;
+import org.apache.jackrabbit.oak.plugins.document.UpdateOp;
 import org.apache.jackrabbit.oak.spi.toggle.Feature;
 import org.apache.jackrabbit.oak.stats.Clock;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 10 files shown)


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -274,6 +277,61 @@ private static String diagsForEntry(Map.Entry<String, PropertyStats> member) {
...
 
+    /**
+     * Produce an {@link UpdateOp} suitable for shrinking branch revision entries for given property in {@link Document}, {@code null} otherwise.
+     * 
+     * @param doc document to inspect for repeated branch commits
+     * @param propertName property to check for
+     * @param revisionChecker filter for revisions (for instance, to check for cluster id)
+     * @return {@link UpdateOp} suitable for shrinking document, {@code null} otherwise
+     */
+    public static @Nullable UpdateOp getShrinkOp(Document doc, String propertyName, Predicate<Revision> revisionChecker) {
```
stefan-egli, line 301: If `doc` could be a `NodeDocument` instead of the generic `Document` some of those `instanceof` could be avoided...

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -274,6 +277,61 @@ private static String diagsForEntry(Map.Entry<String, PropertyStats> member) {
...
+                Revision r = en.getKey();
+                if (revisionChecker.apply(r)) {
+                    String bcv = _bc.get(r);
+                    if ("true".equals(bcv)) {
+                        revs.add(r);
+                    }
+                }
+            }
+            // sort by age
+            Collections.sort(revs, new Comparator<Revision>() {
```
stefan-egli, line 320: wondering if there isn't such a `Comparator` in oak land already?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -274,6 +278,73 @@ private static String diagsForEntry(Map.Entry<String, PropertyStats> member) {
...
+            UpdateOp clean = new UpdateOp(doc.getId(), false);
+            Revision last = null;
+            for (Revision r : revs) {
+                if (last != null) {
+                    if (last.getClusterId() == r.getClusterId()) {
+                        clean.removeMapEntry(propertyName, last);
+                    }
+                }
+                last = r;
+            }
```
stefan-egli, line 341: * this seems to be a bit a more broader GC than expected. It deletes basically all older revisions of the same clusterId. That seems fine - but at that point I'm wondering why restrict it to only branch commits - i.e why specifically only branch commits. Which would lead to the assumption that the idea was to only remove "overwritten branch commits"?
* is the clusterId check here still necessary given the predicate check introduced?
* this doesn't take the usual 24h GC max time nor active checkpoints into account.

stefan-egli: PS:
* it doesn't take late-writes into account (i.e. where traversed state isn't equal head state)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
reschke: I *think* we can do that once we use that from NodeDocumentStore, not DocumentStore...

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
reschke: Not in Revision, as far as  I can tell. I wanted a comparator that sorts by clusterId first; we may not need this if we always filter by cluster id though.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
reschke: - all older revisions that are branch commits (filtered earlier in the code); is there more that we can check? That would require _revisions, but those might be in a different document, right?
- clusterid check - yes, unless we guarantee that the predicate will filter by id
- ack - suggestions?

stefan-egli: some ideas:
* we could restrict it to "overwritten unmerged branch commits" : those we know are garbage whatsoever
* but that might still leave us with a too large doc/prop. we could then try to get the traversed state from "24h ago or the oldest checkpoint" (whichever is older) - then delete anything older than that
* but that still might leave us with a too large doc/prop. then we might have to do an impromptu split and move anything younger than the previous into a split doc ...

stefan-egli: those 3 cases could be .. test cases .. :)

stefan-egli: regarding

> get the traversed state from "24h ago or the oldest checkpoint" (whichever is older)

that might actually be a tricky thing to achieve - and I believe we might not have done that properly in the `DetailedGC` effort so far. I think we might need an actual checkpoint that corresponds to "24h ago"

stefan-egli: > I think we might need an actual checkpoint that corresponds to "24h ago"

... or maybe not a physical checkpoint, but a root revision that corresponds to reading 24h ago : which we might substitute with corresponding revisions (with timestamp 24h minus 1 millisecond) for each known clusterId ... or something like that ...

stefan-egli: > we might not have done that properly in the DetailedGC effort so far

... taking that back .. the difference between `DetailedGC` and this `runtime GC` case here is : in `DetailedGC` we're only looking at documents that have not been modified for 24+ hours. That means, reading their traversed state with headRevision of "now" is fine. But in this `runtime GC` case here, that is not fine (as we need to respect those 24+ hours worth of MVCC)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -274,6 +278,73 @@ private static String diagsForEntry(Map.Entry<String, PropertyStats> member) {
...
+
+    /**
+     * Produce an {@link UpdateOp} suitable for shrinking branch revision entries for given property in {@link Document}, {@code null} otherwise.
+     * 
+     * @param doc document to inspect for repeated branch commits
+     * @param propertName property to check for
+     * @param revisionChecker filter for revisions (for instance, to check for cluster id)
+     * @return {@link UpdateOp} suitable for shrinking document, {@code null} otherwise
+     */
+    public static @Nullable UpdateOp getShrinkOp(Document doc, String propertyName, Predicate<Revision> revisionChecker) {
```
reschke, line 301: @mbaedke - this is where we could check the feature flug for now...

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -274,6 +278,73 @@ private static String diagsForEntry(Map.Entry<String, PropertyStats> member) {
...
+            }
+        }
+        return -1;
+    }
+
+    /**
+     * Produce an {@link UpdateOp} suitable for shrinking branch revision entries for given property in {@link Document}, {@code null} otherwise.
+     * 
+     * @param doc document to inspect for repeated branch commits
+     * @param propertName property to check for
```
mbaedke: Make it @param propertyName, please.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -274,6 +278,73 @@ private static String diagsForEntry(Map.Entry<String, PropertyStats> member) {
...
+                @Override
+                public int compare(Revision r1, Revision r2) {
+                    if (r1.getClusterId() != r2.getClusterId()) {
+                        return r1.getClusterId() - r2.getClusterId();
+                    } else if (r1.getTimestamp() != r2.getTimestamp()) {
+                        return r1.getTimestamp() > r2.getTimestamp() ? 1 : -1;
+                    } else {
+                        return r1.getCounter() - r2.getCounter();
+                    }
+                }});
```
rishabhdaim, line 330:
```suggestion
            revs.sort((c1, c2) -> Comparator.comparing(Revision::getClusterId).thenComparing(Revision::getTimestamp).thenComparing(Revision::getCounter).compare(c1, c2));
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -274,6 +278,73 @@ private static String diagsForEntry(Map.Entry<String, PropertyStats> member) {
...
+     * Produce an {@link UpdateOp} suitable for shrinking branch revision entries for given property in {@link Document}, {@code null} otherwise.
+     * 
+     * @param doc document to inspect for repeated branch commits
+     * @param propertName property to check for
+     * @param revisionChecker filter for revisions (for instance, to check for cluster id)
+     * @return {@link UpdateOp} suitable for shrinking document, {@code null} otherwise
+     */
+    public static @Nullable UpdateOp getShrinkOp(Document doc, String propertyName, Predicate<Revision> revisionChecker) {
+        Object t_bc = doc.get("_bc");
+        Object t_property = doc.get(propertyName);
```
rishabhdaim, line 303: I would use camelCase.

rishabhdaim:
@reschke I wonder if instead of modifying the existing `DocumentStore` classes we create a new wrapper over the DocumentStore (behind a feature flag just like for [throttling](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/ThrottlingDocumentStoreWrapper.java)) and perform all the diagnostic stuff there, wdyt?

PR #1318, opened 2024-02-24
OAK-10662 improve Reproducible Builds

hboutemy:
https://issues.apache.org/jira/browse/OAK-10662

if someone know how `OSGI-INF/l10n/*.properties` are generated, this is one common case for most of remaining issues = Properties save with timestamp: I don't know where this happens and I'd prefer avoiding post-processing

### DIFF
--- a/oak-parent/pom.xml
+++ b/oak-parent/pom.xml
@@ -34,6 +34,9 @@
   <packaging>pom</packaging>
 
   <properties>
+    <!-- build time stamp: should be updated prior to a release -->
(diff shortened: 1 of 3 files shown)


reschke (commented):
Thanks for the input.

I'll open separate issues for the plugin updates, fix them, and link to this issue.

reschke:
I *believe* these are created by the ancient Felix SCR plugin (see [OAK-6741](https://issues.apache.org/jira/browse/OAK-6741))

hboutemy:
everything has already been done in other commits: no need for this PR any more

PR #1328, opened 2024-02-26
OAK-10535: Clean up old revisions in a document

Joscorbe:
Introduce a way to safely detect and delete old revisions in a document. This could be useful to cleanup documents that sometimes grows above the supported size in MongoDB (16MB).

This is still WIP and needs to be integrated into the DetailedGC logic.

This PR should supersede #1191


oak-run/src/main/groovy/org/apache/jackrabbit/oak/console/GroovyConsole.groovy
```diff
@@ -124,6 +124,7 @@ class GroovyConsole {
                 new CheckpointCommand(shell),
                 new LsCommand(shell),
                 new PnCommand(shell),
+                new RevisionCommand(shell),
```
stefan-egli: You might already be aware of this, but just to be sure : do we need a new RevisionCommand when we have the RevisionsCommand?

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -82,6 +97,7 @@ public class RevisionsCommand implements Command {
             "  collect      perform garbage collection",
             "  reset        clear all persisted metadata",
             "  sweep        clean up uncommitted changes",
+            "  cleanup  clean up old/unused revisions",
```
stefan-egli: We might be able to handle this under the existing `collect` command - that one currently runs classic GC, but you can get it to do detailed GC as well, with the corresponding parameter.

(What's then missing is the ability to restrict it to specifically only a path)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentRevisionCleaner.java
```diff
@@ -0,0 +1,270 @@
...
+    private final RevisionPropertiesClassifier revisionClassifier;
+    private final RevisionCleanerUtility revisionCleaner;
+
+    /**
+     * Constructor for NodeDocumentRevisionCleaner.
+     * @param documentStore The DocumentStore instance. Must be writable to perform cleanup.
+     * @param documentNodeStore The DocumentNodeStore instance.
+     * @param workingDocument The document to clean up.
+     */
+    public NodeDocumentRevisionCleaner(DocumentStore documentStore, DocumentNodeStore documentNodeStore,
```
stefan-egli: Is it required to have the DocumentStore passed here? The `DocumentNodeStore.getDocumentStore()` would provide it too - but that way it would be avoided that for some reason the two could not link to the same...

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1053,6 +1053,28 @@ private void collectUnmergedBranchCommits(final NodeDocument doc, final GCPhases
...
+                if (oldRevisions.isEmpty()) {
+                    phases.stop(GCPhase.DETAILED_GC_COLLECT_OLD_REVS);
+                    return;
+                }
+
+
+
+
+
+
```
stefan-egli:
```suggestion
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentRevisionCleaner.java
```diff
@@ -0,0 +1,270 @@
...
+import org.apache.jackrabbit.oak.plugins.document.util.Utils;
+
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Map;
+import java.util.SortedMap;
+import java.util.TreeMap;
+import java.util.TreeSet;
+
+public class NodeDocumentRevisionCleaner {
```
stefan-egli, line 37: Some javadoc to describe what this class is about would be good to have.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -460,6 +494,97 @@ private void sweep(RevisionsOptions options, Closer closer)
...
+            return;
+        }
+
+        DocumentStore documentStore = builder.getDocumentStore();
+        builder.setReadOnlyMode();
+        useMemoryBlobStore(builder);
+        DocumentNodeStore documentNodeStore = builder.build();
+        String id = org.apache.jackrabbit.oak.plugins.document.util.Utils.getIdFromPath(path);
+        NodeDocument workingDocument = documentStore.find(NODES, id);
+        NodeDocumentRevisionCleaner revisionCleaner = new NodeDocumentRevisionCleaner(documentStore, documentNodeStore, workingDocument);
```
stefan-egli: As was discussed offline, we'd need to change this into a way which calls GC in general (or if that is unfeasible at least call DetailedGC in general).

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentRevisionCleanerTest.java
```diff
@@ -0,0 +1,549 @@
...
+    @Mock
+    Checkpoints checkpoints;
+
+    NodeDocument workingDocument;
+
+    NodeDocumentRevisionCleaner nodeDocumentRevisionCleaner;
+
+    @Before
+    public void setUp() {
+        MockitoAnnotations.initMocks(this);
```
stefan-egli: This shows up as deprecated. Would using the suggested alternative (`MockitoAnnotations.openMocks` - plus a later `close()`) work?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentRevisionCleaner.java
```diff
@@ -0,0 +1,270 @@
...
+                        op.removeMapEntry(property, revision);
+                    }
+                }
+
+                op.removeMapEntry("_revisions", revision);
+            }
+        }*/
+    }
+
+    protected class RevisionPropertiesClassifier {
```
stefan-egli: Is there a reason this is not `private` ?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentRevisionCleaner.java
```diff
@@ -0,0 +1,270 @@
...
+                        revisionsModifyingProperty.computeIfAbsent(propertyEntry.getKey(), key ->
+                                new TreeSet<>(StableRevisionComparator.INSTANCE)
+                        ).add(revision);
+                    }
+                }
+            }
+        }
+    }
+
+    protected class RevisionCleanerUtility {
```
stefan-egli: Is there a reason this is not `private` ?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentRevisionCleaner.java
```diff
@@ -0,0 +1,270 @@
...
+
+        private final SortedMap<Integer, TreeSet<Revision>> blockedRevisionsToKeep;
+        private final SortedMap<Integer, TreeSet<Revision>> candidateRevisionsToClean;
+        private final RevisionPropertiesClassifier revisionClassifier;
```
stefan-egli, line 161: While these are internal (presumably `private`) classes, so no expected interference from elsewhere, and given these are not value or holder but more worker type of classes, it might be worth while to avoid direct field access between each other (and instead go via getter methods)

stefan-egli:
It's generally speaking a bit difficult to review as the code doesn't compile at this point (therefore eg you can't debug through it, to understand it better). As I understand this is soon going to be fixed - so I'll wait with reviewing actual functionality for now.

oak-run/src/main/groovy/org/apache/jackrabbit/oak/console/GroovyConsole.groovy
Joscorbe: The intention of this was to be able to run from the interactive console. We would probably don't this anymore, since that console is read-only anyway.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentRevisionCleaner.java
Joscorbe: In the initial version this would receive a read-only DS and the write one. It is now obsolete since I don't think we are going to run this from oak-run console, but from the revisions command.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
Joscorbe: I agree we should at some point combine some of the commands into the same one.
For now, I have changed this to `pathCleanup`. But I think a better name can be found.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollectorIT.java
Joscorbe, line 1: I haven't done any changes in this file.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentRevisionCleaner.java
Joscorbe: Left over from initial testing. I have made it private.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentRevisionCleaner.java
Joscorbe: Left over from initial testing. I have made it private.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentRevisionCleanerTest.java
Joscorbe: Done

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
Joscorbe: This now calls the general detailedGC

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentRevisionCleaner.java
Joscorbe: Done. The class is a prototype, probably we will never use it in production, as deep cleaning in-between checkpointed revisions is probably too much.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -23,6 +23,11 @@
 import java.io.IOException;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
+import java.util.Scanner;
+import java.util.SortedMap;
+import java.util.TreeMap;
+import java.util.TreeSet;
```
rishabhdaim: please remove these un-used imports.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -32,6 +37,7 @@
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
```
rishabhdaim: Not used inside code. Please remove this.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -74,15 +89,18 @@ public class RevisionsCommand implements Command {
...
             "           any modifications",
-            "  collect      perform garbage collection",
-            "  reset        clear all persisted metadata",
-            "  sweep        clean up uncommitted changes",
-            "  detailedGC   perform detailed garbage collection i.e. remove unmerged branch commits, old revisions, deleted properties etc"
+            "  collect        perform garbage collection",
+            "  reset          clear all persisted metadata",
+            "  sweep          clean up uncommitted changes",
+            "  detailedGC     perform detailed garbage collection i.e. remove unmerged branch commits, old revisions, deleted properties etc",
+            "  pathCleanup    clean up old/unused revisions and unmerged branch commits on a specific path"
```
rishabhdaim: we are removing whole detailed garbage here. I would change this description to reflect that it is performing detailedGC on given path.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -460,6 +479,55 @@ private void sweep(RevisionsOptions options, Closer closer)
...
+        DocumentNodeStore documentNodeStore = builder.build();
+        String id = org.apache.jackrabbit.oak.plugins.document.util.Utils.getIdFromPath(path);
+        NodeDocument workingDocument = documentStore.find(NODES, id);
+        //NodeDocumentRevisionCleaner revisionCleaner = new NodeDocumentRevisionCleaner(documentNodeStore, workingDocument);
+
+        VersionGarbageCollector gc = bootstrapVGC(options, closer, true);
+        // Set a LoggingGC monitor that will output the detailedGC operations
+        gc.setGCMonitor(new LoggingGCMonitor(LOG));
+        // Run detailedGC on the given document
+        UpdateOp op = gc.collectGarbageOnDocument(documentNodeStore, workingDocument);
```
rishabhdaim: Returning `UpdateOp` from detailedGC and handling deletion here is not a good idea, cause we might need to duplicate the code written in `DetailedGC.removeGarbage()` OR we might miss something.

In this case, we are missing orphan nodes cause they are handled differently in `detailedGC.removeGarbage()`.
I would delegate the collection & removal of garbage to DetailedGC itself.

rishabhdaim (changes requested):

Joscorbe:
I have squashed the WIP commits, sorry for the forced-PR.

rishabhdaim:
Could you please revert `VersionGarbageCollectorIT` from your PR, it looks like you haven't made any changes in this file and it is reverting changes done in this [PR](https://github.com/apache/jackrabbit-oak/pull/1332/files#diff-b4378c829e2a8761f08cd1cb460bbc3bec86b74f379243bdea64de2bfa27a107)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocumentRevisionCleaner.java
Joscorbe: Done

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
Joscorbe: Done

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
Joscorbe: Done

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -74,15 +80,18 @@ public class RevisionsCommand implements Command {
 
     private static final Logger LOG = LoggerFactory.getLogger(RevisionsCommand.class);
 
+    private static final int REVISION_CAP = getInteger("oak.revision.cap", 250);
```
stefan-egli: (where) Is this still used?

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -74,15 +80,18 @@ public class RevisionsCommand implements Command {
...
+
     private static final String USAGE = Joiner.on(System.lineSeparator()).join(
             "revisions {<jdbc-uri> | <mongodb-uri>} <sub-command> [options]",
             "where sub-command is one of",
             "  info     give information about the revisions state without performing",
```
stefan-egli:
```suggestion
            "  info        give information about the revisions state without performing",
```
(minor: indentation)

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
stefan-egli: alternatively I was wondering if the restriction to a single path could not also be a further command line parameter of `detailedGC` mode? (in which case we wouldn't need a new mode just make `detailedGC` more fancy)

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -460,6 +479,55 @@ private void sweep(RevisionsOptions options, Closer closer)
...
+            return;
+        }
+
+        DocumentStore documentStore = builder.getDocumentStore();
+        builder.setReadOnlyMode();
+        useMemoryBlobStore(builder);
+        DocumentNodeStore documentNodeStore = builder.build();
+        String id = org.apache.jackrabbit.oak.plugins.document.util.Utils.getIdFromPath(path);
+        NodeDocument workingDocument = documentStore.find(NODES, id);
+        //NodeDocumentRevisionCleaner revisionCleaner = new NodeDocumentRevisionCleaner(documentNodeStore, workingDocument);
```
stefan-egli: Is the idea to keep this as a reminder for a future change (or could it be removed now)?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1238,12 +1240,11 @@ private void removeUnmergedBCRevision(final Revision unmergedBCRevision, final N
         }
 
         private void collectOldRevisions(final NodeDocument doc, final GCPhases phases, final UpdateOp updateOp) {
```
stefan-egli: Would suggest to either rename to clarify that this is about old and between-checkpoint-but-not-younger-than-maxRevisionAgeMillis revisions - or to clarify that in a javadoc.

Now, given the old-revision-gc part competes with OAK-10688 we should execute only one of the two, not both. One idea could be to add an static boolean (not a system property, just a plain boolean) that lets you run DetailedGC in either between-checkpoint mode (OAK-10535) or keep-1-revision mode (OAK-10688). Then tests can control that flag at their discretion, we can compare the two and find bugs in the two, but ultimately decide on which one we want.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1398,6 +1399,24 @@ private void delayOnModifications(final long durationMs, final AtomicBoolean can
         }
     }
 
+    public UpdateOp collectGarbageOnDocument(DocumentNodeStore store, NodeDocument doc) {
```
stefan-egli: As an alternative idea to the 2 steps suggested (`collectGarbageOnDocument` then removing the garbage in `RevisionsCommand`) which @rishabhdaim also commented on : what about focusing on executing the [core part](https://github.com/apache/jackrabbit-oak/blob/cf870b532b107b0c243e1d0180f1213d3f6e9197/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java#L736-L782) of `collectDetailedGarbage` in 2 different modes : the current full one, and a new `onPath` one. Eg that would imply refactoring the existing `collectDetailedGarbage`'s inner part out and then more or less directly invoking that on a particular path (perhaps have a small facade method that facilitates that).

The problem I think is indeed, as @rishabhdaim  [already mentioned](https://github.com/apache/jackrabbit-oak/pull/1328/files#r1511381906), that some parts might be missed, if we have a second variant of `removeGarbage` (for example we wouldn't do the embedded verification). Of course doing it "the real way" does require some refactoring, but I think it could be done via that inner loop mentioned. Wdyt?

Joscorbe: I'm working on this approach, will commit the changes soon.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
Joscorbe: Working on this, but I'm refactoring the `collectDetailedGarbage` to run all the logic on a single document. That way I think we will also improve the readability of that code.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
Joscorbe: That's possible, yes, and would be cleaner IMO. Since the `--path` is already a separated argument we can simply apply it to this command.
My only concern with this approach is, if the `path` argument is missing, it could end up running the DetailedGC in all the repository.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
Joscorbe: After some work in that area, I'm abandoning this approach. I initially thought it would be a good idea because that method is already quite long, so splitting it could improve its readability. However, the _real core_ part would be [just these lines](https://github.com/apache/jackrabbit-oak/blob/cf870b532b107b0c243e1d0180f1213d3f6e9197/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java#L752-L756), as the rest of the code is essentially to keep the progress of the iterators.

So I simply [added this part](https://github.com/apache/jackrabbit-oak/blob/cf870b532b107b0c243e1d0180f1213d3f6e9197/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java#L768-L772) to the facade method I already added.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
Joscorbe: Fixed

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
Joscorbe: No, some leftover from old code.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
Joscorbe: No, it's removed now.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
stefan-egli: In theory if someone runs the command they should be aware of what it does - it's an admin command after all.

I guess we're doing it this careful because the new code isn't tested well enough yet.

Could we then - during this testing phase - demand that you either specify `--path` or `--entirerepo` (or something like this)? Then eventually make `--entirerepo` the default and not demand it anymore?

rishabhdaim (approved):

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
Joscorbe: That makes me a little less nervous. I have added that option for now.

stefan-egli (approved):
+1 for merging to the feature branch. Before merging to trunk we should still introduce a flag as to which "old revision cleanup" variant we'll use - but that shouldn't block this PR as of now.

PR #1331, opened 2024-02-27
OAK-10671- [Indexing Job] Improve Mongo regex query: remove condition on non-indexed _path field to speed-up traversal

nfsantos:
Regex path filtering currently is implemented with a condition like:

```
_id in [^[0-9]{1,3}:\Q/foo/bar/\E.*$] OR ('_id' in [^[0-9]{1,3}:h.*$}] AND _path in [^\Q/foo/bar/\E.*$]
```

The second condition is necessary to deal with long path documents, whose _id is an hash instead of the path of the document, and that have an additional `_path` property with the full path of the document. The `_id` field is part of the index used by the query, but `_path` is not indexed. So the performance of this query will be very sensitive to how many time the query condition can be resolved without having to lookup the value of `_path`, which requires retrieving the full document from the column store. If the condition can be evaluated only using the `_id` value, them if there is no match the document should not be retrieved from the column store.

Unfortunately, Mongo does not seem to properly optimize this query and is retrieving the document from the column storage even when `_id` does not match the path /foo/bar and the `_id` is not in the hash format. This leads to very poor performance as both the index and the column store have to be fully read by this query.

We can instead use the following condition:
```
_id in [^[0-9]{1,3}:\Q/foo/bar/\E.*$ , ^[0-9]{1,3}:h.*$}]
```
That is, download the document if the `_id` matches the path or if it is an hash. This has the disadvantage that it will download all long path documents from the repository, many of which might not be needed. However, this query condition only uses the `_id` field so it is guaranteed to be evaluated fully using only the data on the index. And the number of long paths documents is usually very small, some environments don't even have any long path documents, so downloading them should not take much time. And the indexing job will anyway reapply the filter on paths locally, to eliminate the long path documents which are not required by the indexing job.

The gains of scanning only the index can be substantial. For instance, in one experiment, reindexing an index that requires 400K documents out of 29M I got the following results to build the FFS:

- With filter on `_path`: 30 minutes
- No filter on `_path`: 1m30

On another instance, downloading 14M nodes out of 250M:

- With filter on `_path`: 1h30
- No filter on `_path`: 20 minutes.

The gains will be bigger the smaller the fraction of the OAK tree that has to be downloaded, but there should be benefits even when downloading a large fraction of the tree.

### Limitations
Using only the _id field on the filter condition improves the performance significantly when having only `includedPaths`, but it does not change the performance when there are excluded paths or there are custom regex exclude filters. The reason is that Mongo needs to retrieve the document when evaluating a condition like `not regex(_id, "/tmp/.*")`, because a value of `null` also matches this condition and the index does not contain null values. A future PR will address this limitation.

### DIFF
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
@@ -147,13 +147,12 @@ public class PipelinedMongoDownloadTask implements Callable<PipelinedMongoDownlo
      * @param mongoFilterPaths          The paths to be included/excluded in the filter. These define subtrees to be included or excluded.
      *                                  (see {@link MongoFilterPaths} for details)
      * @param customExcludeEntriesRegex Documents with paths matching this regex are excluded from download
-     * @param queryUsesIndexTraversal   Whether the query will use an index to traverse the documents.
      * @return The filter to be used in the Mongo query, or null if no filter is required
      */
-    static Bson computeMongoQueryFilter(@NotNull MongoFilterPaths mongoFilterPaths, String customExcludeEntriesRegex, boolean queryUsesIndexTraversal) {
+    static Bson computeMongoQueryFilter(@NotNull MongoFilterPaths mongoFilterPaths, String customExcludeEntriesRegex) {
         var filters = new ArrayList<Bson>();
 
-        Bson includedFilter = descendantsFilter(mongoFilterPaths.included, queryUsesIndexTraversal);
+        Bson includedFilter = descendantsIncludeFilter(mongoFilterPaths.included);
         if (includedFilter != null) {
             filters.add(includedFilter);
         }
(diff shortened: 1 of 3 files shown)


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -184,65 +181,70 @@ static Bson computeMongoQueryFilter(@NotNull MongoFilterPaths mongoFilterPaths,
...
+        // we could add a condition on the _path field, but this would slow down substantially scanning the DB, because
+        // the _path field is not part of the index used by this query (it's an index on _modified, _id). Therefore,
+        // Mongo would have to retrieve every document from the column store to evaluate the filter condition. So instead
+        // we add below a condition to download all the long path documents. These documents can be identified by the
+        // format of the _id field (<n>:h<hash>), so it is possible to identify them using only the index.
+        // This might download documents for nodes that are not in the included paths, but those documents will anyway
+        // be filtered in the transform stage. And in most repositories, the number of long path documents is very small,
+        // often there are none, so the extra documents downloaded will not slow down by much the download. However, the
+        // performance gains of evaluating the filter of the query using only the index are very significant, especially
+        // when the index requieres only a small number of nodes.
```
fabriziofortino: nitpick: typo
```suggestion
        // when the index requires only a small number of nodes.
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -184,65 +181,70 @@ static Bson computeMongoQueryFilter(@NotNull MongoFilterPaths mongoFilterPaths,
...
+        }
+        ArrayList<Pattern> patterns = new ArrayList<>();
+        for (String path : paths) {
+            if (!path.endsWith("/")) {
+                path = path + "/";
+            }
+            String quotedPath = Pattern.quote(path);
+            patterns.add(Pattern.compile("^[0-9]{1,3}:" + quotedPath + ".*$"));
+        }
+        return patterns;
```
fabriziofortino: just noticed that `descendantsIncludedPatterns` and `descendantsExcludedPatterns` are almost the same. The former only adds an extra condition. I would consider reusing the logic.

fabriziofortino (approved):
LGTM. Just a couple of minors.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
nfsantos: Fixed.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
nfsantos: I created a single method with the common parts of this logic.

steffenvan:
Don't have anything to add LGTM.

PR #1332, opened 2024-02-28
OAK-10676 : used traversed state to find deleted properties

rishabhdaim:


stefan-egli (approved):

rishabhdaim:
thanks for review @stefan-egli

PR #1336, opened 2024-03-01
OAK-10680: tests failures with com.arakelian/docker-junit-rule

mbaedke:


reschke:
Oh, wow.

I'm somewhat concerned by the amount of changes over on the elastic search side. Either these need to be double-checked by the right people (@fabriziofortino ???), or we should extract the bits needed to fix the document-store test and move the other changes to a separate ticket/PR.

reschke (commented):
Oh, wow.

I'm somewhat concerned by the amount of changes over on the elastic search side. Either these need to be double-checked by the right people (@fabriziofortino ???), or we should extract the bits needed to fix the document-store test and move the other changes to a separate ticket/PR.

reschke (changes requested):
we need to break this down into smaller chunks; for instance, we really do not want to introduce shaded Guava into 1.22.

mbaedke:
@reschke,

> we really do not want to introduce shaded Guava into 1.22

Yes, obviously. That one sneaked in with the cherry-pick. I'll remove it.

reschke (approved):
(please make sure we use the same versions of the docker containers as before)

PR #1340, opened 2024-03-04
OAK-10681 - [indexing job] Support custom filters of paths on Mongo

nfsantos:
Add property to specify list of paths to be excluded from Mongo download: `oak.indexer.pipelined.mongoCustomExcludedPaths`.

Custom path filtering will only be applied if the index included/excluded paths resolve to downloading the root with optionally some excluded paths. 

Additional minor change:
- Simplify the regex expressions to filter on paths, remove the trailing `.*$` at the end.


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -126,6 +126,13 @@ public long getDocumentsDownloaded() {
     public static final String OAK_INDEXER_PIPELINED_MONGO_REGEX_PATH_FILTERING_MAX_PATHS = "oak.indexer.pipelined.mongoRegexPathFilteringMaxPaths";
     public static final int DEFAULT_OAK_INDEXER_PIPELINED_MONGO_REGEX_PATH_FILTERING_MAX_PATHS = 20;
 
+    /**
+     * Additional Oak paths to exclude from downloading from Mongo. This is a comma-separated list of paths.
+     * These paths are only filtered if the included paths computed from the idndexes resolve to the root tree (/),
```
fabriziofortino: typo
```suggestion
     * These paths are only filtered if the included paths computed from the indexes resolve to the root tree (/),
```

fabriziofortino (approved):

PR #1352, opened 2024-03-07
OAK-6766: Convert oak-lucene to OSGi R7 annotations

mbaedke:


reschke (commented):
Would be good to get feedback from the Lucene team for this. Maybe the issue of "pirvate" properties can be simplified?

mbaedke:
> Would be good to get feedback from the Lucene team for this.

Definitely

> Maybe the issue of "private" properties can be simplified?

Now that I realized that they can be retrieved directly from the ComponentContext without needing a reference to the ConfigurationAdmin service (see last commit), I don't think it can get simpler anymore.

jsedding:
> I don't think it can get simpler anymore.

Have you tried adding the private property names to the annotation, but leaving the `@Attribute` meta-annotation off? I have a feeling that could work, and you could then access the values more easily. YMMV, I am not 100% certain.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexProviderService.java
```diff
@@ -368,13 +349,19 @@ public class LuceneIndexProviderService {
...
+        enableCopyOnRead = PropertiesUtil.toBoolean(properties.get("enableCopyOnReadSupport"), true);
+        enableCopyOnWrite = PropertiesUtil.toBoolean(properties.get("enableCopyOnWriteSupport"), true);
         asyncIndexesSizeStatsUpdate = new AsyncIndexesSizeStatsUpdateImpl(
-                PropertiesUtil.toLong(config.get(LUCENE_INDEX_STATS_UPDATE_INTERVAL),
-                        LUCENE_INDEX_STATS_UPDATE_INTERVAL_DEFAULT) * 1000); // convert seconds to millis
-        boolean disabled = PropertiesUtil.toBoolean(config.get(PROP_DISABLED), PROP_DISABLED_DEFAULT);
-        hybridIndex = PropertiesUtil.toBoolean(config.get(PROP_HYBRID_INDEXING), PROP_DISABLED_DEFAULT);
+                PropertiesUtil.toLong(config.luceneIndexStatsUpdateInterval(),
+                        PROP_LUCENE_INDEX_STATS_UPDATE_INTERVAL_DEFAULT) * 1000); // convert seconds to millis
+        boolean disabled = PropertiesUtil.toBoolean(config.disabled(), PROP_DISABLED_DEFAULT);
```
kwin: There is no need for a type conversion here. Also the default is always set according to what is defined in the component property type (https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#service.component-component.property.types)

mbaedke:
@jsedding Yes, I tried that :)

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexProviderService.java
mbaedke: Oh yes, missed that one, thx!

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexProviderService.java
```diff
@@ -104,201 +107,178 @@
 import static org.apache.jackrabbit.oak.spi.whiteboard.WhiteboardUtils.scheduleWithFixedDelay;
 
 @SuppressWarnings("UnusedDeclaration")
-@Component(metatype = true, label = "Apache Jackrabbit Oak LuceneIndexProvider")
+@Component(property = { "enableCopyOnReadSupport:Boolean=true", "enableCopyOnWriteSupport:Boolean=true" } )
```
kwin: I would really recommend not mixing component property types with regular properties. Just put the private properties to the Configuration annotation type (but without the metatype annotations on the element). Also the default values only need to be defined once then (and do no longer require a dedicated constant) but can be added as literal on the component property type.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexProviderService.java
```diff
@@ -368,13 +349,19 @@ public class LuceneIndexProviderService {
     private PropertyIndexCleaner cleaner;
     private AsyncIndexesSizeStatsUpdate asyncIndexesSizeStatsUpdate;
 
+    private boolean enableCopyOnRead = true;
+    private boolean enableCopyOnWrite = true;
+
     @Activate
-    private void activate(BundleContext bundleContext, Map<String, ?> config) throws IOException {
+    private void activate(BundleContext bundleContext, ComponentContext componentContext, Configuration config) throws IOException {
```
kwin: As said above mixing regular properties with component property types is rarely necessary, also instead of injecting the ComponentContext you can rely on a Map to inject component properties if it would be necessary (https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#service.component-activation.objects)

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexProviderService.java
mbaedke: Yes, I did that because otherwise the diff tool complains (the properties will not occur in the component.xml). Meanwhile @cziegele enlightened me. I'll fix it.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexProviderService.java
mbaedke: Thx

mbaedke:
Addressed @kwin 's suggestions.

PR #1353, opened 2024-03-07
OAK-10375: remove binary data from logs

steffenvan:
If there is an error while indexing a binary property, we shouldn't log the data. We should just log the property name.

### Current
```
ERROR o.a.j.o.p.i.l.LuceneDocumentMaker - could not index similarity field for property 
haystack0 = <LONG GARBLED BINARY DATA> 
```

### This PR
```
ERROR o.a.j.o.p.i.l.LuceneDocumentMaker - could not index similarity field for property 
haystack0
```


oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -336,6 +338,32 @@ private boolean indexProperty(String path,
         return dirty;
     }
 
+    private String getValueAsTruncatedString(Object val) {
+        String value = val.toString();
+        return value.length() <= 128 ? value
+            : value.substring(0, 128) + " ... [" + value.hashCode() + "]";
```
nfsantos: Extract the cutoff size for truncation into a constant, that avoids hardcoding in multiple places the same numeric value and the constant name works as documentation.

nfsantos: This is going to fail with StringIndexOutOfBoundsException if the value is smaller than 128.

nfsantos: Do not add the three dots if the value is printed in full, add it only if the value is truncated.

nfsantos: What is the use case of printing the hascCode? In what situations will it be useful? I can't think of any right now.

nfsantos (changes requested):
In addition to the fixes to the comments below, please add tests to the truncation logic. The tests should cover cases of binary and non-binary properties, of arrays and of single values, and of strings that are smaller, equal and bigger than the printing threshold.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
steffenvan: I don't see how? The first condition in the ternary returns the value if it's smaller than 128?

steffenvan: Same as the comment above - it won't be printed unless it exceeds the truncation length.

steffenvan: Wasn't sure that it would be. Put it mostly there to have a sense if anyone might think it would be useful. But I'll just remove it.

nfsantos: Ok, it's correct, I got confused by the use of the ternary operator.

nfsantos: Yes, I don't think it adds anything useful.

nfsantos: 👍

steffenvan: Okay done!

reschke (changes requested):
I'm a bit concerned about putting the *contents* of properties into log files. Isn't that a privacy issue by design?

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -336,6 +339,37 @@ private boolean indexProperty(String path,
         return dirty;
     }
 
+    static String getValueAsTruncatedString(Object val) {
+        String value = val.toString();
```
nfsantos: Maybe trim the string before testing below for starting and ending with [ ].

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -336,6 +339,37 @@ private boolean indexProperty(String path,
...
+        boolean isArray = value.startsWith("[") && value.endsWith("]");
+        return value.length() <= LOG_TRUNCATION_LENGTH ? value
```
nfsantos: Personal opinion: in this situation, I find the ternary operator harder to read than an if-then-else. It was also the reason why I misunderstood the code initially, and thought that there was a risk of out of bounds exception. But it's a matter of personal preference, so feel free to leave it as it is.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -336,6 +339,37 @@ private boolean indexProperty(String path,
...
+            Type<?> base = t.getBaseType();
+            StringBuilder sb = new StringBuilder();
+            sb.append("[");
+            for (int i = 0; i < property.count(); i++) {
+                sb.append(property.getValue(base, i));
+                if (i < property.count() - 1) {
+                    sb.append(", ");
+                }
+            }
+            sb.append("]");
```
nfsantos: This would be cleaner in functional style, something like:

```
IntStream.range(0, property.count())
        .mapToObj(i -> property.getValue(Type.STRING, i))
        .collect(Collectors.joining(",", "[", "]"));
```
Maybe less efficient, but this is code is not in a hotspot, so it does not matter.

nfsantos (approved):
Looks good, just a few minor comments and suggestions.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
steffenvan: Yeah I like this better

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
steffenvan: done

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
steffenvan: I've modified the function now - should be more clear.

nfsantos (approved):

chibulcuteanu (approved):

fabriziofortino (commented):
The changes are okay to remove the binary data from the logs but those messages are not very useful. We just report the property name along with the property definition. I think we should also say which node contains this property. What about adding the `path` as well?

steffenvan:
Hmm as long as it doesn't violate any privacy concerns, I guess so. @reschke

reschke:
That's a good point. Paths can contain senstive information as well, but I *believe* we use them alaready almost everywhere, because that information is really essential for debugging.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -317,7 +318,9 @@ private boolean indexProperty(String path,
                                     indexSimilarityStrings(doc, pd, value);
                                 }
                             } catch (Exception e) {
-                                log.error("could not index similarity field for property {} and definition {}", property, pd);
+                                log.error(
+                                    "could not index similarity field for property {} and definition {}",
```
fabriziofortino: we should print the path here as well

steffenvan: Done

fabriziofortino (approved):

PR #1360, opened 2024-03-11
OAK-10703: oak-doc: add branch history to roadmap

reschke:


paulalexwilson:
I'm curious, is the ASCII timeline view of the release history something you wrote by hand, or was a tool used to generate it?

reschke:
:-) it's hand work.

A generic tool would be interesting, but it would need to be highly configurable (because not all changes have the same importance)

PR #1363, opened 2024-03-12
OAK-10708: error-level log failures to write the journal

reschke:
@stefan-egli - afaict, ITs do not seem to trigger this condition.

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Commit.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Commit.java
@@ -355,6 +355,7 @@ public class Commit {
             boolean success = store.create(JOURNAL, singletonList(doc.asUpdateOp(r)));
             if (!success) {
                 LOG.error("Failed to update journal for revision {}", r);
+                LOG.debug("Failed to update journal for revision {} with doc {}", r, doc);
             }
         }
 
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Commit.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Commit.java
@@ -355,7 +355,7 @@ public class Commit {
             boolean success = store.create(JOURNAL, singletonList(doc.asUpdateOp(r)));
             if (!success) {
                 LOG.error("Failed to update journal for revision {}", r);
-                LOG.debug("Failed to update journal for revision {} with doc {}", r, doc);
+                LOG.debug("Failed to update journal for revision {} with doc {}", r, doc.format());
             }
         }
 
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Commit.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Commit.java
@@ -352,7 +352,10 @@ public class Commit {
             JournalEntry doc = JOURNAL.newDocument(store);
             doc.modified(modifiedNodes);
             Revision r = revision.asBranchRevision();
-            store.create(JOURNAL, singletonList(doc.asUpdateOp(r)));
+            boolean success = store.create(JOURNAL, singletonList(doc.asUpdateOp(r)));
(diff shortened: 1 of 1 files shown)


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Commit.java
```diff
@@ -352,7 +352,10 @@ private void applyToDocumentStore(RevisionVector baseBranchRevision)
             JournalEntry doc = JOURNAL.newDocument(store);
             doc.modified(modifiedNodes);
             Revision r = revision.asBranchRevision();
-            store.create(JOURNAL, singletonList(doc.asUpdateOp(r)));
+            boolean success = store.create(JOURNAL, singletonList(doc.asUpdateOp(r)));
+            if (!success) {
+                LOG.error("Failed to update journal for revision {}", r);
```
stefan-egli, line 357: Was wondering if we should also log the `doc` - it would contain the modified paths (perhaps useful in analysing and presumably not leaking sensitive information)?

stefan-egli (commented):
It seems having such a log.error is good, as we're not checking whether `create` was successful or not, so +1 for that.

Meanwhile, what's not quite clear is why we're [catching](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java#L1754-L1755) a plain `MongoException` silently with zero logging - presumably as that happens in other cases where this is expected but there's an outer retry loop (but not for the journal case here)? In which case the new error.log might be anyway good - but actually the swallowed `MongoException` could contain the relevant detail ....... bit of a dead-lock...

reschke:
> Meanwhile, what's not quite clear is why we're [catching](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java#L1754-L1755) a plain MongoException silently with zero logging

Good catch. I'd say that needs a separate ticket.

stefan-egli:
@reschke , now created OAK-10730 for logging the MongoException - it's not entirely clear at what level we should log though - I would suggest to not go log.error right away (and potentially cause unwanted turbulence). wdyt?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Commit.java
reschke: The whole doc, as with "toString()"?

stefan-egli: It would have been the whole doc yes.

Alternatively, we could also do two logs : the existing LOG.error and an additional LOG.debug with the doc... (such that it wouldn't log by default, but it could be enabled)... (but I guess it's either way more of a detail)

reschke:
Like that?

~~~
            if (!success) {
                LOG.error("Failed to update journal for revision {}", r);
                LOG.debug("Failed to update journal for revision {} with doc {}", r, doc);
            }
~~~

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Commit.java
```diff
@@ -352,7 +352,11 @@ private void applyToDocumentStore(RevisionVector baseBranchRevision)
...
+                LOG.debug("Failed to update journal for revision {} with doc {}", r, doc);
```
stefan-egli:
```suggestion
                LOG.debug("Failed to update journal for revision {} with doc {}", r, doc.format());
```

IIUC JournalEntry.toString() only logs the id, so it would probably have to be something like `format()`

stefan-egli (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Commit.java
reschke: ok, done; should I  merge?

PR #1367, opened 2024-03-14
OAK-10710 : added option to reset detailedGC after finishing detailed…

rishabhdaim:
…GC cycle

### DIFF
--- a/oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
+++ b/oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
@@ -17,7 +17,6 @@
 package org.apache.jackrabbit.oak.run;
 
 import org.apache.jackrabbit.guava.common.base.Joiner;
-import org.apache.jackrabbit.guava.common.collect.ImmutableList;
 import org.apache.jackrabbit.guava.common.io.Closer;
 
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


stefan-egli (approved):
looks ok, some things confuse slightly, maybe you could explain: the removal of the `detailedGC / enable detailedGC (collect only)` part, is that related to the reset or is it just coincidence and a general cleanup? also I was wondering how useful this new `reset-after-full-repo-detailedGC` is - I have no concerns having it - just is it really so useful? Couldn't you otherwise just call the reset command only (that would, ok, have to start up oak-run again and connect to mongo for that, but that's it, it would take maybe 10seconds - is it about saving those 10sec? just to understand). Otherwise just wondering if we should in the future aim to do general improvement of code (which I agree we should indeed do) in different PRs (refering to the removal of the guava list thing - which I agree which should do).

rishabhdaim:
`the removal of the detailedGC / enable detailedGC (collect only) part`
* This is done to avoid confusion since we have 2 ways to run detailedGC code i.e. from **_collect_** & **_detailedGC_** command. 

`I was wondering how useful this new reset-after-full-repo-detailedGC`
* I agree that this can be achieved by calling **_reset_** command, but having it inside the **_detailedGC_** command gives us one place/command to control everything around detailedGC. Also, it will save both the time & effort of a developer.

`Otherwise just wondering if we should in the future aim to do general improvement of code`
* I did it since it was only one-liner.

P.S. thanks for the quick review @stefan-egli

PR #1383, opened 2024-03-26
OAK-10724 : Introduce detailed gc mode that only deletes orphan nodes…

stefan-egli:
… and deleted properties

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
@@ -148,12 +148,13 @@ public class VersionGarbageCollector {
     static final String SETTINGS_COLLECTION_DETAILED_GC_DRY_RUN_DOCUMENT_ID_PROP = "detailedGCDryRunId";
 
     static enum RDGCType {
+        NO_OLD_PROP_REV_GC,
         KEEP_ONE_FULL_MODE,
         KEEP_ONE_CLEANUP_USER_PROPERTIES_ONLY_MODE,
         OLDER_THAN_24H_AND_BETWEEN_CHECKPOINTS_MODE
     }
 
-    final static RDGCType revisionDetailedGcType = RDGCType.KEEP_ONE_FULL_MODE;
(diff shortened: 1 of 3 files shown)


stefan-egli:
BranchCommitGCTest now fail due to using the new mode by default - requires fixing first before merging this

Joscorbe (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1007,6 +1008,10 @@ public void collectGarbage(final NodeDocument doc, final GCPhases phases) {
                 // here the node is not orphaned which means that we can reach the node from root
                 collectDeletedProperties(doc, phases, op, traversedState);
                 switch(revisionDetailedGcType) {
+                    case NO_OLD_PROP_REV_GC : {
+                        // this mode does neither unusedproprev, nor unmergedBC
```
rishabhdaim, line 1012: I would add an INFO log here.

rishabhdaim (approved):

rishabhdaim:
PR approved with a minor review comment that can be handled later.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
stefan-egli: this would log it quite frequently no? - what about doing it in the constructor?

stefan-egli: (I'll merge without and we can do it separately as you suggested)

stefan-egli:
> BranchCommitGCTest now fail due to using the new mode by default - requires fixing first before merging this

will be handled in OAK-10726 now (to not block this PR)

log.info we need to follow-up in a new ticket still

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
stefan-egli: PS: created OAK-10727 to follow-up on this

PR #1385, opened 2024-03-27
Issues/OAK-10675

t-rana:
Added service principal support in oak-blob-cloud-azure


oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreAccessManager.java
```diff
@@ -0,0 +1,271 @@
...
+
+import java.net.URISyntaxException;
+import java.security.InvalidKeyException;
+import java.time.Instant;
+import java.util.Date;
+import java.util.EnumSet;
+import java.util.Optional;
+import java.util.Properties;
+
+public class AzureDataStoreAccessManager {
```
smiroslav: Public methods of this class do not resolve access rights. This class rather provides ```CloudBlobContainer``` . I would rename it accordingly.

t-rana: Renamed to AzureBlobContainerProvider

smiroslav (approved):

oak-blob-cloud-azure/pom.xml
```diff
@@ -50,8 +69,23 @@
                         <Embed-Dependency>
                             azure-storage,
                             azure-keyvault-core,
+                            azure-core,
+                            azure-identity,
+                            azure-json,
                             guava,
-                            jsr305
+                            jsr305,
```
reschke, line 76: I'm very concerned about the amount of embeds here. Is this really needed? It will make the nundle bigger, and will make updates harder...

kwin: At least JSR305 is never necessary at run time AFAIK. It is only annotations relevant at compile time.

t-rana: @reschke, we are using azure-identity library for migration to service principals from access key based authentication in azure. All these dependencies came transitively from azure-identity. I initially tried marking some of them as optional which helped in resolving the bundle to active state but when I configured the component to use service principals, it failed with NoClassDefFoundError exceptions meaning that they were indeed required at runtime and cannot be marked as optional.

Attaching the file of errors that I got and the corresponding dependency embedded.
[dependencies_list.pdf](https://github.com/apache/jackrabbit-oak/files/15319377/dependencies_list.pdf)

rootpea: @reschke @kwin do we have anything else to add regarding the embeds list? otherwise I'd suggest to go forward with the PR

cc @t-rana

reschke: I just wanted to make sure we know what we do :-)

PR #1387, opened 2024-03-27
OAK-10727 : log revisionDetailedGcType in VersionGarbageCollector con…

stefan-egli:
…structor


stefan-egli:
@rishabhdaim  wdyt?

Joscorbe (approved):
I think this is a good idea.

rishabhdaim:
I wonder if I shouldn't add this to `VersionGCStats` so it gets logged along with other stats.

rishabhdaim (approved):

stefan-egli:
+1 to adding it also to VersionGCStats - we can still see which one turns out more/how useful and adjust along the way in case

PR #1392, opened 2024-04-02
OAK-10734 : keep _revisions entry for non bcs younger than sweep - added test case to showcase as parentWithGarbageGCChildIndependent

stefan-egli:


stefan-egli:
wrong branch - closing PR

PR #1393, opened 2024-04-02
OAK-10734 : keep _revisions entry for non bcs younger than sweep - ad…

stefan-egli:
…ded test case to showcase as parentWithGarbageGCChildIndependent

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
@@ -1488,7 +1488,7 @@ public class VersionGarbageCollector {
                 }
                 final boolean isRoot = doc.getId().equals(Utils.getIdFromPath(Path.ROOT));
                 // local bcs only considered for removal
-                final boolean isBC = !doc.getLocalBranchCommits().contains(revision);
+                final boolean isBC = doc.getLocalBranchCommits().contains(revision);
                 final boolean newerThanSweep = nodeStore.getSweepRevisions().isRevisionNewer(revision);
                 if (newerThanSweep) {
                     //TODO wondering if we should at all do any DGC on documents newer than sweep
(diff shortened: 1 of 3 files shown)


stefan-egli:
another test class required gccount fixing, done now

stefan-egli:
Note incoming https://github.com/apache/jackrabbit-oak/pull/1396

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollectorIT.java
```diff
@@ -1263,6 +1265,77 @@ private Iterator<NodeDocument> candidates(long fromModified, long toModified, in
...
+
+        // enable the detailed gc flag
+        writeField(gc, "detailedGCEnabled", true, true);
+
+        // wait two hours
+        clock.waitUntil(clock.getTime() + HOURS.toMillis(2));
+
+        // now make a child change so that only parent (and root) gets GCed
+        nb = store1.getRoot().builder();
+        nb.child("parent").child("child").setProperty("ck", "cv2");
```
rishabhdaim: shouldn't this be set to a different value here?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1488,7 +1488,7 @@ private int collectUnusedInternalPropertyRevisions(final NodeDocument doc,
                 }
                 final boolean isRoot = doc.getId().equals(Utils.getIdFromPath(Path.ROOT));
                 // local bcs only considered for removal
-                final boolean isBC = !doc.getLocalBranchCommits().contains(revision);
+                final boolean isBC = doc.getLocalBranchCommits().contains(revision);
```
rishabhdaim, line 1491: good catch.

rishabhdaim (approved):

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollectorIT.java
stefan-egli: good point - it's slightly different though : that code is actually superfluous, it was copied from `parentGCChildIndependent()` and I forgot to delete them. The test case is indeed slightly different in that it wants a revision GCed from the parent that is still needed in a child

I've now removed these lines and added a corresponding comment a few lines above - in https://github.com/apache/jackrabbit-oak/pull/1393/commits/21e8b289447790b69c78291335983b19c9471a7c

PR #1397, opened 2024-04-02
OAK-10736 : collect detailedGC stats for dryRun mode as well

rishabhdaim:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -155,10 +155,10 @@ static enum RDGCType {
         OLDER_THAN_24H_AND_BETWEEN_CHECKPOINTS_MODE
     }
 
-    private static RDGCType revisionDetailedGcType = RDGCType.NO_OLD_PROP_REV_GC;
+    private static final RDGCType REVISION_DETAILED_GC_TYPE = RDGCType.NO_OLD_PROP_REV_GC;
```
stefan-egli: we cannot make this final as the tests overwrite this to test the different modes.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1770,6 +1770,21 @@ public void removeGarbage(final VersionGCStats stats) {
...
+                    stats.deletedPropsCount += deletedPropsCountMap.values().stream().reduce(0, Integer::sum);
+                    stats.deletedInternalPropsCount += deletedInternalPropsCountMap.values().stream().reduce(0, Integer::sum);
+                    stats.deletedPropRevsCount += deletedPropRevsCountMap.values().stream().reduce(0, Integer::sum);
+                    stats.deletedInternalPropRevsCount += deletedInternalPropRevsCountMap.values().stream().reduce(0, Integer::sum);
+                    stats.deletedUnmergedBCCount += deletedUnmergedBCSet.size();
+
+                    // for orphan nodes.
+                    stats.updatedDetailedGCDocsCount += orphanOrDeletedRemovalMap.size();
+                    stats.deletedDocGCCount += orphanOrDeletedRemovalMap.size();
+//                    stats.deletedOrphanNodesCount += orphanOrDeletedRemovalMap.size();
```
stefan-egli, line 1781: Unless there's a reason to keep this comment, I would probably remove it. Or is the idea to introduce a separate orphan count?

rishabhdaim: On local, I do have stats for orphan nodes. I had to comment it out to make it compile.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -148,17 +148,17 @@ public class VersionGarbageCollector {
...
-    static enum RDGCType {
+    enum RDGCType {
         NO_OLD_PROP_REV_GC,
         KEEP_ONE_FULL_MODE,
         KEEP_ONE_CLEANUP_USER_PROPERTIES_ONLY_MODE,
         OLDER_THAN_24H_AND_BETWEEN_CHECKPOINTS_MODE
     }
 
-    private static RDGCType revisionDetailedGcType = RDGCType.NO_OLD_PROP_REV_GC;
+    private static RDGCType REVISION_DETAILED_GC_TYPE = RDGCType.NO_OLD_PROP_REV_GC;
```
stefan-egli: Changing this would fail `VersionGarbageCollectorIT` as it for example overwrites the value in [setUp()](https://github.com/apache/jackrabbit-oak/blob/21e8b289447790b69c78291335983b19c9471a7c/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollectorIT.java#L229)

stefan-egli: as it is non final, does it need to be changed to capitals?

rishabhdaim: ah, good point. Will revert this as well.

stefan-egli:
two tests now fail: `VersionGarbageCollectorIT.testGCDeletedPropsWithDryRunMode`, `VersionGarbageCollectorIT.testDeletedPropsAndUnmergedBCWithCollisionWithDryRunMode`, as they were asserting that all gc counts were zero - while with this change that is no longer the case

rishabhdaim:
It could be the case, since now we are collecting the stats in dry run mode.
Let me update the test cases.

stefan-egli (approved):

PR #1398, opened 2024-04-03
OAK-10733 | Filter out hidden properties while creating FlatFileStore

nit0906:


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeDocumentCodec.java
```diff
@@ -81,7 +81,12 @@ public NodeDocument decode(BsonReader reader, DecoderContext decoderContext) {
         while (reader.readBsonType() != BsonType.END_OF_DOCUMENT) {
             String fieldName = reader.readName();
             Object value = readValue(reader, fieldName);
-            nodeDocument.put(fieldName, value);
+            // Ignore hidden properties (property name starting with :)
+            // Hidden properties are not indexed and also ignored during async index updates.
+            // So it's safe to ignore them while building the FlatFileStore as well.
+            if (fieldName.charAt(0) != ':') {
```
nfsantos: I wonder if `fieldName` can ever be empty? The docs of `readName()` do not exclude this possibility. If it is empty, this would fail with an exception and the whole download would fail. I suggest to add a check for empty, just to be sure we are not surprised by empty names.

nfsantos (approved):
LGTM. Just have a suggestion for an extra check, just to be extra-safe.

fabriziofortino (approved):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeDocumentCodec.java
nit0906: makes sense.

PR #1399, opened 2024-04-03
OAK-10730 : log a warn for a MongoException previously swallowed

stefan-egli:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1752,6 +1752,8 @@ public <T extends Document> boolean create(Collection<T> collection, List<Update
                 }
                 return false;
             } catch (MongoException e) {
+                LOG.warn("Encountered MongoException while inserting documents : {} - exception : {}",
```
reschke:
```suggestion
                LOG.warn("Encountered MongoException while inserting documents: {} - exception: {}",
```

reschke (commented):
I'd remove some whitespace.

We probably should also check that these do not happen in ITs (or if they do, when...?)

stefan-egli:
There is code that asserts that `ds.create()` returns false - eg [BasicDocumentStoreTest#testCreatePartialFailure](https://github.com/apache/jackrabbit-oak/blob/f9776cf827a7c4ceed5d13e9dc97e924e9722195/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BasicDocumentStoreTest.java#L891)

Regarding ITs : you mean upstream ITs or some in oak itself?

reschke:
ITs: I was referring just to Oak ITs...

reschke (approved):

stefan-egli:
thx!

PR #1402, opened 2024-04-04
OAK-10741 : added separate logger for detailed gc

rishabhdaim:


stefan-egli (commented):
Was originally assuming the separate logger would only get audit style messages. At this point however the PR moves some log.error to the separate logger. I think error messages might require different treatment and perhaps stay with the normal logger (or we log to both, but logging only to the audit style logger seems risky as that logger might be configured differently from the default one - eg in a different file where people might not expect errors). wdyt?

rishabhdaim:
Logging into 2 separate log files would be an overkill. We should keep the `error` log in the normal logger.
Added in https://github.com/apache/jackrabbit-oak/pull/1402/commits/ae45ea5935c6d27e7375e6a25ecfd3b920014f77

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1048,8 +1061,9 @@ public void collectGarbage(final NodeDocument doc, final GCPhases phases) {
                     updateOpList.add(op);
                 }
             }
-            if (log.isDebugEnabled()) {
-                log.debug("UpdateOp for {} is {}", doc.getId(), op);
+            if (DETAILED_GC_LOG.isDebugEnabled() && op.hasChanges()) {
+                // only log in case of changes & debug level enabled
+                DETAILED_GC_LOG.debug("UpdateOp for {} is {}", doc.getId(), op);
```
stefan-egli, line 1125: Not sure we should have the entire UpdateOp in the audit log - it can be very noisy. Also not sure if logging the property names is fine here... (we could move it to log.trace, perhaps, not sure..)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1637,8 +1660,8 @@ private int removeUnusedPropertyEntries(NodeDocument doc,
                     // via collectUnmergedBranchCommits. Checking for that next.
                     Operation c = updateOp.getChanges().get(new Key(propertyKey, localRev));
                     if (c == null) {
-                        if (log.isTraceEnabled()) {
-                            log.trace("removeUnusedPropertyEntries : removing property key {} with revision {} from doc {}",
+                        if (DETAILED_GC_LOG.isTraceEnabled()) {
+                            DETAILED_GC_LOG.trace("removeUnusedPropertyEntries : removing property key {} with revision {} from doc {}",
```
stefan-egli, line 1769: Not sure if we should have this in the audit log...?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1613,8 +1636,8 @@ private int removeUnusedPropertyEntries(NodeDocument doc,
                 // could be due to node not existing or current value being in a split
                 // doc - while the former is unexpected, the latter might happen.
                 // in both cases let's skip this property
-                if (log.isDebugEnabled()) {
-                    log.debug("removeUnusedPropertyEntries : no visible revision for property {} in doc {}",
+                if (DETAILED_GC_LOG.isDebugEnabled()) {
+                    DETAILED_GC_LOG.debug("removeUnusedPropertyEntries : no visible revision for property {} in doc {}",
```
stefan-egli, line 1745: I guess we should decide how we treat property names - but this one is similar to the others and I think it should be log.trace

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1664,21 +1687,22 @@ public void close() {
         public void removeGarbage(final VersionGCStats stats) {
 
             if (updateOpList.isEmpty() && orphanOrDeletedRemovalMap.isEmpty()) {
-                if (log.isDebugEnabled() || isDetailedGCDryRun) {
-                    log.debug("Skipping removal of detailed garbage, cause no garbage detected");
+                if (DETAILED_GC_LOG.isDebugEnabled() || isDetailedGCDryRun) {
+                    DETAILED_GC_LOG.debug("Skipping removal of detailed garbage, cause no garbage detected");
```
stefan-egli, line 1796: is this needed in the audit?

stefan-egli:
(Sorry about the conflict I caused with OAK-10743!)

I think we should revisit what we want in the audit log and what in the normal log. Initially thought the audit log should be as compact as possible and not leak any information (eg property names might be risky).

rishabhdaim:
Created a separate jira for audit logger OAK-10753 and merged this PR.

PR #1409, opened 2024-04-08
OAK-10758: Proposal: Add system property to control noCursorTimeout in Mongo iterator

Joscorbe:
This is just a proposal to solve https://issues.apache.org/jira/browse/OAK-10758 

Another proposal PR is coming, that improves exception handling to avoid crashing when this exception is thrown.

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
@@ -247,6 +247,19 @@ public class MongoDocumentStore implements DocumentStore {
             Long.getLong("oak.mongo.maxQueryTimeMS", TimeUnit.MINUTES.toMillis(1));
 
     /**
+     * Flag to control the cursor timeout behavior in MongoDB.
+     * <p>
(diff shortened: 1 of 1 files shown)


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -867,6 +880,8 @@ && canUseModifiedTimeIdx(startValue)) {
                     result.maxTime(maxQueryTime, TimeUnit.MILLISECONDS);
                 }
 
+                result.noCursorTimeout(noCursorTimeout);
```
```suggestion
                if (noCursorTimeout) {
                    result.noCursorTimeout(noCursorTimeout);
                }
```

stefan-egli (commented):
What about only calling `noCursorTimeout()` if it is set to `true` ? (would leave things unchanged if this property is not set or set to `false`)

PS: as per [docu](https://www.mongodb.com/docs/manual/reference/method/cursor.noCursorTimeout/#session-idle-timeout-overrides-nocursortimeout) there could still be a MongoCursorNotFoundException thrown if the _session_ times out (eg after 30min). I.e. if the problem persists even with `noCursorTimeout=true` we might have to shrink the batch size (or introduce explicit `refreshSessions` calls)

or put differently : a smaller batch size could also be something to look into in general

stefan-egli (approved):

rishabhdaim (approved):

PR #1413, opened 2024-04-15
OAK-6762: Convert oak-blob to OSGi R7 annotations

mbaedke:
done


amit-jain (approved):
Looks good to me (with my limited understanding)

PR #1417, opened 2024-04-15
OAK-10753 : added AUDIT logger for DetailedGC

rishabhdaim:


oak-store-spi/src/main/java/org/apache/jackrabbit/oak/spi/state/AbstractNodeState.java
```diff
@@ -50,6 +52,7 @@
  * alternatives.
  */
 public abstract class AbstractNodeState implements NodeState {
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractNodeState.class.getName());
```
stefan-egli: Even though the logging is protected with an `isDebugEnabled()` - as this is quite a core class I'd like to better understand - is this really needed, and what for? could we do without this?

rishabhdaim: This logging will help us in case we are failing in verification during DGC.
Since these logs would only execute in case nodes are not equal and debugging is enabled they shouldn't be causing any noise.

stefan-egli: is this still needed with https://github.com/apache/jackrabbit-oak/pull/1421 ?
(I would like to try to avoid this even if it's protected and thus not noisy - as it is frequently executed code - that is, unless it can't be done any other way)

rishabhdaim: I don't have a strong opinion on keeping this change.
I think we can remove this.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -96,7 +96,8 @@ public class RevisionsCommand implements Command {
     private static final List<String> LOGGER_NAMES = of(
             "org.apache.jackrabbit.oak.plugins.document.VersionGarbageCollector",
             "org.apache.jackrabbit.oak.plugins.document.NodeDocumentSweeper",
-            "org.apache.jackrabbit.oak.plugins.document.VersionGarbageCollector.detailedGC"
+            "org.apache.jackrabbit.oak.plugins.document.VersionGarbageCollector.auditDGC",
+            "org.apache.jackrabbit.oak.spi.state.AbstractNodeState"
```
stefan-egli: is this still needed with he https://github.com/apache/jackrabbit-oak/pull/1417/commits/75eec42dfe4030de30254da688ddb9cbc60c57d7 revert?

rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/1417/commits/d4406db8cfcf27f5197638d235cdeb216944a4f6

stefan-egli (commented):
I did actually have something slightly different in mind wrt an "audit logger", eg something along the lines of:
* `DELETE document 3:/a/b/c`
* `DELETE 3 properties, 7 revisions in 4:/c/d/e/f`

So a bit similar to how this is done in sling.event [here](https://github.com/apache/sling-org-apache-sling-event/blob/master/src/main/java/org/apache/sling/event/impl/jobs/JobManagerImpl.java#L596-L602) : log out only the resulting action and leave the details to the normal logger. 

Wdyt?

rishabhdaim:
Let me make some changes as suggested.

Added in https://github.com/apache/jackrabbit-oak/pull/1417/commits/4b3e2e057af8ff96fa8f5c959e62e2ce37b4fc1d

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1993,17 +1996,18 @@ private boolean verifyViaTraversedState(NodeState traversedState, NodeState trav
...
             final RevisionVector lastRev;
             if (traversedParent == null && newDoc.getPath().isRoot()) {
                 if (!(traversedState instanceof DocumentNodeState)) {
-                    log.error("verify : traversedState not a DocumentNodeState : {}", traversedState.getClass());
+                    log.error("verify : traversedState not a DocumentNodeState : [{}]", traversedState.getClass());
                     return false;
                 }
                 lastRev = ((DocumentNodeState) traversedState).getLastRevision();
             } else {
+                // in case we are reaching here, it means that node is none of parents are null
```
stefan-egli, line 2010: (except in case of a verification error I guess?)

stefan-egli (approved):
ok, let's go with this. It might still be a bit a noisy audit log but we can revisit and adjust later as well.

PR #1418, opened 2024-04-15
OAK-10763 : fixed the skipping of orphan nodes in case parent is not …

rishabhdaim:
…existing


stefan-egli:
also note my alternative PR suggestion at https://github.com/apache/jackrabbit-oak/pull/1421

rishabhdaim:
With https://github.com/apache/jackrabbit-oak/pull/1421  PR, we are skipping the opportunity to clean other types of garbage.
If we want to go via https://github.com/apache/jackrabbit-oak/pull/1421 , we will keep this garbage till we roll out non-gap orphan deletion.

IMO, even with https://github.com/apache/jackrabbit-oak/pull/1421, the logic in this PR is correct i.e. we should be able to delete stuff in case parent/ancestors don't exist.

stefan-egli:
As discussed offline I'd argue that an orphan node should be removed entirely - if a gc mode prevents that, it should be skipped. That makes verification stronger and keeps the stats correct.

PR #1419, opened 2024-04-15
OAK-10751 : added compaction support in oak-run

rishabhdaim:


oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -349,28 +357,24 @@ private void printInfo(VersionGarbageCollector gc, RevisionsOptions options)
...
+                System.out.println("Stopping Revision GC...");
+                running.set(false);
+                gc.cancel();
+                finished.acquireUninterruptibly();
+                System.out.println("Stopped Revision GC.");
             }));
             if (options.isContinuous()) {
                 while (running.get()) {
-                    long lastRun = System.currentTimeMillis();
+                    long lastRun = currentTimeMillis();
```
stefan-egli, line 377: Just as a comment, I don't see this as an improvement, the old variant for me at least is more readable and thus more user friendly - the other issue with this is that it increases the change surface of this change (and of the feature branch), i.e. we'll then have code changes that are unrelated (unnecessary) and it makes merging/backporting/source-lookups more difficult.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -349,28 +357,24 @@ private void printInfo(VersionGarbageCollector gc, RevisionsOptions options)
...
-                    finished.acquireUninterruptibly();
-                    System.out.println("Stopped Revision GC.");
-                }
+            Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+                System.out.println("Detected QUIT signal.");
+                System.out.println("Stopping Revision GC...");
+                running.set(false);
+                gc.cancel();
+                finished.acquireUninterruptibly();
+                System.out.println("Stopped Revision GC.");
```
stefan-egli, line 373: Same thing here - this might be something to discuss in general in oak, but I fail to see the advantage of doing these changes (see other comment also).

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -320,15 +330,13 @@ private VersionGarbageCollector bootstrapVGC(RevisionsOptions options,
         return gc;
     }
 
-    private void info(RevisionsOptions options, Closer closer)
-            throws IOException {
+    private void info(RevisionsOptions options, Closer closer) throws IOException {
```
stefan-egli, line 333: Similar here - these are formatting changes that increase change surface making porting effort more expensive.

stefan-egli (approved):
added a few general comments but they are not blocking

PR #1427, opened 2024-04-19
OAK-10771 - Add disk cache size stats and issue warning if evicted segment has zero length

ahanikel:
The persistent disk cache computes its size internally by adding/subtracting the size of added/purged segments. We would like to be able to see if that computation is correct, by having both the computed size and the effective size on disk in the metrics.

The background for this is that in a few rare instances, the disk cache evicted too many items (down to almost zero) instead of the expected 65% of its max size.

Also we want to make sure that no evictions happen on items that have a length of zero. We're not sure if that happens or not, so the change is to refuse eviction in that case and log a warning instead.

### DIFF
--- a/oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
+++ b/oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
@@ -204,7 +204,16 @@ public class PersistentDiskCache extends AbstractPersistentCache {
                     }
                     if (cacheSize.get() > maxCacheSizeBytes * 0.66) {
                         File segment = segmentCacheEntry.getPath().toFile();
-                        cacheSize.addAndGet(-segment.length());
+                        long length = segment.length();
+                        if (length == 0) {
+                            if (logger.isDebugEnabled()) {
+                                logger.debug("Avoiding removal of zero-sized file {}", segmentCacheEntry.getPath());
+                            } else {
+                                logger.warn("Avoiding removal of zero-sized file.");
+                            }
+                            return;
+                        }
+                        cacheSize.addAndGet(-length);
                         segment.delete();
                         evictionCount.incrementAndGet();
                     } else {
--- a/oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
+++ b/oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
@@ -61,7 +61,7 @@ public class PersistentDiskCache extends AbstractPersistentCache {
 
     private final File directory;
     private final long maxCacheSizeBytes;
-    private final IOMonitor diskCacheIOMonitor;
+    private final DiskCacheIOMonitor diskCacheIOMonitor;
     /**
      * Wait time before attempting to clean up orphaned temp files
      */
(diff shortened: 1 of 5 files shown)


oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
```diff
@@ -204,7 +204,17 @@ private void cleanUpInternal() {
...
+                        if (length == 0) {
+                            if (logger.isDebugEnabled()) {
+                                logger.debug("Avoiding removal of zero-sized file {}", segmentCacheEntry.getPath());
+                            } else {
+                                logger.warn("Avoiding removal of zero-sized file.");
+                            }
+                            return;
+                        }
+                        long cacheSizeAfter = cacheSize.addAndGet(-length);
+                        diskCacheIOMonitor.updateCacheSize(cacheSizeAfter, directory.length());
```
smiroslav: https://docs.oracle.com/javase/8/docs/api/java/io/File.html#length--

API docs says that File.length returns unspecified value for directories.

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
```diff
@@ -204,7 +204,17 @@ private void cleanUpInternal() {
                     }
                     if (cacheSize.get() > maxCacheSizeBytes * 0.66) {
                         File segment = segmentCacheEntry.getPath().toFile();
-                        cacheSize.addAndGet(-segment.length());
+                        long length = segment.length();
+                        if (length == 0) {
+                            if (logger.isDebugEnabled()) {
```
smiroslav, line 210: Why both debug and warn are needed.

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
joerghoh: It probably should be this:

```suggestion
                        diskCacheIOMonitor.updateCacheSize(cacheSizeAfter, segment.length());
```

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
ahanikel: @smiroslav The debug logs the path which could mean a lot of logging if those zero-sized files occur often. The warnings are always the same string without the file name, so if they occur frequently, they are suppressed and logged as "last warning was repeated nnn times" (or something like that).

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
ahanikel: You're both right, the directory size can only be calculated by traversing the directory, the size attribute only shows the size of the directory structure itself on disk. I'm going to change this to measure the calculated size and the size change instead.

ahanikel:
@smiroslav Could you have another look at this one so we can also have it in the 1.64 release?

smiroslav (approved):

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/package-info.java
```diff
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-@Version("1.0.0")
```
reschke, line 17: Is this a public API?

reschke: If it's for use inside Oak only, it should be marked "@Internal".

PR #1435, opened 2024-04-25
OAK-10778 - Support downloading from Mongo in parallel. 

nfsantos:
A Mongo cluster (`REPLICA_SET`) usually consists of one primary and two secondaries. This PR adds support to download in parallel from the two secondaries.

Adds new boolean system property: `oak.indexer.pipelined.mongoParallelDump` defaults to false.

Parallel download can be enabled only when `oak.indexer.pipelined.retryOnConnectionErrors` is true. (Parallel download requires ordered traversals on Mongo which are enabled by retry on connection errors.) 

## Goals of this PR ##
- Do not download from the primary because this is more likely to slow down the operation of the Mongo and affect the other workloads (read/writes by Oak). 
-  Use one and only one connection at a time to each secondary to avoid overloading any given replica. More on this in the notes below.
- Gracefully handle scale up/down operations and failures in general, pausing the second download when there is only one secondary available and reconnecting when all secondaries are up.   

## Implementation
One of the difficulties of parallel downloading is to partition the range of documents among the download threads. We do not know in advance the distribution of the documents over the range of keys `(_modified, _id)` which are used to download, so it is challenging to distribute them evenly among the download threads. This PR sidesteps this problem by having one thread download in ascending and the other in descending order. This has the limitation that we can only have 2 parallel threads, but this fits nicely the typical configuration of a Mongo cluster with two secondaries. Adding more parallel downloads would provide a smaller increase in overall download speed and would risk overloading the replicas.

The two download threads coordinate to check when the ranges they have downloaded have crossed, to stop the download at this point.

To distribute the threads among replicas this PR creates a custom implementation of [ServerSelector](https://mongodb.github.io/mongo-java-driver/3.12/javadoc/com/mongodb/selector/ServerSelector.html). The Mongo Java driver calls the registered ServerSelectors whenever it needs to open a connection to a Mongo cluster to get a list of eligible servers. The default implementations are taken from the `readPreference` settings, but it is possible to create and register a custom implementation. This PR uses an implementation that allows connections only to secondaries and keeps track of which thread last received a given secondary, so that this secondary is not given to any other thread. 

There is also a mechanism to disconnect from a replica who was promoted from secondary to primary. This happens whenever there is a scale up/down. Mongo will first take one secondary down and replace it by a new one, then the other secondary, and before taking down the primary, it promotes one of the new secondaries to primary. Therefore, we may have connected to a secondary which in the meantime was promoted to primary. Since is very likely to happen during a scale up/down, it was important to detect promotions of secondaries to primary and disconnect. This is done by listening to [ClusterListener](https://mongodb.github.io/mongo-java-driver/3.12/javadoc/com/mongodb/event/ClusterListener.html) events and having each download thread periodically query to check if the replica it is using is now the Primary. If it is, the replica disconnects and tries to establish a new connection, which will be redirected to a secondary.

### System 1
Sequential download:
```
Timings:
  Mongo dump: 12:09:11
  Merge sort: 00:05:28
  Build FFS (Dump+Merge): 12:14:40
```

Parallel download
```
Timings:
  Mongo dump: 06:14:16
  Merge sort: 00:05:26
  Build FFS (Dump+Merge): 06:19:46
```

### System 2

Sequential download:
```
Timings:
  Mongo dump: 00:14:10
  Merge sort: 00:02:19
  Build FFS (Dump+Merge): 00:16:29
```

Parallel download
```
Timings:
  Mongo dump: 00:06:41
  Merge sort: 00:02:13
  Build FFS (Dump+Merge): 00:08:54
```

## Note on risk of slowing down the Mongo cluster
When parallel download is enabled, both secondaries will be under load. There may be a concern if this will slow down writes, which need an ack from the primary and a secondary. I do not think this would be a problem in the common cases. A single download connection should not be enough to saturate a node. Mongo seems to allocate only one thread to handle a connection, so as long as the Mongo node has at least 2 cores, the download query will not take up all the CPU. The pressure on IOPS, disk throughput and network bandwidth created by the downloader may in some situations get close to the limit, but even there this pressure is not kept all the time. The protocol used by the Mongo Java driver is synchronous request/response, and the client only sends the request for the next batch of results after the current one is parsed and iterated over, so during this time the Mongo server is idle. I have observed that the IOPS and disk throughput usually stay below 80% on Mongo, which gives some spare headroom to process other queries. 

# Additional changes
- Improve the test of recovery from disconnections from Mongo. The previous test was relying on mockito to simulate a connection failure. This was very complex and tedious to write. This PR replaces that test by one that uses a real Mongo server inside a Docker container and toxiproxy to simulate a connection failure. 
- Create a parametrized version of PipelinedIT tests which test all combinations of: regex path filtering, parallel download and retry on connection errors. 

This PR has a large change set in part because it refactors the PipelinedMongoDownloadTask in order to make it more concise and cohesive:
- It moves out the logic to handle regex filtering to a separate class
- It creates a new DownloadTask class that includes the logic to do the actual download from Mongo, trying to separate it from the logic to setup and launch download threads.

### DIFF
--- a/oak-run-commons/pom.xml
+++ b/oak-run-commons/pom.xml
@@ -204,16 +204,21 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>com.github.stefanbirkner</groupId>
-            <artifactId>system-rules</artifactId>
-            <version>1.19.0</version>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>toxiproxy</artifactId>
+            <version>${testcontainers.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
-            <!-- see https://issues.apache.org/jira/browse/OAK-7787 -->
-            <groupId>javax.activation</groupId>
-            <artifactId>activation</artifactId>
-            <version>1.1.1</version>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>mongodb</artifactId>
+            <version>${testcontainers.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.github.stefanbirkner</groupId>
+            <artifactId>system-rules</artifactId>
+            <version>1.19.0</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoDownloaderRegexUtils.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoDownloaderRegexUtils.java
@@ -150,13 +150,11 @@ class MongoDownloaderRegexUtils {
         TreeSet<String> ancestors = new TreeSet<>();
         for (String child : paths) {
             String parent = child;
-            while (true) {
+            while (!PathUtils.denotesRoot(parent)) {
                 ancestors.add(parent);
-                if (PathUtils.denotesRoot(parent)) {
-                    break;
-                }
                 parent = PathUtils.getParentPath(parent);
             }
+            ancestors.add(parent); // add the root path as well
         }
         return new ArrayList<>(ancestors);
     }
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoDownloaderRegexUtils.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoDownloaderRegexUtils.java
@@ -1,0 +1,152 @@
+package org.apache.jackrabbit.oak.index.indexer.document.flatfile.pipelined;
+
+import com.mongodb.client.model.Filters;
+import org.apache.jackrabbit.oak.commons.PathUtils;
+import org.apache.jackrabbit.oak.plugins.document.NodeDocument;
+import org.apache.jackrabbit.oak.plugins.document.util.Utils;
+import org.bson.conversions.Bson;
+import org.jetbrains.annotations.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.TreeSet;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+class MongoDownloaderRegexUtils {
+    private static final Logger LOG = LoggerFactory.getLogger(PipelinedMongoDownloadTask.class);
+
+    final static Pattern LONG_PATH_ID_PATTERN = Pattern.compile("^[0-9]{1,3}:h");
+
+    /**
+     * Creates the filter to be used in the Mongo query
+     *
+     * @param mongoFilterPaths          The paths to be included/excluded in the filter. These define subtrees to be included or excluded.
+     *                                  (see {@link MongoRegexPathFilterFactory.MongoFilterPaths} for details)
+     * @param customExcludeEntriesRegex Documents with paths matching this regex are excluded from download
+     * @return The filter to be used in the Mongo query, or null if no filter is required
+     */
+    static Bson computeMongoQueryFilter(@NotNull MongoRegexPathFilterFactory.MongoFilterPaths mongoFilterPaths, String customExcludeEntriesRegex) {
+        var filters = new ArrayList<Bson>();
+
+        List<Pattern> includedPatterns = compileIncludedDirectoriesRegex(mongoFilterPaths.included);
+        if (!includedPatterns.isEmpty()) {
+            // The conditions above on the _id field is not enough to match all JCR nodes in the given paths because nodes
+            // with paths longer than a certain threshold, are represented by Mongo documents where the _id field is replaced
+            // by a hash and the full path is stored in an additional field _path. To retrieve these long path documents,
+            // we could add a condition on the _path field, but this would slow down substantially scanning the DB, because
+            // the _path field is not part of the index used by this query (it's an index on _modified, _id). Therefore,
+            // Mongo would have to retrieve every document from the column store to evaluate the filter condition. So instead
+            // we add below a condition to download all the long path documents. These documents can be identified by the
+            // format of the _id field (<n>:h<hash>), so it is possible to identify them using only the index.
+            // This might download documents for nodes that are not in the included paths, but those documents will anyway
+            // be filtered in the transform stage. And in most repositories, the number of long path documents is very small,
+            // often there are none, so the extra documents downloaded will not slow down by much the download. However, the
+            // performance gains of evaluating the filter of the query using only the index are very significant, especially
+            // when the index requires only a small number of nodes.
+            var patternsWithLongPathInclude = new ArrayList<>(includedPatterns);
+            patternsWithLongPathInclude.add(LONG_PATH_ID_PATTERN);
+            filters.add(Filters.in(NodeDocument.ID, patternsWithLongPathInclude));
+        }
+
+        // The Mongo filter returned here will download the top level path of each excluded subtree, which in theory
+        // should be excluded. That is, if the tree /a/b/c is excluded, the filter will download /a/b/c but none of
+        // its descendants.
+        // This is done because excluding also the top level path would add extra complexity to the filter and
+        // would not have any measurable impact on performance because it only downloads a few extra documents, one
+        // for each excluded subtree. The transform stage will anyway filter out these paths.
+        List<Pattern> filterPatterns = compileExcludedDirectoriesRegex(mongoFilterPaths.excluded);
+        filterPatterns.forEach(p -> filters.add(Filters.regex(NodeDocument.ID, p)));
+
+        Pattern customRegexExcludePattern = compileCustomExcludedPatterns(customExcludeEntriesRegex);
+        if (customRegexExcludePattern != null) {
+            filters.add(Filters.regex(NodeDocument.ID, customRegexExcludePattern));
+        }
+
+        if (filters.isEmpty()) {
+            return null;
+        } else if (filters.size() == 1) {
+            return filters.get(0);
+        } else {
+            return Filters.and(filters);
+        }
+    }
+
+    private static List<Pattern> compileIncludedDirectoriesRegex(List<String> includedPaths) {
+        return compileDirectoryRegex(includedPaths, false);
+    }
+
+    private static List<Pattern> compileExcludedDirectoriesRegex(List<String> excludedPaths) {
+        return compileDirectoryRegex(excludedPaths, true);
+    }
+
+    private static List<Pattern> compileDirectoryRegex(List<String> paths, boolean negate) {
+        if (paths.isEmpty()) {
+            return List.of();
+        }
+        if (paths.size() == 1 && paths.get(0).equals("/")) {
+            return List.of();
+        }
+        ArrayList<Pattern> patterns = new ArrayList<>();
+        for (String path : paths) {
+            if (!path.endsWith("/")) {
+                path = path + "/";
+            }
+            String regex = "^[0-9]{1,3}:" + Pattern.quote(path);
+            Pattern pattern;
+            if (negate) {
+                pattern = compileExcludedDirectoryRegex(regex);
+            } else {
+                pattern = Pattern.compile(regex);
+            }
+            patterns.add(pattern);
+        }
+        return patterns;
+    }
+
+    static Pattern compileExcludedDirectoryRegex(String regex) {
+        // https://stackoverflow.com/questions/1240275/how-to-negate-specific-word-in-regex
+        return Pattern.compile("^(?!" + regex + ")");
+    }
+
+    static Pattern compileCustomExcludedPatterns(String customRegexPattern) {
+        if (customRegexPattern == null || customRegexPattern.trim().isEmpty()) {
+            LOG.info("Mongo custom regex is disabled");
+            return null;
+        } else {
+            LOG.info("Excluding nodes with paths matching regex: {}", customRegexPattern);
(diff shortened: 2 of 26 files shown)


oak-run-commons/pom.xml
```diff
@@ -204,16 +204,21 @@
...
-            <artifactId>system-rules</artifactId>
-            <version>1.19.0</version>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>toxiproxy</artifactId>
+            <version>${testcontainers.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
-            <!-- see https://issues.apache.org/jira/browse/OAK-7787 -->
-            <groupId>javax.activation</groupId>
```
nit0906, line 214: Want to confirm if the dependency removal was not an oversight ?

nfsantos: It was intentional. It's a test dependency added as a workaround to an old issue. And the tests pass without it, so I don't think it's needed anymore.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoServerSelector.java
```diff
@@ -0,0 +1,179 @@
...
+        lastSeenClusterDescription.getServerDescriptions().stream()
+                .filter(ServerDescription::isPrimary)
+                .map(ServerDescription::getAddress)
+                .forEach(primaryAddress -> {
+                    for (var entry : serverAddressHashMap.entrySet()) {
+                        if (entry.getValue().equals(primaryAddress)) {
+                            connectedToPrimaryThreads.add(entry.getKey());
+                        }
+                    }
+                });
```
fabriziofortino, line 151: to be consistent in terms of style, you can write this logic in a fully functional way. This would require `connectedToPrimaryThreads` to be set as non final though. It's a matter of taste.

```
private Set<Long> connectedToPrimaryThreads = new HashSet<>();
```

```suggestion
        connectedToPrimaryThreads = lastSeenClusterDescription.getServerDescriptions().stream()
    .filter(ServerDescription::isPrimary)
    .map(ServerDescription::getAddress)
    .flatMap(primaryAddress -> serverAddressHashMap.entrySet().stream()
        .filter(entry -> primaryAddress.equals(entry.getValue()))
        .map(Map.Entry::getKey))
    .collect(Collectors.toSet());
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoParallelDownloadCoordinator.java
```diff
@@ -0,0 +1,138 @@
...
+ */
+package org.apache.jackrabbit.oak.index.indexer.document.flatfile.pipelined;
+
+import org.apache.jackrabbit.oak.plugins.document.NodeDocument;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Coordinates the two parallel download streams used to download from Mongo when parallelDump is enabled. One stream
+ * downloads in ascending order the other in descending order. This class keeps track of the top limit of the ascending
+ * stream and of the bottom limit of the descending stream, and determines if the streams have crossed. This indidcates
```
fabriziofortino: typo
```suggestion
 * stream and of the bottom limit of the descending stream, and determines if the streams have crossed. This indicates
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoRegexPathFilterFactory.java
```diff
@@ -170,9 +170,9 @@ static List<String> mergeIndexAndCustomExcludePaths(List<String> indexExcludedPa
             return indexExcludedPaths;
         }
 
-        var excludedUnion = new HashSet<>(indexExcludedPaths);
```
fabriziofortino, line 173: out of curiosity: I have seen you have replaced vars in multiple places. Any specific reason?

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoTestBackend.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.jackrabbit.oak.index.indexer.document.flatfile.pipelined;
+
```
fabriziofortino, line 20: nitpick: extra line
```suggestion
```

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelineITUtil.java
```diff
@@ -0,0 +1,216 @@
...
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import static org.junit.Assert.assertEquals;
+
```
fabriziofortino, line 52: nitpick: extra line
```suggestion
```

fabriziofortino (approved):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoRegexPathFilterFactory.java
nfsantos: No specific reason, I'm still unsure about when to use var vs explicit type. In many cases keeping the type annotation makes the code more clear and does not significantly increase verbosity, for instance:
```
var i = 1;

int i = 1;
```
The second option makes it clear that it is an int, while the first is ambiguous.

In other cases, even if the type annotation is longer, I still somewhat prefer that it appears at least one time. For instance, here I feel it's better to use var:
```
var bi = new DownloadPosition(batch[i].getModified(), batch[i].getId());
```
Because the type is explicit on the right hand side.

But here:
```
            FindIterable<NodeDocument> mongoIterable = dbCollection
                    .find(findQuery)
                    .sort(sortOrder);
```
I think it's nicer to have the type annotation because the right hand side does not contain the type.

But it's just my gut feeling of what seems more clear and easy to read, I'm not following any set of best-practices. It would be interesting to have a discussion about the use of var.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoServerSelector.java
nfsantos: I initially wrote the code all as streams, but I did not like the stream-within-a-stream nesting and the use of flatMap. I feel that the forEach loop is a bit more clear.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoDownloaderRegexUtils.java
```diff
@@ -0,0 +1,170 @@
...
+            while (true) {
+                ancestors.add(parent);
+                if (PathUtils.denotesRoot(parent)) {
+                    break;
+                }
+                parent = PathUtils.getParentPath(parent);
+            }
+        }
+        return new ArrayList<>(ancestors);
+    }
```
steffenvan: Since we only use the `TreeSet` to maintain a sorted order in the end (and not intermittently) I think we could use a hash set and sort in the end, but that's just a minor thing. In theory it's still the same O-complexity, but in practice I _think_ it should be faster because of the cache locality benefits of sorting an `ArrayList`. 

But I would prefer to change the `while(true)` to condition on the the root check instead. 

```suggestion
static List<String> getAncestors(List<String> paths) {
    HashSet<String> ancestors = new HashSet<>();
    for (String child : paths) {
        String parent = child;
        while (!PathUtils.denotesRoot(parent)) {
            ancestors.add(parent);
            parent = PathUtils.getParentPath(parent);
        }
        ancestors.add(parent); // add the root path as well
    }
    List<String> sortedAncestors = new ArrayList<>(ancestors);
    Collections.sort(sortedAncestors);
    
    return sortedAncestors;
}
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -581,94 +498,295 @@ private MongoFilterPaths getPathsForRegexFiltering() {
...
+                    numberOfFailures = 0;
                 }
-            } catch (MongoException e) {
-                if (e instanceof MongoInterruptedException || e instanceof MongoIncompatibleDriverException) {
-                    // Non-recoverable exceptions
+                try {
+                    while (cursor.hasNext()) {
+                        NodeDocument next = cursor.next();
+                        String id = next.getId();
+                        this.nextLastModified = next.getModified();
```
steffenvan, line 650: My editor warns me that this might produce a `NullPointerException`. And we use this a few other places.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoParallelDownloadCoordinator.java
```diff
@@ -0,0 +1,138 @@
...
+     * Updates the lower range top to b[i].
+     *
+     * @param batch The batch of documents to be added to the lower range, must be in ascending order
+     */
+    public synchronized int extendLowerRange(NodeDocument[] batch, int sizeOfBatch) {
+        // batch must be in ascending order
+        int i = sizeOfBatch - 1;
+        // Start by the highest value in the range and compare it with the bottom of the upper range.
+        while (i >= 0) {
+            var bi = new DownloadPosition(batch[i].getModified(), batch[i].getId());
```
steffenvan, line 93: the `batch[i].getModified()` might also produce a `NullPointerException`.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoParallelDownloadCoordinator.java
```diff
@@ -0,0 +1,138 @@
...
+     * the first/highest document in this batch that was already downloaded by the ascending download thread.
+     *
+     * @param batch The batch of documents to be added to the upper range, must be in descending order
+     */
+    public synchronized int extendUpperRange(NodeDocument[] batch, int sizeOfBatch) {
+        // batch must be in descending order
+        int i = sizeOfBatch - 1;
+        // Find the highest value in the batch that is not yet on the upper range of values downloaded
+        while (i >= 0) {
+            var bi = new DownloadPosition(batch[i].getModified(), batch[i].getId());
```
steffenvan, line 119: Same here

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeDocumentCodec.java
```diff
@@ -98,8 +99,7 @@ public Class<NodeDocument> getEncoderClass() {
         return NodeDocument.class;
     }
 
-    private Object readValue(BsonReader reader, String fieldName) {
-        estimatedSizeOfCurrentObject += 16 + fieldName.length();
+    private Object readValue(BsonReader reader, String fieldName, int[] estimatedSizeOfCurrentObject) {
```
steffenvan: Question: 
Why is `estimatedSizeOfCurrentObject` an `int[]`? I get that we pass it around so that we can modify it as a side effect in `readValue`. And I'm not sure this is the best approach from a design perspective.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoDownloaderRegexUtils.java
nfsantos: This code is not on the critical path, so I prefer to use the TreeSet to both sort and remove duplicates. I have simplified the while loop as suggested.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeDocumentCodec.java
nfsantos: I wanted to avoid creating a new class. But I agree it is not elegant. I have changed to use the MutableInt in Apache Commons, which is already in the classpath.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
nfsantos: Good catch, this would indeed be a problem when doing a column traversal because the query used for that was downloading all documents, even those without the `_modified` field. We do not need these documents to build the FFS, so it is safe to filter them on the Mongo query. Like this, there is not need to have a null check here. This code is on the critical path, so it should be kept as lean as possible.

PR #1438, opened 2024-05-03
OAK-10784: o.a.j.o.plugins.migration.version.VersionableEditor should…

mbaedke:
… create the version storage node, if needed

Implemented; improved logging.


reschke (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/migration/version/VersionableEditor.java
```diff
@@ -79,7 +80,7 @@ public class VersionableEditor extends DefaultEditor {
 
     private VersionableEditor(Provider provider, NodeBuilder rootBuilder) {
         this.rootBuilder = rootBuilder;
-        this.versionStorage = getVersionStorage(rootBuilder);
+        this.versionStorage = createVersionStorage(rootBuilder);
```
reschke, line 83: Hm, on secomd thought... I'd like to understand under what circumstances there can be an Oak repository without version storage. Isn't it always created by default?

mbaedke: Yes, but the migration works completely on the nodestore API.

PR #1441, opened 2024-05-09
OAK-10780: add azure access token refresh logic

t-rana:
Azure Access token is valid for 1 hour only. Added code to check for token expiry every 4 minutes, and refresh the token when the same is close to expiry.


oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
```diff
@@ -207,6 +230,61 @@ public void write(@NotNull byte[] bytes, int offset, int length) {
...
+            this.clientSecretCredential = clientSecretCredential;
+            this.accessToken = accessToken;
+            this.storageCredentialsToken = storageCredentialsToken;
+        }
+
+        @Override
+        public void run() {
+            try {
+                log.debug("Checking for azure access token expiry at: {}", LocalDateTime.now());
+                OffsetDateTime tokenExpiryThreshold = OffsetDateTime.now().plusMinutes(5);
```
rootpea, line 266:
```suggestion
                OffsetDateTime tokenExpiryThreshold = OffsetDateTime.now().plusMinutes(20);
```

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
```diff
@@ -207,6 +230,61 @@ public void write(@NotNull byte[] bytes, int offset, int length) {
...
+                if (accessToken.getExpiresAt() != null && accessToken.getExpiresAt().isBefore(tokenExpiryThreshold)) {
+                    log.info("Access token is about to expire (5 minutes or less) at: {}. New access token will be generated",
```
rootpea, line 268:
```suggestion
                    log.info("Access token is about to expire (20 minutes or less) at: {}. New access token will be generated",
```

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
t-rana: azure identity serves the token from cache, and will only refresh the token when there are 5 minutes left in token expiry. so keeping a 20 minutes threshold will not help as the token will be served from cache only and will not be refreshed. hence i have taken a threshold of 5 minutes similar to what is being used in azure identity.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
t-rana: same as above.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
rootpea: There is no cache in IdentityClient.authenticateWithClientSecret.
So the token will definitely be refreshed even at 20 minutes.
With 5 minutes there is a risk of a single transient failure causing token expiration.

t-rana: I guess you are referring to an older version of azure identity.
we are using azure-identity@1.11.3. The token is first looked up in cache, and after the cache miss, a new token is generated.

https://github.com/Azure/azure-sdk-for-java/blob/azure-identity_1.11.3/sdk/identity/azure-identity/src/main/java/com/azure/identity/ClientSecretCredential.java#L126


https://github.com/Azure/azure-sdk-for-java/blob/azure-identity_1.11.3/sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentitySyncClient.java#L176

rootpea: Thanks for the clarification. In this case we should reduce the TOKEN_REFRESHER_DELAY to 1 minute so that even in the case of transient failures we will still successfully refresh before the token expires.

rootpea: Alternatively instead of ClientSecretCredential, you can create directly a ConfidentialClientApplication and set skipCache(true) in the parameters

```
// configure once, then store these for later use in the tokenrefresher:
        ConfidentialClientApplication confidentialClientApplication = ConfidentialClientApplication
                .builder(clientId, ClientCredentialFactory.createFromSecret(clientSecret))
                .authority("https://login.microsoftonline.com/" + tenantId + "/")
                .build();

        ClientCredentialParameters clientCredentialParameters =
                ClientCredentialParameters.builder(Collections.singleton(AZURE_DEFAULT_SCOPE)).skipCache(true)
                        .tenant(IdentityUtil
                                .resolveTenantId(tenantId, new TokenRequestContext().addScopes(AZURE_DEFAULT_SCOPE), new IdentityClientOptions())).build();

// refreshing anytime:
        IAuthenticationResult authResult = confidentialClientApplication.acquireToken(clientCredentialParameters).join();
        AccessToken accessToken = new AccessToken(authResult.accessToken(), OffsetDateTime.ofInstant(authResult.expiresOnDate().toInstant(), ZoneOffset.UTC));

```

t-rana: I would prefer the first way here since that is tested. Reduced the token refresh interval to 1 minute

rootpea (approved):

PR #1439, opened 2024-05-06
OAK-10786: oak-lucene: use copy of lucene-core 4.7.2 source code

reschke:
I have structured the PR so that each step should be easy to follow.

PR #1443, opened 2024-05-11
OAK-10787: oak-lucene: backport fix for lucene-core vulnerability

reschke:
This backports the "official" fix (https://github.com/apache/lucene/issues/11537) to the 4.7.2 code base.


oak-lucene/src/main/java/org/apache/lucene/util/automaton/RegExp.java
```diff
@@ -872,23 +874,39 @@ private boolean check(int flag) {
...
+  @FunctionalInterface
+  private interface MakeRegexGroup {
+    RegExp get(RegExp exp1, RegExp exp2);
+  }
+
+  final RegExp iterativeParseExp(
+      Supplier<RegExp> gather, BooleanSupplier stop, MakeRegexGroup associativeReduce)
+      throws IllegalArgumentException {
+    RegExp result = gather.get();
+    while (stop.getAsBoolean() == true) {
```
steffenvan, line 905: We can simplify to: `while (stop.getAsBoolean())`

oak-lucene/pom.xml
```diff
@@ -107,7 +107,7 @@
             <_exportcontents>
               !org.apache.lucene.queryparser.xml,
               !org.apache.lucene.queryparser.xml.builders,
-              org.apache.lucene.*;version=${lucene.version}-oak1
+              org.apache.lucene.*;version=${lucene.version}-oak2
```
steffenvan: What is `oak1` and `oak2`?

oak-lucene/src/main/java/org/apache/lucene/util/automaton/RegExp.java
reschke: We could, but I wanted to minimize the amount of changes compared to the "official" Lucene patch. (https://github.com/apache/lucene/pull/12462/files#diff-93255512c837705492a154d5794b61f7132ae4264a292d224d5944de35f45847R1100)

oak-lucene/pom.xml
reschke: lucene.version is 4.7.2. We could call all of this 4.7.3, but I assume the Lucene project would object. Thus the "oak" qualifier that should make clear that this is "our" version. (oak1 is the copy of 4.7.2 with minimal changes for JDK 11 compliance - see https://github.com/apache/jackrabbit-oak/pull/1439/commits/bca1e9c63c470967e3bee363d22c2026219d49e8 - "oak2" is this fix).

rishabhdaim (approved):

mbaedke (approved):

oak-lucene/src/main/java/org/apache/lucene/util/automaton/RegExp.java
steffenvan: Hmm okay fair makes sense.

PR #1445, opened 2024-05-13
OAK-10796 - Avoid creation of intermediate StringBuilder in JsopBuilder

nfsantos:
Since `invokedynamic` was added to Java, this code in the JsopBuilder class:

```
StringBuilder buff = new StringBuilder(length + 2);
return buff.append('\"').append(s).append('\"').toString(); 
```

can be written more succinctly and efficiently like this:
```
 return '\"' + s + '\"';
```

https://www.baeldung.com/java-string-concatenation-invoke-dynamic

Results of JsopTest micro benchmark with 100M iterations:

Baseline performance:
```
19.02 seconds dummy: 1210065408
18.67 seconds dummy: 1210065408
18.62 seconds dummy: 1210065408
18.65 seconds dummy: 1210065408
18.62 seconds dummy: 1210065408
```
With String concatenation instead of StringBuilder:
```
17.84 seconds dummy: 1210065408
17.68 seconds dummy: 1210065408
17.68 seconds dummy: 1210065408
17.72 seconds dummy: 1210065408
17.69 seconds dummy: 1210065408
```


oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsopBuilder.java
```diff
@@ -28,7 +28,7 @@ public class JsopBuilder implements JsopWriter {
 
     private static final boolean JSON_NEWLINES = false;
 
-    private StringBuilder buff = new StringBuilder();
+    private final StringBuilder buff = new StringBuilder();
```
reschke: what is this change about?

reschke (changes requested):

reschke:
FWIW, if this is merged, by all means use "squash + merge".

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsopBuilder.java
nfsantos: The field is not changed, so it should be final to convey the intention that it should not change and to get the compiler to enforce this invariant. This was causing a warning in IntelliJ and as it was in the same file that I was modifying, I added this small improvement.

steffenvan (approved):

fabriziofortino (approved):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsopBuilder.java
reschke: Please: no. Changes for a ticket should only for the issue desribed in the ticket (or extend the ticket description).

nfsantos: reverted.

reschke (approved):

PR #1447, opened 2024-05-13
OAK-10760: oak-core should check namespace registry consistency on st…

mbaedke:
…artup

Implemented.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadOnlyNamespaceRegistry.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadOnlyNamespaceRegistry.java
@@ -49,13 +49,17 @@ public class ReadOnlyNamespaceRegistry
 
     private static final Logger LOG = LoggerFactory.getLogger(ReadOnlyNamespaceRegistry.class);
 
+    private static volatile boolean CONSISTENCY_CHECKED;
+
     protected final Tree namespaces;
     protected final Tree nsdata;
 
     public ReadOnlyNamespaceRegistry(Root root) {
         this.namespaces = root.getTree(NAMESPACES_PATH);
         this.nsdata = namespaces.getChild(REP_NSDATA);
-        checkConsistency();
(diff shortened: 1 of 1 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadOnlyNamespaceRegistry.java
```diff
@@ -16,8 +16,8 @@
  */
 package org.apache.jackrabbit.oak.plugins.name;
 
-import static org.apache.jackrabbit.guava.common.collect.Iterables.toArray;
 import static java.util.Collections.emptyList;
+import static org.apache.jackrabbit.guava.common.collect.Iterables.toArray;
```
reschke: Can we remove the Guava-ism while we are at it?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadOnlyNamespaceRegistry.java
```diff
@@ -112,4 +120,53 @@ public String getPrefix(String uri) throws NamespaceException {
                 "No namespace prefix registered for URI " + uri);
     }
 
+    protected void checkConsistency() {
+        final String jcrPrimaryType = "jcr:primaryType";
+        List<String> prefixes = Arrays.asList(getPrefixes());
+        List<String> uris = Arrays.asList(getURIs());
+        if (prefixes.size() != uris.size()) {
+            LOG.error("The namespace registry is inconsistent: found {} registered prefixes and {} registered URIs. The numbers have to be equal.", prefixes.size(), uris.size());
```
reschke: In the log messages, I'd replace "URI" by either "namespace name" or "namespace URI", and "prefix" to "namespace prefix".

reschke (commented):
The other important thing is that these checks and the associated log messages should only hapen once per repository startup. Is that the case?

reschke (dismissed):

reschke (approved):

PR #1448, opened 2024-05-13
OAK-10798: Introduce functions to clean nodes based on regular expression

Joscorbe:
This adds a few functions to the [oak-mongo.js](https://github.com/apache/jackrabbit-oak/blob/ad7e429b02f724a05679ca0e96ab2c45d0e2a6e7/oak-run/src/main/js/oak-mongo.js) in [oak-run](https://github.com/apache/jackrabbit-oak/tree/trunk/oak-run) to perform some low-level operations.

### DIFF
--- a/oak-run/src/main/js/oak-mongo.js
+++ b/oak-run/src/main/js/oak-mongo.js
@@ -348,10 +348,10 @@ var oak = (function(global){
      * Use regexFind to find the nodes that match the pattern prior deletion.
      *
      * @memberof oak
-     * @method removeDescendantsMatching
+     * @method removeDescendantsAndSelfMatching
      * @param {string} pattern the pattern to match the nodes to be removed.
      */
-    api.removeDescendantsMatching = function(pattern) {
+    api.removeDescendantsAndSelfMatching = function(pattern) {
         var count = 0;
         db.nodes.find({_id: {$regex: pattern}}).forEach(function(doc) {
             print("Removing " + doc._id + " and its children");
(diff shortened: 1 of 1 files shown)


oak-run/src/main/js/oak-mongo.js
```diff
@@ -329,6 +329,59 @@ var oak = (function(global){
...
     };
 
+    /**
+     * Helper method to find nodes based on Regular Expression.
+     *
+     * @memberof oak
+     * @method regexFind
+     * @param {string} pattern the pattern to match the nodes.
+     */
+	api.regexFind = function(pattern) {
```
stefan-egli: indenting is not beautified (tabs instead of spaces)

oak-run/src/main/js/oak-mongo.js
```diff
@@ -329,6 +329,59 @@ var oak = (function(global){
...
+
+    /**
+     * Remove the complete subtree of all the nodes matching a regex pattern.
+     * Use regexFind to find the nodes that match the pattern prior deletion.
+     *
+     * @memberof oak
+     * @method removeDescendantsMatching
+     * @param {string} pattern the pattern to match the nodes to be removed.
+     */
+    api.removeDescendantsMatching = function(pattern) {
```
stefan-egli:
```suggestion
    api.removeDescendantsAndSelfMatching = function(pattern) {
```

naming : should we use a similar name to `removeDescendantsAndSelf` to stress that `self` will also be deleted?

stefan-egli (commented):
Generally looks good, just two detail aspects noted. Also noticed that the original `removeDescendantsAndSelf` doesn't stress anything about such low level operations potentially breaking certain constraints - so I don't think we need to add that now neither.

oak-run/src/main/js/oak-mongo.js
Joscorbe: Done

rishabhdaim (approved):

stefan-egli (approved):

PR #1451, opened 2024-05-13
OAK-10801: Introduce functions to remove nodes that are older than a timestamp

Joscorbe:
This function removes all the children of a node that are older than a given timestamp. To know if a node was modified recently, the value of the internal property _modified is used.

The function doesn't remove the given path. Only its children and their subtrees.

A helper method is added to find the candidates that would be deleted.

### DIFF
--- a/oak-run/src/main/js/oak-mongo.js
+++ b/oak-run/src/main/js/oak-mongo.js
@@ -330,6 +330,53 @@ var oak = (function(global){
     };
 
     /**
+     * Finds all the descendants with a _modified timestamp older than the
+     * given timestamp.
+     *
+     * @memberof oak
+     * @method findDescendantsOlderThan
+     * @param {string} path of the parent node to search.
+     * @param {number} timestamp to compare with.
+     */
+    api.findDescendantsOlderThan = function(path, timestamp) {
+        var count = 0;
+        var depth = pathDepth(path);
+        var prefix = path + "/";
+        depth++;
+        var query = { _id: pathFilter(depth, prefix), _modified: {$lt: timestamp}};
+        db.nodes.find(query).forEach(function(doc) {
+            print(doc._id);
+            count++;
+        });
+        print("nCandidates : " + count);
+    }
+
+    /**
+     * Removes all the descendants with a _modified timestamp older than the
+     * given timestamp.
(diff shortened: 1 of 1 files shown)


stefan-egli (commented):
Trying to understand the context a bit more. Would it be possible to add some comment here or in the jira as to use cases for which this is needed?

Naming wise : I'm wondering if the name is clear enough. On the one hand it uses the term "descendants" which implies it removes the entire subtree (which it ultimately does, for those matching the query conditions, as it uses removeDescendantsAndSelf). On the other hand it actually checks only direct children (`depth++`) and _then_ removes _their_ descendants. So this combination of descendants-of-matching-direct-children I'm not sure if the name reflects that entirely. Maybe something like removeDescendantsOf(Direct)ChildrenOlderThan (but yes, the name gets ... long)

Was also wondering if the two methods could be combined into one - to reduce risk of bugs and maintenance. I.e. there could be.an internal "worker function" that accepts a dryRun flag - and two caller functions that is exposed in the `api.` that calls either dryRun true or false?

oak-run/src/main/js/oak-mongo.js
```diff
@@ -329,6 +329,53 @@ var oak = (function(global){
...
+            print(doc._id);
+            count++;
+        });
+        print("nCandidates : " + count);
+    }
+
+    /**
+     * Removes all the descendants with a _modified timestamp older than the
+     * given timestamp.
+     * Use removeDescendantsAndSelf() first to get the list of nodes to remove.
```
stefan-egli, line 357:
```suggestion
     * Use findDescendantsOlderThan() first to get the list of nodes to remove.
```

Was this indeed meant to refer to removeDescendantsAndSelf, or not findDescendantsOlderThan?

rishabhdaim:
Do we need to add that we use the `_modified` field to find documents older than the given timestamp inside the documentation?

IIUC, this is an internal detail and can be avoided, @stefan-egli wydt?

rishabhdaim (approved):

stefan-egli:
Good point @rishabhdaim , we don't need to refer to the underlying key - what we might want to mention is that the timestamp is at 5 sec resolution.

PR #1458, opened 2024-05-14
OAK-10792 : rename DetailedGC to FullGC

stefan-egli:


Joscorbe (approved):

rishabhdaim (approved):

rishabhdaim:
@stefan-egli there are a couple more places in the VersionGarbageCollector class inside JavaDocs & loggers where we are still using `Detailed` instead of `Full`.

stefan-egli:
ok, let's do them as they come.. ;)

PR #1461, opened 2024-05-15
OAK-10804 - Optimize check for hidden paths and clear warnings in NodeStateUtils class.

nfsantos:


oak-store-spi/src/main/java/org/apache/jackrabbit/oak/spi/state/NodeStateUtils.java
```diff
@@ -53,12 +52,7 @@ public static boolean isHidden(@NotNull String name) {
      * @return true if one of the nodes is hidden
      */
     public static boolean isHiddenPath(@NotNull String path) {
-        for (String n : PathUtils.elements(path)) {
-            if (isHidden(n)) {
-                return true;
-            }
-        }
-        return false;
+        return (!path.isEmpty() && path.charAt(0) == ':') || path.contains("/:");
```
nit0906, line 55: There's one behaviour change introduced by this PR - isHiddenPath no longer checks if the path is valid or not. Earlier this used to happen when calling PathUtils.elements.

This is not documented for this method, so in theory (might) be ok as well. Otherwise the change looks ok to me.

nit0906 (commented):
Minor comment.

oak-store-spi/src/main/java/org/apache/jackrabbit/oak/spi/state/NodeStateUtils.java
nfsantos: Good point. But the check in `PathUtils.elements(path)` is done by an `assert`. And very likely asserts will be disabled in production (the default in the JVM is disabled). And in any case, it is bad practice to rely on asserts in production. So I think it is ok to change the behavior.

anchela (approved):
@nfsantos the change looks good to me and i agree that a check that is executed with an 'assert' should not be considered a binding behavior.

i checked the usage of the isHiddenPath in https://github.com/apache/jackrabbit-oak/blob/trunk/oak-authorization-cug/src/main/java/org/apache/jackrabbit/oak/spi/security/authorization/cug/impl/CugPermissionProvider.java#L180 and i don't think this relies on the mentioned path-validation of the original impl of isHiddenPath.

so good to go from my pov.

PR #1465, opened 2024-05-16
OAK-10812: DocumentNodeStore#diffManyChildren(...) may produce incorr…

mbaedke:
…ect results in readonly mode

Added test case.

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
@@ -296,11 +296,6 @@ public final class DocumentNodeStore
     private final int clusterId;
 
     /**
-     * The configured cluster id.
-     */
-    private final int configuredClusterId;
-
-    /**
      * Map of known cluster nodes and the last known state updated
      * by {@link #updateClusterState()}.
      * Key: clusterId, value: ClusterNodeInfoDocument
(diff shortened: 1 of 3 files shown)


mbaedke:
Preliminary test. Please review if it makes any sense.

stefan-egli (approved):
+1 for adding the test. It indeed highlights the problem in `Utils.getMinTimestampForDiff` ie in `RevisionVector.difference when in read-only mode.

Perhaps we can migrate it to use the virtual clock before the test gets unignored.

mbaedke:
Potential fix: if in DocumentNodeStore#getMinExternalRevisions() we replace getClusterId() - which is 0 in readonly mode - with the configured cluster id, the problem disappears.

stefan-egli:
I'm wondering if that wouldn't introduce a regression elsewhere .. there must have been a reason this was set to 0 for read-only?

mbaedke:
@stefan-egli , I understand that it's not supposed to collide with a simultaneously running read-write node. The proposal is not to change the cluster id, but to add an additional property to store the configured cluster id.
That would probably not cover all scenarios, though, because the real cluster id may differ from the configured one.

stefan-egli:
(maybe we could add a test case to reproduce real being different than configured id - and then the result might be two additional properties would be required, real and configured..)

mbaedke:
@stefan-egli , see ticket for an explanation.

stefan-egli:
PS: [commented](https://issues.apache.org/jira/browse/OAK-10812?focusedCommentId=17852849&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17852849) on the ticket

ionutzpi:
@mbaedke adding a check to handle the case where fromRev.pmin(minRevisions) results in an empty vector could fix the issue

reschke (commented):
I really can't comment without diving deep into this area.

PR #1473, opened 2024-05-20
OAK-10803 - Modified DocumentPropertyState.java to compress values ba…

ionutzpi:
Added compression for values in DocumentPropertyState


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +43,59 @@
...
 
     private final String name;
 
     private final String value;
 
     private PropertyState parsed;
+    private byte[] compressedValue;
+    private final Compression compression = Compression.GZIP;
+
+    public static final int DEFAULT_COMPRESSION_THRESHOLD = Integer.getInteger("oak.mongo.compressionThreshold", 1024);
```
rishabhdaim: I would keep this private.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +43,59 @@
...
 
     DocumentPropertyState(DocumentNodeStore store, String name, String value) {
         this.store = store;
         this.name = name;
-        this.value = value;
+
+        int size = value.getBytes().length;
+        if (size > DEFAULT_COMPRESSION_THRESHOLD) {
+            compressedValue =  compress(value.getBytes());
```
rishabhdaim: If there is an exception while compressing, we should return the uncompressed value rather than throwing the exception.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +43,59 @@
...
+
+    private byte[] compress(byte[] value) {
+        try {
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+            OutputStream compressionOutputStream = compression.getOutputStream(out);
+            compressionOutputStream.write(value);
+            compressionOutputStream.close();
+            return out.toByteArray();
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to compress data", e);
```
rishabhdaim: please log the exception.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +43,59 @@
...
+        }
+    }
+
+    private String decompress(byte[] value) {
+        try {
+            return new String(compression.getInputStream(new ByteArrayInputStream(value)).readAllBytes());
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to decompress data", e);
```
rishabhdaim: please log the exception.

rishabhdaim (changes requested):

rishabhdaim:
@ionutzpi could you please add test cases for this change.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -38,6 +39,19 @@
 public class DocumentPropertyStateTest {
 
     private static final int BLOB_SIZE = 16 * 1024;
+    public static final String TEST = "test";
```
stefan-egli: naming .. but maybe it could indicate what it is used for, eg `TEST_NODE` ?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -38,6 +39,19 @@
...
+    public static final String STRING_HUGEVALUE = "dummyalgjalegaafdajflalsdddkajf;kdfjakdfjadlsfjalkdsfjakldsfjkladsfjalkdsfjadlk;" +
```
stefan-egli: would org.apache.commons.lang3.RandomStringUtils (maybe using the variant that accepts a `Random` that can be seeded with a fixed value) be a useful alternative for this hard coded string?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -68,10 +82,10 @@ public void multiValuedBinarySize() throws Exception {
         for (int i = 0; i < 3; i++) {
             blobs.add(builder.createBlob(new RandomStream(BLOB_SIZE, i)));
         }
-        builder.child("test").setProperty("p", blobs, Type.BINARIES);
+        builder.child(TEST).setProperty("p", blobs, Type.BINARIES);
         TestUtils.merge(ns, builder);
 
-        PropertyState p = ns.getRoot().getChildNode("test").getProperty("p");
+        PropertyState p = ns.getRoot().getChildNode(TEST).getProperty("p");
```
stefan-egli: minor: could also leave this method unchanged as it is unrelated and only cosmetic

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +43,52 @@
...
+
+    private byte[] compress(byte[] value) {
+        try {
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+            OutputStream compressionOutputStream = compression.getOutputStream(out);
+            compressionOutputStream.write(value);
+            compressionOutputStream.close();
+            return out.toByteArray();
+        } catch (IOException e) {
+            LOG.warn("Failed to compress data: ", value);
```
stefan-egli: I wouldn't log the value as that could leak sensitive data into the log file. While we don't have the path, perhaps just log the property name plus the IOException message?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -116,7 +149,16 @@ public int count() {
...
     String getValue() {
-        return value;
+        return value != null ? value : decompress(this.compressedValue);
+    }
+
+    private String decompress(byte[] value) {
+        try {
+            return new String(compression.getInputStream(new ByteArrayInputStream(value)).readAllBytes());
+        } catch (IOException e) {
+            LOG.warn("Failed to decompress data.", value);
```
stefan-egli: an IOException on decompress is bad, as that would fail the reading of the value (while the same on compression is not a problem, then it would just not compress). So I'd argue this would justify a `LOG.error` here.

problem might then be, that in IOException case this might become a noisy logger .. not sure if we should already consider that now though...

but also : I wouldn't log the value itself.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +43,52 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(DocumentPropertyState.class);
+
     private final DocumentNodeStore store;
 
     private final String name;
 
     private final String value;
 
     private PropertyState parsed;
+    private byte[] compressedValue;
```
stefan-egli: could/should this be `final` as well? (just one more `null` assignment necessary..)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +43,52 @@
...
+    private static final int DEFAULT_COMPRESSION_THRESHOLD = Integer.getInteger("oak.mongo.compressionThreshold", 1024);
 
     DocumentPropertyState(DocumentNodeStore store, String name, String value) {
         this.store = store;
         this.name = name;
-        this.value = value;
+
+        int size = value.getBytes().length;
+        if (size > DEFAULT_COMPRESSION_THRESHOLD) {
+            compressedValue =  compress(value.getBytes());
```
stefan-egli:
```suggestion
            compressedValue = compress(value.getBytes());
```
formatting

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +43,52 @@
...
+    private byte[] compress(byte[] value) {
+        try {
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+            OutputStream compressionOutputStream = compression.getOutputStream(out);
+            compressionOutputStream.write(value);
+            compressionOutputStream.close();
+            return out.toByteArray();
+        } catch (IOException e) {
+            LOG.warn("Failed to compress data: ", value);
+            return value;
```
stefan-egli: with now returning the original uncompressed `value` here in exception case, that then likely leads to subsequent issues with uncompressing. It might need more complex logic in exception cases I think.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
rishabhdaim: we could replace this  `"<any string>".repeat(<count>)`.

rishabhdaim: this could be made private as well.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +43,59 @@
...
-        this.value = value;
+
+        int size = value.getBytes().length;
+        if (size > DEFAULT_COMPRESSION_THRESHOLD) {
+            try {
+                compressedValue = compress(value.getBytes());
+                this.value = null;
+            } catch (Exception e) {
+                // if compression fails, store the value as is
+                this.value = value;
```
rishabhdaim: please log a `WARN` message here.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -161,8 +210,9 @@ private PropertyState parsed() {
 
     /**
      * Read a {@code PropertyState} from a {@link JsopReader}
-     * @param name  The name of the property state
-     * @param reader  The reader
+     *
+     * @param name   The name of the property state
+     * @param reader The reader
```
rishabhdaim, line 215: please revert these un-necessary changes.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -172,8 +222,8 @@ PropertyState readProperty(String name, JsopReader reader) {
     /**
      * Read a {@code PropertyState} from a {@link JsopReader}.
      *
-     * @param name the name of the property state
-     * @param store the store
+     * @param name   the name of the property state
+     * @param store  the store
```
rishabhdaim, line 226: same as above.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +43,59 @@
...
+    }
+
+    private byte[] compress(byte[] value) {
+        try {
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+            OutputStream compressionOutputStream = compression.getOutputStream(out);
+            compressionOutputStream.write(value);
+            compressionOutputStream.close();
+            return out.toByteArray();
+        } catch (IOException e) {
```
stefan-egli: I would suggest to create a test case with compression/decompression throwing exceptions

Joscorbe:
I'm fine with the tests, but both `stringBelowThresholdSize()` and `stringAboveThresholdSize()` are essentially the same - which is correct from a black box test point of view.

However, I wonder if you could verify (with mockito for example) that the compressing/uncompressing was actually done in one of the tests and not in the other. Otherwise I feel these tests would pass even without using the new implemented code at all.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -81,4 +88,83 @@ public void multiValuedBinarySize() throws Exception {
...
+
+        PropertyState p = ns.getRoot().getChildNode(TEST_NODE).getProperty("p");
+        assertEquals(Type.STRING, Objects.requireNonNull(p).getType());
+        assertEquals(1, p.count());
+
+        reads.clear();
+        assertEquals(5, p.size(0));
+        // must not read the string via stream
+        assertEquals(0, reads.size());
+    }
```
Joscorbe, line 133: I'm fine with these tests, but both stringBelowThresholdSize() and stringAboveThresholdSize() are essentially the same - which is correct from a black box test point of view.

However, I wonder if you could verify (with mockito for example) that the compressing/uncompressing was actually done in one of the tests and not in the other. Otherwise I feel these tests would pass even without using the new implemented code at all.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -216,7 +264,7 @@ static PropertyState readProperty(String name, DocumentNodeStore store, JsopRead
     /**
      * Read a multi valued {@code PropertyState} from a {@link JsopReader}.
      *
-     * @param name the name of the property state
+     * @param name   the name of the property state
```
rishabhdaim: Could you please remove the format changes.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -227,8 +275,8 @@ PropertyState readArrayProperty(String name, JsopReader reader) {
     /**
      * Read a multi valued {@code PropertyState} from a {@link JsopReader}.
      *
-     * @param name the name of the property state
-     * @param store the store
+     * @param name   the name of the property state
+     * @param store  the store
```
rishabhdaim: Same as above.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -16,14 +16,17 @@
  */
 package org.apache.jackrabbit.oak.plugins.document;
 
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
```
rishabhdaim: Please avoid `*` imports and import only required packages.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -116,7 +155,20 @@ public int count() {
...
+    private String decompress(byte[] value) {
+        try {
+            return new String(compression.getInputStream(new ByteArrayInputStream(value)).readAllBytes());
+        } catch (IOException e) {
+            LOG.error("Failed to decompress property {} value: ", getName(), e);
+            return "\"{}\"";
+        }
+    }
+
+    public byte[] getCompressedValue() {
```
stefan-egli: This could probably be package protected?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -75,10 +82,97 @@ public void multiValuedBinarySize() throws Exception {
...
+    public void uncompressValueThrowsException() throws CommitFailedException, IOException {
+
+        DocumentNodeStore mockDocumentStore = mock(DocumentNodeStore.class);
+        Compression mockCompression = mock(Compression.class);
+        OutputStream mockOutputStream= mock(OutputStream.class);
+        when(mockCompression.getOutputStream(any(OutputStream.class))).thenReturn(mockOutputStream);
+        when(mockCompression.getInputStream(any(InputStream.class))).thenThrow(new IOException("Compression failed"));
+
+        DocumentPropertyState documentPropertyState = new DocumentPropertyState(mockDocumentStore, "p", STRING_HUGEVALUE, mockCompression);
+        documentPropertyState.getValue(Type.STRING);
```
stefan-egli: could you wrap this in a `assertEquals()` or similar?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
stefan-egli: +1 to assert that compression actually happened (which is indirectly already done by asserting null/non-null for compressedValue, but it could in addition check that it did some compression indeed and not leave the value unchanged - which btw could also be a nice test case, to pass a compression type that doesn't actually compress - it should then still also work)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
reschke: 1. Can we avoid creating a stream wrapper here, even if this needs changes in compression?
2. Constructing a string from bytes should be done with an explicit character encoding, otherwise it's platform/locale dependant.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +43,58 @@
...
+
+    DocumentPropertyState(DocumentNodeStore store, String name, String value, Compression compression) {
         this.store = store;
         this.name = name;
-        this.value = value;
+        this.compression = compression;
+        int size = value.getBytes().length;
+        if (compression != null && size > DEFAULT_COMPRESSION_THRESHOLD ) {
+            try {
+                compressedValue = compress(value.getBytes());
```
reschke: value.getBytes() should specify the character encoding to use (note UTF-8 is not ncessarily the default).

reschke (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -160,14 +161,14 @@ String getValue() {
...
-            return new String(compression.getInputStream(new ByteArrayInputStream(value)).readAllBytes());
+            return new String(compression.getInputStream(new ByteArrayInputStream(value)).readAllBytes(), StandardCharsets.UTF_8);
         } catch (IOException e) {
             LOG.error("Failed to decompress property {} value: ", getName(), e);
             return "\"{}\"";
         }
     }
 
-    public byte[] getCompressedValue() {
+    byte[] getCompressedValue() {
```
reschke: private maybe?

ionutzpi: No, I need it in other class.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done point 2. For 1. I wouldn't add changes in compression interface, only if indeed required.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
stefan-egli: the revert for this wasn't done though?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
stefan-egli: the revert for this wasn't done though?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
stefan-egli: thx
(nitpick: I think then you could remove this line as they getValue is now called twice)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: The revert was done.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: The revert was done.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done. I reverted readArrayProperty doc changes. Thanks.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done. I reverted also readArrayProperty doc changes. As per Stefan suggestion, thanks

stefan-egli (approved):
+1, lgtm. I see @Joscorbe  and @reschke have to yet comment on their concerns.

ionutzpi:
Hi @Joscorbe  and @reschke could you please check the PR and approve it if is fine?

Joscorbe:
+1 from me.

Joscorbe (approved):

ionutzpi:
Hi @reschke , @Joscorbe, @stefan-egli could we close this PR? @reschke it need only your approval. Thanks

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -33,11 +37,16 @@
 
 import static org.apache.jackrabbit.guava.common.collect.Lists.newArrayList;
 import static org.apache.jackrabbit.guava.common.collect.Sets.newHashSet;
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
```
reschke: no wildcard imports, please

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
reschke: Just in the test class? Can we avoid exposing it just for that purpose?

reschke (changes requested):

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: I think that the access modifier is fine as default for this method.

ionutzpi: Done. I made the modifier for the method private.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +44,58 @@
...
+                this.value = null;
+            } catch (IOException e) {
+                LOG.warn("Failed to compress property {} value: ", name, e);
+                this.value = value;
+                this.compressedValue = null;
+            }
+        } else {
+            this.value = value;
+            compressedValue = null;
+        }
```
stefan-egli:
```suggestion
        String localValue = value;
        String localCompressedValue = null;
        if (compression != null && size > DEFAULT_COMPRESSION_THRESHOLD ) {
            try {
                localCompressedValue = compress(value.getBytes(StandardCharsets.UTF_8));
                localValue = null;
            } catch (IOException e) {
                LOG.warn("Failed to compress property {} value: ", name, e);
            }
        }
        this.value = localValue;
        this.compressedValue = localCompressedValue;
```

A last-minute idea : what about making `value` and `compressedValue` `final` again and assign them like shown above? Having them final would make it somewhat stronger and clearer in my view. Wdyt?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +44,58 @@
...
     private final String name;
 
-    private final String value;
+    private String value;
 
     private PropertyState parsed;
+    private byte[] compressedValue;
+    private final Compression compression;
+
+    private static final int DEFAULT_COMPRESSION_THRESHOLD = Integer.getInteger("oak.mongo.compressionThreshold", 1024);
```
stefan-egli: Two more things came to mind:
* we should have a way to disable compression - eg by making this value `-1`. Currently that is not the case.
* are we sure we want the default to be 1024? We could choose -1 instead until we are confident this change should be on by default.

stefan-egli (commented):
Two more last minute comments added

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done. I set default to -1

reschke:
FWIW, you don't *need* my approval. What you need is an Oak committer willing to merge this on your behalf.

I didn't have time to review this properly yet.

stefan-egli (approved):
+1, PR looks good to me now, thx for the changes

PR #1474, opened 2024-05-21
Issue/oak 8848

shodaaan:


shodaaan:
- included the unit test created by Adelina Marian, which reproduces the bug
- added a fix to VersionEditor that allows a node with mix:versionable to be moved over a deleted node

stefan-egli (commented):
Would it be possible to try to reproduce a constraint violation exception when replacing a node with another node when both have a version history? (to see if this is can be reproduced - and is the same or a different issue)

mbaedke:
> Would it be possible to try to reproduce a constraint violation exception when replacing a node with another node when both have a version history? (to see if this is can be reproduced - and is the same or a different issue)

@stefan-egli I don't understand, isn't that exactly what the new test does?

shodaaan:
@stefan-egli: I created another unit test where the 2 nodes have version histories and the code flow is the same and throws ConstraintViolationException.

So if the 2 nodes have only "mix:versionable" or version histories, the behaviour and result is the same. Apparently the "jcr:versionHistory" property is created automatically for mix:versionable nodes even if no check-out / check-in operations are performed on them, and the VersionEditor.propertyChanged method throws 
the exception for the after node's "jcr:versionHistory" property.

The fix I propose in the latest commit works for both cases and was tested in unit tests using Segment, DocumentNodeStore (Mongo) and MemoryNodeStore as underlying storage when running the unit tests.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/MoveVersionableNodeWithMemoryRepositoryTest.java
```diff
@@ -0,0 +1,26 @@
+package org.apache.jackrabbit.oak.jcr;
```
reschke: So, IIUC, you have created multiple test classes so that the different mixtures are covered automatically. That was certainly helpful for testing, but it wouldn't scale if we did this everywhere.

So I would propose to simplify that again; and we also should wortk on the orthogonal issue of documenting and potentially fixing the CI pipelines.

reschke (changes requested):

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/MoveVersionableNodeWithMemoryRepositoryTest.java
shodaaan: fixed in latest commit

stefan-egli:
@shodaaan `MoveVersionableNodeWithNodeRepositoryTest` is missing the license header

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionEditor.java
```diff
@@ -152,20 +153,37 @@ public void propertyChanged(PropertyState before, PropertyState after)
                 vMgr.checkin(node);
             }
         } else if (propName.equals(JCR_BASEVERSION)) {
-            String baseVersion = after.getValue(Type.REFERENCE);
-            if (baseVersion.startsWith(RESTORE_PREFIX)) {
-                baseVersion = baseVersion.substring(RESTORE_PREFIX.length());
-                node.setProperty(JCR_BASEVERSION, baseVersion, Type.REFERENCE);
+            // OAK-8848: skip restore if this is an overwrite of a node and a restore operation is not in progress
+            if (!nodeWasMovedOrCopied() || after.getValue(Type.REFERENCE).startsWith(RESTORE_PREFIX)) {
```
stefan-egli: I'm a bit lost at this condition here and wonder if it needs to be at this complexity. I'm also wondering if all permutations of this are covered in unit tests. Let's say nodeWasMovedOrCopied is false, why is the second part of the condition needed then? Can you please elaborate a bit on this?

shodaaan: The reason why the nodeWasMovedOrCopied check is necessary in the code path that performs the version restore (starting with if (propName.equals(JCR_BASEVERSION))
) is that without it a bug occurs in certain conditions when doing the move with overwrite that is required to work in the definition of done of OAK-8848.

I will describe the flow that leads to the bug.

Prerequisites:
Start with two nodes, and create different versions for both of them. I tested this with image files that I then modify at least once, therefore creating versions for them.  

Then run the curl command for overwriting one of the nodes with the other. An example of using the Sling curl command for moving a node named Testfile.png from the parent sourceDir, under the parent destinationDir, where it overwrites another node with the same name is: 
curl -u admin:admin -X MOVE http://localhost:4502/api/assets/sourceDir/Testfile.png -H "X-Destination: http://localhost:4502/api/assets/destinationDir/Testfile.png" -H "X-Overwrite: T"  

If both of the nodes have version histories, i.e. they both have the jcr:versionHistory property set, as well as the jcr:baseVersion property set, the following flow will happen in the JCR layer of Oak, as a result of the session.save() call:

Step 1: ModifiedNodeState.compareAgainstBaseState will be called for the node Testfile.png (the moved node that will overwrite the previous node), leading to calls to EditorDiff.propertyChanged for all properties of the node.
This in turn will lead to VersionEditor.propertyChanged to be called for all properties of the  Testfile.png node.

Step 2: Eventually VersionEditor.propertyChanged will be called for the jcr:versionHistory property of the Testfile.png node. 
In my integration tests, this was called before being called for the jcr:baseVersion property. This order is very important, as will be shown in the following steps.
Since nodeWasMovedOrCopied will return true for Testfile.png, the throwProtected method will not be called.
This is the fix for the ConstraintViolationException that is described in OAK-8848.

Step 3: Since the node has multiple versions, it has the jcr:baseVersion property. Therefore, VersionEditor.propertyChanged will be called for this property as well.
When this happens, without the nodeWasMovedOrChanged check, the restore code will run, leading to versionManager.restore to be called for the Testfile.png node.

This logic is wrong, since there is no restore of a version of the node that should take place here.
Further, the VersionableState.restore method will be called for the Testfile.png node, leading to the call checkNotNull(versionable).setProperty(JCR_ISCHECKEDOUT, false, Type.BOOLEAN), which will set the jcr:isCheckedOut property for the node to false.
This is wrong behavior, since the move of a node over another should not result in the node being automatically checked in.

The fix I propose in the PR is to also add the check for nodeWasMovedOrChanged for the code that performs the restore. Since the check returns true, the restore logic will not run when the overwrite is performed.


Regarding the after.getValue(Type.REFERENCE).startsWith(RESTORE_PREFIX)) condition:
Debugging the code when restoring previous versions of a node led me to understand that the jcr:baseVersion property is first modified to have the prefix “restore-” to the original UUID, and then this prefix is used in the restore code section in VersionEditor.propertyChanged to set the jcr:baseVersion property back to its’ original UUID, when the method runs for the jcr:baseVersion property.
This leads me to believe that the check for startsWith(RESTORE_PREFIX) should also be added for clarity.


Regarding the unit tests, I agree that the bug described above should have a unit test that reproduces it. I will create it and update the PR.

stefan-egli: Thx for the very detailed description!

> Regarding the after.getValue(Type.REFERENCE).startsWith(RESTORE_PREFIX)) condition:

As the code is now `restore()` is only called if this condition is true. Before this change however, it was also called in other cases. Now if those other cases were wrong, then that's a bug. I would however prefer to look into that and have a test cases that proofs that is a bug. I'm suspecting that with this change we'll introduce a regression...

stefan-egli:
Regarding the refactoring of `propertyChanged` : so far I am not seeing the advantage of the refactoring. That code isn't reused elsewhere, it is merely cutting a larger method into 4 parts that have historically been together. I think the same could be achieved with adding more // comments where useful, the additional methods for me at least aren't improving things.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/MoveVersionableNodeWithNodeRepositoryTest.java
```diff
@@ -0,0 +1,230 @@
...
+
+        //get the version histories for the removed node and the moved node
+        try {
+            VersionHistory sourceNodeHistory = (VersionHistory) session.getNode(delNodeVHPath);
+
+            versionIterator = sourceNodeHistory.getAllVersions();
+            while (versionIterator.hasNext()) {
+                sourceNodeVersions.add(versionIterator.nextVersion());
+            }
+        }catch (Exception ex) {
```
reschke:
```suggestion
        } catch (Exception ex) {
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionEditor.java
```diff
@@ -145,27 +146,73 @@ public void propertyChanged(PropertyState before, PropertyState after)
...
+        // Updates the checked-out / checked-in state of the currently processed node when the JCR_ISCHECKEDOUT property
+        // change is processed.
         if (propName.equals(JCR_ISCHECKEDOUT)) {
             if (wasCheckedIn()) {
                 vMgr.checkout(node);
             } else {
                 vMgr.checkin(node);
             }
+
+       /*Completes the restore of a version from version history.
```
reschke:
```suggestion
       /* Completes the restore of a version from version history.
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionEditor.java
```diff
@@ -145,27 +146,73 @@ public void propertyChanged(PropertyState before, PropertyState after)
...
+       When the JCR_BASEVERSION property is processed, a check is made for the current base version property.
+       If a restore is currently in progress for the current base version (the check for this is that the
+       current base version name has the format "restore-[UUID of the version to restore to]"), then the restore
+       is completed for the current node to the version specified by the UUID.
+
+       If a node that was moved or copied to the location of a deleted node is currently being processed
+       (see OAK-8848 for context), the restore operation must NOT be performed when the JCR_BASEVERSION property change
+       is processed for the node.
+
+       TODO: check if this code ever runs when after.getValue(Type.REFERENCE).startsWith(RESTORE_PREFIX) is FALSE,
```
reschke: Is this still a TODO?

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/MoveVersionableNodeWithNodeRepositoryTest.java
```diff
@@ -0,0 +1,230 @@
...
+import static java.util.Collections.singleton;
+import static org.apache.jackrabbit.oak.commons.FixturesHelper.Fixture.DOCUMENT_NS;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Test for moving versionable nodes over deleted versionable nodes using the DOCUMENT_NS fixture.
+ */
+public class MoveVersionableNodeWithNodeRepositoryTest extends AbstractRepositoryTest {
+
+    private static final String mixVersionable = "mix:versionable";
```
reschke: we should have a constant for this elsewhere... org.apache.jackrabbit.JcrConstants.MIX_VERSIONABLE;

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/security/authorization/VersionManagementTest.java
```diff
@@ -45,7 +45,7 @@ public class VersionManagementTest extends AbstractEvaluationTest {
 
     @Override
     @Before
-    protected void setUp() throws Exception {
+    public void setUp() throws Exception {
```
reschke: why are changes needed in this class?

reschke (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionEditor.java
shodaaan: Done.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/MoveVersionableNodeWithNodeRepositoryTest.java
shodaaan: Done.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionEditor.java
shodaaan: Done.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/MoveVersionableNodeWithNodeRepositoryTest.java
shodaaan: I used JcrConstants.MIX_VERSIONABLE. Done.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/security/authorization/VersionManagementTest.java
shodaaan: I reverted all of the changes, since the new unit tests are now in a different class.

reschke (commented):
Thanks for the changes.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionEditor.java
shodaaan: You are right, thank you for the comment. I will remove the second check.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionEditor.java
shodaaan: No, it is not, I will remove the comment. I will remove the second check in the version restore code block - the one for after.getValue(Type.REFERENCE).startsWith(RESTORE_PREFIX) - since it is not necessary in order for all of the restore logic and the fix to work fine.

stefan-egli:
two more comments that I have drafted in [this temporary branch](https://github.com/apache/jackrabbit-oak/compare/OAK-8848-review-illustration?expand=1) for illustration:
* we could move the `nodeWasMovedOrCopied()` check for the 2nd case into that if block, to make it more limited to just that case. Previously it was throwing the exception - now I think we want it to not throw it anymore. With the check being in the outer if there is possibility it might still throw if the next if case hits (perhaps unlikely, but by doing it the above drafted way we'd reduce that risk)
* some reformatting suggestion to use `//` instead of `/*` - I think that's a bit more following how inline comments are frequently done

wdyt?

shodaaan:
> two more comments that I have drafted in [this temporary branch](https://github.com/apache/jackrabbit-oak/compare/OAK-8848-review-illustration?expand=1) for illustration:
> 
> * we could move the `nodeWasMovedOrCopied()` check for the 2nd case into that if block, to make it more limited to just that case. Previously it was throwing the exception - now I think we want it to not throw it anymore. With the check being in the outer if there is possibility it might still throw if the next if case hits (perhaps unlikely, but by doing it the above drafted way we'd reduce that risk)
> * some reformatting suggestion to use `//` instead of `/*` - I think that's a bit more following how inline comments are frequently done
> 
> wdyt?

Thank you for the suggestions. I have made the changes in the latest commit.

stefan-egli:
Thx for the update @shodaaan 

I'm reviewing the changes and actually would like to find a test that verifies the opposite of the new MoveVersionableNodeWithNodeRepositoryTest to better understand the context etc. Might take me some more time. If someone else wants to chime in and approve in the meantime, pls feel free.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/version/VersionEditor.java
```diff
@@ -145,27 +146,71 @@ public void propertyChanged(PropertyState before, PropertyState after)
...
         } else if (isReadOnly && getOPV(after) != OnParentVersionAction.IGNORE) {
             throwCheckedIn("Cannot change property " + after.getName()
                     + " on checked in node");
         }
     }
 
+    /**
+     * Returns true if and only if the given node was moved or copied from another location.
+     */
+    private boolean nodeWasMovedOrCopied() {
```
stefan-egli: @shodaaan a question about this one : I'm not sure this is also applicable to copy - I can see the move case, as that is covered through `:source-path` - but in which case is `:source-path` also added for copy?

shodaaan: I just tested the copy instead of move case and it can not be done in one session, because WorkspaceDelegate.copy throws a javax.jcr.ItemExistsException. 
Copy was not mentioned in the ticket requirement either, it was a wrong assumption on my part that copy will behave the same as move in this case.

I will rename the check method to nodeWasMoved.

RELEASE-NOTES.txt
```diff
@@ -1,4 +1,4 @@
-Release Notes -- Apache Jackrabbit Oak -- Version 1.64.0
+Release Notes -- Apache Jackrabbit Oak -- Version 1.62.0
```
stefan-egli: this file shouldn't be part of the PR

shodaaan: Fixed it, thank you for the observation.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/MoveVersionableNodeRepositoryTest.java
```diff
@@ -0,0 +1,229 @@
...
+
+        session.removeItem(destNode.getPath());
+
+        session.move(sourceNode.getPath(), destPath);
+        session.save();
+
+        // check setting property - via NodeImpl - on moved node
+        setNodePropertyAndCheckIn(sourceNode, "testProp", "testValue");
+
+        // check version history - TODO HORIA: should be in separate method - EXTRACT
```
stefan-egli: I would avoid personal TODOs and just have it without username..

stefan-egli (approved):
+1, lgtm now - one very minor comment that should block though.
I'll merge tomorrow to leave room for a potential other review

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/MoveVersionableNodeRepositoryTest.java
shodaaan: Fixed it, thank you for the catch.

stefan-egli:
jenkins failure seems unrelated to me and I'll go ahead and merge

in both oak-jcr and oak-store-document cases it is a failed fork :
```
[ERROR] org.apache.maven.surefire.booter.SurefireBooterForkException: The forked VM terminated without properly saying goodbye. VM crash or System.exit called?

[ERROR] Command was /bin/sh -c cd /home/jenkins/workspace/Jackrabbit_oak-trunk-pr_PR-1474/oak-jcr && /usr/local/asfpackages/java/adoptium-jdk-11.0.16.1+1/bin/java -javaagent:/home/jenkins/maven-repositories/1/org/jacoco/org.jacoco.agent/0.8.12/org.jacoco.agent-0.8.12-runtime.jar=destfile=/home/jenkins/workspace/Jackrabbit_oak-trunk-pr_PR-1474/oak-jcr/target/coverage-reports/jacoco-ut.exec -Xmx512m -XX:+HeapDumpOnOutOfMemoryError -Dupdate.limit=100 -Djava.awt.headless=true -jar /home/jenkins/workspace/Jackrabbit_oak-trunk-pr_PR-1474/oak-jcr/target/surefire/surefirebooter13470384701666492214.jar /home/jenkins/workspace/Jackrabbit_oak-trunk-pr_PR-1474/oak-jcr/target/surefire 2024-06-18T14-46-01_816-jvmRun1 surefire319862193111735272tmp surefire_012291488413358130927tmp
[ERROR] Error occurred in starting fork, check output in log
[ERROR] Process Exit Code: 143
[ERROR] Crashed tests:
[ERROR] org.apache.jackrabbit.oak.jcr.query.QueryTest
[ERROR] 	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork(ForkStarter.java:669)
```

PR #1477, opened 2024-05-22
OAK-10823 : replaced db.collection.remove with db.collection.deleteOn…

rishabhdaim:
…e/deleteMany to make it compatible with both mongo & mongosh

### DIFF
--- a/oak-run/src/main/js/oak-mongo.js
+++ b/oak-run/src/main/js/oak-mongo.js
@@ -352,10 +352,10 @@ var oak = (function(global){
      * @param {string} pattern the pattern to match the nodes to be removed.
      */
     api.removeDescendantsAndSelfMatching = function(pattern) {
-        let count = 0;
+        var count = 0;
         db.nodes.find({_id: {$regex: pattern}}, {_id: 1}).forEach(function(doc) {
             print("Removing " + doc._id + " and its children");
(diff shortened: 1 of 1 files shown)


oak-run/src/main/js/oak-mongo.js
```diff
@@ -352,12 +352,12 @@ var oak = (function(global){
      * @param {string} pattern the pattern to match the nodes to be removed.
      */
     api.removeDescendantsAndSelfMatching = function(pattern) {
-        var count = 0;
+        let count = 0;
```
stefan-egli: It seems like this change is a no-op? What would make it still useful to have?

stefan-egli (approved):

oak-run/src/main/js/oak-mongo.js
rishabhdaim: the variables defined with `let` are block scoped and hence cause less confusion and are less error-prone. But I agree this can be reverted.

Will remove this.

rishabhdaim: > the variables defined with `let` are block scoped and hence cause less confusion and are less error-prone. But I agree this can be reverted.
> 
> Will remove this.

Removed in https://github.com/apache/jackrabbit-oak/pull/1477/commits/30672ee37040d8d280ade40dddc35a7bab5063f5

stefan-egli (approved):

PR #1480, opened 2024-05-23
OAK-10711 : replaced preconditions.checkNotNull with Objects.requireNotNull

rishabhdaim:
…otNull

### DIFF
--- a/oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/DynamicSyncContext.java
+++ b/oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/DynamicSyncContext.java
@@ -52,7 +52,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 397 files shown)


stefan-egli:
no actual review just a comment : should we wait until after the release (with such a large change, even though seeming trivial) ?

reschke:
I'd also like to take the opportunity to move the new imports to the "right" place, if only for consistency with all the other changes we already made.

Let me have a look at this later this week.

reschke:
Using...

~~~
for i in $(grep -rl "Objects.requireNonNull" .) ; do num=$(grep "import static" $i | wc -l) ; [ $num -gt 1 ] && (grep "import static" $i > x1 ; sort x1 > x2 ; cmp -s x1 x2 ||  ( diff -u x1 x2 | (grep -q "+import static java.util.Objects.requireNonNull") && (echo ; echo $num; echo ; echo $i ; echo ; diff -u x1 x2 ))) ; done
~~~

I identified files where the static imports now needed some reshuffling.

Note that while doing this, I also found a case where the "new" import was already present, so we got a warning about duplicate imports.

reschke (approved):

PR #1485, opened 2024-05-24
Issues/oak 10781

t-rana:
Azure Access token is valid for 1 hour only. Added code to check for token expiry every 1 minutes, and refresh the token when the same is close to expiry.


rootpea (approved):

rootpea:
@t-rana the jenkins test failed, could you please re-run

rootpea (approved):

dulceanu (approved):

PR #1511, opened 2024-06-07
OAK-10866 - Avoid creating an intermediate String when appending a string value to a JsopBuilder.

nfsantos:
Provides a ~10% improvement when generating Json.

Microbenchmark adapted from JsopStreamTest:
```
public static void main(String... args) {
    for (int k = 0; k < 5; k++) {
        String s = "Hello \"World\" Hello \"World\" Hello \"World\" Hello \"World\" Hello \"World\" Hello \"World\" ";
        StopWatch timer = new StopWatch();
        JsopBuilder w = new JsopBuilder() ;
        for (int i = 0; i < 50000000; i++) {
            w.value(s);
            if (i % 100 == 0) {
                w.resetWriter();
            }
        }
        System.out.println(w.getClass() + ": " + timer.seconds());
    }
}
```

Base line:
```
class org.apache.jackrabbit.oak.commons.json.JsopBuilder: 9.27 seconds
class org.apache.jackrabbit.oak.commons.json.JsopBuilder: 9.19 seconds
class org.apache.jackrabbit.oak.commons.json.JsopBuilder: 9.21 seconds
class org.apache.jackrabbit.oak.commons.json.JsopBuilder: 9.23 seconds
class org.apache.jackrabbit.oak.commons.json.JsopBuilder: 9.20 seconds
```

Optimized `JsopBuilder.value(String)`
```
class org.apache.jackrabbit.oak.commons.json.JsopBuilder: 8.09 seconds
class org.apache.jackrabbit.oak.commons.json.JsopBuilder: 8.04 seconds
class org.apache.jackrabbit.oak.commons.json.JsopBuilder: 8.12 seconds
class org.apache.jackrabbit.oak.commons.json.JsopBuilder: 8.22 seconds
class org.apache.jackrabbit.oak.commons.json.JsopBuilder: 8.16 seconds

### DIFF
--- a/oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsopBuilder.java
+++ b/oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsopBuilder.java
@@ -28,7 +28,7 @@ public class JsopBuilder implements JsopWriter {
 
     private static final boolean JSON_NEWLINES = false;
 
-    private final StringBuilder buff = new StringBuilder();
+    private StringBuilder buff = new StringBuilder();
     private boolean needComma;
     private int lineLength, previous;
 
@@ -286,7 +286,7 @@ public class JsopBuilder implements JsopWriter {
      * @param buff the target buffer
      * @return the target buffer with the encoded string appended
      */
-    public static StringBuilder encode(String s, StringBuilder buff) {
+    private static StringBuilder encode(String s, StringBuilder buff) {
         if (s == null) {
             return buff.append("null");
         }
--- a/oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsopBuilder.java
+++ b/oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsopBuilder.java
@@ -197,6 +197,7 @@ public class JsopBuilder implements JsopWriter {
      * @param value the value
      * @return this
      */
+    @Override
     public JsopBuilder value(String value) {
         optionalCommaAndNewline(strLength(value));
         encode(value, buff);
@@ -267,7 +268,7 @@ public class JsopBuilder implements JsopWriter {
         }
         for (int i = 0; i < length; i++) {
             char c = s.charAt(i);
-            if (c == '\"' || c == '\\' || c < ' ' || (c >= 0xd800 && c <= 0xdbff)) {
+            if (shouldEscape(c)) {
                 StringBuilder buff = new StringBuilder(length + 2 + length / 8);
                 buff.append('\"');
                 escape(s, length, buff);
@@ -296,7 +297,7 @@ public class JsopBuilder implements JsopWriter {
         }
         for (int i = 0; i < length; i++) {
             char c = s.charAt(i);
-            if (c == '\"' || c == '\\' || c < ' ' || (c >= 0xd800 && c <= 0xdbff)) {
+            if (shouldEscape(c)) {
                 buff.append('\"');
                 escape(s, length, buff);
                 return buff.append('\"');
@@ -313,6 +314,10 @@ public class JsopBuilder implements JsopWriter {
      */
     public static void escape(String s, StringBuilder buff) {
         escape(s, s.length(), buff);
+    }
+
+    private static boolean shouldEscape(char c) {
+        return c == '\"' || c == '\\' || c < ' ' || (c >= 0xd800 && c <= 0xdbff);
     }
 
     /**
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by fabriziofortino — APPROVED** [review]

File: oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsopBuilder.java [hunk 1]
```diff
@@ -275,6 +277,34 @@ public static String encode(String s) {
...
+    /**
+     * Similar to #encode(String) but appends the result to the given buffer. This method exists to avoid the allocation
+     * of intermediate objects to represent the encoded String when the caller already has a buffer where the result
+     * should be appended.
+     *
+     * @param s the text to convert
+     * @param buff the target buffer
+     * @return the target buffer with the encoded string appended
+     */
+    private static StringBuilder encode(String s, StringBuilder buff) {
```
reschke, line 290: maybe return the StringBuffer so this can be used fluently?

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsopBuilder.java
```diff
@@ -275,6 +277,34 @@ public static String encode(String s) {
...
+        if (s == null) {
+            return buff.append("null");
+        }
+        int length = s.length();
+        if (length == 0) {
+            return buff.append("\"\"");
+        }
+        for (int i = 0; i < length; i++) {
+            char c = s.charAt(i);
+            if (c == '\"' || c == '\\' || c < ' ' || (c >= 0xd800 && c <= 0xdbff)) {
```
reschke: Maybe extract the logic that detects what needs to be escaped?

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsopBuilder.java
```diff
@@ -197,9 +197,11 @@ public JsopBuilder value(boolean value) {
      * @param value the value
      * @return this
      */
-    @Override
     public JsopBuilder value(String value) {
-        return encodedValue(encode(value));
+        optionalCommaAndNewline(strLength(value));
```
reschke, line 202: why do we loose the "@Override" here?

reschke (changes requested):
Also: maybe this should to into the interface as well?

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsopBuilder.java
nfsantos: Missed that one. Fixed.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsopBuilder.java
nfsantos: The method does return the StringBuilder received as an argument.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsopBuilder.java
nfsantos: Done.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/json/JsopBuilder.java
```diff
@@ -275,6 +278,34 @@ public static String encode(String s) {
...
+     * @param buff the target buffer
+     * @return the target buffer with the encoded string appended
+     */
+    private static StringBuilder encode(String s, StringBuilder buff) {
+        if (s == null) {
+            return buff.append("null");
+        }
+        int length = s.length();
+        if (length == 0) {
+            return buff.append("\"\"");
```
reschke, line 296: I personally find it easier to *read*, when all returns are on the same level of nesting (so this would require else/if etc). But this may be a matter of taste.

nfsantos: As this method is a variation of `StringBuilder encode(String s)`, I kept it as close as possible to the other method with only the changes needed to use an existing `StringBuilder`, without doing any stylist changes, as this seems to be outside the scope of this PR.

reschke: good point

reschke (approved):

PR #1513, opened 2024-06-10
OAK-10867: add service principal support in oak-run commons

t-rana:


smiroslav (approved):

nit0906:
Check on a couple of things - 

Currently the azureBlobContainerProvider is not getting closed - this should be handled in the AzureBlobStoreBacked#close.

Another point , if the executor service being used for the token refresh stalls the jvm from closure - will the shutdown hook be actually called ?

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/Utils.java
```diff
@@ -19,38 +19,61 @@
...
+import java.time.format.DateTimeFormatter;
+import java.util.Properties;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 
 public final class Utils {
 
-    public static final String DEFAULT_CONFIG_FILE = "azure.properties";
+    static {
```
amit-jain, line 65: This should not be needed. Instead the close() method should be implemented as earlier and property called in the DataStore/Backend#close() e.g. https://github.com/apache/jackrabbit-oak/blob/trunk/oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java#L518

t-rana:
Closing this PR for now, will raise a new one.

PR #1516, opened 2024-06-10
OAK-10873: store-spi: remove use of Guava CharMatcher in test case

reschke:


oak-store-spi/src/main/java/org/apache/jackrabbit/oak/json/JsonDeserializer.java
```diff
@@ -240,9 +239,10 @@ boolean hasOrderableChildren(NodeBuilder builder) {
         }
 
         private boolean hasSingleColon(String jsonString) {
-            //In case the primaryType was encoded then it would be like nam:oak:Unstructured
-            //So check if there is only one occurrence of ';'
-            return CharMatcher.is(':').countIn(jsonString) == 1;
+            // In case the primaryType was encoded then it would be like
+            // "nam:oak:Unstructured". So check if there is only one occurrence
+            // of ':'.
+            return jsonString.replace(":", "").length() == jsonString.length() - 1;
```
nfsantos: This is going to create an intermediate String, which could have a negative performance impact if this method is called very often. Couldn't this be rewritten without creating intermediate objects?

reschke: (I somehow thought that this was just test code, but I was wrong; thanks for pointing it out)

Of course we could just walk the string until the end and count the colons.

nfsantos: I don't think this is test code. This method is called by `inferPropertyType()` in the same class, which is called by `JsonDeserializer.readProperty()` and `JsonDeserializer.readArrayProperty()`. IIUC, the method seems to be on the hot path. We use this class heavily in the indexing job. Is there a better option than this class?

There are more compact ways of doing the same using streams, but if the method is indeed hot as I think it is, I would suggest something like this:
```
            int colons = 0;
            for (int i = 0; i < jsonString.length(); i++) {
                if (jsonString.charAt(i) == ':') {
                    colons++;
                }
            }
            return colons == 1;
```
Could be optimized to abort early if it finds more than a colon, but probably the size of the argument will be small, so it won't matter.

reschke: Yes, this is almost exactly what I came up with and which I'm now testing.

nfsantos: Actually, it was copilot who wrote that code. :)

nfsantos (approved):

PR #1518, opened 2024-06-11
Issues/oak 10781

t-rana:
Closing the executor service which was preventing the JVM from shutting down.


chibulcuteanu:
It would be good to create a test for this.

t-rana:
@chibulcuteanu , azurite currently does not support service principals, creating a test would not be possible I guess. There is another test for checking  successful authentication via service principals.
https://github.com/apache/jackrabbit-oak/blob/trunk/oak-blob-cloud-azure/src/test/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackendTest.java#L155

amit-jain (approved):

amit-jain:
> @chibulcuteanu , azurite currently does not support service principals, creating a test would not be possible I guess. There is another test for checking successful authentication via service principals. https://github.com/apache/jackrabbit-oak/blob/trunk/oak-blob-cloud-azure/src/test/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackendTest.java#L155

@t-rana The scheduled executor is opened in all cases and not just for service principals, so a test particularly to test proper cleanup can be added

t-rana:
Please do not merge this for now, I am still testing some newly done changes

t-rana:
> Please do not merge this for now, I am still testing some newly done changes

Changes done and tested, please re-review the PR.

Following changes are done:

1. A new access token was being generated on every read / write leading to multiple access token refresh executor. Now after the changes, only one access token will exist per class instance, and there will only be one token refresh exectuor that will check and refresh the access token.
2. Added check in test case to ensure proper cleanup is done.

@nit0906 @amit-jain

amit-jain (approved):

reschke:
when merging please make sure to squash and to have the JIRA ticket properly in the commit message

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobContainerProvider.java
```diff
@@ -205,19 +212,33 @@ private CloudBlobContainer getBlobContainerFromServicePrincipals(@Nullable BlobR
...
+            isAccessTokenGenerated = true;
+        }
+
+        Objects.requireNonNull(storageCredentialsToken, "storage credentials token cannot be null");
+
+        // start refresh token executor only when the access token is first generated
+        if (isAccessTokenGenerated) {
+            log.info("starting refresh token task at: {}", OffsetDateTime.now());
+            TokenRefresher tokenRefresher = new TokenRefresher();
+            executorService.scheduleWithFixedDelay(tokenRefresher, TOKEN_REFRESHER_INITIAL_DELAY, TOKEN_REFRESHER_DELAY, TimeUnit.MINUTES);
```
nit0906, line 240: Can you add a comment here describing why the 1 minute delay is needed.

nit0906 (approved):

PR #1520, opened 2024-06-12
Oak 10880 -- Logging for hitting an orphaned UUIDs and reference indexes

ionutzpi:


stefan-egli:
was this meant to be a continuation of OAK-10803? I see many commits coming from that work.

regarding the changes being in ReferenceCheck : that one is only used in the context of oak-run, eg when running the consistency check through oak-run. What the intention of Oak 10880 was however is that we should get a log.warn in a productive oak instance. So for example when [getNodeByUUID](https://jackrabbit.apache.org/oak/docs/apidocs/org/apache/jackrabbit/oak/jcr/session/SessionImpl.html#getNodeByUUID(java.lang.String)) is called (not saying it needs to be in SessionImpl, it might be deeper down, haven't checked, so this is just for illustration).

ionutzpi:
Done changes. @stefan-egli no, it is for the new task(probably some issues on my local).

stefan-egli:
[commented](https://issues.apache.org/jira/browse/OAK-10880?focusedCommentId=17854696&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17854696) on OAK-10880

PR #1522, opened 2024-06-13
Oak-10874 | Add jmx function for bringing forward a delayed async lane to a latest checkpoint.

nit0906:
For scenarios where the async indexing lanes might be lagging for a large amount of time (let's assume days) and it might not be possible for the lane to catchup normally on it's own due to -
1. Large content diff to process leading to OOM and continuous failed async cycles
2. Possibility of old reference checkpoint no longer being present in the system and the async lane thus doing a complete reindex of the whole repo.

In such cases, it might be required to make the async lane think that it is up to date by setting it's reference checkpoint to a more recent one. 
This PR introduces a jmx function to - 
1. Pause and Abort the lagging async lane
2. Clear it's lease
3. Create a new checkpoint 
4. Set the new checkpoint as reference checkpoint for the lane
5. delete the old checkpoint
6. resume the async indexing on the lane.

If this operation is performed the async lane should start async indexing from the newly created checkpoint. But the diff of the content that was missed would need to be reindexed separately using oak-run's offline reindexing. 

To prevent unintentional misuse of this function, 

1. It will specifically ask the user to enter "CONFIRM" before proceeding so that the user knows what he is doing.
2. It will not skip in case the lane on which it is called is not failing.

### DIFF
--- a/oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/IndexStatsMBean.java
+++ b/oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/IndexStatsMBean.java
@@ -128,8 +128,8 @@ public interface IndexStatsMBean {
      */
     String getReferenceCheckpoint();
 
-    @Description("Force update the indexing lane to a latest checkpoint. This will abort and pause the running lane, release it's lease and set the reference checkpoint to a latest one." +
-            "Any content changes b/w the old reference checkpoint and the new one will be not be indexed and a reindexing would be required." +
+    @Description("Force update the indexing lane to a checkpoint created during execution of this function. This will abort and pause the running lane, release it's lease and set the reference checkpoint to a latest one." +
+            "Any content changes between the old reference checkpoint and the new one will be not be indexed and a reindexing would be required." +
             "Only use this operation if you are sure that the lane is stuck and not updated since many days and cannot catchup on it's own." +
             "Once this operation is completed, reindexing for all indexes on the lane is required.")
     String forceIndexLaneCatchup(
--- a/oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/IndexStatsMBean.java
+++ b/oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/IndexStatsMBean.java
@@ -128,9 +128,9 @@ public interface IndexStatsMBean {
      */
     String getReferenceCheckpoint();
 
-    @Description("Force update the indexing lane to a latest checkpoint. This will abort and pause the running lane, release it's lease and set the reference checkpoint to a latest one." +
+    @Description("Force update the indexing lane to a latest checkpoint. This will abort and pause the running lane, release its lease and set the reference checkpoint to a latest one." +
             "Any content changes b/w the old reference checkpoint and the new one will be not be indexed and a reindexing would be required." +
-            "Only use this operation if you are sure that the lane is stuck and not updated since many days and cannot catchup on it's own." +
+            "Only use this operation if you are sure that the lane is stuck and not updated for many days and cannot catch up on its own." +
             "Once this operation is completed, reindexing for all indexes on the lane is required.")
     String forceIndexLaneCatchup(
             @Name("Confirmation Message")
--- a/oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/IndexStatsMBean.java
+++ b/oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/IndexStatsMBean.java
@@ -20,6 +20,7 @@ package org.apache.jackrabbit.oak.api.jmx;
 import javax.management.openmbean.CompositeData;
 import javax.management.openmbean.TabularData;
 
+import org.apache.jackrabbit.oak.api.CommitFailedException;
 import org.osgi.annotation.versioning.ProviderType;
 
 @ProviderType
@@ -126,6 +127,18 @@ public interface IndexStatsMBean {
      * @return the reference checkpoint
      */
     String getReferenceCheckpoint();
+
+    /**
+     * Used to set the reference checkpoint for the async indexer to the provided checkpoint value.
+     */
+    @Description("Force update the indexing lane to a latest checkpoint. This will abort and pause the running lane, release it's lease and set the reference checkpoint to a latest one." +
+            "Any content changes b/w the old reference checkpoint and the new one will be not be indexed and a reindexing would be required." +
+            "Only use this operation if you are sure that the lane is stuck and not updated since many days and cannot catchup on it's own." +
+            "Once this operation is completed, reindexing for all indexes on the lane is required.")
+    void pretendIndexLaneCatchup(
+            @Name("Confirmation Message")
+            @Description("Enter 'CONFIRM' to confirm the operation")
+            String confirmationMessage) throws CommitFailedException;
 
     /**
      * Returns the processed checkpoint used by the async indexer. If this index
--- a/oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/IndexStatsMBean.java
+++ b/oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/IndexStatsMBean.java
@@ -128,14 +128,11 @@ public interface IndexStatsMBean {
      */
     String getReferenceCheckpoint();
 
-    /**
-     * Used to set the reference checkpoint for the async indexer to the provided checkpoint value.
-     */
     @Description("Force update the indexing lane to a latest checkpoint. This will abort and pause the running lane, release it's lease and set the reference checkpoint to a latest one." +
             "Any content changes b/w the old reference checkpoint and the new one will be not be indexed and a reindexing would be required." +
             "Only use this operation if you are sure that the lane is stuck and not updated since many days and cannot catchup on it's own." +
             "Once this operation is completed, reindexing for all indexes on the lane is required.")
-    void pretendIndexLaneCatchup(
+    String forceIndexLaneCatchup(
             @Name("Confirmation Message")
             @Description("Enter 'CONFIRM' to confirm the operation")
             String confirmationMessage) throws CommitFailedException;
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
@@ -1252,12 +1252,12 @@ public class AsyncIndexUpdate implements Runnable, Closeable {
             }
 
             if (!this.isFailing()) {
-                String msg = "The lane is not failing. This operation should only be performed if the lane is failing, it should first be allowed to catchup on it's own.";
+                String msg = "The lane is not failing. This operation should only be performed if the lane is failing, it should first be allowed to catchup on its own.";
                 log.warn(msg);
                 return msg;
             }
 
-            log.info("Running a forced catchup for indexing lane [{}]. ", name);
+            log.info("Running a forced catchup for indexing lane [{}]", name);
             // First we need to abort and pause the running indexing task
             this.abortAndPause();
             log.info("Aborted and paused async indexing for lane [{}]", name);
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
@@ -1291,9 +1291,10 @@ public class AsyncIndexUpdate implements Runnable, Closeable {
             } catch (Exception e) {
                 log.error("Exception while trying to force update the indexing lane [{}]", name, e);
                 if (this.isPaused()) {
+                    this.resume();
                     log.info("Resuming the lane [{}] as it was paused during the operation", name);
                 }
-                return "Unable to complete the force update due to" + e.getMessage() + "Please check logs for more details";
+                return "Unable to complete the force update due to " + e.getMessage() + ".Please check logs for more details";
             }
         }
 
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
@@ -1261,7 +1261,7 @@ public class AsyncIndexUpdate implements Runnable, Closeable {
                     "created", now(),
                     "thread", Thread.currentThread().getName(),
                     "name", name + "-forceModified"));
-            String existingReferenceCheckpoint = referenceCp;
+            String existingReferenceCheckpoint = this.referenceCp;
             log.info("Modifying the referred checkpoint for lane [{}] from {} to {}." +
                     " This means that any content modifications b/w these checkpoints will not reflect in the indexes on this lane." +
                     " Reindexing would be needed to get this content indexed.", name, existingReferenceCheckpoint, newReferenceCheckpoint);
@@ -1269,7 +1269,11 @@ public class AsyncIndexUpdate implements Runnable, Closeable {
             builder.child(ASYNC).setProperty(name, newReferenceCheckpoint);
             this.referenceCp = newReferenceCheckpoint;
             // Remove the existing reference checkpoint
-            store.release(existingReferenceCheckpoint);
+            if (store.release(existingReferenceCheckpoint)) {
+                log.info("Old reference checkpoint {} removed or didn't exist", existingReferenceCheckpoint);
+            } else {
+                log.warn("Unable to remove old reference checkpoint {}. This can result in orphaned checkpoints and would need to be removed manually.", existingReferenceCheckpoint);
+            }
             mergeWithConcurrencyCheck(store, validatorProviders, builder, null, null, name);
 
             // Resume the paused lane;
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
@@ -1257,37 +1257,44 @@ public class AsyncIndexUpdate implements Runnable, Closeable {
                 return msg;
             }
 
-            log.info("Running a forced catch-up for indexing lane [{}]. ", name);
-            // First we need to abort and pause the running indexing task
-            this.abortAndPause();
-            log.info("Aborted and paused async indexing for lane [{}]", name);
-            // Release lease for the paused lane
-            this.releaseLeaseForPausedLane();
-            log.info("Released lease for paused lane [{}]", name);
-            String newReferenceCheckpoint = store.checkpoint(lifetime, Map.of(
-                    "creator", AsyncIndexUpdate.class.getSimpleName(),
-                    "created", now(),
-                    "thread", Thread.currentThread().getName(),
-                    "name", name + "-forceModified"));
-            String existingReferenceCheckpoint = this.referenceCp;
-            log.info("Modifying the referred checkpoint for lane [{}] from {} to {}." +
-                    " This means that any content modifications between these checkpoints will not reflect in the indexes on this lane." +
-                    " Reindexing is needed to get this content indexed.", name, existingReferenceCheckpoint, newReferenceCheckpoint);
-            NodeBuilder builder = store.getRoot().builder();
-            builder.child(ASYNC).setProperty(name, newReferenceCheckpoint);
-            this.referenceCp = newReferenceCheckpoint;
-            // Remove the existing reference checkpoint
-            if (store.release(existingReferenceCheckpoint)) {
-                log.info("Old reference checkpoint {} removed or didn't exist", existingReferenceCheckpoint);
-            } else {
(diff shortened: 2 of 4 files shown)


oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/IndexStatsMBean.java
```diff
@@ -127,6 +128,15 @@ public interface IndexStatsMBean {
      */
     String getReferenceCheckpoint();
 
+    @Description("Force update the indexing lane to a latest checkpoint. This will abort and pause the running lane, release it's lease and set the reference checkpoint to a latest one." +
```
fabriziofortino: typo
```suggestion
    @Description("Force update the indexing lane to a latest checkpoint. This will abort and pause the running lane, release its lease and set the reference checkpoint to a latest one." +
```

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/IndexStatsMBean.java
```diff
@@ -127,6 +128,15 @@ public interface IndexStatsMBean {
...
+            "Any content changes b/w the old reference checkpoint and the new one will be not be indexed and a reindexing would be required." +
+            "Only use this operation if you are sure that the lane is stuck and not updated since many days and cannot catchup on it's own." +
```
fabriziofortino: typos
```suggestion
            "Only use this operation if you are sure that the lane is stuck and not updated for many days and cannot catch up on its own." +
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1242,6 +1242,54 @@ public String getReferenceCheckpoint() {
...
+        public String forceIndexLaneCatchup(String confirmMessage) throws CommitFailedException {
+
+            if (!"CONFIRM".equals(confirmMessage)) {
+                String msg = "Please confirm that you want to force the lane catchup by passing 'CONFIRM' as argument";
+                log.warn(msg);
+                return msg;
+            }
+
+            if (!this.isFailing()) {
+                String msg = "The lane is not failing. This operation should only be performed if the lane is failing, it should first be allowed to catchup on it's own.";
```
fabriziofortino: typo
```suggestion
                String msg = "The lane is not failing. This operation should only be performed if the lane is failing, it should first be allowed to catchup on its own.";
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1242,6 +1242,54 @@ public String getReferenceCheckpoint() {
...
+                log.warn(msg);
+                return msg;
+            }
+
+            log.info("Running a forced catchup for indexing lane [{}]. ", name);
```
fabriziofortino: removed extra space and `.` to make it consistent with other logs in the same function
```suggestion
            log.info("Running a forced catchup for indexing lane [{}]", name);
```

fabriziofortino (approved):
LGTM. Just a few minors.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1242,6 +1242,54 @@ public String getReferenceCheckpoint() {
...
+            }
+
+            log.info("Running a forced catchup for indexing lane [{}]", name);
+            // First we need to abort and pause the running indexing task
+            this.abortAndPause();
+            log.info("Aborted and paused async indexing for lane [{}]", name);
+            // Release lease for the paused lane
+            this.releaseLeaseForPausedLane();
+            log.info("Released lease for paused lane [{}]", name);
+            String newReferenceCheckpoint = store.checkpoint(lifetime, ImmutableMap.of(
```
nfsantos: Can we use the JDK `Map` instead of the Guava one? `Map.of(...` should be equivalent.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/IndexStatsMBean.java
```diff
@@ -127,6 +128,15 @@ public interface IndexStatsMBean {
      */
     String getReferenceCheckpoint();
 
+    @Description("Force update the indexing lane to a latest checkpoint. This will abort and pause the running lane, release its lease and set the reference checkpoint to a latest one." +
+            "Any content changes b/w the old reference checkpoint and the new one will be not be indexed and a reindexing would be required." +
```
nfsantos: I would prefer to not use abbreviations in user facing documentation, so use `between` instead of `b/w`. I'm not even sure that `b/w` is meant to mean `between`, I'm just guessing.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/IndexStatsMBean.java
```diff
@@ -127,6 +128,15 @@ public interface IndexStatsMBean {
      */
     String getReferenceCheckpoint();
 
+    @Description("Force update the indexing lane to a latest checkpoint. This will abort and pause the running lane, release its lease and set the reference checkpoint to a latest one." +
```
nfsantos: Can you be more specific of what latest mean? Exact values if possible.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1242,6 +1242,54 @@ public String getReferenceCheckpoint() {
...
+                    "creator", AsyncIndexUpdate.class.getSimpleName(),
+                    "created", now(),
+                    "thread", Thread.currentThread().getName(),
+                    "name", name + "-forceModified"));
+            String existingReferenceCheckpoint = this.referenceCp;
+            log.info("Modifying the referred checkpoint for lane [{}] from {} to {}." +
+                    " This means that any content modifications b/w these checkpoints will not reflect in the indexes on this lane." +
```
nfsantos:
```suggestion
                    " This means that any content modifications between these checkpoints will not reflect in the indexes on this lane." +
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1242,6 +1242,54 @@ public String getReferenceCheckpoint() {
...
+                    " Reindexing would be needed to get this content indexed.", name, existingReferenceCheckpoint, newReferenceCheckpoint);
```
nfsantos:
```suggestion
                    " Reindexing is needed to get this content indexed.", name, existingReferenceCheckpoint, newReferenceCheckpoint);
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1242,6 +1242,54 @@ public String getReferenceCheckpoint() {
...
+                log.info("Old reference checkpoint {} removed or didn't exist", existingReferenceCheckpoint);
+            } else {
+                log.warn("Unable to remove old reference checkpoint {}. This can result in orphaned checkpoints and would need to be removed manually.", existingReferenceCheckpoint);
+            }
+            mergeWithConcurrencyCheck(store, validatorProviders, builder, null, null, name);
+
+            // Resume the paused lane;
+            this.resume();
+            log.info("Resumed async indexing for lane [{}]", name);
+            return ("Lane successfully forced to catchup. New reference checkpoint is " + newReferenceCheckpoint + " . Please make sure to perform reindexing to get the diff content indexed.");
```
nfsantos: I think the parenthesis are unnecessary.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1242,6 +1242,54 @@ public String getReferenceCheckpoint() {
...
+                return msg;
+            }
+
+            if (!this.isFailing()) {
+                String msg = "The lane is not failing. This operation should only be performed if the lane is failing, it should first be allowed to catchup on its own.";
+                log.warn(msg);
+                return msg;
+            }
+
+            log.info("Running a forced catchup for indexing lane [{}]", name);
```
nfsantos: Just a minor grammar detail. `catchup` is incorrect, the word does not exist. Either
- catch up -> verb
- catch-up -> adjective/noum
So here it should be catch-up. There are other incorrect uses of `catchup` in this PR. It's not too important, but as this user facing documentation I think we should try to use correct grammar.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdateTest.java
```diff
@@ -589,6 +589,83 @@ public void run() {
...
     }
 
+    @Test
+    public void testForceUpdateAsyncLane() throws CommitFailedException {
+        NodeStore store = new MemoryNodeStore();
+        IndexEditorProvider provider = new PropertyIndexEditorProvider();
+
+        NodeBuilder builder = store.getRoot().builder();
+        createIndexDefinition(builder.child(INDEX_DEFINITIONS_NAME),
+                "rootIndex", true, false, ImmutableSet.of("foo"), null)
```
nfsantos: Use JDK Set instead of Guava.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdateTest.java
```diff
@@ -589,6 +589,83 @@ public void run() {
...
+        async.run();
+        root = store.getRoot();
+
+        // first check that the index content nodes exist
+        checkPathExists(root, INDEX_DEFINITIONS_NAME, "rootIndex",
+                INDEX_CONTENT_NODE_NAME);
+        assertFalse(root.getChildNode(INDEX_DEFINITIONS_NAME).hasChildNode(
+                ":conflict"));
+        PropertyIndexLookup lookup = new PropertyIndexLookup(root);
+        assertEquals(ImmutableSet.of("testRoot", "testRoot1"), find(lookup, "foo", "abc"));
```
nfsantos: Again Guava vs JDK.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdateTest.java
```diff
@@ -589,6 +589,83 @@ public void run() {
...
+        async.getIndexStats().forceIndexLaneCatchup("CONFIRM");
+        async.run();
+        root = store.getRoot();
+
+        checkPathExists(root, INDEX_DEFINITIONS_NAME, "rootIndex",
+                INDEX_CONTENT_NODE_NAME);
+        assertFalse(root.getChildNode(INDEX_DEFINITIONS_NAME).hasChildNode(
+                ":conflict"));
+        lookup = new PropertyIndexLookup(root);
+        assertEquals(ImmutableSet.of("testRoot", "testRoot1", "testRoot2"), find(lookup, "foo", "abc"));
```
nfsantos: Guava vs JDK.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdateTest.java
```diff
@@ -589,6 +589,83 @@ public void run() {
...
+        root = store.getRoot();
+
+        checkPathExists(root, INDEX_DEFINITIONS_NAME, "rootIndex",
+                INDEX_CONTENT_NODE_NAME);
+        assertFalse(root.getChildNode(INDEX_DEFINITIONS_NAME).hasChildNode(
+                ":conflict"));
+        lookup = new PropertyIndexLookup(root);
+        // testRoot3 will not be indexed, because it was created after the last successfully run index update and before the forceUpdate was called.
+        // So it lands in the missing content diff that needs to be reindexed.
+        assertEquals(ImmutableSet.of("testRoot", "testRoot1", "testRoot2", "testRoot4"), find(lookup, "foo", "abc"));
```
nfsantos: Guava vs JDK

nfsantos (commented):
This feature will be very useful. 👍  Overall I think it's good, but I have some high level questions:

- the `forceIndexLaneCatchup` is not thread safe. What would happen if it is called concurrently? Or if while it is being called, there are other calls being performed by JMX to this object? I know this is unlikely to happen, but maybe it's worth to make calls into this bean thread safe.

- Do all the intermediate operations done in this method take effect immediately after they are invoked? I'm asking because I had the idea that when doing the manual steps, some operations would take sometime to be applied. Just a few seconds, but if that is the case, then it could be a problem when doing all the steps in succession without waiting between them. 

- There is no good error handling in case some of the intermediate operations fail. It would be good if there was a failure halfway through this method, we could revert all the changes. Or at least show a message to the user explaining what was changed and what should be done to restore the system to a proper state.

oak-api/src/main/java/org/apache/jackrabbit/oak/api/jmx/IndexStatsMBean.java
nit0906: it would be to a checkpoint created during this method execution - I will update that.

nit0906:
@nfsantos  - I took care of the grammar related issues in the logs.

I will try and answer the other questions here - 

> the forceIndexLaneCatchup is not thread safe. What would happen if it is called concurrently? Or if while it is being called, there are other calls being performed by JMX to this object? I know this is unlikely to happen, but maybe it's worth to make calls into this bean thread safe.

The beans object is created while initialising the AsyncIndexUpdate object (which is a scheduled thread). It's taken care that the AsyncIndexUpdate will only run on author leader and cannot run concurrently (that is basically using the lease check mechanism - but we do clear the lease as part of this). Anyhow, the only actual resource that this bean modifies is the :async node's checkpoint detail which is committed using a mergeConcurrencyCheck (it basically will check before committing that the checkpoint for the lane is the same as the old referred checkpoint in this method - otherwise will throw a concurrent update exception and this will fail) - I suppose that should be enough for the use case ? 

I will check if the clear lease is really needed in this operation or not - if not I will remove that too.



> There is no good error handling in case some of the intermediate operations fail.

The main point of failure can be during the commit merge - if it fails there - the lane is left in a paused state with the old reference checkpoint. Given that the lane was already failing - shouldn't be much of a problem.

If there's a failure post the commit merge - it can be while deleting the old checkpoint (there's a log added for that failure). In this case we are left with an orphaned checkpoint (However this should not lead to an exception - so the flow will continue).

I have still added a basic try catch, and resumed the lane in the catch (if it was left in a paused state). I think this should be sufficient - wdyt ? 


> Just a few seconds, but if that is the case, then it could be a problem when doing all the steps in succession without waiting between them.

That's a good point, the operations don't take immediate effect - since these setup the flags (paused, releaseLease etc), These are read during index update callback - so basically if I abort a lane - it will take either reflect on the next async indexing run, or on an index update processed during the running indexing cycle - which would eventually throw an interrupt and abort the lane. In case of just pausing the lane - it will take effect only on the next cycle run.

Either way, what we need to make sure is that the new run that starts post this abort - starts with the new referred checkpoint, which would be the case here irrespective of the timing of these events - for instance, 

Let's say we placed a request for pause and abort and then cleared the lease  - assuming that this actually didn't happen till now and we went ahead and changed the reference checkpoint on the :async node - If an async run is already in a place and has already picked up the older checkpoint - it will run till it gets interrupted , in which case when the lane will resume, it will pick up the new checkpoint (as desired) and if the async run has yet to pick up a checkpoint - it will pick up the new one - which is ok (even though this wouldn't be the case, since we are only running this on failing lanes under the assumption that they have been delayed for quite a while).

Let me know if you think any of the above doesn't make sense. 

I would also request @thomasmueller  to review the PR and my responses above once .

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1242,6 +1242,61 @@ public String getReferenceCheckpoint() {
...
+                // Resume the paused lane;
+                this.resume();
+                log.info("Resumed async indexing for lane [{}]", name);
+                return "Lane successfully forced to catch-up. New reference checkpoint is " + newReferenceCheckpoint + " . Please make sure to perform reindexing to get the diff content indexed.";
+            } catch (Exception e) {
+                log.error("Exception while trying to force update the indexing lane [{}]", name, e);
+                if (this.isPaused()) {
+                    log.info("Resuming the lane [{}] as it was paused during the operation", name);
+                }
+                return "Unable to complete the force update due to" + e.getMessage() + "Please check logs for more details";
```
nfsantos:
```suggestion
                return "Unable to complete the force update due to " + e.getMessage() + ". Please check logs for more details";
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1242,6 +1242,61 @@ public String getReferenceCheckpoint() {
...
+                    log.warn("Unable to remove old reference checkpoint {}. This can result in orphaned checkpoints and would need to be removed manually.", existingReferenceCheckpoint);
+                }
+                // Resume the paused lane;
+                this.resume();
+                log.info("Resumed async indexing for lane [{}]", name);
+                return "Lane successfully forced to catch-up. New reference checkpoint is " + newReferenceCheckpoint + " . Please make sure to perform reindexing to get the diff content indexed.";
+            } catch (Exception e) {
+                log.error("Exception while trying to force update the indexing lane [{}]", name, e);
+                if (this.isPaused()) {
+                    log.info("Resuming the lane [{}] as it was paused during the operation", name);
```
nfsantos, line 1295: Should there be a call to resume here?

nfsantos (approved):
Thanks for the clarifications, I think I understand how this works a bit better.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
nit0906: good catch. Thanks.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1242,6 +1242,62 @@ public String getReferenceCheckpoint() {
...
+        @Override
+        public String forceIndexLaneCatchup(String confirmMessage) throws CommitFailedException {
+
+            if (!"CONFIRM".equals(confirmMessage)) {
+                String msg = "Please confirm that you want to force the lane catch-up by passing 'CONFIRM' as argument";
+                log.warn(msg);
+                return msg;
+            }
+
+            if (!this.isFailing()) {
```
thomasmueller, line 1254: In theory (not now, but maybe later) we could add an additional check: only allow if the indexing lane is behind more than 30 minutes, or something like that.

But again, not needed right now.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1242,6 +1242,62 @@ public String getReferenceCheckpoint() {
...
+                String msg = "The lane is not failing. This operation should only be performed if the lane is failing, it should first be allowed to catch up on its own.";
+                log.warn(msg);
+                return msg;
+            }
+
+            try {
+                log.info("Running a forced catch-up for indexing lane [{}]. ", name);
+                // First we need to abort and pause the running indexing task
+                this.abortAndPause();
```
thomasmueller, line 1263: I wouldn't mind if this is just an additional preconditions... instead of doing it here, just verify it's done. But it's OK, no need to change it now.

(the idea is: We don't want this feature is used a lot, so making it a little bit harder to use might help that it is used less often... The "CONFIRM" is a great example of that!... Do not make it easy to delete stuff, so that it's less likely that stuff is deleted by mistake.)

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1242,6 +1242,62 @@ public String getReferenceCheckpoint() {
...
+                log.info("Aborted and paused async indexing for lane [{}]", name);
+                // Release lease for the paused lane
+                this.releaseLeaseForPausedLane();
+                log.info("Released lease for paused lane [{}]", name);
+                String newReferenceCheckpoint = store.checkpoint(lifetime, Map.of(
```
thomasmueller, line 1268: Ah, DEFAULT_LIFETIME is 1000 days! could you change that to 100 days please? That should still be plenty, but (hopefully) will result in slightly less "orphaned checkpoints".

thomasmueller (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
nit0906: done.

PR #1523, opened 2024-06-13
OAK-10876: Oak 1.22 does not compile with JDK 11 anymore - remove animal sniffer plugin, require JDK 11+ for compilation, target JRE 8

reschke:


oak-parent/pom.xml
```diff
@@ -71,12 +71,15 @@
...
-    <java.version.signature>java18</java.version.signature>
-    
-   <!-- specifies on which fixture to run the integration testing tests. 
-      override in profiles or provide from command line to change behaviour. Provide 
-      more fixtures space separated. See org.apache.jackrabbit.oak.jcr.FixturesHelper#AVAILABLE_FIXTURES 
+    <!-- determines the bytecode version (i.e. the minimum JRE required to run the build artifact) -->
+    <javaTargetVersion>8</javaTargetVersion>
+    <maven.compiler.release>${javaTargetVersion}</maven.compiler.release>
+    <maven.compiler.target>${javaTargetVersion}</maven.compiler.target>
+    <minimalJavaBuildVersion>${javaTargetVersion}</minimalJavaBuildVersion>
```
kwin: This should be 11 as < 9 does not support the release option.

reschke: Hmmm.

a) We need to create Java 8 class files, do we not.
b) It seems to exactly do what we wanted. It builds with Java11+, but generates class files for Java 8

What am I missing?

kwin: `minimalJavaBuildVersion` should be 11, otherwise the enforcement of the minimum build JDK is wrong!

reschke: ok: I can do that but it doesn't build with JDK 8 anyway:

~~~
[INFO] --- maven-enforcer-plugin:3.1.0:enforce (enforce-java-version) @ oak-parent ---
[ERROR] Rule 0: org.apache.maven.plugins.enforcer.RequireJavaVersion failed with message:
Detected JDK Version: 1.8.0-341 is not in the allowed range 8.
~~~

kwin: > Detected JDK Version: 1.8.0-341 is not in the allowed range 8. 

would be a very confusing error, right?

kwin (approved):

oak-parent/pom.xml
reschke: Indeed.

kwin:
The readme needs to be adjusted as well: https://github.com/apache/jackrabbit-oak/tree/1.22?tab=readme-ov-file#getting-started

kwin (approved):

PR #1526, opened 2024-06-14
OAK-10803 -- compress/uncompress property

ionutzpi:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +44,66 @@
...
 
     private final String name;
 
     private final String value;
 
     private PropertyState parsed;
+    private final byte[] compressedValue;
+    private final Compression compression;
+
+    private static final int DEFAULT_COMPRESSION_THRESHOLD = Integer.getInteger("oak.mongo.compressionThreshold", -1);
```
reschke: please use SystemPropertySupplier

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +44,66 @@
...
+            this.name = name;
+            this.value = value;
+            this.compression = null;
+            this.compressedValue = null;
+        } else {
+            this.store = store;
+            this.name = name;
+            this.compression = compression;
+
+            int size = value.getBytes().length;
```
reschke: That serializes the String into a byte array, obtains the lenght, and then throws the byte array away.

reschke: (also it needs to specify the charset)

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -18,26 +18,43 @@
...
-import org.junit.Before;
-import org.junit.Rule;
-import org.junit.Test;
+import org.junit.*;
+import org.junit.runners.MethodSorters;
 
 import static org.apache.jackrabbit.guava.common.collect.Lists.newArrayList;
 import static org.apache.jackrabbit.guava.common.collect.Sets.newHashSet;
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
```
reschke: please no wildcard imports

reschke:
I believe the main concern is that we don't know yet exactly why we are doing this. IMHO it would be better to first collect data on the size of the properties.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +44,66 @@
...
     DocumentPropertyState(DocumentNodeStore store, String name, String value) {
-        this.store = store;
-        this.name = name;
-        this.value = value;
+        this(store, name, value, Compression.GZIP);
+    }
+
+    DocumentPropertyState(DocumentNodeStore store, String name, String value, Compression compression) {
+        if (DEFAULT_COMPRESSION_THRESHOLD == -1) {
+            this.store = store;
```
reschke: things that are the same should be moved out of the condition

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +45,64 @@
...
 
     private final String name;
 
     private final String value;
 
     private PropertyState parsed;
+    private final byte[] compressedValue;
+    private final Compression compression;
+
+    private static final int DEFAULT_COMPRESSION_THRESHOLD =
```
reschke: I'd set the logger here as well ("logTo()")

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -128,7 +188,7 @@ public boolean equals(Object object) {
         } else if (object instanceof DocumentPropertyState) {
             DocumentPropertyState other = (DocumentPropertyState) object;
             return this.name.equals(other.name)
-                    && this.value.equals(other.value);
+                    && this.getValue().equals(other.getValue());
```
reschke: Hm. That means the "equals()" method would need to uncompress. That seems performance critical.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -81,4 +100,179 @@ public void multiValuedBinarySize() throws Exception {
...
+        long duration = (endTime - startTime);  // divide by 1000000 to get milliseconds.
+        long usedMemory = endMemory - startMemory; // Calculate the used memory during the operation
+
+        System.out.println("Time taken for compression: " + duration + " nanoseconds");
+        System.out.println("Memory used for compression: " + usedMemory + " bytes");
+
+        // Assert that the compressed value is not null or empty
+        assertTrue(compressedValue != null && !compressedValue.isEmpty());
+    }
+}
```
reschke, line 394: We absolutely need to test string values which would fail with a pedantic UTF-8 encoder. See https://issues.apache.org/jira/browse/OAK-10803?focusedCommentId=17855529&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17855529.

reschke (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +45,64 @@
...
     private final String name;
 
     private final String value;
 
     private PropertyState parsed;
+    private final byte[] compressedValue;
+    private final Compression compression;
+
+    private static final int DEFAULT_COMPRESSION_THRESHOLD =
+            SystemPropertySupplier.create("oak.mongo.compressionThreshold", -1).get();
```
stefan-egli: the property isn't mongo specific. this isn't handled entirely consistent in the existing code base neither - perhaps it should be `oak.documentMK.` ?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +45,64 @@
...
+        if (DEFAULT_COMPRESSION_THRESHOLD == -1) {
+            this.value = value;
+            this.compression = null;
+            this.compressedValue = null;
+        } else {
+            this.compression = compression;
+            int size = value.length();
+            String localValue = value;
+            byte[] localCompressedValue = null;
+            if (compression != null && size > DEFAULT_COMPRESSION_THRESHOLD) {
```
stefan-egli: I think the `compression != null` part could be move to the `if` above?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -81,4 +100,179 @@ public void multiValuedBinarySize() throws Exception {
...
+
+        assertEquals(documentPropertyState.getValue(Type.STRING), "{}");
+
+        verify(mockCompression, times(1)).getInputStream(any(InputStream.class));
+
+        compressionThreshold.set(null, -1);
+    }
+
+    @Test
+    public void adefaultValueSetToMinusOne() throws NoSuchMethodException, InvocationTargetException, IllegalAccessException, NoSuchFieldException {
```
stefan-egli: was wondering if this method should also test that `-1` indeed "does nothing" - i.e. it wouldn't call length()/getBytes()/compress() - not sure if we need to cover all these cases, but adding something that would have failed in the previous iteration might be useful. PS: is there reason the method starts with adefault?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Yes.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Starts with a for execute this method first.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: The method test that for -1, the compression is not made.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done.

reschke: IIRC, the test fails, right? That means that for certain property values, compression would be lossy, right?

ionutzpi: The test failed. The compression didn't work properly.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +45,64 @@
...
     private final String name;
 
     private final String value;
 
     private PropertyState parsed;
+    private final byte[] compressedValue;
+    private final Compression compression;
+
+    private static final int DEFAULT_COMPRESSION_THRESHOLD = SystemPropertySupplier
+            .create("oak.documentMK.stringCompressionThreshold ", -1).loggingTo(LOG).get();
```
stefan-egli: I would suggest to refactor this. 
One aspect is that this variable shouldn't be called "default", as it is actually the real configured value - the default is always -1, that can be a separate, hard-code const.
A second aspect is how this variable can be controlled from tests. This has been a controversial topic before - in my opinion if we like to test different values of say "compressionThreshold", then there should be a clean way to achieve that. A `final` in my view isn't ideal, as compilers can in theory inline final variables if they so wish. What I think can be an option is to add an explicit (static) setter method that is only used for testing - and that allows overwriting this field explicitly. With this, the need for explicit test order plus the various reflections in the junit class disappear.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -38,24 +45,64 @@
...
 
     DocumentPropertyState(DocumentNodeStore store, String name, String value) {
+        this(store, name, value, Compression.GZIP);
+    }
+
+    DocumentPropertyState(DocumentNodeStore store, String name, String value, Compression compression) {
         this.store = store;
         this.name = name;
-        this.value = value;
+        if (compression != null && DEFAULT_COMPRESSION_THRESHOLD == -1) {
```
stefan-egli: if `compression` is null, then it would go into `else` and run into a NPE

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
stefan-egli: is there a way that equals wouldn't have to uncompress?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -18,26 +18,45 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
 
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
 public class DocumentPropertyStateTest {
 
     private static final int BLOB_SIZE = 16 * 1024;
+    public static final int COMPRESSED_SIZE = 543;
```
stefan-egli: nitpick : could probably be private?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -116,7 +163,20 @@ public int count() {
...
+    private String decompress(byte[] value) {
+        try {
+            return new String(compression.getInputStream(new ByteArrayInputStream(value)).readAllBytes(), StandardCharsets.UTF_8);
+        } catch (IOException e) {
+            LOG.error("Failed to decompress property {} value: ", getName(), e);
+            return "\"{}\"";
+        }
+    }
+
+    private byte[] getCompressedValue() {
```
stefan-egli: this method is currently unused but instead the junit class does reflection to access `compressedValue`. If this method was made less protected it could actually replace that reflection use.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -81,4 +100,225 @@ public void multiValuedBinarySize() throws Exception {
...
+        Compression mockCompression = mock(Compression.class);
+        when(mockCompression.getOutputStream(any(OutputStream.class))).thenThrow(new IOException("Compression failed"));
+
+        Field compressionThreshold = DocumentPropertyState.class.getDeclaredField("DEFAULT_COMPRESSION_THRESHOLD");
+        compressionThreshold.setAccessible(true);
+        Field modifiersField = Field.class.getDeclaredField("modifiers");
+        modifiersField.setAccessible(true);
+        modifiersField.setInt(compressionThreshold, compressionThreshold.getModifiers() & ~Modifier.FINAL);
+
+        compressionThreshold.set(null, DEFAULT_COMPRESSION_THRESHOLD);
```
stefan-egli: see other comments, this reflection could be avoided with corresponding changes in the target class (same applies to other test methods)

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -81,4 +100,225 @@ public void multiValuedBinarySize() throws Exception {
...
+
+        DocumentPropertyState documentPropertyState = new DocumentPropertyState(mockDocumentStore, "p", "\"" + STRING_HUGEVALUE + "\"", mockCompression);
+
+        assertEquals(documentPropertyState.getValue(Type.STRING), STRING_HUGEVALUE);
+
+        verify(mockCompression, times(1)).getOutputStream(any(OutputStream.class));
+
+        compressionThreshold.set(null, -1);
+
```
stefan-egli: nitpick : I'd consider removing some of the newlines (here and in other test methods), some seem unnecessary.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
stefan-egli: I'd suggest to avoid relying on test order in this case - and ensure the test class doesn't leave `DocumentPropertyState` in a changed state. Otherwise other test classes will be affected unintentionally.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
stefan-egli: (assuming this was originally referring to `testInterestingStrings`)

> The test failed. The compression didn't work properly.

So is the plan to still fix this? I don't think this test should be expected to throw `ComparisonFailure`

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -81,4 +100,225 @@ public void multiValuedBinarySize() throws Exception {
...
+        DocumentPropertyState state = new DocumentPropertyState(store, "propertyName", testString, Compression.GZIP);
+
+        String value = state.getValue(Type.STRING);
+
+        assertEquals(testString, value);
+
+    }
+
+    @Test
+    public void performanceTest() throws NoSuchFieldException, IllegalAccessException {
```
stefan-egli: I fail to see the use of this versus running oak-benchmark tests with different values of `-Doak.documentMK.stringCompressionThreshold=1024` ?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
reschke: It actually was my proposal not to expose it. Maybe we can just *copy* or *move* it to the test class?

stefan-egli: ah, now I remember. Your comment wasn't visible in this particular review anymore, so I didn't connect the dots.

I'm actually a fan of facilitating testing, and that can mean to introduce methods in productive classes that are there solely for testing purpose. That method can still be non-public but package-protected for example. That in my view makes things more explicit than using reflection in tests which is brittle. Moving this method to the test class I assume implies opening up the actual variable, which seems less favourable.

ionutzpi: Yes. I did the same idea as Stefan. But Julian insisted to make it private(it was package-protected). Also, this approach complicated many tests(trying to solve with reflection)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
reschke: We could special-case the compressed case, and compare the byte arrays. Still expensive.

Maybe we could keep the String's hashCode in order to avoid comparing the byte arrays in most cases.

stefan-egli: Right, I think there should be distinction between non-compressed, compressed, mixed cases. Only the compressed one is slightly more involved. For that one we could use a combination of length comparison, hashcode (calculated when needed) and/or plain Arrays.equals(byte[],byte[])..

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -127,8 +188,25 @@ public boolean equals(Object object) {
             return true;
         } else if (object instanceof DocumentPropertyState) {
             DocumentPropertyState other = (DocumentPropertyState) object;
-            return this.name.equals(other.name)
-                    && this.value.equals(other.value);
+            // Compare names and raw un-parsed values
+            if (!this.name.equals(other.name) || !this.getValue().equals(other.getValue())) {
```
reschke: This will always decompress, no?

reschke (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Not necessarily.

reschke: "getValue()" will uncompress when the value is compressed, no?

ionutzpi: Yes, but only then, not always.

reschke (commented):
I note that we delegate "hashCode" to AbstractPropertyState, which only takes the property name into account (this is unrelated to the PR). Do we have a problem here?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
reschke: This doesn't make sense. If "equals()" uncompresses anyway, all the subsequent code isn't needed. But the idea was to avoid that de-compression happens.

stefan-egli:
[looks](https://github.com/apache/jackrabbit-oak/blob/14c357120e49d83281dc473d810ca485deb94501/oak-store-spi/src/main/java/org/apache/jackrabbit/oak/plugins/memory/AbstractPropertyState.java#L103-L106) like the original intention was that properties shouldn't or wouldn't be used as keys in hash maps... if that's the case then it should be fine..

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -60,6 +73,11 @@ public void before() throws Exception {
         ns = builderProvider.newBuilder().setBlobStore(bs).getNodeStore();
     }
 
+    @After
+    public void tearDown() {
+        ns.dispose();
```
stefan-egli:
```suggestion
        try {
            ns.dispose();
        } finally {
            DocumentPropertyState.setCompressionThreshold(DISABLED_COMPRESSION);
        }
```

to avoid side-effects, the test should ensure it doesn't leak a non default compression threshold to other tests. DISABLED_COMPRESSION is for illustration to refer to -1 - ideally the -1 would only be defined in the source class...

ionutzpi: Done

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
stefan-egli: [Commented](https://issues.apache.org/jira/browse/OAK-10803?focusedCommentId=17856243&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17856243) in the ticket - to me it looks like this `assertEquals(value,"");` is not what we'd aim for.

ionutzpi: The compression did work properly even for interesting strings. Done

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -81,4 +102,133 @@ public void multiValuedBinarySize() throws Exception {
...
+        assertEquals(state.getValue(Type.STRING), STRING_HUGEVALUE);
+        assertEquals(STRING_HUGEVALUE, state.getValue(Type.STRING));
+    }
+
+    @Test
+    public void testInterestingStrings() {
+
+        DocumentNodeStore store = mock(DocumentNodeStore.class);
+        String testString = "\"\"simple:foo\", \"cr:a\\n\\b\", \"dquote:a\\\"b\", \"bs:a\\\\b\", \"euro:a\\u201c\", \"gclef:\\uD834\\uDD1E\",\n" +
+                "                \"tab:a\\tb\", \"nul:a\\u0000b\", \"brokensurrogate:\\ud800\"";
```
stefan-egli: Was it the intention to convert the [original string array](https://github.com/apache/jackrabbit-oak/blob/0e327a27eb988a0f7656535275f3735ef8474e5d/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BasicDocumentStoreTest.java#L754-L755) into 1 string with escaped double quotes? Shouldn't it be something like below - plus then iterated over each case? 

```suggestion
        String[] tests = new String[] { "simple:foo", "cr:a\n\b", "dquote:a\"b", "bs:a\\b", "euro:a\u201c", "gclef:\uD834\uDD1E",
                "tab:a\tb", "nul:a\u0000b", "brokensurrogate:\ud800" };
```

stefan-egli: (in which case it will fail)

ionutzpi: No. The string can have interesting characters in it(it can be string, not string array).

stefan-egli: sure, it can, but if you test with the original strings the test will fail

ionutzpi: The test didn't failed for the big string, it compress and uncompress properly.

stefan-egli: Again, I'd suggest you try the string array as per the original - and try each one individually

ionutzpi: Done.

stefan-egli: Maybe we need to clarify first whether we want to keep supporting broken surrogates (for RDBDocumentStore). Before this PR they are supported. With this PR they will no longer be supported, if that particular property value goes through compression/decompression. I don't think the way `isValidSurrogatePair` is currently introduce in the test is correct.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
stefan-egli: Right, but IIUC the idea was to avoid uncompressing - and that's actually what the code does after this check. So either we don't aim to avoid uncompressing (but then the lines after this aren't necessary), or we want to avoid it (then this condition here needs to be adjusted). Currently we do a bit of both.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
reschke: Back then (2018) my proposal was to reject string values that do not have an UTF-8 representation, but met resistance from the segment team.

This question affects both segment store and document store, and it would be cool if we could come to consensus about this.

WRT to *this* PR we should just ensure that compression does not affect how MongoDS currently behaves.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -218,22 +218,27 @@ public void stringAboveThresholdSizeNoCompression() {
     public void testInterestingStrings() {
         DocumentNodeStore store = mock(DocumentNodeStore.class);
         String[] tests = new String[] { "simple:foo", "cr:a\n\b", "dquote:a\"b", "bs:a\\b", "euro:a\u201c", "gclef:\uD834\uDD1E",
-                "tab:a\tb", "nul:a\u0000b", "brokensurrogate:\ud800"};
+                "tab:a\tb", "nul:a\u0000b", "brokensurrogate:dfsa\ud800"};
 
         DocumentPropertyState.setCompressionThreshold(1);
         for (String test : tests) {
             if (isValidSurrogatePair(test.split(":")[1])) {
```
reschke: I'm not sure why we check that here; can't we test all strings?

reschke (changes requested):

stefan-egli:
> WRT to this PR we should just ensure that compression does not affect how MongoDS currently behaves.

Why is this a MongoDB specific aspect? For MongoDB we know it is not supported (see [this skip here](https://github.com/apache/jackrabbit-oak/blob/0e327a27eb988a0f7656535275f3735ef8474e5d/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BasicDocumentStoreTest.java#L760)) - but for RDB it _is_ supported. So shouldn't this PR keep the support for RDB?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: If I test with broken surrogate it will fail, so I filter that value.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: I support you proposal to reject string values that do not have an UTF-8 representation.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done.

reschke:
> Why is this a MongoDB specific aspect? For MongoDB we know it is not supported (see [this skip here](https://github.com/apache/jackrabbit-oak/blob/0e327a27eb988a0f7656535275f3735ef8474e5d/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BasicDocumentStoreTest.java#L760)) - but for RDB it is supported. So shouldn't this PR keep the support for RDB?

I believe we should come up with consistent behaviour for all types of persistences. My preference would be to disallow Strings with broken UTF-8 serializations, but that would be a change for segment persistence, and that's why I faced resistence when I proposed that back then.

For *this* PR, it will have to handle what the current system allows.

stefan-egli:
With the default set to disable compression, in theory this PR shouldn't do no harm and could be merged, as changing the default is a conscious decision that could be done based on the requirement of the system and which DocumentStore is used.

Having said that, I think the current state of `testInterestingStrings` isn't ideal, as it lists a broken surrogate in a test string, but then doesn't test it. So I'd say either it uses that broken surrogate in a test, or then shouldn't list it at all. Except, I think now that this problem with broken surrogate was identified, we need to address it, so we probably should have it in the test indeed.

What we might do is have the test use different DocumentStore fixtures and test on memory, mongo and rdb *with* the broken surrogate property and verify that the behavior with or without compression is the same.

That could mean that for mongo the fact that compression doesn't support broken surrogates turns out as a non-issue as mongo already doesn't support it. It would just be good to have a test that verifies that.

For RDB it could be different: it could mean that if you want to use compression for RDB you'd have to accept that broken surrogates are not supported (unless we find a fast way to support it in compression).

ionutzpi:
@reschke @reschke I created methods to check the behavior with broken surrogate before and after introducing compression. After compression the test failed.

reschke:
bq. For RDB it could be different: it could mean that if you want to use compression for RDB you'd have to accept that broken surrogates are not supported (unless we find a fast way to support it in compression).

The elephant in the room is segment persistence, which IMHO allows *any* Java string. I don't believe discussing the differences of document store backends is helpful here; we should come to an agreement what the expections for Oak in *general* are. Having document store and segment store behave differently IMHO is a problem we should solve (by fixing segment store, would be my preference).

ionutzpi:
@reschke I vote for your preference(fixing segment store)

stefan-egli:
Turns out that it might look slightly different altogether:

In the latest test from @ionutzpi where it passes a broken surrogate through NodeBuilder (except it shouldn't also go via DocumentPropertyState there), it can be seen that when it applies a property to the commit, as part of [persist](https://github.com/apache/jackrabbit-oak/blob/a5300d8088eacf7d5662ef5b8a7a64abad1ce7bf/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBranch.java#L313) / [compareAgainstBaseState](https://github.com/apache/jackrabbit-oak/blob/a5300d8088eacf7d5662ef5b8a7a64abad1ce7bf/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBranch.java#L289), it does go through [JsonSerializer](https://github.com/apache/jackrabbit-oak/blob/a5300d8088eacf7d5662ef5b8a7a64abad1ce7bf/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CommitDiff.java#L182-L183), which handles broken surrogates properly.

So in short : DocumentNodeStore does actually handle broken surrogates in property values fine.

What I think happened in that original [testInterestingStrings](https://github.com/apache/jackrabbit-oak/blob/0e327a27eb988a0f7656535275f3735ef8474e5d/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BasicDocumentStoreTest.java#L750), is that it the DocumentStore API directly - instead of the DocumentNodeStore API...

Wrt this PR I think there are a few things still open though :
* beautify the newly introduced tests - they have a few issue, namely : 
  * `getBrokenSurrogateAndInitializeDifferentStores` calls `createPropAndCheckValue` 3 times dependent on the DocumentStore used, except there's no difference between the 3 calls, can do that in 1 call. 
  * the way it uses the fixture is different from how it is classically done through a parameterized test. either I'd suggest to consider a parameterized test (where it would dispose the fixture at the end) or make sure to dispose the store/documentNodeStore created. The way it is done now leaves the DocumentNodeStore undisposed, which means follow-up tests hit an active lease and wait for lease expiration, which extends the test unnecessarily
  * `createPropAndCheckValue` as mentioned should go via `DocumentPropertyState` when setting the property value. that is not how it happens in existing code. `DocumentPropertyState` is only used for reading. Writing happens via builder which uses things like ModifiedDocumentNodeState, goes via jcr ValueImpl, StringPropertyState etc. 
* additionally though I believe there is a bigger issue : also based on the new test, when debugging through compression/decompression it can be seen that immediately after a `DocumentPropertyState` is created and with compression the value is compressed, it goes ahead and wants to put it into the cache and for that wants to know its size, so it goes an calls `parsed()` - several times - as part of which it calls decompression - several times. I think this highlights a bigger problem, that we should see if/how compression is performant. It seems the way it is now would add a lot of decompression overhead. This needs further improvement in my view.

wdyt?

ionutzpi:
After @stefan-egli  insights and looking on performance implication for whole properties compression/decompression process, my opinion is that it doesn't brink to much value to continue.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -297,13 +297,11 @@ private void getBrokenSurrogateAndInitializeDifferentStores(DocumentStoreFixture
...
-        } else if (store instanceof RDBDocumentStore) {
-            createPropAndCheckValue(nodeStore, test, compressionEnabled);
-        } else {
-            createPropAndCheckValue(nodeStore, test, compressionEnabled);
-        }
+
+        createPropAndCheckValue(nodeStore, test, compressionEnabled);
+
+        nodeStore.dispose();
+        store.dispose();
```
stefan-egli: I would suggest to do that in a teardown or at least a finally block

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -18,26 +18,50 @@
...
 import org.apache.jackrabbit.oak.api.Blob;
+import org.apache.jackrabbit.oak.api.CommitFailedException;
 import org.apache.jackrabbit.oak.api.PropertyState;
 import org.apache.jackrabbit.oak.api.Type;
+import org.apache.jackrabbit.oak.commons.Compression;
+import org.apache.jackrabbit.oak.plugins.document.mongo.MongoDocumentStore;
+import org.apache.jackrabbit.oak.plugins.document.mongo.MongoTestUtils;
+import org.apache.jackrabbit.oak.plugins.document.rdb.RDBDocumentStore;
+import org.apache.jackrabbit.oak.plugins.document.util.MongoConnection;
+import org.apache.jackrabbit.oak.plugins.memory.PropertyStates;
```
stefan-egli: minor : these aren't needed

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -18,26 +18,50 @@
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
+import java.lang.reflect.InvocationTargetException;
 import java.util.List;
+import java.util.Map;
```
stefan-egli: minor: this isn't needed

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -81,4 +114,289 @@ public void multiValuedBinarySize() throws Exception {
...
+
+    @Test
+    public void testInterestingStringsWithoutCompression() {
+        DocumentNodeStore store = mock(DocumentNodeStore.class);
+        String[] tests = new String[] { "simple:foo", "cr:a\n\b", "dquote:a\"b", "bs:a\\b", "euro:a\u201c", "gclef:\uD834\uDD1E",
+                "tab:a\tb", "nul:a\u0000b"};
+
+        for (String test : tests) {
+                DocumentPropertyState state = new DocumentPropertyState(store, "propertyName", test, Compression.GZIP);
+                assertEquals(test, state.getValue());
```
stefan-egli: minor : indented too far

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -81,4 +114,289 @@ public void multiValuedBinarySize() throws Exception {
...
+
+    }
+
+    private void createPropAndCheckValue(DocumentNodeStore nodeStore, String test, boolean compressionEnabled) throws CommitFailedException {
+        NodeBuilder builder = nodeStore.getRoot().builder();
+        if (compressionEnabled) {
+            DocumentPropertyState.setCompressionThreshold(1);
+        }
+        DocumentPropertyState documentPropertyState = new DocumentPropertyState(nodeStore, "p", test, Compression.GZIP);
+        builder.child(TEST_NODE).setProperty("p", documentPropertyState.getValue(), Type.STRING);
```
stefan-egli:
```suggestion
        builder.child(TEST_NODE).setProperty("p", test, Type.STRING);
```

As mentioned in the [previous comment](https://github.com/apache/jackrabbit-oak/pull/1526#issuecomment-2236322219), going via DocumentPropertyState when setting a property on a NodeBuilder is not correct (I might have made that point not fully clear there).

With this fix, the tests actually succeed

stefan-egli (commented):
After some further offline discussion the suggested plan is as follows:

* fix the last bits of the PR (mainly the interestingString test buglet where it goes via DocumentStateProperty but shouldn't, the rest are nitpicks).
* then merge, however this time the default should really be that it is disabled and does no harm by default
* before activation of the compression however, performance tests are necessary. At the moment the assumption is that enabling compression comes at cost of CPU (it calls getValue() several times right after getNodeAtRevision, which would seem inefficient and needs to be improved).

**tldr** let's get this finished and merged, but before using compression, (more) performance tests and improvements are needed, the suggestion is to not yet use this feature as is (unless the consequences are understood)

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done. Thanks.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
ionutzpi: Done

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -81,4 +110,288 @@ public void multiValuedBinarySize() throws Exception {
...
+
+            createPropAndCheckValue(nodeStore, test, compressionEnabled);
+
+        } finally {
+            if (nodeStore != null) {
+                nodeStore.dispose();
+            }
+            if (store != null) {
+                store.dispose();
+            }
```
stefan-egli:
```suggestion
```

`nodeStore.dispose` already disposes the `DocumentStore`, calling it twice here results in a test failure:

```
[ERROR] testBrokenSurrogateWithoutCompressionForRDB(org.apache.jackrabbit.oak.plugins.document.DocumentPropertyStateTest)  Time elapsed: 0.19 s  <<< ERROR!
java.lang.IllegalStateException: Connection handler is already closed (0ms ago)
	at org.apache.jackrabbit.oak.plugins.document.rdb.RDBConnectionHandler.getDataSource(RDBConnectionHandler.java:144)
	at org.apache.jackrabbit.oak.plugins.document.rdb.RDBConnectionHandler.getConnection(RDBConnectionHandler.java:154)
	at org.apache.jackrabbit.oak.plugins.document.rdb.RDBConnectionHandler.getRWConnection(RDBConnectionHandler.java:76)
	at org.apache.jackrabbit.oak.plugins.document.rdb.RDBDocumentStore.dispose(RDBDocumentStore.java:797)
	at org.apache.jackrabbit.oak.plugins.document.DocumentPropertyStateTest.getBrokenSurrogateAndInitializeDifferentStores(DocumentPropertyStateTest.java:308)
```

ionutzpi: Done.

stefan-egli (approved):
+1, lgtm now. I'll (squash and) merge. Feature should this time hopefully be disabled thus have no effect.

PR #1535, opened 2024-06-17
OAK-10890 - Added log warning

ionutzpi:


reschke (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexEditor.java
```diff
@@ -328,6 +332,7 @@ private void checkUniquenessConstraints() throws CommitFailedException {
                     String msg = String.format(
                             "Uniqueness constraint violated property %s having value %s",
                             propertyNames, failed);
+                    log.warn(msg);
```
stefan-egli: I'm wondering if we should add a prefix to the log. Without the prefix it might be hard for down-stream logging to distinguish between the two sources of this message (the log.warn and the exception), as they contain a large common part. What about something like below (or maybe a better variant of it) ?
```suggestion
                    log.warn("checkUniquenessConstraints: " + msg);
```

stefan-egli:
+1 other than the minor comment. 
PS: there's another case of 0030 in [UniquenessConstraintValidator](https://github.com/apache/jackrabbit-oak/blob/0e327a27eb988a0f7656535275f3735ef8474e5d/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/property/UniquenessConstraintValidator.java#L81) but I'd go with this PropertyIndexEditor as a start. Was also wondering if the fact that it contains the effected paths is a leak of sorts - but given the CommitFailedException is frequently logged anyway, that doesn't seem to be a problem.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexEditor.java
ionutzpi: Done

stefan-egli (approved):

stefan-egli:
@ionutzpi should I go ahead and merge, i.e. is the PR ready?

ionutzpi:
> @ionutzpi should I go ahead and merge, i.e. is the PR ready?

yes, the PR is ready

stefan-egli:
ok - just checking why the build fails - restarted it now at https://ci-builds.apache.org/job/Jackrabbit/job/oak-trunk-pr/job/PR-1535/3/

PR #1539, opened 2024-06-17
OAK-10896: - Add OSGI configuration variable for removal of deleted properties and orphaned nodes

shodaaan:
- create class FullGCOptions which will hold the different parameters used for running full GC
- added variables to Configuration.java for 2 full GC modes: GAP_ORPHANS and EMPTY_PROPERTIES
- updated VersionGarbageCollector constructor, as well as DocumentNodeStore Builder / Service and DocumentNodeStore to get values received from Configuration via the FullGCOptions class

- applied the new full GC mode options in VersionGarbageCollector via the fullGCOptions parameter passed in the constructor
- added unit test methods for testing feature toggle and configuration options set in DocumentNodeStoreBuilder, based on existing testing of feature toggles and configuration settings in unit tests
- fixed CommitBuilderTest test failures by changing expected exception type


stefan-egli (commented):
I think the main aspect of this work is the question whether we want actual feature (toggles) for the new config options or not (despite the jira title). In my opinion we wouldn't automatically need a full-fledged feature toggle for each new config we introduce, as they probably have a higher cost. With "gc mode" related ones, I'd vote to treat them as plain osgi configuration only.

Another aspect is the question of having a separate `FullGCOptions` class or merging it with the existing `VersionGCOptions`. We had a history of treating things separate (eg `VersionGarbageCollector.FullGC`) or combined (`VersionGCRecommendations`) - so we weren't consistent. For `FullGCOptions` case I'm not sure there is advantages of having it separate though? One argument that comes to mind for combining is that full-gc already relies on the existing options, so from that point of view it doesn't live separate. This is just my opinion though...

PS: please avoid wildcard imports

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -38,22 +38,11 @@
...
-import org.apache.jackrabbit.oak.plugins.document.DocumentNodeStore;
-import org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreBuilder;
-import org.apache.jackrabbit.oak.plugins.document.DocumentStore;
-import org.apache.jackrabbit.oak.plugins.document.FormatVersion;
-import org.apache.jackrabbit.oak.plugins.document.MissingLastRevSeeker;
-import org.apache.jackrabbit.oak.plugins.document.NodeDocument;
-import org.apache.jackrabbit.oak.plugins.document.RevisionContextWrapper;
-import org.apache.jackrabbit.oak.plugins.document.SweepHelper;
-import org.apache.jackrabbit.oak.plugins.document.VersionGCSupport;
+import org.apache.jackrabbit.oak.plugins.document.*;
```
rishabhdaim: Please avoid `*` imports.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -69,11 +58,7 @@
 import static org.apache.jackrabbit.oak.plugins.document.Collection.NODES;
 import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreHelper.createVersionGC;
 import static org.apache.jackrabbit.oak.plugins.document.FormatVersion.versionOf;
-import static org.apache.jackrabbit.oak.plugins.document.util.Utils.getIdFromPath;
-import static org.apache.jackrabbit.oak.plugins.document.util.Utils.getRootDocument;
-import static org.apache.jackrabbit.oak.plugins.document.util.Utils.isFullGCEnabled;
-import static org.apache.jackrabbit.oak.plugins.document.util.Utils.isEmbeddedVerificationEnabled;
-import static org.apache.jackrabbit.oak.plugins.document.util.Utils.timestampToString;
+import static org.apache.jackrabbit.oak.plugins.document.util.Utils.*;
```
rishabhdaim: Please avoid `*` imports.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
```diff
@@ -34,9 +34,7 @@
 import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreBuilder.DEFAULT_NODE_CACHE_PERCENTAGE;
 import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreBuilder.DEFAULT_PREV_DOC_CACHE_PERCENTAGE;
 import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreBuilder.DEFAULT_UPDATE_LIMIT;
-import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreService.DEFAULT_FULL_GC_ENABLED;
-import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreService.DEFAULT_EMBEDDED_VERIFICATION_ENABLED;
-import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreService.DEFAULT_THROTTLING_ENABLED;
+import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreService.*;
```
rishabhdaim: Please avoid `*` imports.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
```diff
@@ -661,7 +655,9 @@ public DocumentNodeStore(DocumentNodeStoreBuilder<?> builder) {
         this.branches = new UnmergedBranches();
         this.asyncDelay = builder.getAsyncDelay();
         this.versionGarbageCollector = new VersionGarbageCollector(
-                this, builder.createVersionGCSupport(), isFullGCEnabled(builder), false, isEmbeddedVerificationEnabled(builder));
+                this, builder.createVersionGCSupport(), isFullGCEnabled(builder), false,
+                isEmbeddedVerificationEnabled(builder),
+                new FullGCOptions(isFullGCModeGapOrphansEnabled(builder), Utils.isFullGCModeEmptyPropertiesEnabled(builder)));
```
rishabhdaim:
```suggestion
                new FullGCOptions(isFullGCModeGapOrphansEnabled(builder), isFullGCModeEmptyPropertiesEnabled(builder)));
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -312,6 +316,20 @@ public boolean isEmbeddedVerificationEnabled() {
         return this.embeddedVerificationEnabled;
     }
 
+    public T setFullGCModeGapOrphansEnabled(boolean b) {
+        this.fullGCModeGapOrphansEnabled = b;
+        return thisBuilder();
+    }
+
+    public boolean isFullGCModeGapOrphansEnabled() { return this.fullGCModeGapOrphansEnabled; }
```
rishabhdaim:
```suggestion
    public boolean isFullGCModeGapOrphansEnabled() { 
    return this.fullGCModeGapOrphansEnabled; 
    }
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -312,6 +316,20 @@ public boolean isEmbeddedVerificationEnabled() {
...
+
+    public T setFullGCModeEmptyPropertiesEnabled(boolean b) {
+        this.fullGCModeEmptyPropertiesEnabled = b;
+        return thisBuilder();
+    }
+
+    public boolean isFullGCModeEmptyPropertiesEnabled() { return this.fullGCModeEmptyPropertiesEnabled; }
```
rishabhdaim:
```suggestion
    public boolean isFullGCModeEmptyPropertiesEnabled() { 
    return this.fullGCModeEmptyPropertiesEnabled; 
    }
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -380,6 +398,22 @@ public Feature getDocStoreEmbeddedVerificationFeature() {
         return docStoreEmbeddedVerificationFeature;
     }
 
+    public T setDocStoreFullGCModeGapOrphansFeature(@Nullable Feature docStoreFullGCModeGapOrphans) {
+        this.docStoreFullGCModeGapOrphansFeature = docStoreFullGCModeGapOrphans;
+        return thisBuilder();
+    }
+
+    @Nullable
+    public Feature getDocStoreFullGCModeGapOrphansFeature() { return this.docStoreFullGCModeGapOrphansFeature;    }
```
rishabhdaim:
```suggestion
    public Feature getDocStoreFullGCModeGapOrphansFeature() { 
    return this.docStoreFullGCModeGapOrphansFeature;    
    }
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -380,6 +398,22 @@ public Feature getDocStoreEmbeddedVerificationFeature() {
...
+
+    public T setDocStoreFullGCModeEmptyPropertiesFeature(@Nullable Feature docStoreFullGCModeEmptyProperties) {
+        this.docStoreFullGCModeEmptyPropertiesFeature = docStoreFullGCModeEmptyProperties;
+        return thisBuilder();
+    }
+
+    @Nullable
+    public Feature getDocStoreFullGCModeEmptyPropertiesFeature() { return this.docStoreFullGCModeEmptyPropertiesFeature; };
```
rishabhdaim:
```suggestion
    public Feature getDocStoreFullGCModeEmptyPropertiesFeature() { 
    return this.docStoreFullGCModeEmptyPropertiesFeature; 
    }
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
```diff
@@ -199,6 +201,16 @@ public class DocumentNodeStoreService {
      */
     private static final String FT_NAME_EMBEDDED_VERIFICATION = "FT_EMBEDDED_VERIFICATION_OAK-10633";
 
+    /**
+     * Feature toggle name to enable full GC mode GAP_ORPHANS cleanup for Mongo Document Store
+     */
+    private static final String FT_NAME_FULL_GC_MODE_GAP_ORPHANS = "FT_FULL_GC_MODE_GAP_ORPHANS";
```
rishabhdaim: please add the OAK ticket id in the feature toggle name.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
```diff
@@ -199,6 +201,16 @@ public class DocumentNodeStoreService {
...
+
+    /**
+     * Feature toggle name to enable full GC mode EMPTY_PROPERTIES cleanup for Mongo Document Store
+     */
+    private static final String FT_NAME_FULL_GC_MODE_EMPTY_PROPERTIES = "FT_FULL_GC_MODE_EMPTY_PROPERTIES";
```
rishabhdaim: same as above

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilder.java
```diff
@@ -139,6 +163,25 @@ public Feature getDocStoreFullGCFeature() {
...
     }
 
+    public RDBDocumentNodeStoreBuilder setDocStoreFullGCModeGapOrphansFeature(@Nullable Feature docStoreFullGCModeGapOrphans) {
+        // fullGC modes are not supported for RDB
+        return thisBuilder();
+    }
+
+    @Override
+    @Nullable
+    public Feature getDocStoreFullGCModeGapOrphansFeature() { return null; }
```
rishabhdaim:
```suggestion
    public Feature getDocStoreFullGCModeGapOrphansFeature() { 
    return null; 
    }
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilder.java
```diff
@@ -139,6 +163,25 @@ public Feature getDocStoreFullGCFeature() {
...
+
+    @Override
+    public RDBDocumentNodeStoreBuilder setDocStoreFullGCModeEmptyPropertiesFeature(@Nullable Feature docStoreFullGCModeEmptyProperties) {
+        // fullGC modes are not supported for RDB
+        return thisBuilder();
+    }
+
+    @Override
+    @Nullable
+    public Feature getDocStoreFullGCModeEmptyPropertiesFeature() { return null; };
```
rishabhdaim:
```suggestion
    public Feature getDocStoreFullGCModeEmptyPropertiesFeature() { 
    return null; 
    }
```

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilderTest.java
```diff
@@ -16,9 +16,7 @@
  */
 package org.apache.jackrabbit.oak.plugins.document.rdb;
 
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
```
rishabhdaim: please avoid `*` imports.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -1009,6 +1009,28 @@ public static boolean isEmbeddedVerificationEnabled(final DocumentNodeStoreBuild
         return builder.isEmbeddedVerificationEnabled() || (docStoreEmbeddedVerificationFeature != null && docStoreEmbeddedVerificationFeature.isEnabled());
     }
 
+    /**
+     * Check whether full GC mode gap orphans is enabled or not for document store.
+     *
+     * @param builder instance for DocumentNodeStoreBuilder
+     * @return true if full GC is enabled else false
+     */
+    public static boolean isFullGCModeGapOrphansEnabled(final DocumentNodeStoreBuilder<?> builder) {
```
rishabhdaim: please add unit cases for this method.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -1009,6 +1009,28 @@ public static boolean isEmbeddedVerificationEnabled(final DocumentNodeStoreBuild
...
+        return builder.isFullGCModeGapOrphansEnabled() || (docStoreFullGCModeGapOrphansFeature != null && docStoreFullGCModeGapOrphansFeature.isEnabled());
+    }
+
+    /**
+     * Check whether full GC mode gap orphans is enabled or not for document store.
+     *
+     * @param builder instance for DocumentNodeStoreBuilder
+     * @return true if full GC is enabled else false
+     */
+    public static boolean isFullGCModeEmptyPropertiesEnabled(final DocumentNodeStoreBuilder<?> builder) {
```
rishabhdaim: please add unit cases for this method.

rishabhdaim:
> I think the main aspect of this work is the question whether we want actual feature (toggles) for the new config options or not (despite the jira title). In my opinion we wouldn't automatically need a full-fledged feature toggle for each new config we introduce, as they probably have a higher cost. With "gc mode" related ones, I'd vote to treat them as plain osgi configuration only.
> 
> Another aspect is the question of having a separate `FullGCOptions` class or merging it with the existing `VersionGCOptions`. We had a history of treating things separate (eg `VersionGarbageCollector.FullGC`) or combined (`VersionGCRecommendations`) - so we weren't consistent. For `FullGCOptions` case I'm not sure there is advantages of having it separate though? One argument that comes to mind for combining is that full-gc already relies on the existing options, so from that point of view it doesn't live separate. This is just my opinion though...
> 
> PS: please avoid wildcard imports

I agree with @stefan-egli that we don't need feature toggles for each FULL_GC_MODE, just one feature toggle is enough to enable/disable FULL_GC.

However, I recommend using an OSGI config (with values limited from an enum of FULL_GC_MODE) to select the FULL_GC_MODE.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
stefan-egli: +1

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
```diff
@@ -35,13 +35,7 @@
...
 import static org.apache.jackrabbit.oak.plugins.document.NodeDocument.PATH;
 import static org.apache.jackrabbit.oak.plugins.document.Path.ROOT;
-import static org.apache.jackrabbit.oak.plugins.document.util.Utils.alignWithExternalRevisions;
-import static org.apache.jackrabbit.oak.plugins.document.util.Utils.getIdFromPath;
-import static org.apache.jackrabbit.oak.plugins.document.util.Utils.getModuleVersion;
-import static org.apache.jackrabbit.oak.plugins.document.util.Utils.isFullGCEnabled;
-import static org.apache.jackrabbit.oak.plugins.document.util.Utils.isEmbeddedVerificationEnabled;
-import static org.apache.jackrabbit.oak.plugins.document.util.Utils.isThrottlingEnabled;
-import static org.apache.jackrabbit.oak.plugins.document.util.Utils.pathToId;
+import static org.apache.jackrabbit.oak.plugins.document.util.Utils.*;
```
stefan-egli: no wildcard imports

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
```diff
@@ -140,6 +140,8 @@ public class DocumentNodeStoreService {
     static final boolean DEFAULT_THROTTLING_ENABLED = false;
     static final boolean DEFAULT_FULL_GC_ENABLED = false;
     static final boolean DEFAULT_EMBEDDED_VERIFICATION_ENABLED = true;
+    static final int DEFAULT_FULL_GC_MODE = 0;
+    static final boolean DEFAULT_FULL_GC_EMPTY_PROPERTIES_ENABLED = false;
```
stefan-egli: this ls unused

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
```diff
@@ -199,6 +201,16 @@ public class DocumentNodeStoreService {
...
+    /**
+     * Feature toggle name to enable full GC mode GAP_ORPHANS cleanup for Mongo Document Store
+     */
+    private static final String FT_NAME_FULL_GC_MODE_GAP_ORPHANS = "FT_FULL_GC_MODE_GAP_ORPHANS_OAK-10896";
+
+    /**
+     * Feature toggle name to enable full GC mode EMPTY_PROPERTIES cleanup for Mongo Document Store
+     */
+    private static final String FT_NAME_FULL_GC_MODE_EMPTY_PROPERTIES = "FT_FULL_GC_MODE_EMPTY_PROPERTIES_OAK-10896";
+
```
stefan-egli: I was under the impression we decided not to use feature toggles, in which case this (and related code) is not necessary

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
```diff
@@ -276,6 +290,8 @@ protected void activate(ComponentContext context, Configuration config) throws E
         cancelInvalidationFeature = Feature.newFeature(FT_NAME_CANCEL_INVALIDATION, whiteboard);
         docStoreFullGCFeature = Feature.newFeature(FT_NAME_FULL_GC, whiteboard);
         docStoreEmbeddedVerificationFeature = Feature.newFeature(FT_NAME_EMBEDDED_VERIFICATION, whiteboard);
+        docStoreFullGCModeGapOrphansFeature = Feature.newFeature(FT_NAME_FULL_GC_MODE_GAP_ORPHANS, whiteboard);
+        docStoreFullGCModeEmptyPropertiesFeature = Feature.newFeature(FT_NAME_FULL_GC_MODE_EMPTY_PROPERTIES, whiteboard);
```
stefan-egli: (same as above comment - I'll not repeat this comment further though)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1092,61 +1124,63 @@ public void collectGarbage(final NodeDocument doc, final GCPhases phases) {
             }
 
             if (!isDeletedOrOrphanedNode(traversedState, greatestExistingAncestorOrSelf, phases, doc)) {
+
```
stefan-egli: please avoid inserting empty lines unless say demarking new code for readability etc

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1092,61 +1124,63 @@ public void collectGarbage(final NodeDocument doc, final GCPhases phases) {
...
                 // here the node is not orphaned which means that we can reach the node from root
-                switch(fullGcMode) {
-                    case NONE : {
+                switch (fullGcMode) {
+                    case NONE: {
```
stefan-egli: please avoid unrelated reformatting

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1092,61 +1124,63 @@ public void collectGarbage(final NodeDocument doc, final GCPhases phases) {
...
+            // only add if there are changes for this doc
+            if (op.hasChanges()) {
+                op.equals(MODIFIED_IN_SECS, doc.getModified());
+                garbageDocsCount++;
+                totalGarbageDocsCount++;
+                monitor.info("Collected [{}] garbage count in [{}]", op.getChanges().size(), doc.getId());
+                AUDIT_LOG.info("<Collected> [{}] garbage count  in [{}]", op.getChanges().size(), doc.getId());
+                updateOpList.add(op);
+            }
+
```
stefan-egli: why was this moved? doesn't look correct

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1226,7 +1260,10 @@ private boolean isDeletedOrOrphanedNode(final NodeState traversedState, final Pa
                 phases.stop(GCPhase.FULL_GC_COLLECT_ORPHAN_NODES);
                 return false;
             }
+
+            // first check if gapOrphansModeEnabled is set in fullGCOptions, then check in fullGCMode enum
             if (fullGcMode == GAP_ORPHANS || fullGcMode == GAP_ORPHANS_EMPTYPROPS) {
+
```
stefan-egli: (as mentioned above, we should only include related changes, this new empty line doesn't seem necessary)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1226,7 +1260,10 @@ private boolean isDeletedOrOrphanedNode(final NodeState traversedState, final Pa
                 phases.stop(GCPhase.FULL_GC_COLLECT_ORPHAN_NODES);
                 return false;
             }
+
+            // first check if gapOrphansModeEnabled is set in fullGCOptions, then check in fullGCMode enum
```
stefan-egli: not sure I see the purpose of this comment, would suggest to remove

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilder.java
```diff
@@ -128,6 +128,18 @@ public RDBDocumentNodeStoreBuilder setFullGCEnabled(boolean b) {
         return thisBuilder();
     }
 
+    @Override
+    public RDBDocumentNodeStoreBuilder setFullGCMode(int v) {
+        // fullGC modes are not supported for RDB
```
stefan-egli, line 163: should we then throw something like an `IllegalArgumentException` here?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CommitBuilderTest.java
```diff
@@ -83,8 +83,8 @@ public void buildWithNullRevision() {
         CommitBuilder builder = new CommitBuilder(ns, null);
         try {
             builder.build(null);
-            expectNPE();
-        } catch (NullPointerException e) {
+            expectIllegalArgumentException();
+        } catch (IllegalArgumentException e) {
```
stefan-egli: looks wrong, please revert

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CommitBuilderTest.java
```diff
@@ -119,8 +119,8 @@ public void addNodePathNull() {
         CommitBuilder builder = new CommitBuilder(ns, null);
         try {
             builder.addNode((Path) null);
-            expectNPE();
-        } catch (NullPointerException e) {
+            expectIllegalArgumentException();
+        } catch (IllegalArgumentException e) {
```
stefan-egli: same

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CommitBuilderTest.java
```diff
@@ -302,6 +302,10 @@ public void defaultStartRevisions() {
         assertEquals(new RevisionVector(), c.getStartRevisions());
     }
 
+    private static void expectIllegalArgumentException() {
+        fail("IllegalArgumentException expected");
+    }
+
     private static void expectNPE() {
```
stefan-egli: I didn't check all cases but the first I checked was still a NPE - why was this change applied?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollectorIT.java
```diff
@@ -308,6 +310,9 @@ private void createSecondaryStore(LeaseCheckMode leaseCheckNode, boolean withFai
                 .setLeaseCheckMode(leaseCheckNode)
                 .setDocumentStore(ds2).setAsyncDelay(0);
         store2 = documentMKBuilder2.getNodeStore();
+        // custom fullGcMode needs to be set after each node store creation, since the VersionGarbageCollector
+        // constructor sets the fullGcMode to the value read from OSGI Configuration - method setFullGcMode
+        writeStaticField(VersionGarbageCollector.class, "fullGcMode", fullGcMode, true);
```
stefan-egli, line 316: can't we use `setFullGcMode` here?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilderTest.java
```diff
@@ -16,9 +16,10 @@
  */
 package org.apache.jackrabbit.oak.plugins.document.rdb;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
-import static org.junit.Assert.fail;
```
stefan-egli: please avoid unnecessary import reorgs

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -211,11 +211,31 @@ static FullGCMode getFullGcMode() {
...
+                fullGcMode = NONE;
+                break;
+            case 2:
+                fullGcMode = GAP_ORPHANS;
+                break;
+            case 3:
+                fullGcMode = GAP_ORPHANS_EMPTYPROPS;
+                break;
+        }
+    }
```
stefan-egli, line 235: I would suggest to add a `default` catch all section here that logs the fact that a not support mode was used and that the default mode is used instead

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -211,11 +211,31 @@ static FullGCMode getFullGcMode() {
...
+
     private final DocumentNodeStore nodeStore;
     private final DocumentStore ds;
     private final boolean fullGCEnabled;
     private final boolean isFullGCDryRun;
     private final boolean embeddedVerification;
+    private final int fullGCModeConfigValue;
```
stefan-egli: this seems unused

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -228,13 +248,25 @@ static FullGCMode getFullGcMode() {
                             final boolean fullGCEnabled,
                             final boolean isFullGCDryRun,
                             final boolean embeddedVerification) {
+        this(nodeStore, gcSupport, fullGCEnabled, isFullGCDryRun, embeddedVerification, 0);
```
stefan-egli: please use `Configuration.DEFAULT_FULL_GC_MODE` instead of 0

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -228,13 +248,25 @@ static FullGCMode getFullGcMode() {
...
+    }
+
+    VersionGarbageCollector(DocumentNodeStore nodeStore,
+                            VersionGCSupport gcSupport,
+                            final boolean fullGCEnabled,
+                            final boolean isFullGCDryRun,
+                            final boolean embeddedVerification,
+                            final int fullGCModeConfigVal) {
```
stefan-egli: what about using a shorter name like below?

```suggestion
                            final int fullGCMode) {
```

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGCTest.java
```diff
@@ -531,6 +531,51 @@ public void testResetFullGCDryRunMode() throws Exception {
 
     // OAK-10370 END
 
+    // OAK-10896
+
+    @Test
+    public void testVersionGCLoadGCModeConfigurationNotApplicable() throws Exception {
+        int FULL_GC_MODE_NOT_ALLOWED_VALUE = 5;
```
stefan-egli: variable names shouldn't be all capital

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGCTest.java
```diff
@@ -531,6 +531,51 @@ public void testResetFullGCDryRunMode() throws Exception {
...
+        VersionGarbageCollector gc = new VersionGarbageCollector(
+                ns, new VersionGCSupport(store), true, false, false,
+                FULL_GC_MODE_NOT_ALLOWED_VALUE);
+
+        assertEquals("Starting VersionGarbageCollector with not applicable / not allowed value" +
+                "will set fullGcMode to default NONE", gc.getFullGcMode(), VersionGarbageCollector.FullGCMode.NONE);
```
stefan-egli: this test should actually have failed,a s the coded default `VersionGarbageCollector.FullGCMode` is `GAP_ORPHANS_EMPTYPROPS` - the test only succeeds as a side-effect of `setUp` building a `DocumentNodeStore` which uses `DocumentNodeStoreBuilder.getFullGCMode` which uses an uninitialized thus default value for `fullGCMode`. 

I' suggest to change the test to first set a valid mode (eg 2) an then set an invalid mode (eg the 5) and then it should have been set to 0

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -171,6 +171,7 @@ public class DocumentNodeStoreBuilder<T extends DocumentNodeStoreBuilder<T>> {
     private boolean throttlingEnabled;
     private boolean fullGCEnabled;
     private boolean embeddedVerificationEnabled = DocumentNodeStoreService.DEFAULT_EMBEDDED_VERIFICATION_ENABLED;
+    private int fullGCMode;
```
stefan-egli: please initialize with the default

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
stefan-egli: (same applies to other case statements below)

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollectorIT.java
stefan-egli: ok, I see, we had it like this previously .. I'd still vote to introducing a dedicated testing-only setter (`static void setFullGcMode(FullGCMode)`) instead of reflection (for this and a few other similar cases for fullgcmode in this and the other test class) - makes code maintenance easier imv

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
stefan-egli: (I see that we're not consistent with this .. eg as with the enabled booleans above, which default to false and that happens to be what is also the default ... perhaps we can ignore that fact plus given this is an `int`, it might be good to set it explicitly)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
stefan-egli: (same applies to different places)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
```diff
@@ -35,8 +35,9 @@
 import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreBuilder.DEFAULT_PREV_DOC_CACHE_PERCENTAGE;
 import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreBuilder.DEFAULT_UPDATE_LIMIT;
 import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreService.DEFAULT_FULL_GC_ENABLED;
-import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreService.DEFAULT_EMBEDDED_VERIFICATION_ENABLED;
```
rishabhdaim, line 38: please revert this un-intended change.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
```diff
@@ -324,4 +325,12 @@
                     ". Note that this value can be overridden via framework " +
                     "property 'oak.documentstore.embeddedVerificationEnabled'")
     boolean embeddedVerificationEnabled() default DEFAULT_EMBEDDED_VERIFICATION_ENABLED;
+
+    @AttributeDefinition(
+            name = "Document Node Store Full GC Mode",
+            description = "Boolean value indicating which Full GC mode should be enabled for " +
```
rishabhdaim:
```suggestion
            description = "Integer value indicating which Full GC mode should be used for " +
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
```diff
@@ -324,4 +325,12 @@
...
+                    "document node store or not. The Default value is " + DEFAULT_FULL_GC_MODE +
```
rishabhdaim:
```suggestion
                    "document node store. The Default value is " + DEFAULT_FULL_GC_MODE +
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
```diff
@@ -324,4 +325,12 @@
                     ". Note that this value can be overridden via framework " +
                     "property 'oak.documentstore.embeddedVerificationEnabled'")
     boolean embeddedVerificationEnabled() default DEFAULT_EMBEDDED_VERIFICATION_ENABLED;
+
+    @AttributeDefinition(
```
rishabhdaim, line 347: please mention that `fullGC` can be entirely disabled via `fullGCEnabled`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilder.java
rishabhdaim: Since `fullGC` is supported for RDB, logging a `WARN` message should be sufficient.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -211,11 +211,31 @@ static FullGCMode getFullGcMode() {
         return fullGcMode;
     }
 
+    /**
+     * Set the full GC mode to be used according to the provided configuration value.
+     * The configuration value will be ignored if it is set to any other values than the supported ones.
+     * @param fullGcModeConfig
+     */
+    static void setFullGcMode(int fullGcModeConfig) {
```
rishabhdaim: We decided to put this logic inside the `FullGCMode` enum.
`FullGCMode` enum should have a constructor with both `int` & `string` value and int value should match with the OSGI config.

rishabhdaim (changes requested):

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
shodaaan: fixed

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
shodaaan: fixed

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
shodaaan: fixed

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
shodaaan: fixed

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
shodaaan: fixed

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilder.java
shodaaan: fixed

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilder.java
shodaaan: fixed

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilderTest.java
shodaaan: fixed

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
shodaaan: no longer necessary

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
shodaaan: no longer necessary

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
shodaaan: fixed

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
shodaaan: fixed.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
shodaaan: fixed, thank you for the comment.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
shodaaan: fixed, thank you for the comment.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
shodaaan: fixed, thank you for the comment. I used DocumentNodeStoreService.DEFAULT_FULL_GC_MODE

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
shodaaan: fixed, thank you for the comment

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
shodaaan: fixed, thank you for the comment

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilder.java
shodaaan: added log.warn, thank you for the suggestion.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CommitBuilderTest.java
shodaaan: The tests fail because an IllegalArgumentException is actually thrown, so I have modified them to pass.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CommitBuilderTest.java
shodaaan: The tests fail because an IllegalArgumentException is actually thrown, so I have modified them to pass.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CommitBuilderTest.java
shodaaan: The tests fail because an IllegalArgumentException is actually thrown, so I have modified them to pass.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGCTest.java
shodaaan: I updated the test, thank you for the suggestion.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollectorIT.java
shodaaan: I propose we make that change when we do a larger scale refactoring. Changing the way we set fullGCMode in unit tests now would also require updating all of the related unit tests (for consistency), and there are a lot of them.

We should do refactoring in a separate, dedicated ticket, not as a part of this feature, IMHO.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
shodaaan: removed, thank you for the observation.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
shodaaan: Done, thank you for the suggestion.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
shodaaan: I set it to the default value, thank you for the suggestion.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
shodaaan: Done, thank you for the suggestion.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
shodaaan: I think this is outside the scope of this ticket.
Doing this change now would also require refactoring the unit tests that use fullGCMode. 
I suggest that we do that in a separate refactoring ticket, since there are a lot of tests involved and we may also consider other changes to the unit tests when we do the refactoring.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
shodaaan: Actually, this is the suggested way to format switch statements, in fact it was automatically suggested by the IntelliJ Java text editor.
Also, other switch statements in the Oak project use this formatting, with no space between the value and ":".

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGCTest.java
shodaaan: Fixed, thank you for the comment

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
```diff
@@ -692,7 +694,7 @@ public int getMemory() {
         diffCache = builder.getDiffCache(this.clusterId);
 
         // check if root node exists
-        NodeDocument rootDoc = store.find(NODES, Utils.getIdFromPath(ROOT));
+        NodeDocument rootDoc = store.find(NODES, getIdFromPath(ROOT));
```
stefan-egli: I think we should try to avoid making unrelated changes like these, they make the PR change surface unnecessarily larger (and in this particular case I don't think it is an improvement - but I guess that is a controversial topic)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
```diff
@@ -716,7 +718,7 @@ public int getMemory() {
             setRoot(head);
             // make sure _lastRev is written back to store
             backgroundWrite();
-            rootDoc = store.find(NODES, Utils.getIdFromPath(ROOT));
+            rootDoc = store.find(NODES, getIdFromPath(ROOT));
```
stefan-egli: (same as above)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
```diff
@@ -2603,7 +2605,7 @@ private void cleanOrphanedBranches() {
     }
 
     private void cleanRootCollisions() {
-        String id = Utils.getIdFromPath(ROOT);
+        String id = getIdFromPath(ROOT);
```
stefan-egli: (same as above throughout this class)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
```diff
@@ -238,6 +239,8 @@ static DocumentStoreType fromString(String type) {
     private Feature cancelInvalidationFeature;
     private Feature docStoreFullGCFeature;
     private Feature docStoreEmbeddedVerificationFeature;
+    private Feature docStoreFullGCModeGapOrphansFeature;
+    private Feature docStoreFullGCModeEmptyPropertiesFeature;
```
stefan-egli: I think we aimed to remove the features here?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
stefan-egli: still, this in unrelated and we should not increase PR change surface when that is not necessary or part of the PR

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
stefan-egli: but the comment is still here?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CommitBuilderTest.java
stefan-egli: it fails with a NPE for me though

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGCTest.java
```diff
@@ -531,6 +531,57 @@ public void testResetFullGCDryRunMode() throws Exception {
...
+        // set fullGcMode to allowed value that is different than NONE
+        VersionGarbageCollector.setFullGcMode(fullGcModeGapOrphans);
+
+        // reinitialize VersionGarbageCollector with not allowed value
+        VersionGarbageCollector gc = new VersionGarbageCollector(
+                ns, new VersionGCSupport(store), true, false, false,
+                fullGcModeNotAllowedValue);
+
+        assertEquals("Starting VersionGarbageCollector with not applicable / not allowed value" +
+                "will set fullGcMode to default NONE", gc.getFullGcMode(), VersionGarbageCollector.FullGCMode.NONE);
```
stefan-egli:
```suggestion
                "will set fullGcMode to default NONE", VersionGarbageCollector.getFullGcMode(), VersionGarbageCollector.FullGCMode.NONE);
```

static method should be accessed in a static way

stefan-egli (commented):
I think the unrelated changes in particular in DocumentNodeStore should be avoided. Also, the feature code is not fully removed yet

stefan-egli:
PS: I think we should adjust jira/pr titles to talk about osgi config instead of feature toggle (to avoid future confusion)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
shodaaan: This is consistent with how we are using static method calls. The consistent approach would be IMHO to actually do this everywhere, but I think a refactoring ticket is the best place for that.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
shodaaan: This is consistent with how we are using static method calls. The consistent approach would be IMHO to actually do this everywhere, but I think a refactoring ticket is the best place for that.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
shodaaan: Fixed, thank you for the comment.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGCTest.java
shodaaan: Fixed the comment, unit tests and static change, thank you for the suggestions

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
stefan-egli: but it's in a random unrelated code - which is against our goal to not do unrelated changes

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -47,13 +47,13 @@
...
 import org.apache.jackrabbit.oak.plugins.document.SweepHelper;
+import org.apache.jackrabbit.oak.plugins.document.VersionGCOptions;
 import org.apache.jackrabbit.oak.plugins.document.VersionGCSupport;
+import org.apache.jackrabbit.oak.plugins.document.VersionGarbageCollector;
 import org.apache.jackrabbit.oak.plugins.document.VersionGarbageCollector.VersionGCInfo;
 import org.apache.jackrabbit.oak.plugins.document.VersionGarbageCollector.VersionGCStats;
 import org.apache.jackrabbit.oak.plugins.document.util.MongoConnection;
 import org.apache.jackrabbit.oak.run.commons.Command;
-import org.apache.jackrabbit.oak.plugins.document.VersionGCOptions;
-import org.apache.jackrabbit.oak.plugins.document.VersionGarbageCollector;
```
rishabhdaim: please revert these import changes.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -69,10 +69,11 @@
 import static org.apache.jackrabbit.oak.plugins.document.Collection.NODES;
 import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreHelper.createVersionGC;
 import static org.apache.jackrabbit.oak.plugins.document.FormatVersion.versionOf;
+import static org.apache.jackrabbit.oak.plugins.document.util.Utils.getFullGCMode;
 import static org.apache.jackrabbit.oak.plugins.document.util.Utils.getIdFromPath;
 import static org.apache.jackrabbit.oak.plugins.document.util.Utils.getRootDocument;
-import static org.apache.jackrabbit.oak.plugins.document.util.Utils.isFullGCEnabled;
```
rishabhdaim: same as above.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -211,6 +211,29 @@ static FullGCMode getFullGcMode() {
         return fullGcMode;
     }
 
+    /**
+     * Set the full GC mode to be used according to the provided configuration value.
+     * The configuration value will be ignored and fullGCMode will be reset to NONE
+     * if it is set to any other values than the supported ones.
+     * @param fullGcModeConfig
+     */
+    static void setFullGcMode(int fullGcModeConfig) {
```
rishabhdaim: I would suggest renaming variable name to `fullGcMode`

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
```diff
@@ -1009,6 +1009,16 @@ public static boolean isEmbeddedVerificationEnabled(final DocumentNodeStoreBuild
         return builder.isEmbeddedVerificationEnabled() || (docStoreEmbeddedVerificationFeature != null && docStoreEmbeddedVerificationFeature.isEnabled());
     }
 
+    /**
+     * Returns the full GC mode value for the DocumentNodeStore.
+     *
+     * @param builder instance for DocumentNodeStoreBuilder
+     * @return true if full GC is enabled else false
+     */
+    public static int getFullGCMode(final DocumentNodeStoreBuilder<?> builder) {
```
rishabhdaim: This utility method doesn't make sense, it only returns `fullGcMode` from the `builder` object. I would remove this.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/util/UtilsTest.java
```diff
@@ -236,6 +237,30 @@ public void fullGCDisabledForRDB() {
         assertFalse("Full GC is disabled for RDB Document Store", fullGCEnabled);
     }
 
+    @Test
+    public void fullGCModeDefaultValue() {
+        int fullGCModeDefaultValue = getFullGCMode(newDocumentNodeStoreBuilder());
+        final int FULL_GC_MODE_NONE = 0;
```
rishabhdaim: please use camelCase here.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/util/UtilsTest.java
```diff
@@ -236,6 +237,30 @@ public void fullGCDisabledForRDB() {
...
+        assertEquals("Full GC mode has NONE value by default", fullGCModeDefaultValue, FULL_GC_MODE_NONE);
+    }
+
+    @Test
+    public void fullGCModeSetViaConfiguration() {
+        DocumentNodeStoreBuilder<?> builder = newDocumentNodeStoreBuilder();
+        final int FULL_GC_MODE_GAP_ORPHANS = 2;
```
rishabhdaim: same as above.

rishabhdaim (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
shodaaan: Fixed.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
shodaaan: Fixed.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
shodaaan: done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/Utils.java
shodaaan: Fixed.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/util/UtilsTest.java
shodaaan: Done.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/util/UtilsTest.java
shodaaan: Done.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
```diff
@@ -324,4 +325,14 @@
...
+            name = "Document Node Store Full GC Mode",
+            description = "Integer value indicating which Full GC mode should be enabled for " +
+                    "document node store. The Default value is " + DEFAULT_FULL_GC_MODE +
+                    ". Note that this value can be overridden via framework " +
+                    "property 'oak.documentstore.fullGCMode'." +
+                    "FullGC can be entirely enabled / disabled with the variable fullGCEnabled, unless fullGCEnabled" +
+                    "is set to true, the fullGCMode will be ignored.")
```
stefan-egli:
```suggestion
                    "property 'oak.documentstore.fullGCMode'. " +
                    "FullGC can be entirely enabled / disabled with the variable fullGCEnabled, unless fullGCEnabled " +
                    "is set to true, the fullGCMode will be ignored.")
```

stefan-egli (commented):
an osgi config formatting detail added - plus the conflicts with trunk would need to be resolved - otherwise looking good

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentNodeStoreBuilderTest.java
```diff
@@ -111,4 +110,11 @@ public void collectionCompressionDisabled() {
         MongoDocumentNodeStoreBuilder builder = new MongoDocumentNodeStoreBuilder();
         assertNull(builder.getCollectionCompressionType());
     }
+
+    @Test
+    public void fullGCModeDefaultValue() {
+        MongoDocumentNodeStoreBuilder builder = new MongoDocumentNodeStoreBuilder();
+        final int FULL_GC_MODE_NONE = 0;
```
rishabhdaim: please use camelCase here.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
shodaaan: Done.

stefan-egli (approved):
+1 (there's some unrelated newlines, but let's leave them for now)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilder.java
```diff
@@ -154,6 +157,18 @@ public RDBDocumentNodeStoreBuilder setFullGCExcludePaths(@NotNull String[] exclu
...
+        log.warn("FullGC modes are not supported for RDB");
```
rishabhdaim, line 164: this should return the `builder` object.

shodaaan: Fixed.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilder.java
```diff
@@ -143,7 +147,6 @@ public RDBDocumentNodeStoreBuilder setFullGCIncludePaths(@NotNull String[] inclu
         return thisBuilder();
     }
 
-    @Override
```
rishabhdaim: please revert this.

rishabhdaim (approved):

rishabhdaim (approved):

stefan-egli:
+1, I'll do a squash-n-merge next. was wondering why `DocumentNodeStoreConfigTest` fails - but that's unrelated, fails on trunk too, so this is good to go ahead.

PR #1541, opened 2024-06-18
OAK-10901 - Bypass the DocumentNodeState cache when resolving paths downloaded from Mongo.

nfsantos:
The indexing job needs to resolve each of the paths found in documents downloaded from Mongo into DocumentNodeState instances. It does so using the DocumentNodeStore, which internally keeps a cache with previously resolved documents DocumentNodeState instances. But in the case of the indexing job, none of the paths that will be resolved will be found in the cache, because as part of the traversal of the repository, it resolves each path once and only once. Therefore, not only it is wasteful to check the cache for the presence of the path/revision pair, but it is also wasteful to put the resolved node in the cache. 

This PR bypasses this cache, by calling the internal method `DocumentNodeStore#readNode(Path, RevisionVector)` directly.  

For more details see here: https://github.com/apache/jackrabbit-oak/pull/1533

### DIFF
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedTransformTask.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedTransformTask.java
@@ -24,6 +24,7 @@ import org.apache.jackrabbit.oak.commons.IOUtils;
 import org.apache.jackrabbit.oak.index.indexer.document.flatfile.NodeStateEntryWriter;
 import org.apache.jackrabbit.oak.plugins.document.DocumentNodeState;
 import org.apache.jackrabbit.oak.plugins.document.DocumentNodeStore;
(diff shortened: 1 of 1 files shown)


fabriziofortino (commented):
Is this also true when the indexing job needs to reindex more than one index? Let's say we reindex 2 versions of the same index that need the same nodes, would we hit the cache in this case?

nfsantos:
> Is this also true when the indexing job needs to reindex more than one index? Let's say we reindex 2 versions of the same index that need the same nodes, would we hit the cache in this case?

This change affects only the dump phase of the indexing job. At his phase, we are only converting the Mongo representation of a node (contains metadata of revisions, and maybe children of the node if node bundling is used) into a DocumentNodeState which is the Oak node at a give revision. This is what is written to the FFS. There is no indexing being done at this phase, so each path is looked up only once.

The reindex is done reading from the FFS, it's a separate code path, with an implementation of the NodeStore that reads from the FFS.

fabriziofortino (approved):

thomasmueller (approved):

PR #1543, opened 2024-06-19
OAK-10748: Improve statistics to collect which type of garbage is sent/deleted

Joscorbe:
It effectively adds 4 metrics, all of them are candidates for deletion. They may later be skipped at deletion, but that's intended, so we have the difference between collected and deleted:
- Properties
- Revisions + internal revisions
- Documents

The statistics are stored per phase, so when we get the stats we will know which phase collected each amount of properties/revisions/documents. Yes, there are some objects that are only possible to be collected in one phase (like documents, only detected during orphan cleanup).

Apart from this, I introduced a few improvements. For example, the output was flooded of messages saying "collected 0 revisions". Those messages will be logged only if more than 0 objects were collected. This should reduce the logging by a big factor.

It also fixes the initialization of a statistics provider for FullGC statistics, since it was a NOOP at least when executed via oak-run revisions command.


Joscorbe:
Will integrate this into tests in a future iteration (as stated in the ticket). This statistics are just informative and we could introduce potential flaky tests.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -535,8 +544,8 @@ private void collectDocument(RevisionsOptions options, Closer closer, String pat
         }
         gc.collectGarbageOnDocument(documentNodeStore, workingDocument, options.isVerbose());
 
-        //TODO: Probably we should output some details of fullGCStats. Could be done after OAK-10378
-        //gc.getFullGCStats();
+        System.out.println("Full GC Stats:");
+        System.out.println(gc.getFullGCStatsReport());
```
stefan-egli: just curious : above the report is indented with 4 spaces but not here - is this on purpose?

stefan-egli (approved):
+1, just a question added

stefan-egli:
actually, there are compilation errors:
```
[INFO] -------------------------------------------------------------
Error:  COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
Error:  /home/runner/work/jackrabbit-oak/jackrabbit-oak/oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java:[367,11] setStatisticsProvider(org.apache.jackrabbit.oak.stats.StatisticsProvider) is not public in org.apache.jackrabbit.oak.plugins.document.VersionGarbageCollector; cannot be accessed from outside package
Error:  /home/runner/work/jackrabbit-oak/jackrabbit-oak/oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java:[538,11] setStatisticsProvider(org.apache.jackrabbit.oak.stats.StatisticsProvider) is not public in org.apache.jackrabbit.oak.plugins.document.VersionGarbageCollector; cannot be accessed from outside package
[INFO] 2 errors 
```

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
Joscorbe: Not on purpose, I will add them here too.

Joscorbe:
> actually, there are compilation errors:
> 
> ```
> [INFO] -------------------------------------------------------------
> Error:  COMPILATION ERROR : 
> [INFO] -------------------------------------------------------------
> Error:  /home/runner/work/jackrabbit-oak/jackrabbit-oak/oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java:[367,11] setStatisticsProvider(org.apache.jackrabbit.oak.stats.StatisticsProvider) is not public in org.apache.jackrabbit.oak.plugins.document.VersionGarbageCollector; cannot be accessed from outside package
> Error:  /home/runner/work/jackrabbit-oak/jackrabbit-oak/oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java:[538,11] setStatisticsProvider(org.apache.jackrabbit.oak.stats.StatisticsProvider) is not public in org.apache.jackrabbit.oak.plugins.document.VersionGarbageCollector; cannot be accessed from outside package
> [INFO] 2 errors 
> ```

I will double-check this, sounds weird, I built it and ran all the integration tests locally...

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -240,7 +240,21 @@ static FullGCMode getFullGcMode() {
...
+     * the mongo query to include a '_path' condition - which disallows
+     * mongo from using the '_modified_id' index. IOW long paths
+     * would result in full scans - which results in bad performance.
+     */
+    void setFullGCPaths(@NotNull Set<String> includes, @NotNull Set<String> excludes) {
+        this.fullGCIncludePaths = requireNonNull(includes);
+        this.fullGCExcludePaths = requireNonNull(excludes);
+    }
+
+    void setStatisticsProvider(StatisticsProvider provider) {
```
Joscorbe: Oh, the public was lost here.

Joscorbe:
I will squash this PR into a single commit once approved.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/FullGCStatsCollector.java
```diff
@@ -31,6 +32,34 @@ public interface FullGCStatsCollector {
      */
     void documentRead();
 
+    /**
+     * Total No. of properties deleted during a given GC phase
+     * @param mode GC phase
+     * @param numProps no. of properties deleted in current cycle
+     */
+    void collectedPropertiesDeleted(GCPhase mode, long numProps);
```
rishabhdaim: this method is a duplicate of `void propertiesDeleted(long numProps);`. I don't get the rationale behind this.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/FullGCStatsCollector.java
```diff
@@ -31,6 +32,34 @@ public interface FullGCStatsCollector {
...
+
+    /**
+     * Total No. of documents updated during a given GC phase
+     * @param mode GC phase
+     * @param numCommits no. of unmerged branch commits in current cycle
+     */
+    void collectedUnmergedBranchCommits(GCPhase mode, long numCommits);
```
rishabhdaim: same as above. Duplicate of `void unmergedBranchCommitsDeleted(long numCommits);`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1316,6 +1316,7 @@ private void collectDeletedProperties(final NodeDocument doc, final GCPhases pha
                         .sum();
 
                 deletedPropsCountMap.put(doc.getId(), deletedPropsCount);
+                fullGCStats.collectedPropertiesDeleted(GCPhase.FULL_GC_COLLECT_PROPS, deletedPropsCount);
```
rishabhdaim: this stats is wrong because we might end up in a situation where the property collected for deletion during the `COLLECT` phase is not at all deleted.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1355,6 +1356,7 @@ private void collectUnmergedBranchCommits(final NodeDocument doc, final GCPhases
             
             olderUnmergedBranchCommits.forEach(bcRevision -> removeUnmergedBCRevision(bcRevision, doc, updateOp));
             deletedUnmergedBCSet.addAll(olderUnmergedBranchCommits);
+            fullGCStats.collectedUnmergedBranchCommits(GCPhase.FULL_GC_COLLECT_UNMERGED_BC, olderUnmergedBranchCommits.size());
```
rishabhdaim: same as above, collecting deletion stats here is wrong.

rishabhdaim:
I understand that this PR is trying to get the stats of fullGC in terms of GCPhase, but the place is wrong in this PR.

We are collecting deletion stats while collecting potential garbage but while deletion, we might skip those documents/properties; thus, the actual deletion no. could be different.

cc @stefan-egli

rishabhdaim (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/FullGCStatsCollector.java
rishabhdaim: I would rename this to `propertiesCollected` or `collectedProperties`

Joscorbe:
Right now we are only getting the aggregated number of garbage that gets deleted, but we don't know which of the phases found it.
This PR focuses on the garbage that gets found by each of the phases, that's why the statistics are collected during the collection phase, not at deletion.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/FullGCStatsCollector.java
Joscorbe: Yes, makes more sense

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
Joscorbe: They may seem duplicated but are not, the new methods count the amounts of garbage that would be deleted by each of the phases, and the idea is to eventually use this as a reference to know which phases of FullGC would bring more benefit.

Joscorbe: Probably all the methods could be renamed to state this better.

rishabhdaim (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/FullGCStatsCollector.java
```diff
@@ -31,6 +32,34 @@ public interface FullGCStatsCollector {
...
+     * @param numProps no. of garbage properties found in current cycle
+     */
+    void candidateProperties(GCPhase mode, long numProps);
+
+    /**
+     * Total No. of documents detected as garbage during a given GC phase
+     * @param mode GC phase
+     * @param numCommits no. of garbage documents found in current cycle
+     */
+    void candidateDocuments(GCPhase mode, long numCommits);
```
stefan-egli: What is the use case of this counter? I see it is currently only used in `collectUnmergedBranchCommits` hence seems a bit asymmetric. There already is a counter for how many documents are read (`documentRead`), is there an advantage of having `candidateDocuments` too?

Joscorbe: This is wrong, yes. `collectUnmergedBranchCommits` deletes revisions, so I will change this to the one that counts deleted revisions. Probably this one is unneeded then, as there is already one that counts deletedDocuments (orphan).

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
Joscorbe: I have changed the names of the statistics to clarify it is counting the candidates to delete. Those numbers can then mismatch the actually deleted statistics, since they could be skipped in later stages.

stefan-egli (approved):
+1, I don't have full overview of how many metrics this will introduce (i.e. whether that number might be excessive), but I think that's something that we'll see and can handle downstream, in case.

Joscorbe:
> +1, I don't have full overview of how many metrics this will introduce (i.e. whether that number might be excessive), but I think that's something that we'll see and can handle downstream, in case.

I have added a proper description to the PR. Thanks for double checking, will merge this end of the day.

PR #1545, opened 2024-06-19
OAK-10904: Close token refresh executor service after access token is no longer needed

t-rana:


oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
```diff
@@ -74,7 +74,12 @@ public final class AzureUtilities {
     private static final long TOKEN_REFRESHER_DELAY = 1L;
 
     private static final Logger log = LoggerFactory.getLogger(AzureUtilities.class);
-    private static final ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor();
+
+    private static final ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor(runnable -> {
```
amit-jain: Making daemon is fine but is there no way we can call close on this?

amit-jain: Not closing it properly could cause random issues in tests for example

t-rana: > Making daemon is fine but is there no way we can call close on this?

This [function](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java#L148) is part of the Util class and is being called from several places including other Util classes, closing this is not as straightforward, hence I opted for using daemon threads and adding a shutdown hook to close the executor.

t-rana: > Not closing it properly could cause random issues in tests for example

I did not face any issues in the test suite run, can you please elaborate so that I can handle the same? I guess the framework running test terminates the JVM and shutdown method is executed which will close this exectuor.

nfsantos: This implementation is problematic not only for tests but also for production use. If I understood correctly, every call to `storageCredentialAccessTokenFrom(...)` will start a new refresh tasks. There is no mechanism to close this task when the token is no longer needed, so this can lead to the accumulation of many tasks that will keep on executing until the end of the lifetime of the JVM. This is problematic for several reasons:

- Memory leak of tasks
- Many unnecessary calls to Azure. This can cause issues with API request quotas in Azure.
- The execution of obsolete tasks may interfere with the ones that are still required, blocking them from being executed, which can lead to tokens not being refreshed on time.

A better design would be for the refresh service to be explicitly managed by the caller, requiring the caller to create and stop it when the token is no longer needed.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/tool/ToolUtils.java
```diff
@@ -202,63 +186,33 @@ public static CloudBlobDirectory createCloudBlobDirectory(String path, Environme
...
+                    Iterator<String> revisionIterator = Iterators.transform(journalReader, new Function<JournalEntry, String>() {
+                        @NotNull
+                        @Override
+                        public String apply(JournalEntry entry) {
+                            return entry.getRevision();
+                        }
+                    });
+                    return newArrayList(revisionIterator);
+                } catch (Exception e) {
+                    e.printStackTrace();
```
rishabhdaim, line 203: please add an error log here.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureUtilities.java
t-rana: Changed the current logic, moved the code to generate token to a concrete class, where each instance of a class will have a single access token and the same will be refresh periodically. No unnecessary call will be made to azure for token generation each time a blob is requested.

The caller of this class will be responsible for closing the token refresh executor service when the token is no longer needed.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/tool/ToolUtils.java
t-rana: done

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureStorageCredentialManager.java
```diff
@@ -0,0 +1,146 @@
...
+        final String clientId = environment.getVariable(AZURE_CLIENT_ID);
+        final String clientSecret = environment.getVariable(AZURE_CLIENT_SECRET);
+        final String tenantId = environment.getVariable(AZURE_TENANT_ID);
+
+        if (StringUtils.isNoneBlank(clientId, clientSecret, tenantId)) {
+            try {
+                return getStorageCredentialAccessTokenFromServicePrincipals(accountName, clientId, clientSecret, tenantId);
+            } catch (IllegalArgumentException | StringIndexOutOfBoundsException e) {
+                throw new IllegalArgumentException(
+                        "Could not connect to the Azure Storage. Please verify if AZURE_CLIENT_ID, AZURE_CLIENT_SECRET and AZURE_TENANT_ID environment variables are correctly set!");
```
nfsantos, line 68: You should log the original exception or else propagate it, so that the cause of the problem is displayed on the logs.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureStorageCredentialManager.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        log.warn("AZURE_CLIENT_ID, AZURE_CLIENT_SECRET and AZURE_TENANT_ID environment variables empty or missing. Switching to authentication with AZURE_SECRET_KEY.");
+
+        String key = environment.getVariable(AZURE_SECRET_KEY);
+        try {
+            return new StorageCredentialsAccountAndKey(accountName, key);
+        } catch (IllegalArgumentException | StringIndexOutOfBoundsException e) {
+            throw new IllegalArgumentException(
+                    "Could not connect to the Azure Storage. Please verify if AZURE_SECRET_KEY environment variable is correctly set!");
```
nfsantos, line 80: Same thing here.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureStorageCredentialManager.java
```diff
@@ -0,0 +1,146 @@
...
+        if (accessToken == null) {
+            clientSecretCredential = new ClientSecretCredentialBuilder()
+                    .clientId(clientId)
+                    .clientSecret(clientSecret)
+                    .tenantId(tenantId)
+                    .build();
+            accessToken = clientSecretCredential.getTokenSync(new TokenRequestContext().addScopes(AZURE_DEFAULT_SCOPE));
+            if (accessToken == null || StringUtils.isBlank(accessToken.getToken())) {
+                log.error("Access token is null or empty");
+                throw new IllegalArgumentException("Could not connect to azure storage, access token is null or empty");
```
nfsantos: I suggest removing the log.error from here, because the exception contains the same information.

nfsantos (approved):
The new changes seem to fix the issues that I had identified with thread leaks, the new logic seems good, it leaves up to the caller the responsibility of creating and shutting down the token refresh service.
I added just a few minor comments about logging.
I only looked carefully at the threading logic, I did not go into detail about the other aspects of the PR.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureStorageCredentialManager.java
t-rana: Done

dulceanu (approved):

PR #1546, opened 2024-06-19
OAK-10903 - When parallel download is enabled, the dump may miss documents updated during the download

nfsantos:
When using parallel download, after the ascending and descending tasks terminate, do another query to download all documents that were modified since the download started. Documents modified during the download might be missed because their new modified value will be higher than what the descending download thread saw at the start of the download, so if they are modified before the download thread downloaded them, they will not be included in the initial traversals. 
The final catchup query downloads in ascending order all the documents with `_modified >= H`, where H is the highest modified value seen by the descending traversal thread, that is, the first element that it downloaded. This will include some documents twice, both those with `_modified=H`, but also those that were downloaded by the initial traversals and then were modified before the download completed. This is not a problem for correctness, as the merge-sort phase de-duplicates entries, but will increase the number of documents and entries processed.

Add new configuration properties:
- `oak.indexer.pipelined.mongoParallelDump.secondariesOnly` - enables/disables the mongo replica selection policy that restricts parallel downloads to secondaries only, one connection per secondary. Disabling this is useful for testing parallel download without having to setup a replicated cluster.
- `oak.indexer.pipelined.mongoBatchSize` - *internal property, for testing only*. Sets the limit for the number of elements in a batch returned by Mongo during iteration over query results. This is useful for the connection interruption tests, to ensure that we get a first response from Mongo before the connection is broken.

### DIFF
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
@@ -218,6 +218,10 @@ public class PipelinedMongoDownloadTask implements Callable<PipelinedMongoDownlo
     private Instant lastDelayedEnqueueWarningMessageLoggedTimestamp = Instant.now();
     private MongoParallelDownloadCoordinator mongoParallelDownloadCoordinator;
 
+    // For testing purposes
+    static final String OAK_INDEXER_PIPELINED_MONGO_BATCH_SIZE = "oak.indexer.pipelined.mongoBatchSize";
+    private final int mongoBatchSize = Integer.getInteger(OAK_INDEXER_PIPELINED_MONGO_BATCH_SIZE, -1);
+
     public PipelinedMongoDownloadTask(MongoClientURI mongoClientURI,
                                       MongoDocumentStore docStore,
                                       int maxBatchSizeBytes,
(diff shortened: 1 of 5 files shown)


nit0906:
I have one confusion here - I am not very familiar with the Pipeline code, but in theory the usage of a checkpointed state should prevent any changes in the repo post the checkpoint to affect the final outcome of the FFS.

thomasmueller (approved):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -533,17 +581,33 @@ public boolean downloadInAscendingOrder() {
...
     private class DownloadTask {
+        private final Stopwatch downloadTaskStart = Stopwatch.createUnstarted();
         private final DownloadOrder downloadOrder;
         private final DownloadStageStatistics downloadStatics;
+        private final long fromModified;
+        private final long toModified;
+        private final MongoParallelDownloadCoordinator parallelDownloadCoordinator;
         private long documentsDownloadedTotalBytes;
         private long documentsDownloadedTotal;
         private long totalEnqueueWaitTimeMillis;
```
fabriziofortino: this is initialized to 0 and updated at line 830 but never used. Can we remove it?

fabriziofortino (approved):

nfsantos:
> I have one confusion here - I am not very familiar with the Pipeline code, but in theory the usage of a checkpointed state should prevent any changes in the repo post the checkpoint to affect the final outcome of the FFS.

I have described a scenario where this may happen in the Jira ticket of this issue:
https://issues.apache.org/jira/browse/OAK-10903
Please have a look and let me know if the explanation is not clear. You can also verify that this indeed occurs by commenting out the code that does the catch-up after the two parallel streams finish and running the test: `PipelinedMongoConnectionFailureIT#mongoDisconnectTest()`

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
nfsantos: Done.

PR #1547, opened 2024-06-19
OAK-10742 : add support for defining include/exclude paths for fullGC

stefan-egli:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -216,6 +216,8 @@ static FullGCMode getFullGcMode() {
     private final boolean fullGCEnabled;
     private final boolean isFullGCDryRun;
     private final boolean embeddedVerification;
+    private Set<String> fullGCIncludePaths;
```
reschke: Maybe make these "@NonNull"?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
```diff
@@ -132,6 +133,49 @@ public CloseableIterable<NodeDocument> getPossiblyDeletedDocs(final long fromMod
...
+            final List<Bson> ors = new ArrayList<>(includes.size());
+            for (String incl : includes) {
+                ors.add(Filters.regex(ID, ":" + incl));
+            }
+            inclExcl = or(ors);
+        }
+        if (excludes != null && !excludes.isEmpty()) {
+            final List<Bson> ands = new ArrayList<>(excludes.size());
+            for (String excl : excludes) {
+                ands.add(not(Filters.regex(ID, ":" + excl)));
```
reschke, line 167: Is this the part where we match oak paths to the identifiers used in a DocumentStore ("n:...")? In that case a comment would be good.

Also, do we need to consider the special identifiers for long paths? If we can't, excluding these documents in a layer above might work; but sure what that would mean for "includePaths" though.

reschke (commented):
It might be do

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
```diff
@@ -132,6 +133,49 @@ public CloseableIterable<NodeDocument> getPossiblyDeletedDocs(final long fromMod
...
+     * @param query the query with which to do an AND
+     * @return the combined bson with include/exclude path prefixes
+     * AND the provided query
+     */
+    private Bson withIncludeExcludes(Set<String> includes, Set<String> excludes, Bson query) {
+        Bson inclExcl = null;
+        if (includes != null && !includes.isEmpty()) {
+            final List<Bson> ors = new ArrayList<>(includes.size());
+            for (String incl : includes) {
+                ors.add(Filters.regex(ID, ":" + incl));
```
rishabhdaim, line 160: Don't we need to add `level` before `:` ?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCSupport.java
```diff
@@ -98,12 +98,15 @@ public Iterable<NodeDocument> getPossiblyDeletedDocs(final long fromModified, fi
      * @return matching documents.
      */
     public Iterable<NodeDocument> getModifiedDocs(final long fromModified, final long toModified, final int limit,
-                                                  @NotNull final String fromId) {
+                                                  @NotNull final String fromId, final Set<String> includePaths,
```
rishabhdaim: I would suggest to make `includePaths` & `excludePaths` as `@NotNull`. It would avoid us in all the null checks in the later code.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
stefan-egli: Yes, it is for the id syntax. Good point about long paths - the includes/excludes wouldn't have taken those into account as per current state of the PR. I'll look into that. (Worst case could be a comment saying that include/excludes aren't supported for long paths, but that would be unfortunate..)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
stefan-egli: the idea is that any string before `:` matches. So any level would be fine. But also any split document in theory - except they don't get included in the outer query as they have no `_modified`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
stefan-egli: done in https://github.com/apache/jackrabbit-oak/pull/1547/commits/14ba933fad9b8dadcd51d00aa2e186a0b3e39568

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
stefan-egli: added a comment about what that regex is doing
plus supported long-paths
done in https://github.com/apache/jackrabbit-oak/pull/1547/commits/8f176f6b20c5066f377cd3433cfc0e1e51cb1d2d

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCSupport.java
stefan-egli: done in https://github.com/apache/jackrabbit-oak/pull/1547/commits/14ba933fad9b8dadcd51d00aa2e186a0b3e39568

stefan-egli:
@reschke , @rishabhdaim , addressed your findings, please feel free to re-review, thx!

rishabhdaim (approved):

stefan-egli:
One thing I wonder now is : with including `_path` for long path, the query will no longer be served from the `_modified_id` index. IOW the resulting performance will be dramatically lower.

Wondering if we should indeed support long paths with include/excludes ... or make that .. configurable ..

stefan-egli:
I'd suggest to not support long paths with regards to include/excludes in mongo for now. They will kill performance. Yes, we could have a config that would control if one wants long paths or not - but I think that's going a bit too far. The include/exclude path idea was meant to be a potentially short-/mid-term option to fine-tune the rollout of Full-GC. Whether or not full paths are 100% correctly filtered as per include/exclude I think isn't that important to justify the performance impact.

I've tentatively reverted the long-path support via https://github.com/apache/jackrabbit-oak/pull/1547/commits/6cf86edbc74c2f22e0a8e0bf27af41360517a873

stefan-egli:
Please note that I've re-added support for long-paths for excludes, client-side. I think that part could be quite useful, to specifically not do full-GC on certain paths, even if they are long ones.

Joscorbe:
The inclusion/exclusion paths I can think about are usually just a few levels below root, like `jcr:system`, `oak:index`, etc. Don't see the need (at least for now) to need support for long paths by default.

Joscorbe (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
```diff
@@ -132,6 +133,56 @@ public CloseableIterable<NodeDocument> getPossiblyDeletedDocs(final long fromMod
...
+            }
+            inclExcl = or(ors);
+        }
+        if (excludes != null && !excludes.isEmpty()) {
+            final List<Bson> ands = new ArrayList<>(excludes.size());
+            for (String excl : excludes) {
+                ands.add(not(Filters.regex(ID, ":" + excl)));
+            }
```
nfsantos, line 168: Mongo will not use the index `_modified_id` if the query includes a negative condition. The reason is that `not(Filters.regex(ID, ":" + excl))` matches documents with `_id=null` and the indexes do not include null values. I had the same problem when implement include/exclude paths in the indexing job, the performance when using only include paths was much better than when there were exclude paths. The fix I came up with is very ugly, but it works and provides a massive speed-up when downloading with excluded paths:

https://github.com/apache/jackrabbit-oak/blob/dbc728fc638fe1f6158165a23f24b418e200790e/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/MongoDownloaderRegexUtils.java#L132-L141

https://issues.apache.org/jira/browse/OAK-10682

stefan-egli: Great, thanks for the heads-up @nfsantos ! I'll need to see how to fix this (worst case would I guess be to disallow exclude paths until that's improved)

stefan-egli: @nfsantos , have you tried/considered using an index hint, eg `"_modified_1__id_1"` in the query? It looks like it might accept that (from a test against a local/empty db)

PR #1548, opened 2024-06-19
OAK-10845 : disabling more flaky fullgc test combinations

stefan-egli:


stefan-egli:
created https://issues.apache.org/jira/browse/OAK-10906 to track the sonar issue

PR #1551, opened 2024-06-21
OAK-10912 : make fullGc include/exclude paths configurable via OSGI config

rishabhdaim:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilder.java
```diff
@@ -128,6 +132,28 @@ public RDBDocumentNodeStoreBuilder setFullGCEnabled(boolean b) {
...
     }
 
+    @Override
+    public Set<String> getFullGCIncludePaths() {
+        return of();
+    }
+
+    @Override
+    public RDBDocumentNodeStoreBuilder setFullGCIncludePaths(@NotNull String[] includePaths) {
+        // fullGC is non supported for RDB
```
shodaaan: minor typo: fullGC is not supported

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilder.java
```diff
@@ -128,6 +132,28 @@ public RDBDocumentNodeStoreBuilder setFullGCEnabled(boolean b) {
...
+    }
+
+    @Override
+    public Set<String> getFullGCExcludePaths() {
+        return of();
+    }
+
+    @Override
+    public RDBDocumentNodeStoreBuilder setFullGCExcludePaths(@NotNull String[] excludePaths) {
+        // fullGC is non supported for RDB
```
shodaaan: minor typo: fullGC is not supported

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
```diff
@@ -499,6 +499,8 @@ private void configureBuilder(DocumentNodeStoreBuilder<?> builder) {
                 setDocStoreEmbeddedVerificationFeature(docStoreEmbeddedVerificationFeature).
                 setThrottlingEnabled(config.throttlingEnabled()).
                 setFullGCEnabled(config.fullGCEnabled()).
+                setFullGCIncludePaths(config.fullGCIncludes()).
```
shodaaan: Please consider adding a check and WARN message if a path is part of both include and exclude lists. 
I am thinking exclude would take priority in this case and the path would be excluded, but it is something we can discuss if you think it is relevant.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceConfigurationTest.java
```diff
@@ -115,6 +118,22 @@ public void fullGCEnabled() throws Exception {
         assertEquals(fullGCDocStore, config.fullGCEnabled());
     }
 
+    @Test
```
shodaaan, line 121: Please consider also testing the logic when a path element appears in both the includes and excludes list.
Depending on what we decide the logic should be in this case (I propose that exclude takes priority), the test should reflect that.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentNodeStoreBuilderTest.java
```diff
@@ -60,6 +62,22 @@ public void fullGCFeatureToggleDisabled() {
         assertNull(builder.getDocStoreFullGCFeature());
     }
 
+    @Test
```
shodaaan, line 65: Please consider unit testing for invalid paths - maybe use a RegEx or a Utils method that validates the path string.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilderTest.java
```diff
@@ -67,6 +68,20 @@ public void fullGCDisabled() {
         assertFalse(builder.isFullGCEnabled());
     }
 
+    @Test
```
shodaaan, line 71: Same here, please consider unit testing for invalid paths.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
rishabhdaim: That's a good catch. I have tested with a path in both the `include` & `exclude` list and it turns out that in that case, we won't be fetching any document from db.
cc @stefan-egli

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceConfigurationTest.java
rishabhdaim: I believe both lists should be different, with even one common path, nothing gets fetched from DB.
cc @stefan-egli

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
rishabhdaim: I have tested on local and in case a path is present in both `include` & `exclude`, it won't be selected.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceConfigurationTest.java
rishabhdaim: > I believe both lists should be different, with even one common path, nothing gets fetched from DB. cc @stefan-egli

my above is wrong, only common path is excluded not all.

rishabhdaim: Check for invalid paths is added here: https://github.com/apache/jackrabbit-oak/pull/1551/commits/772c8559d4fdfa7e9fd8d251272f18606325ca04

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/rdb/RDBDocumentNodeStoreBuilderTest.java
rishabhdaim: added in https://github.com/apache/jackrabbit-oak/pull/1551/commits/167b46c5eaa518d39bad952853ac55bed2e0b7a5

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentNodeStoreBuilderTest.java
rishabhdaim: added in https://github.com/apache/jackrabbit-oak/pull/1551/commits/167b46c5eaa518d39bad952853ac55bed2e0b7a5

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
```diff
@@ -260,6 +260,27 @@
                     "are separated with '::'. Example: -Doak.documentstore.persistentCacheIncludes=/content::/var")
     String[] persistentCacheIncludes() default {"/"};
 
+    @AttributeDefinition(
+            name = "Full GC Include Paths",
+            description = "Paths which should be included in full garbage collection." +
```
stefan-egli: formatting is broken as some spaces are missing after dot

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
```diff
@@ -260,6 +260,27 @@
...
+                    "Empty value means all paths are included. " +
```
stefan-egli: I find this potentially confusing.

Currently when no include path is specified, it results in "include everything" - so specifying an "include everything" explicitly wouldn't be necessary. But if we wanted to support this, then I think it shouldn't be an empty string, but rather eg `"/"` as that is more explicit (so if someone would specify an empty string, we could rather log a warn and skip that entry).

wdyt?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
```diff
@@ -260,6 +260,27 @@
...
+                    "Any path which is added to both include & exclude paths, " +
+                    "would be removed from included paths." +
```
stefan-egli: Not sure we need to point out someone configuring duplicate paths specifically.

What about making this comment more generic into something like "Include and exclude paths can overlap. Exclude paths will take precedence" ?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
```diff
@@ -260,6 +260,27 @@
...
+                    "Note that this value can be overridden with a system property " +
+                    "'oak.documentstore.fullGCIncludes' where paths " +
+                    "are separated with '::'. Example: -Doak.documentstore.fullGCIncludes=/content::/var")
```
stefan-egli: I see we sometimes state it can be overwritten via `-Doak.mongo` and sometimes `-Doak.documentstore` ... does it have this magic indeed or is one or the other wrong?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
```diff
@@ -260,6 +260,27 @@
...
+
+    @AttributeDefinition(
+            name = "Full GC Exclude Paths",
+            description = "Paths which should be excluded from full Garbage collection." +
+                    "Empty value means no paths are excluded." +
+                    "Any path added to excluded list would be removed from include paths (if present)." +
+                    "Note that this value can be overridden with a system property " +
+                    "'oak.documentstore.fullGCExcludes' where paths " +
+                    "are separated with '::'. Example: -Doak.documentstore.fullGCExcludes=/content::/var")
+    String[] fullGCExcludes() default {};
```
stefan-egli: was wondering if the config should be called `fullGCExcludePaths` - a bit more explicit, wdyt?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -170,6 +174,8 @@ public class DocumentNodeStoreBuilder<T extends DocumentNodeStoreBuilder<T>> {
     private boolean clusterInvisible;
     private boolean throttlingEnabled;
     private boolean fullGCEnabled;
+    private Set<String> fullGCIncludePaths = of();
+    private Set<String> fullGCExcludePaths = of();
```
stefan-egli: btw, just a side-comment, I find this less readable than say `Set.of()` (but I think this might be controversial as IDEs visualize it magically - but if you read the raw file it is less readable for me)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
stefan-egli: Configuring the identical path in both include and exclude seems weird - but the code does behaves correctly, namely exclude has precedence, so it's "as if that path wasn't there". But I don't think we need to add an extra warn for this ..

stefan-egli (commented):
lgtm in general, mainly added some comments about config names used, and config descriptions

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
rishabhdaim: This magic happens here: https://github.com/apache/jackrabbit-oak/blob/trunk/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceConfiguration.java#L123

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/1551/commits/55f310e2d65cfe3208ff7688ce93787eccb0cfba

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/1551/commits/55f310e2d65cfe3208ff7688ce93787eccb0cfba

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/1551/commits/55f310e2d65cfe3208ff7688ce93787eccb0cfba

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/1551/commits/55f310e2d65cfe3208ff7688ce93787eccb0cfba

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/1551/commits/55f310e2d65cfe3208ff7688ce93787eccb0cfba

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/1551/commits/55f310e2d65cfe3208ff7688ce93787eccb0cfba

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceConfigurationTest.java
```diff
@@ -83,7 +84,9 @@ public void defaultValues() throws Exception {
         assertEquals("MONGO", config.documentStoreType());
         assertEquals(DocumentNodeStoreService.DEFAULT_BUNDLING_DISABLED, config.bundlingDisabled());
         assertEquals(DocumentMK.Builder.DEFAULT_UPDATE_LIMIT, config.updateLimit());
-        assertEquals(Arrays.asList("/"), Arrays.asList(config.persistentCacheIncludes()));
+        assertEquals(of("/"), Arrays.asList(config.persistentCacheIncludes()));
```
stefan-egli, line 87: nitpick : looks unrelated/unnecessary?

stefan-egli (approved):
+1 with just a nitpick about an unrelated change :

Joscorbe (approved):

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceConfigurationTest.java
rishabhdaim: Yes, it can be avoided, but IntelliJ insisted on making this change.

shodaaan (approved):

PR #1553, opened 2024-06-24
OAK-10914 : disable mongo-server-side exclude paths until query perfo…

stefan-egli:
…rmance is fixed


rishabhdaim (approved):

rishabhdaim:
I think we might need to update the test for this as well.

Joscorbe (approved):

stefan-egli:
> I think we might need to update the test for this as well.

Client-side incl/exclude always also applies (perhaps something to improve at some point), so if mongo doesn't do server-side exclude, the client-side still applies, so tests still run fine.

PR #1556, opened 2024-06-24
OAK-10915: Avoid deleting expired checkpoints when running FullGC on a read-only NodeStore

Joscorbe:
I'm introducing a similar method `getLastAliveRevision` in **VersionGCRecommendations**. Alternatively, I could modify the logic or add another method directly to the `Checkpoints` class, but since the scope of this change is very focused on oak-run FullGC I preferred to keep most of the logic directly in **VersionGCRecommendations**.

I'm open to change it if you think it should be done differently.

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Checkpoints.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Checkpoints.java
@@ -204,6 +204,10 @@ class Checkpoints {
         return rv;
     }
 
+    DocumentNodeStore getNodeStore() {
+        return nodeStore;
+    }
+
     void setInfoProperty(@NotNull String checkpoint, @NotNull String key, @Nullable String value) {
         Revision r = Revision.fromString(checkNotNull(checkpoint));
         Info info = getCheckpoints().get(r);
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Checkpoints.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Checkpoints.java
@@ -118,7 +118,7 @@ class Checkpoints {
      */
     @SuppressWarnings("unchecked")
     @Nullable
-    public Revision getOldestRevisionToKeep() {
+    public Revision getOldestRevisionToKeep(boolean performCleanup) {
         //Get uncached doc
         SortedMap<Revision, Info> checkpoints = getCheckpoints();
 
@@ -145,7 +145,7 @@ class Checkpoints {
             }
         }
 
-        if (op.hasChanges()) {
+        if (performCleanup && op.hasChanges()) {
             try {
                 store.findAndUpdate(Collection.SETTINGS, op);
                 LOG.debug("Purged {} expired checkpoints", op.getChanges().size());
@@ -156,6 +156,17 @@ class Checkpoints {
         }
 
         return lastAliveRevision;
+    }
+
+    /**
+     * Backwards compatible method to get the oldest revision to keep
+     * with cleanup.
+     *
+     * @return oldest valid checkpoint registered. Might return null if no valid
+     * checkpoint found
+     */
+    public Revision getOldestRevisionToKeep() {
+        return getOldestRevisionToKeep(true);
     }
 
     @SuppressWarnings("unchecked")
@@ -202,10 +213,6 @@ class Checkpoints {
             rv = expand(r);
         }
         return rv;
-    }
-
-    DocumentNodeStore getNodeStore() {
-        return nodeStore;
     }
 
     void setInfoProperty(@NotNull String checkpoint, @NotNull String key, @Nullable String value) {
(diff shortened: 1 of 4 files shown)


Joscorbe:
Note: The new method is a copy of [getOldestRevisionToKeep()](https://github.com/apache/jackrabbit-oak/blob/2b734cfba6e8300e9846097c459398c65d75cdb0/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Checkpoints.java#L121) without deleting expired checkpoints.

However, as it ignores expired checkpoints, the FullGC is able to proceed, as the method will return the oldest **valid** checkpoint.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
```diff
@@ -321,6 +328,32 @@ public void evaluate(VersionGCStats stats) {
         }
     }
 
+    @Nullable
+    private Revision getLastAliveRevision(Clock clock, Checkpoints checkpoints) {
```
stefan-egli: this duplicates quite a bit of code (the `Checkpoints.getOldestRevisionToKeep`).

what about alternatively introducing a boolean, eg `cleanupExpiredCheckpoints` to `getOldestRevisionToKeep` to have just 1 method? Then the current `getOldestRevisionToKeep()` could just call `getOldestRevisionToKeep(true)`, and the `VersionGCRecommendations` could call `getOldestRevisionToKeep(false)`?

stefan-egli: for example
```
        if (op.hasChanges()) {
```
in `getOldestRevisionToKeep` could be changed to
```
        if (cleanupExpiredCheckpoints && op.hasChanges()) {
```

Joscorbe: Ok, didn't do that because that method is used in 15 places, and thought it may be too many unnecessary changes.

stefan-egli: You can also keep the original method and have that invoke the new one with the default being `true`. That is probably good as changing 15 places seems problematic indeed

Joscorbe: Yes, did that already [in this commit](https://github.com/apache/jackrabbit-oak/pull/1556/commits/50b9eb6b12b876630923350d9d3b0caf56940d3e), it was easier than changing the 15 other places.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Checkpoints.java
```diff
@@ -158,6 +158,17 @@ public Revision getOldestRevisionToKeep() {
         return lastAliveRevision;
     }
 
+    /**
+     * Backwards compatible method to get the oldest revision to keep
+     * with cleanup.
```
stefan-egli: I would suggest to keep the original comment 

```suggestion
     * Returns the oldest valid checkpoint registered.
     *
     * <p>It also performs cleanup of expired checkpoint
```

not sure it necessarily needs the comment about backwards compatible though?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Checkpoints.java
```diff
@@ -118,7 +118,7 @@ public void release(String checkpoint) {
      */
     @SuppressWarnings("unchecked")
     @Nullable
-    public Revision getOldestRevisionToKeep() {
+    public Revision getOldestRevisionToKeep(boolean performCleanup) {
```
stefan-egli, line 120: Would suggest to add a javadoc description of the new `@param`

stefan-egli (commented):
Just added some javadoc comments.

What I was wondering though is whether it would be worth while to have a new test (that would verify that passing a read-only document store would now work, if `false` is passed to `getOldestRevisionToKeep`)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Checkpoints.java
rishabhdaim: IMO, a better name would be `isReadOnlyRepo`

Joscorbe: I'm adding the `@param`. 

But I don't think we should change the name to `isReadOnlyRepo` because then it may be misleading. We could know in this method if the repository is read-only simply calling:
`nodeStore.isReadOnlyMode();`

In our use-case we want to avoid the checkpoint cleanup when the repository is read-only, but I don't think we should do that in general because we would be changing the logic of other places where this method is already used.

stefan-egli (approved):

rishabhdaim (approved):

PR #1559, opened 2024-06-26
OAK-10918 : added logging of fullgc query explain once per 24h

stefan-egli:


Joscorbe (approved):
Looks good to me. I like the idea of reducing the length of the log message depending on the logging level.

stefan-egli:
test failures were : sonar : the usual java 17 issue - jenkins : long test duration or some other condition killing the tests after 1h+ - but they looked fine .. going to merge as these seem orthogonal issue

PR #1560, opened 2024-06-27
OAK-10905 | Add a configurable  async checkpoint creator service

nit0906:
This is to create a configurable job in oak that will create checkpoints after a said interval of time and ensure a min number of concurrent checkpoints in the system at a given point of time. 

The AsyncCheckpointService is defined with a factory config, meaning that client can configure this service with multiple configurations, each of would create a separate execution.

The time interval after which the task registered with this service will run is configured via timeIntervalBetweenCheckpoints -> this would be the time duration b/w creation of 2 checkpoints.

The lifetime of a checkpoint is configured via checkpointLifetime - this is the time till which a checkpoint will reside in the system if not deleted explicitly.

minConcurrentCheckpoints -> this is used to define the min number of checkpoints by a particular task of this service that should be kept in the system. It will be ensured that the system always has at least these many number of checkpoints.

More details of the configurations can be found in the AsyncCheckpointService class which then registers tasks of the Runnable AsyncCheckpointCreator.


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.jackrabbit.guava.common.collect.ImmutableMap;
+import org.apache.jackrabbit.oak.spi.state.NodeStore;
+import org.apache.jackrabbit.util.ISO8601;
+import org.jetbrains.annotations.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.util.Calendar;
+import java.util.Set;
+
+public class AsyncCheckpointCreator implements Runnable {
```
steffenvan, line 39: Can you write a description for this class that explains more in depth why it's necessary, what responsibility it has and where it's used?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public String getName() {
+        return name;
+    }
+
+    protected long getCheckpointLifetimeInSeconds() {
+        return checkpointLifetimeInSeconds;
+    }
+
+    protected long getMinConcurrentCheckpoints() {
```
steffenvan, line 82: Is there a reason this is protected rather than package-private?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
```diff
@@ -0,0 +1,100 @@
...
+        return minConcurrentCheckpoints;
+    }
+
+    @Override
+    public void run() {
+        Calendar cal = Calendar.getInstance();
+        String creationTimeStamp = String.valueOf(cal.getTimeInMillis());
+        String creationTimeISOFormat = ISO8601.format(cal);
+        String checkpoint = store.checkpoint(checkpointLifetimeInSeconds * 1000, ImmutableMap.of(
```
steffenvan: Since we are removing Guava, I think we can just replace this with `Map.of()` which is anyway immutable.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointService.java
```diff
@@ -0,0 +1,94 @@
...
+
+        @AttributeDefinition(
+                name = "Time Interval",
+                description = "Time interval between consecutive job runs in seconds. This would be the time interval between two consecutive checkpoints creation."
+        )
+        long timeIntervalBetweenCheckpoints() default 60 * 15;
+
+    }
+
+    private final List<Registration> regs = Lists.newArrayList();
```
steffenvan: Also removing Guava here.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/IndexUtilsTest.java
```diff
@@ -18,11 +18,20 @@
...
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.Set;
+import java.util.Map;
+import java.util.LinkedHashMap;
+import java.util.Calendar;
+
 import static org.apache.jackrabbit.guava.common.collect.Lists.newArrayList;
 import static org.apache.jackrabbit.oak.plugins.memory.EmptyNodeState.EMPTY_NODE;
 import static org.junit.Assert.*;
```
steffenvan: Wildcard import.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointServiceTest.java
```diff
@@ -0,0 +1,121 @@
...
+        AsyncCheckpointCreator checkpointCreator3 = getCheckpointCreator("checkpoint-async-test-3");
+        assertEquals(4, checkpointCreator3.getMinConcurrentCheckpoints());
+        assertEquals(60 * 24, checkpointCreator3.getCheckpointLifetimeInSeconds());
+        MockOsgi.deactivate(service, context.bundleContext());
+        assertNull(context.getService(Runnable.class));
+    }
+
+
+    private AsyncCheckpointCreator getCheckpointCreator(String name) {
+        return (AsyncCheckpointCreator) context.getServices(Runnable.class, "(oak.checkpoint.async="+name+")")[0];
```
steffenvan, line 106: What is the `[0]` element here?

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointServiceTest.java
```diff
@@ -0,0 +1,121 @@
...
+public class AsyncCheckpointServiceTest {
+
+    @Rule
+    public final OsgiContext context = new OsgiContext();
+
+    private final MemoryNodeStore nodeStore = new AsyncCheckpointServiceTest.FakeClusterableMemoryNodeStore();
+    private final AsyncCheckpointService service = new AsyncCheckpointService();
+
+    @Test
+    public void asyncReg() {
```
steffenvan, line 49: I don't fully understand how this test works. Could you write a description of what this test is doing?

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreatorTest.java
```diff
@@ -0,0 +1,59 @@
...
+ * under the License.
+ */
+package org.apache.jackrabbit.oak.plugins.index;
+
+import org.apache.jackrabbit.oak.plugins.memory.MemoryNodeStore;
+import org.apache.jackrabbit.oak.spi.state.NodeStore;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.*;
```
steffenvan, line 26: wildcard import

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointService.java
```diff
@@ -0,0 +1,94 @@
...
+    @Reference
+    private NodeStore nodeStore;
+
+    @Activate
+    public void activate(BundleContext bundleContext, AsyncCheckpointService.Configuration config) {
+        Whiteboard whiteboard = new OsgiWhiteboard(bundleContext);
+        if (config.enable()) {
```
steffenvan, line 84: Are we setting this to true anywhere? I see that the default value is `false` but don't see this set anywhere.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
```diff
@@ -0,0 +1,100 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.util.Calendar;
+import java.util.Set;
+
+public class AsyncCheckpointCreator implements Runnable {
+
+    /**
+     * Name of service property which determines the name of Async task
```
steffenvan: I don't fully understand this comment - could you please clarify?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointService.java
nit0906: Yes , default false is expected - it would be upto the downstream OSGI project to provide a configuration with enabled option. Oak is not responsible to set these values anywhere.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
nit0906: You can consider this an an identifier for the Async task / service. You can find it's usage in the corresponding test, where we filter based on this property.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
nit0906: Majorly to be able to verify in unit tests.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
steffenvan: Perhaps we could reformulate this to be:
`Name of service property which determines the name of *this* Async task`

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUtils.java
```diff
@@ -281,4 +287,35 @@ public static String getCaller(@Nullable String[] ignoredJavaPackages) {
...
+                .collect(Collectors.toMap(str -> str,
+                        store::checkpointInfo))
+                .entrySet()
+                .stream()
+                .filter(predicate)
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    }
+
+    /**
+     * Returns a map with checkpoint name as key and checkpoint metadata map as value, sorted on the basis of particular key in the metadata map.
```
steffenvan, line 309: could you rewrite what this could sort based on? Imo: `sorted on the basis of particular key in the metadata map` isn't super clear. And an example would be very helpful.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointServiceTest.java
nit0906: It would be an instance of the AsyncCheckpointCreator ..getServices basically returns a list (even if the match is a single item).

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
```diff
@@ -0,0 +1,100 @@
...
+        return checkpointLifetimeInSeconds;
+    }
+
+    protected long getMinConcurrentCheckpoints() {
+        return minConcurrentCheckpoints;
+    }
+
+    @Override
+    public void run() {
+        Calendar cal = Calendar.getInstance();
```
nfsantos: What is the timezone used to store the time? Shouldn't we specify it?
And more generally, the Calendar class is old, there is a new time API in Java. Maybe it would be better to use the equivalent classes on this new API.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
```diff
@@ -0,0 +1,100 @@
...
+        String creationTimeISOFormat = ISO8601.format(cal);
+        String checkpoint = store.checkpoint(checkpointLifetimeInSeconds * 1000, ImmutableMap.of(
+                CHECKPOINT_CREATOR_KEY, AsyncCheckpointCreator.class.getSimpleName(),
+                CHECKPOINT_CREATED_KEY, creationTimeISOFormat,
+                CHECKPOINT_CREATED_TIMESTAMP_KEY, creationTimeStamp,
+                CHECKPOINT_THREAD_KEY, Thread.currentThread().getName(),
+                CHECKPOINT_NAME_KEY, name));
+        log.info("[{}] Created checkpoint {} with creation time {}", name, checkpoint, creationTimeISOFormat);
+
+        // Get a sorted checkpoint set on created-epoch field from the list of filtered checkpoint based on the creating process.
```
nfsantos: This sentence is not clear.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
```diff
@@ -0,0 +1,100 @@
...
+            if (counter > minConcurrentCheckpoints) {
+                if(store.release(cp) ) {
+                    log.info("[{}] Deleted checkpoint {}", name, cp);
+                } else {
+                    log.warn("[{}] Unable to delete checkpoint {}. Removal will be attempted again in next run.", name, cp);
+                }
+            } else {
+                break;
+            }
+            counter --;
```
nfsantos, line 138: The counter is decreased even if the checkpoint was not deleted. Is it ok to delete fewer than the intended number of checkpoints?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
```diff
@@ -0,0 +1,100 @@
...
+                CHECKPOINT_CREATED_TIMESTAMP_KEY, creationTimeStamp,
+                CHECKPOINT_THREAD_KEY, Thread.currentThread().getName(),
+                CHECKPOINT_NAME_KEY, name));
+        log.info("[{}] Created checkpoint {} with creation time {}", name, checkpoint, creationTimeISOFormat);
+
+        // Get a sorted checkpoint set on created-epoch field from the list of filtered checkpoint based on the creating process.
+        Set<String> sortedCheckpointSet = IndexUtils.getSortedCheckpointMap(IndexUtils.getFilteredCheckpoints(store,
+                entry -> name.equals(entry.getValue().get(CHECKPOINT_NAME_KEY))), CHECKPOINT_CREATED_TIMESTAMP_KEY).keySet();
+        int counter = sortedCheckpointSet.size();
+        // Iterate over the sortedCheckpointSet and delete any checkpoints more than concurrentCheckpoints
```
nfsantos: The second part of the sentence is not clear. I guess you mean to say that it keeps the number of checkpoints to `concurrentCheckpoints`, deleting old checkpoints until there are only this number.

nfsantos (commented):
I'm missing a more detailed description of what the checkpoint service should do in the PR description. Like what is the configuration, when does it take checkpoints, how many it keeps, when checkpoints are deleted, what is the maximum time that a checkpoint can be kept. Otherwise, it's hard to evaluate the PR.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
nit0906: Yes - this was done on purpose - we need to keep a *minimum* number of checkpoints in the system. 

We are iterating on a sorted checkpoint set - > oldest first, if one of older checkpoints fail to get deleted and we do not decrease the counter - a comparative newer checkpoint will be deleted (which would be wrong). In this case - we simply get left with an additional older checkpoint, which should get deleted on the next job run.

I will try and add a comment to explain this better.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
nit0906: Added a more verbose comment to make it clear.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
nit0906: Tried to explain it better.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
nit0906: done.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointServiceTest.java
nit0906: added

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
nit0906: ok

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUtils.java
nit0906: >sorted on the basis of particular key in the metadata map.

every checkpoint has an associated metadata map and this sort can be based on any of the keys of the metadata map. 

Added an example to try and make it more clear.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
nit0906: I have added the a specific time zone as UTC , but I am not entirely sure if it's really needed (unless there could be a scenario where the author leader's pod could possibly get created in different timezones ? in which case, the default time zone would change).

Either way, I specified it explicitly now. 

Currently not moving to the new Time api , since using oak's ISO8601 for formatting, and would need to change that class as well to support the new api.

nit0906:
> I'm missing a more detailed description of what the checkpoint service should do in the PR description. Like what is the configuration, when does it take checkpoints, how many it keeps, when checkpoints are deleted, what is the maximum time that a checkpoint can be kept. Otherwise, it's hard to evaluate the PR.

@nfsantos  - I have tried to add more details in comments in the code and in the description. The main configurations of the service are defined in the AsyncCheckpointService class with descriptions. But I have added the same in the PR description as well.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointService.java
```diff
@@ -0,0 +1,95 @@
+package org.apache.jackrabbit.oak.plugins.index;
+
+
```
fabriziofortino: nit: multiple blank lines
```suggestion


```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointService.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import static org.apache.jackrabbit.oak.spi.whiteboard.WhiteboardUtils.ScheduleExecutionInstanceTypes.RUN_ON_LEADER;
+import static org.apache.jackrabbit.oak.spi.whiteboard.WhiteboardUtils.scheduleWithFixedDelay;
+
+@Component(
+        configurationPolicy = ConfigurationPolicy.REQUIRE,
+        service = {})
```
fabriziofortino: `service = {}` is redundant
```suggestion
@Component(configurationPolicy = ConfigurationPolicy.REQUIRE)
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
```diff
@@ -0,0 +1,121 @@
...
+        // Iterate over the sortedCheckpointSet which is sorted based on the creation timestamp -> the oldest first
+        // We try and delete the checkpoints as long as the counter is greater than minConcurrentCheckpoints
+        // This ensures that the system always has concurrent checkpoints equal to or greater than the configured minConcurrentCheckpoints
+        for (String cp : sortedCheckpointSet) {
+            // Delete the checkpoint as long as the checkpoint count is greater than concurrentCheckpoints
+            if (counter > minConcurrentCheckpoints) {
+                if(store.release(cp) ) {
+                    log.info("[{}] Deleted checkpoint {}", name, cp);
+                } else {
+                    log.warn("[{}] Unable to delete checkpoint {}. Removal will be attempted again in next run.", name, cp);
```
fabriziofortino, line 129: If we are not able to release a checkpoint we retry in the next run. That's fine.
What if this condition persists for several runs? In this case, we will keep adding checkpoints without deleting them.
Should we add some logic before creating a new checkpoint which checks there are not too many checkpoints?

nit0906: @fabriziofortino  - the point to deliberate here is - what is worse - 

1. Having more number of orphaned checkpoints in the system
2. Having to do a larger checkpoint diff (because we didn't create newer checkpoints because older ones didn't get deleted).

Please note that the checkpoint lifetime is configurable, with default suggested value as 24 hours, so if a checkpoint somehow manages to not get deleted in consequent runs, it should get cleaned up after 24 hours itself (if we keep that value in the configurations).

nit0906: I can still possibly add a check to not create any checkpoints if the number of checkpoints in the system right now are let's say 10 times than that of the minConcurrentCheckpoints value ? 
Or we could make it configurable with a value of maxConcurentCheckpoints as well

fabriziofortino: > Or we could make it configurable with a value of maxConcurentCheckpoints as well
+1

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
```diff
@@ -0,0 +1,121 @@
...
+                }
+            } else {
+                break;
+            }
+            // Decrement the counter irrespective of the outcome of the checkpoint deletion.
+            // If we don't decrement the counter in case of a failure while trying to delete a checkpoint,
+            // the next iteration will try to delete a comparatively newer checkpoint and keep the older one in the system (which we don't one).
```
nfsantos:
```suggestion
            // the next iteration will try to delete a comparatively newer checkpoint and keep the older one in the system (which we don't want).
```
I'm not totally sure you mean "want".

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointService.java
```diff
@@ -0,0 +1,95 @@
...
+
+
+
+import org.apache.jackrabbit.oak.osgi.OsgiWhiteboard;
+import org.apache.jackrabbit.oak.spi.state.NodeStore;
+import org.apache.jackrabbit.oak.spi.whiteboard.CompositeRegistration;
+import org.apache.jackrabbit.oak.spi.whiteboard.Registration;
+import org.apache.jackrabbit.oak.spi.whiteboard.Whiteboard;
+import org.osgi.framework.BundleContext;
+import org.osgi.service.component.annotations.*;
```
nfsantos: Do not use wildcard imports

nfsantos (approved):
Just a few minor comments.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
nit0906: done

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
```diff
@@ -0,0 +1,143 @@
...
+
+    @Override
+    public void run() {
+        Map<String, Map<String, String>> filteredCheckpointMap = IndexUtils.getFilteredCheckpoints(store,
+                entry -> name.equals(entry.getValue().get(CHECKPOINT_NAME_KEY)));
+        // If the number of checkpoints created by this task are equal to or greater than the maxConcurrentCheckpoints, skip
+        // creation of a new checkpoint.
+        // This could happen in case the deletion of older checkpoints failed in multiple previous runs.
+        if (filteredCheckpointMap.size() >= maxConcurrentCheckpoints) {
+            log.info("[{}] Skipping checkpoint creation as the number of concurrent checkpoints is already at max limit", name);
```
fabriziofortino: I would print this as a WARN
```suggestion
            log.warn("[{}] Skipping checkpoint creation as the number of concurrent checkpoints is already at max limit", name);
```

fabriziofortino (approved):
LGTM

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreator.java
nfsantos: I would also suggest indicating what is the max limit, it can be useful for debugging.

PR #1562, opened 2024-06-27
OAK-10921 : fixed race condition where fullGC database variables gets…

rishabhdaim:
… overridden if they are reset by external components

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
@@ -274,10 +274,11 @@ public class VersionGCRecommendations {
         } else if (!stats.canceled && !stats.ignoredGCDueToCheckPoint && !isFullGCDryRun) {
             // success, we would not expect to encounter revisions older than this in the future
             setVGCSetting(SETTINGS_COLLECTION_OLDEST_TIMESTAMP_PROP, scope.toMs);
-            updateVGCSetting(new HashMap<>() {{
-                put(SETTINGS_COLLECTION_FULL_GC_TIMESTAMP_PROP, stats.oldestModifiedDocTimeStamp);
-                put(SETTINGS_COLLECTION_FULL_GC_DOCUMENT_ID_PROP, stats.oldestModifiedDocId);
-            }});
+
+            final Map<String, Object> updateVGCMap = new HashMap<>();
+            updateVGCMap.put(SETTINGS_COLLECTION_FULL_GC_TIMESTAMP_PROP, stats.oldestModifiedDocTimeStamp);
+            updateVGCMap.put(SETTINGS_COLLECTION_FULL_GC_DOCUMENT_ID_PROP, stats.oldestModifiedDocId);
+            updateVGCSetting(updateVGCMap);
 
             int count = stats.deletedDocGCCount - stats.deletedLeafDocGCCount;
             double usedFraction;
@@ -359,9 +360,9 @@ public class VersionGCRecommendations {
      * @see VersionGCRecommendations#setVGCSetting(Map)
      */
     private void setVGCSetting(final String propName, final Object val) {
-        setVGCSetting(new HashMap<>() {{
-            put(propName, val);
-        }});
+        final Map<String, Object> vgcMap = new HashMap<>();
+        vgcMap.put(propName, val);
+        setVGCSetting(vgcMap);
     }
 
     /**
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
@@ -179,6 +179,9 @@ public class VersionGCRecommendations {
                         () -> oldestModifiedDocTimeStamp.set(0L));
                 oldestModifiedDocId = MIN_ID_VALUE;
                 log.info("fullGCTimestamp found: {}", timestampToString(oldestModifiedDocTimeStamp.get()));
+                // initialize the fullGC database variables i.e. fullGCTimestamp and fullGCId
+                setVGCSetting(of(SETTINGS_COLLECTION_FULL_GC_TIMESTAMP_PROP, oldestModifiedDocTimeStamp.get(),
+                        SETTINGS_COLLECTION_FULL_GC_DOCUMENT_ID_PROP, oldestModifiedDocId));
             } else {
                 oldestModifiedDocTimeStamp.set(fullGCTimestamp);
             }
@@ -266,13 +269,15 @@ public class VersionGCRecommendations {
             long nextDuration = Math.max(precisionMs, scope.getDurationMs() / 2);
             gcmon.info("Limit {} documents exceeded, reducing next collection interval to {} seconds",
                     this.maxCollect, TimeUnit.MILLISECONDS.toSeconds(nextDuration));
-            setLongSetting(VersionGarbageCollector.SETTINGS_COLLECTION_REC_INTERVAL_PROP, nextDuration);
+            setVGCSetting(VersionGarbageCollector.SETTINGS_COLLECTION_REC_INTERVAL_PROP, nextDuration);
             stats.needRepeat = true;
         } else if (!stats.canceled && !stats.ignoredGCDueToCheckPoint && !isFullGCDryRun) {
             // success, we would not expect to encounter revisions older than this in the future
-            setLongSetting(of(SETTINGS_COLLECTION_OLDEST_TIMESTAMP_PROP, scope.toMs,
-                    SETTINGS_COLLECTION_FULL_GC_TIMESTAMP_PROP, stats.oldestModifiedDocTimeStamp));
-            setStringSetting(SETTINGS_COLLECTION_FULL_GC_DOCUMENT_ID_PROP, stats.oldestModifiedDocId);
+            setVGCSetting(SETTINGS_COLLECTION_OLDEST_TIMESTAMP_PROP, scope.toMs);
+            updateVGCSetting(new HashMap<>() {{
+                put(SETTINGS_COLLECTION_FULL_GC_TIMESTAMP_PROP, stats.oldestModifiedDocTimeStamp);
+                put(SETTINGS_COLLECTION_FULL_GC_DOCUMENT_ID_PROP, stats.oldestModifiedDocId);
+            }});
 
             int count = stats.deletedDocGCCount - stats.deletedLeafDocGCCount;
             double usedFraction;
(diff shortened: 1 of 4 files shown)


oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollectorIT.java
```diff
@@ -94,7 +94,6 @@
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
-import static org.junit.Assume.assumeThat;
```
rishabhdaim, line 98: removed it, cause it was an unused import.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollectorIT.java
```diff
@@ -1478,6 +1477,77 @@ private Iterator<NodeDocument> candidates(long fromModified, long toModified, in
...
+
+        //Remove property
+        NodeBuilder b2 = store1.getRoot().builder();
+        for (int i = 0; i < 5; i++) {
+                b2.getChildNode("z"+i).removeProperty("prop");
+        }
+        store1.merge(b2, EmptyHook.INSTANCE, CommitInfo.EMPTY);
+        store1.runBackgroundOperations();
+
+        final AtomicReference<VersionGarbageCollector> gcRef = Atomics.newReference();
```
stefan-egli: what is `gcRef` used for? Maybe it was meant for some multi threading but doesn't look like it's used in any way?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollectorIT.java
```diff
@@ -1478,6 +1477,77 @@ private Iterator<NodeDocument> candidates(long fromModified, long toModified, in
...
+        final VersionGCSupport gcSupport = new VersionGCSupport(store1.getDocumentStore()) {
+
+            @Override
+            public Iterable<NodeDocument> getModifiedDocs(long fromModified, long toModified, int limit, @NotNull String fromId,
+                                                          final @NotNull Set<String> includePaths, final @NotNull Set<String> excludePaths) {
+                // reset fullGC variables externally while GC is running
+                store1.getDocumentStore().remove(SETTINGS, SETTINGS_COLLECTION_ID);
```
stefan-egli: It looks like we have two variations of reset :
* [reset|https://github.com/apache/jackrabbit-oak/blob/65e2ac78fb618915d6783033c61cd724fac54b36/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java#L382-L384]
* [reset fullGC|https://github.com/apache/jackrabbit-oak/blob/65e2ac78fb618915d6783033c61cd724fac54b36/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java#L386-L391]

IIUC then this just tests one of the two. Might be good to actually test both (whether in same or different test method)?

Joscorbe (approved):
Looks good to me, assuming there will be another test case as @stefan-egli already proposed.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollectorIT.java
```diff
@@ -191,7 +190,7 @@ public GCCounts(FullGCMode mode, int deletedDocGCCount, int deletedPropsCount,
     public DocumentStoreFixture fixture;
 
     @Parameter(1)
-    public FullGCMode fullGcMode;
+    FullGCMode fullGcMode;
```
rishabhdaim: reduced the scope for `FullGCMode` to avoid leaking this object.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollectorIT.java
stefan-egli: I see the new `resetFullGCFromOakRunWhileRunning` is virtually identical to `resetGCFromOakRunWhileRunning` except for `reset()` vs `resetFullGC()` - which was the idea of course - but I would suggest to consider refactoring these two to avoid (test) code duplication - drafted this (without testing) in https://github.com/apache/jackrabbit-oak/pull/1572 to illustrate

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
```diff
@@ -266,13 +269,15 @@ public void evaluate(VersionGCStats stats) {
...
-            setLongSetting(VersionGarbageCollector.SETTINGS_COLLECTION_REC_INTERVAL_PROP, nextDuration);
+            setVGCSetting(VersionGarbageCollector.SETTINGS_COLLECTION_REC_INTERVAL_PROP, nextDuration);
             stats.needRepeat = true;
         } else if (!stats.canceled && !stats.ignoredGCDueToCheckPoint && !isFullGCDryRun) {
             // success, we would not expect to encounter revisions older than this in the future
-            setLongSetting(of(SETTINGS_COLLECTION_OLDEST_TIMESTAMP_PROP, scope.toMs,
-                    SETTINGS_COLLECTION_FULL_GC_TIMESTAMP_PROP, stats.oldestModifiedDocTimeStamp));
-            setStringSetting(SETTINGS_COLLECTION_FULL_GC_DOCUMENT_ID_PROP, stats.oldestModifiedDocId);
+            setVGCSetting(SETTINGS_COLLECTION_OLDEST_TIMESTAMP_PROP, scope.toMs);
+            updateVGCSetting(new HashMap<>() {{
```
stefan-egli: subclassing HashMap seems heavy, what about using `Map.of()` as was done in some other cases?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
```diff
@@ -345,20 +350,31 @@ private Map<String, Object> getVGCSettings() {
         return settings;
     }
 
-    private void setLongSetting(String propName, long val) {
-        setLongSetting(of(propName, val));
+    private void setVGCSetting(final String propName, final Object val) {
+        setVGCSetting(new HashMap<>() {{
```
stefan-egli: (same as previous)
subclassing HashMap seems heavy, what about using `Map.of()` as was done in some other cases?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
```diff
@@ -345,20 +350,31 @@ private Map<String, Object> getVGCSettings() {
...
-        vgc.getDocumentStore().createOrUpdate(Collection.SETTINGS, updateOp);
+    private void setUpdateOp(final Map<String, Object> propValMap, final UpdateOp updateOp) {
+        propValMap.forEach((k, v) -> {
+            if (v instanceof Number) updateOp.set(k, ((Number) v).longValue());
+            if (v instanceof String) updateOp.set(k, (String) v);
+            if (v instanceof Boolean) updateOp.set(k, (Boolean) v);
+        });
+    }
+
+    private void updateVGCSetting(final Map<String, Object> propValMap) {
```
stefan-egli, line 394: Noticed a couple issues with this:
* it is not clear in which situation `updateVGCSetting` is used. It seems to be targeted for `evaluate`. But that method has a complex (if-) structure and it's not easy to figure out in which cases `updateVGCSetting` is called. Maybe this is more of a readability comment. Perhaps adding a comment would be a good improvement thouhg.
* `VersionGCInitTest.lazyInitialize` currently fails since `updateVGCSetting` _is_ invokved but with a condition that `fullGCTimeStamp` and `fullGCId` are set. Except it seems those are not initialized. So the initialization of these must somehow be better covered, I'd suggest first in a separate test case perhaps, and then in the code.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
rishabhdaim: `Map.of()` doesn't allow `null` values and during some test cases, some values happen to be `null` and cause the test failures.

stefan-egli: Got it. Alternatively, you could of course create the hasmap first, then add the key value. Would be slightly more code but avoid the subclass.

rishabhdaim: oh, didn't know the performance aspect of this. Will make the change.

stefan-egli:
Started rereviewing and must admit that it's perhaps not immediately obvious why we use `update` in one case and `set` in the other.

Then I noticed that the jira as well doesn't much go into details as to what exactly the race condition is. The PR does : 

>  (fixed) race condition where fullGC database variables gets overridden if they are reset by external components

What about adding a comment (or the actual description) to the jira that explains what goes wrong when "the race condition" happens? I guess it is : when you do a reset using oak-run while an oak instance is executing a regular fullGC (not dry run), then the reset would be ignored? And at the same time perhaps why dry run is different from wet run in this regard? Just might make code reading a bit easier.

PS: also noticed that due to the force-push I can't review the diff between my last review and this one - so can't follow what has changed since.

stefan-egli:
Related to that, currently `resetGCFromOakRunWhileRunning` and `resetFullGCFromOakRunWhileRunning` test doing a reset (in the 2 variants) against a wet fullGC run. Would having a a combination with a dry fullGC run (and the 2 reset variants) be good to have as well?

Perhaps in combination with explaining why dry and wet are different when they race with the different resets, could make the code a bit easier to understand.

rishabhdaim:
The flow is like this:

1. Full GC is running normally.
2. We reset the variables externally let's say via oak-run.
3. Oak run removes the variables **fullGCTimeStamp & fullGCId** from the database.
4. Once the normal cycle of fullGC finishes, it will again add/update those variables, which would nullify the effect of reset.

 
To fix this, we need to replace the updateAndCreate API (to update full GC variables after each full GC cycle) with findAndModify, this will ensure that in case they are already removed, they won't get added again.

 

Once the new cycle of GC starts, it will add them again with values of the oldest modified document & its ID.

 

Note: This race condition won't come in case of dry run because dry run is not supported in normal flow. We can dry run with only the oak run command.

stefan-egli (approved):
+1, thx for the explanation in the jira, helpful indeed.

PR #1570, opened 2024-07-03
OAK-10933 : skip timestamp calculation if fullgc is not enabled

stefan-egli:


stefan-egli:
PS: waiting for test to confirm no regression introduced by this - but ideally this would be included asap

stefan-egli:
tests look "good"

rishabhdaim:
IMO, this is an incomplete fix. We need to add `fullGCEnabled` check before `isFullGCDryRun` at https://github.com/apache/jackrabbit-oak/blob/trunk/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java#L152 & https://github.com/apache/jackrabbit-oak/blob/trunk/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java#L152, just like we did here: https://github.com/apache/jackrabbit-oak/blob/trunk/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java#L304

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
```diff
@@ -170,7 +170,7 @@ public class VersionGCRecommendations {
             } else {
                 oldestModifiedDryRunDocTimeStamp.set(fullGCDryRunTimestamp);
             }
-        } else {
+        } else if (fullGCEnabled) {
```
rishabhdaim, line 173: We should move this check before the `isFullGCDryRun` check.

rishabhdaim (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
stefan-egli: I was assuming dry run mode wouldn't be enabled if fullgc is disabled - but probably better to protect from that indeed. now done in https://github.com/apache/jackrabbit-oak/pull/1570/commits/0777573384e8d6504ab06eda073d0805a4e8a53a

stefan-egli:
agree this is incomplete - it is merely a first fix for the most noisy logger. we likely will have to exclude more - but that shouldn't prevent this one from going through.

rishabhdaim (approved):

PR #1577, opened 2024-07-11
OAK-10341 Tree store

thomasmueller:
Ready for review.


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/tree/TreeStore.java
```diff
@@ -0,0 +1,169 @@
...
+import org.apache.jackrabbit.oak.index.indexer.document.NodeStateEntry.NodeStateEntryBuilder;
+import org.apache.jackrabbit.oak.index.indexer.document.flatfile.NodeStateEntryReader;
+import org.apache.jackrabbit.oak.index.indexer.document.tree.store.Session;
+import org.apache.jackrabbit.oak.index.indexer.document.tree.store.Store;
+import org.apache.jackrabbit.oak.index.indexer.document.tree.store.StoreBuilder;
+import org.apache.jackrabbit.oak.index.indexer.document.tree.store.utils.Cache;
+import org.apache.jackrabbit.oak.plugins.memory.EmptyNodeState;
+import org.apache.jackrabbit.oak.spi.state.NodeState;
+
+public class TreeStore implements Iterable<NodeStateEntry>, Closeable {
```
nit0906: I see that this is draft right now - but we should probably implement the IndexStore interface to make it easier to integrate.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/tree/TreeStore.java
```diff
@@ -166,4 +171,28 @@ public Store getStore() {
...
+    }
+
+    @Override
+    public String getIndexStoreType() {
+        return STORE_TYPE;
+    }
+
+    @Override
+    public boolean isIncremental() {
+        return true;
```
nit0906: This should be false. 

The TreeStore should be the base implementation - that builds the a tree representation of the whole repository, and then we can have something like an IncrementalTreeStore, which builds the diff b/w 2 checkpoints to be appended to the base TreeStore.

Or maybe simply use the diff by IncrementalFlatFileStore and merge that in the base TreeStore.

Refer - 

IndexStore -> FlatFileStore -> IncrementalFlatFileStore

thomasmueller: To update the tree store incrementally, we can use log-structured merges, using Session.checkpoint() / Session.flush(). In theory, this is much more efficient than the IncrementalFlatFileStore, if done correctly. Specially if there are few changes. This is already implemented in the tree store, but needs to be "wired" correctly. 

But I think I will not yet do that just yet, but instead concentrate on getting the basics working correctly (right now, e.g. the cache size is in number of entries instead of memory used... this can lead to OOME, or then in too low memory usage and so bad performance, if we have very unevenly sizes pages -- and we probably have this). Also, I need to port some unit tests.

So I'll set it to "false" currently.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -239,6 +239,12 @@ nodeStore, getMongoDocumentStore(), traversalLog))
         return storeList;
     }
 
+    public IndexStore buildTreeStore() throws IOException, CommitFailedException {
+        System.setProperty(FlatFileNodeStoreBuilder.OAK_INDEXER_SORT_STRATEGY_TYPE,
```
nfsantos: This will leave this system property behind, even after finishing building the tree store, so it changes global state. Couldn't this cause unintended problems?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -19,17 +19,32 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
```
nfsantos: The order of imports is different from the one of InteliJ. We do not have an agreed formatting style, but since most of us are using InteliJ and the file is already with the InteliJ ordering, I'd prefer we leave it the current order.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -283,7 +324,7 @@ public List<IndexStore> buildList(IndexHelper indexHelper, IndexerSupport indexe
     private IndexStoreFiles createdSortedStoreFiles() throws IOException, CompositeException {
         // Check system property defined path
         String sortedFilePath = System.getProperty(OAK_INDEXER_SORTED_FILE_PATH);
-        if (StringUtils.isNotBlank(sortedFilePath)) {
+        if (sortedFilePath != null && !sortedFilePath.isEmpty()) {
```
nfsantos: This is not equivalent. If the string contains only whitespace, `StringUtils::isNotBlank` returns false, while this test here returns true, as `#isEmpty` tests that the String contains no characters.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedTransformTask.java
```diff
@@ -160,7 +160,7 @@ public Result call() throws Exception {
                     for (NodeDocument nodeDoc : nodeDocumentBatch) {
                         statistics.incrementMongoDocumentsTraversed();
                         mongoObjectsProcessed++;
-                        if (mongoObjectsProcessed % 50000 == 0) {
+                        if (mongoObjectsProcessed % 200_000 == 0) {
```
nfsantos: Why the change? I don't find that this log line is printed too often. But adding a thousand separator to the `50000` would be nice.

nfsantos (approved):
I have only looked at the parts that interact with existing code, did not review the TreeStore or IndexStore logic.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
thomasmueller: I didn't think about that... I'll try to fix it.

I think it's the same way as the "reuse existing flat file store" works btw. we might also revisit these other usages.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
thomasmueller: I'll revert this... I couldn't compile this part of the code due to a problem with the dependency.

(I don't think the usage of "StringUtils.isNotBlank" is intentional btw.)

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedTransformTask.java
thomasmueller: In my view we log way too much in general... but OK, I'll change it back.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
thomasmueller: Fixed here.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
thomasmueller: I think everyone uses a slightly different setting... but yes it makes sense to avoid such diffs.

This case should be resolved now.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
thomasmueller: Fixed.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedTransformTask.java
thomasmueller: Fixed.

steffenvan:
Just my opinion, but I see no comments that address the logic or readability of the TreeStore. Wouldn't it have made sense to have some reviews on that before merging?

thomasmueller:
> I see no comments that address the logic or readability of the TreeStore

Just documenting the tree store (alone) wont resolve that problem. I think we are missing high-level documentation of the implementation for many many features. Including the blob store, the segment store, the document store, oak-core, oak-jcr, etc. I think we should discuss what implementation level-documentation is needed _in general_. I view that as an independent problem.

PR #1582, opened 2024-07-16
OAK-10849: query-spi: remove deprecated slf4j.event.Level based method

reschke:


stefan-egli:
I'm not overly familiar with this bundle, so would like others to comment as well.

What I was wondering however, it seems like a somewhat short time between increasing to `ERROR` (in OAK-10847) and removing it? (I assume it has been deprecated for much longer, but the time window between additional noise level and removal seems a bit short?)

reschke:
ok, maybe schedule for 1.70.0 then...

PR #1595, opened 2024-07-22
OAK-10966 - Indexing job: create optimized version of PersistedLinkedList

nfsantos:
Adds a new implementation of PersistedLinkedList with several optimizations. These are the main differences compared to the current implement:

- The in-memory cache can be bounded both by number of elements and memory usage of those elements. The current implementation only supports bounding by number of elements. Bounding by memory is safer because entries can vary widely in size, so any limit only on the number of entries has to be very conservative to avoid OOME with large entries, and by being conservative will underperform in the common case where entries are small. By setting a bound in memory, we can use as much memory as available for caching without risking OOME. 
- Entries added to the list are either put in the in-memory cache or in the persistent store, but not in both. Currently, when an entry is added to the list it is put in the cache and written to the store, which incurs a serialization cost. In the common case that an entry fits in the cache, this serialization is wasted overhead.  
- Entries are not moved between the persistent store and the in-memory cache. This improves the cache hit rate. More on this later. 

# Performance improvement

### Experiment 1 - Medium size repository (70M nodes), with moderate size aggregations, up to 5K children

In an experiment indexing a FlatFileStore with more than 70M nodes and an index with aggregations, PersistedLinkedListV2 is 29% faster than the current implementation:

- PersistedLinkedList: 713s (11m53s)
- PersistedLinkedListV2: 514s (8m34) - **29% faster**

In both cases, the cache size was large enough that there were no cache misses, so the results reflect only the gains of not writing the entries to the persistent store when they fit in the cache.

To test the improvements in the cache miss rate, I tested with a different repository and with `cacheSize=1000`. This results in cache misses, because during indexing the persisted linked list reaches a maximum of almost 4000 entries. The PersistedLinkedListV2 outperforms by 19% the current implementation and has 39% fewer cache misses.

- PersistedLinkedList: 83s, 555K cache misses.
- PersistedLinkedListV2: 70s, 340K cache misses - **19% faster**

Note that the performance impact of the cache misses is not greater because the number of entries written to the persistent store at any one time is likely small enough to fit in the file system cache or in the cache of the persistent store.

### Experiment 2: Small repository (8M nodes) but with deep nesting in aggregated nodes (up to 40K children)

- PersistedLinkedList: 1h30m  
- PersistedLinkedListV2: 5m - **94% faster**

# Implementation

The in-memory cache is bounded by the following parameters 
- `cacheSize`: the maximum number of elements to keep in the in-memory cache</li>
- `cacheSizeMB`:  the maximum size of the in-memory cache in MB</li>

The recommended configuration is to bound the cache with a memory limit that ensures that the process will not run out of memory, and set a very high limit for the number of elements that can be stored in the cache. In experiments, I have observed that `cacheSize=50000` and `cacheSizeMB=64` is enough to avoid any cache miss even for repositories with deep nested documents and indexes with complex aggregations. 

Each element is stored either in the cache or in the persistent store, but not in both. And elements are not moved  between the two tiers, so even if there is a cache miss, that element will remain in the persistent store. For the access pattern of the indexer, this policy has a lower rate of cache misses than if we move to the cache an element after a miss.

To understand why, let's assume we want to traverse the children of a node P that is at line/position 100 in the FFS. When we call getChildren on P, this creates an iterator that scans from position 100 for all the children. If we call recursively getChildren on a child C of P, this will also create a new iterator that will start also at position 100. Therefore the iterators will frequently scan from 100 down. Let's assume the cache can only hold 10 nodes and that we use a policy of moving to the cache the last node that was accessed. Then if an iterator scans from 100 to, let's say 150, when it finishes iterating, the nodes 141 to 150 will be the only ones in the cache. The next iterator that is scanning from 100 will have cache misses for all the nodes until 140. And this will repeat for every new iterator. On the other hand, if we keep in the cache the nodes from 100 to 109, every iterator starting from 100 will at least have 10 cache hits, which is better than having a cache miss for all elements.


oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -606,7 +599,7 @@ public void onResult(Aggregate.PropertyIncludeResult result) {
      */
     private boolean indexAggregatedNode(String path, D doc, Aggregate.NodeIncludeResult result) {
         //rule for node being aggregated might be null if such nodes
-        //are not indexed on there own. In such cases we rely in current
+        //are not indexed on their own. In such cases we relly in current
```
nit0906: relly -> rely*

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreIterator.java
```diff
@@ -19,34 +19,44 @@
...
+    public static final String PERSISTED_LINKED_LIST_CACHE_SIZE = "oak.indexer.persistedLinkedList.cacheSize";
+    public static final int DEFAULT_PERSISTED_LINKED_LIST_CACHE_SIZE = 1000;
+
+    public static final String PERSISTED_LINKED_LIST_V2_CACHE_SIZE = "oak.indexer.persistedLinkedListV2.cacheSize";
+    public static final int DEFAULT_PERSISTED_LINKED_LIST_V2_CACHE_SIZE = 10000;
+
+    public static final String PERSISTED_LINKED_LIST_V2_MEMORY_CACHE_SIZE_MB = "oak.indexer.persistedLinkedListV2.cacheMaxSizeMB";
+    public static final int DEFAULT_PERSISTED_LINKED_LIST_V2_MEMORY_CACHE_SIZE_MB = 8;
+
+    public static final String PERSISTED_LINKED_LIST_USE_V2 = "oak.indexer.persistedLinkedList.useV2";
```
nit0906, line 58: Since we are using versioning convention here - would it make more sense to actually configure the version that is to be used instead of a flag to use/not use a particular version ?

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
nfsantos: fixed.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreIterator.java
nfsantos: I don't expect that there will be more versions of PersistedLinkedList. My intention is that for some time we keep both versions around, with the possibility to chose one or the other via configuration. It is a safe-guard for the case we find some issue with the new implementation. But once we feel confident that the PersistedLinkedListV2 is working properly, we should remove the old version as it offers no advantage over the new one, and also remove this configuration property.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/linkedList/PersistedLinkedListV2.java
```diff
@@ -0,0 +1,293 @@
...
+        } else {
+            cacheHits++;
+        }
+        return result;
+    }
+
+    private void addEntryToCache(Long index, NodeStateEntry entry) {
+        long now = System.currentTimeMillis();
+        long newCacheSizeBytes = cacheSizeEstimationBytes + entry.estimatedMemUsage();
+        if (cache.size() == cacheSizeLimit || newCacheSizeBytes > cacheSizeLimitBytes) {
```
thomasmueller: I would use the more conservative "cache.size() >= cacheSizeLimit".

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/linkedList/PersistedLinkedListV2.java
```diff
@@ -0,0 +1,293 @@
...
+            entryRemoved = reader.read(mapEntry);
+        } else {
+            cacheHits++;
+            cacheSizeEstimationBytes -= entryRemoved.estimatedMemUsage();
+        }
+
+        headIndex++;
+        totalEntries--;
+        if (totalEntries == 0) {
+            map.clear();
```
thomasmueller, line 174: We could log a warning here if cacheSizeEstimationBytes != 0

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/linkedList/PersistedLinkedListV2.java
```diff
@@ -0,0 +1,293 @@
...
+            String s = map.get(index);
+            result = reader.read(s);
+            LOG.trace("Cache miss: {}={}", index, result.getPath());
+        } else {
+            cacheHits++;
+        }
+        return result;
+    }
+
+    private void addEntryToCache(Long index, NodeStateEntry entry) {
```
thomasmueller: I would use "long" instead of "Long" here, because index may not be null. (I'm aware it can be null in the internal hash map.)

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreIterator.java
thomasmueller: (As always when using caches with a "size" limit and not a "memory" limit) there was a risk of running out-of-memory. This risk existed even before, with 1000. 

Now we have 10'000, but there is another "memory" limit at 8 MB. 8 MB is s bit small in my view, I would use 32 MB at least. What about using just the memory limit, and not a size limit? That would simplify the code, and simpler code typically has less bugs. I understand that on the other hand, if there is a bug with the memory calculation, then there is a risk as well. But I think the memory calculation must not have bugs (of the form, memory is smaller or equal 0): if it has, then we are in trouble anyway.

So I would opt for the simpler code!

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/linkedList/PersistedLinkedListV2.java
nfsantos: done

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/linkedList/PersistedLinkedListV2.java
nfsantos: Done.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/linkedList/PersistedLinkedListV2.java
nfsantos: Done. I initially chose Long in an earlier version of this method to reduce boxing. But in the current version, it already does only one boxing in all possible execution paths, so indeed this optimization is no longer necessary.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreIterator.java
nfsantos: Both of these design choices are based on learnings from developing the Pipelined strategy, which internally also keeps several buffers that ideally should be as big as possible while remaining within the limits.

- 8MB default - This is intentionally low to avoid failures when running with small heaps. In the past, I had test failures in some CI agents because they were configured with only 512MB of RAM and I had set a default buffer size that was taking up a significant portion of the available memory. The intention is that in any deployment of Oak, the value will be set to 128MB or greater, but by default, I prefer to be very conservative. And 8MB is not so small, based on my experiments, 1000 entries usually take less than 1MB of RAM and this is the default for the previous implementation. 
- Limit on size and count - This is to account for the case where entries are all very small. The memory estimation is not perfect, because it does not take in consideration the memory required by the internal structures of the HashMap or any other per-entry overhead that we may not be considering. Once again, I had issues with unexpected OOME with the buffers used in the Pipelined strategy when I did not set a limit on the number of entries. And having a limit on the number of entries gives us another lever to adjust in case we find some unexpected issues in production.

thomasmueller: > This is to account for the case where entries are all very small.

What I usually do is add some offset to the memory calculation... basically "+ 100" or so. (in the past I also verified that the memory calculation is somewhat correct, by have a small program that runs out-of-memory... but I understand this can't be a unit test, and we want to support multiple JVM versions without having to test each time... so the memory estimation of course won't ever be accurate... but it doesn't have to be)

thomasmueller (approved):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreIterator.java
nfsantos: The memory estimation is conservative because it assumes that all strings will be encoded with 2 bytes per char. In practice, in most cases the majority of strings will be encoded with 1 byte per char, so the memory estimation will almost double than the memory effectively used. And if using string interning or string deduplication, the difference might be even bigger. But of course we have to do the calculation for the worst case scenario. And as you say, it does not need to be precise. The goal here is just to prevent OOME while ensuring that the cache is big enough to give us the majority of the performance benefits.

PR #1596, opened 2024-07-22
OAK-10966 - Avoid object creation in PathUtils.isAncestor

nfsantos:
The indexing logic makes a large number of calls to `PathUtils.isAncestor` when traversing the FFS. This was showing up in a JFR profiling of an indexing task as a significant percentage of the total execution time and of the total allocations. It is fairly easy to avoid allocating a new string inside this method, and this optimization will benefit many other areas of Oak.

### DIFF
--- a/oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/PathUtils.java
+++ b/oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/PathUtils.java
@@ -367,12 +367,13 @@ public final class PathUtils {
         if (denotesRoot(ancestor)) {
             if (denotesRoot(path)) {
                 return false;
+            } else {
+                return path.startsWith(ancestor);
             }
         } else {
             // Equivalent to path.startsWith(ancestor + "/") but avoids allocating a new string
             return path.startsWith(ancestor) && path.length() > ancestor.length() && path.charAt(ancestor.length()) == '/';
         }
-        return path.startsWith(ancestor);
     }
 
     /**
--- a/oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/PathUtils.java
+++ b/oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/PathUtils.java
@@ -369,7 +369,8 @@ public final class PathUtils {
                 return false;
             }
         } else {
-            ancestor += "/";
+            // Equivalent to path.startsWith(ancestor + "/") but avoids allocating a new string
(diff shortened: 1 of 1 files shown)


fabriziofortino (approved):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/PathUtils.java
```diff
@@ -367,11 +367,13 @@ public static boolean isAncestor(String ancestor, String path) {
         if (denotesRoot(ancestor)) {
             if (denotesRoot(path)) {
                 return false;
+            } else {
+                return path.startsWith(ancestor);
```
reschke, line 371: Wouldn't that make "/a" and ancestor of "/aa"? Maybe we need more unit tests here?

nfsantos: There is a test for that case:
https://github.com/apache/jackrabbit-oak/blob/4354285d4a69efc0242a9332064b4c3f91197226/oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/PathUtilsTest.java#L233

The line 371 is inside a block that is executed only when `ancestor = "/"`. So this test is in fact just checking that path starts with "/". This is the same test that is done in the original test in this case.

nfsantos: The method has no detailed spec, for instance, it does not say what should be the result with relative paths. And it seems to assume that paths are given in the format "/a/b/c", not as "/a/b/c/". I tried to preserve the semantics of the method.

reschke: ah, ok.

reschke (approved):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/PathUtils.java
reschke: Unless there's test coverage for relative paths, I'd assume that the function is not supposed to be used with relative paths. We should document that, and reject relative paths.

I'm also not that happy with the use of "assert", but that's a separate issue.

Maybe we need to open a ticket to cover the relative path case.

nfsantos: I agree that the use of assert in this class is confusing, probably a misguided attempt at reducing runtime overhead. Maybe it made sense 10 years ago, but now I don't think it is justifiable. It should probably be replaced with standard error checking using Preconditions or something similar.

PR #1597, opened 2024-07-24
OAK-10971 - Add a method to test if a path is a direct ancestor of another: PathUtils.isDirectAncestor()

nfsantos:


reschke (commented):
So that's a new method, right?

Shouldn't this be called "isParent" or "isParentOf"?

Where would it be used?

nfsantos:
> So that's a new method, right?

Not quite. This method is to replace an internal implementation here:
https://github.com/apache/jackrabbit-oak/blob/2ef302327e5eb0b19c273a7a794e7008a5ba4fcb/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/ChildNodeStateProvider.java#L127-L129

But it would be a new method in the public API.

> Shouldn't this be called "isParent" or "isParentOf"?

I chose `isDirectAncestor` to be aligned with the existing method `isAncestor`. But I'm open to using a different name.

> Where would it be used?

See the link above, it would replace the internal method in ChildNodeStateProvider.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/PathUtils.java
```diff
@@ -374,6 +374,32 @@ public static boolean isAncestor(String ancestor, String path) {
...
+        }
+        int idx = path.lastIndexOf('/');
+        if (idx == 0) {
+            // path is a top level path. So it is only an immediate child of root
+            return PathUtils.denotesRoot(ancestor);
+        } else {
+            return idx == ancestor.length() && path.regionMatches(0, ancestor, 0, idx);
+        }
+    }
+
```
mbaedke, line 404: Why not simply

```
public static boolean isDirectAncestor(String ancestor, String path) {
    return isAncestor(ancestor, path) && path.lastIndexOf('/') == ancestor.length();
}

```
?

mbaedke (approved):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/PathUtils.java
nfsantos: The proposed implementation would fail with isDirectAncestor("/", "/foo"). 
- `isAncestor(ancestor, path) -> true`
- `path.lastIndexOf('/')` = 0
- `ancestor.length() = 1`

We could add another condition to check if the ancestor is `/`,  for instance:

```
        if (isAncestor(ancestor, path)) {
            int idx = path.lastIndexOf('/');
            if (idx == 0) {
                return PathUtils.denotesRoot(ancestor);
            } else {
                return idx == ancestor.length();
            }
        } else {
            return false;
        }
```
But this has almost the same complexity as the standalone implementation with the disadvantage of relying on a separate method.

nfsantos:
I have messed up the git history of this PR, so I opened a new one that is clean:
Please review this one:
https://github.com/apache/jackrabbit-oak/pull/1598

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/PathUtils.java
mbaedke: Ah yes, that is an edge case. What about

```
public static boolean isDirectAncestor(String ancestor, String path) {
	int idx = path.lastIndexOf('/');
    return ( ancestor.length() == 1 && idx == 0 || idx == ancestor.length() ) && isAncestor(ancestor, path);
}

```

PR #1603, opened 2024-07-25
OAK-10976 - Avoid unnecessary call to PathUtils.getName in IndexDefinition

nfsantos:
In NamePattern.matches(), do not compute the property name if the parent path does not match. This avoids an unnecessary call to PathUtils.getName() in the parentPath do not match. This logic is in the critical path of indexing.


reschke:
I'd prefer to split this into two PRs, if possible.

See https://jackrabbit.apache.org/oak/docs/participating.html#making-changes

nfsantos:
> I'd prefer to split this into two PRs, if possible.
> 
> See https://jackrabbit.apache.org/oak/docs/participating.html#making-changes

Done. PR with the cosmetic changes: https://github.com/apache/jackrabbit-oak/pull/1604

fabriziofortino (approved):

nit0906 (approved):

PR #1604, opened 2024-07-25
OAK-10977 - Cleanup IndexDefinition class

nfsantos:
- replace Guava usages with JDK equivalents 
- fix typos
- improve formatting

### DIFF
--- a/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/IndexDefinition.java
+++ b/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/IndexDefinition.java
@@ -19,29 +19,11 @@
 
 package org.apache.jackrabbit.oak.plugins.index.search;
 
-import java.io.InputStream;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Set;
-import java.util.UUID;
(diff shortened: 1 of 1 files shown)


reschke:
Be careful when replacing `ImmutableList.of()`- in Guava, it preservers iteration ordering...

nfsantos:
> Be careful when replacing `ImmutableList.of()`- in Guava, it preservers iteration ordering...

In the JDK, `List.of` and `List.copyOf` also preserve the order by which the elements are added. The problem might be in replacing a Guava by an JDK Set. In Guava, the iteration order is deterministic, but in the JDK the iteration order changes each time the JVM is restarted (they use some kind of randomness to ensure that users do not rely on a fixed iteration order in sets). 

Tomorrow I will look more carefully at the usages of the sets in this class to make sure no code is depending on the iteration order of sets, which anyway would be a bug.

nfsantos:
> > Be careful when replacing `ImmutableList.of()`- in Guava, it preservers iteration ordering...
> 
> In the JDK, `List.of` and `List.copyOf` also preserve the order by which the elements are added. The problem might be in replacing a Guava by an JDK Set. In Guava, the iteration order is deterministic, but in the JDK the iteration order changes each time the JVM is restarted (they use some kind of randomness to ensure that users do not rely on a fixed iteration order in sets).
> 
> Tomorrow I will look more carefully at the usages of the sets in this class to make sure no code is depending on the iteration order of sets, which anyway would be a bug.

I looked at all the cases of usages of Map or Set, and i did not find any that seemed to be relying on the order of iterations being always the same.

thomasmueller (approved):

PR #1605, opened 2024-07-29
OAK-10875 - unmerged bc on root can make it appear merged

ionutzpi:
If an oak instance crashes during a branch commit, which in turns leaves the bc unmerged - and that branch contained changes on root - it can later result in that unmerged bc to be considered as merged. If that values is then cached before other changes of the branch are read, those other changes are then considered merged even though they never have.


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocument.java
```diff
@@ -1943,6 +1943,9 @@ public static void setBranchCommit(@NotNull UpdateOp op,
     public static void removeBranchCommit(@NotNull UpdateOp op,
                                           @NotNull Revision revision) {
         checkNotNull(op).removeMapEntry(BRANCH_COMMITS, revision);
+        if (op.getId().equals(Utils.getIdFromPath("/"))) {
+            checkNotNull(op).removeMapEntry(COMMIT_ROOT, revision);
+        }
```
stefan-egli: I think removing the unmerged revision from `_commitRoot` is indeed what we should do. I do however think there is more required than just this. For example the question, what happens to a later resolving of a revision that is then older than sweep but all the info about the revision having been a branch commit is by then gone. That I think requires more testing and dependent on that perhaps another change.

For example what could be done as an alternative to removing `_commitRoot` *and* `_bc` would be to not remove neither - in case of an OAK-10875-type unmerged-bc. That would be somewhat of a step back, but it would clearly mark the revision as unmerged for good. (But I'd prefer a solution where we could remove `_commitRoot` as you did above).

Now one thing we'd need to change though, if we wanted to remove the `_commitRoot` : that shouldn't be done inside `removeBranchCommit` - as these static methos here on NodeDocument are basic operations on the UpdateOp that do only exactly what the method name says, that's a bit the contract of those methods. And also removing a `_commitRoot` as part of removing a `_bc` seems to break that contract.

stefan-egli (commented):
I think the fix should contain 2 main parts:

1. put the repository back into a state where things a proper, in case of a OAK-10875 style "inconsistency". For that I think removing the `_commitRoot` could be a part of it. Added more details to alternatives in the commend below (we could also not remove neither `_commitRoot` nor `_bc`).

2. as a second part though, we'd also have to handle cases where the inconsistency is already done and it's too late to clean up, as we've gone passed that moment (recovery was already done, unmerged bcs already removed). For this I think there could be two sub options:

2a. introduce a type of purgeUnmergedRevisions where we are able to clean up such an inconsistency even if it happened months ago (before the fix we want to introduce here), and thus the `_bc` is already gone. If this were possible, that would be nice, I think.

2b. make the "commit value resolving code" (which is spread over several classes) aware of this type of inconsistency and handle it properly. So that we wouldn't depend on a cleanup.

3. I'm thinking FullGC might also have to be made capable - or could be made capable - of cleaning up this inconsistency. Perhaps as alternative to 2a.

stefan-egli:
Also it might be good to modify the original test to have 2 kinds of tests:

* one that tests normally (basically the current test)
* one that tests when the repository was brought into the corrupted state, without any fix applied yet. Worst case the repository state would need to be created "hard coded" - eg by going through the steps as 1, but then explicitly force the state where `_bc` is removed but `_commitRoot` is not. This would be used to simulate suggested part 2. above.

(perhaps we can come up with a test that ensures that a property created as unmerged bc on root is never treated as committed - even though it is passed sweep rev and no branch commit info is remaining)

ionutzpi:
For the first part I removed the unmerged revision from _commitRoot(I think this is the proper fix to solve the bug).
For the second part, I will create another ticket to restore the state of repository if it is needed.

stefan-egli:
What about splitting the two parts, as you suggested, into two different PRs? I would tend to prefer if OAK-10875 would stay open as I believe both parts are required? But they could certainly be split into two PRs to simplify things. Wdyt?

Btw: there's something odd going on with this PR as it seems to duplicate the original : eg [unmergedBCOnRoot](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BranchTest.java#L117) is already there in trunk, but this PR adds it again, same with the changes of FailingDocumentStore. The [CI](https://ci-builds.apache.org/job/Jackrabbit/job/oak-trunk-pr/job/PR-1605/5/cloudbees-pipeline-explorer/?filter=52) currently complains with 
```
[ERROR] /home/jenkins/jenkins-agent/workspace/Jackrabbit_oak-trunk-pr_PR-1605/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BranchTest.java:[317,18] method invalidateCommitValueResolverCache(org.apache.jackrabbit.oak.plugins.document.DocumentNodeStore) is already defined in class org.apache.jackrabbit.oak.plugins.document.BranchTest
```

I think originally the PR was complaining it can't merge due to these duplications - now it doesn't do that anymore. No idea why...

ionutzpi:
Yes. We can have 2 PR for this ticket. The first part can be reviewed. I added a test method for the case where we have unmergedBC but no property set on root in the same BC(it didn't failed).

stefan-egli:
looks like there's some remaining merge conflict

ionutzpi:
It is fine now.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocument.java
```diff
@@ -610,6 +610,9 @@ int purgeUncommittedRevisions(final int clusterId, final int batchSize,
             if (!Utils.isCommitted(commitValue) && r.getClusterId() == clusterId && olderThanLastWrittenRootRevPredicate.test(r)) {
                 uniqueRevisions.add(r);
                 removeBranchCommit(op, r);
+                if (op.getId().equals(Utils.getIdFromPath("/"))) {
+                    removeCommitRoot(op, r);
+                }
```
stefan-egli:
```suggestion
                if (op.getId().equals(Utils.getIdFromPath(Path.ROOT))
                        && getLocalCommitRoot().containsKey(r)) {
                    removeCommitRoot(op, r);
                }
```

While this is somewhat of a detail, I think it would make it less intrusive if `removeCommitRoot` was only called when the revision `r` is actually on that document (`this`). Otherwise this is added to many unnecessary cases (where it wouldn't necessarily do harm, but be unnecessary). Wdyt?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BranchTest.java
```diff
@@ -56,6 +56,12 @@ public class BranchTest {
     @Rule
     public DocumentMKBuilderProvider builderProvider = new DocumentMKBuilderProvider();
 
+    @Before
+    public void before() {
```
stefan-egli:
```suggestion
    @After
    public void after() {
```

Shouldn't this be in an after method rather than before?

stefan-egli (commented):
I'd have one suggestion for consideration, otherwise looking good.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BranchTest.java
stefan-egli: (or there could be before and after, but only a before leaves state changed for other upcoming unit tests)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocument.java
ionutzpi: Done. It is fine for me.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BranchTest.java
ionutzpi: Done. I put it in after method.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/NodeDocument.java
stefan-egli: (going to close this conversation here, which was primarily about part 1 - for which now I think the way the PR is now is fine)

stefan-egli:
waiting for tests to succeed (or only have unrelated flakyness) before approving/merging next

stefan-egli:
failure seems unrelated :
```
[ERROR] Process Exit Code: 143
[ERROR] Crashed tests:
[ERROR] org.apache.jackrabbit.oak.jcr.RepositoryTest
[ERROR] org.apache.maven.surefire.booter.SurefireBooterForkException: The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
[ERROR] Command was /bin/sh -c cd /home/jenkins/workspace/Jackrabbit_oak-trunk-pr_PR-1605/oak-jcr && /usr/local/asfpackages/java/adoptium-jdk-11.0.16.1+1/bin/java -javaagent:/home/jenkins/maven-repositories/1/org/jacoco/org.jacoco.agent/0.8.12/org.jacoco.agent-0.8.12-runtime.jar=destfile=/home/jenkins/workspace/Jackrabbit_oak-trunk-pr_PR-1605/oak-jcr/target/coverage-reports/jacoco-ut.exec -Xmx512m -XX:+HeapDumpOnOutOfMemoryError -Dupdate.limit=100 -Djava.awt.headless=true -jar /home/jenkins/workspace/Jackrabbit_oak-trunk-pr_PR-1605/oak-jcr/target/surefire/surefirebooter9881064834527679463.jar /home/jenkins/workspace/Jackrabbit_oak-trunk-pr_PR-1605/oak-jcr/target/surefire 2024-08-21T14-23-22_802-jvmRun1 surefire1194720512709980725tmp surefire_011467686415861265474tmp
```

stefan-egli (approved):

PR #1612, opened 2024-07-30
OAK-10989 - Performance improvements to JSON parsing

nfsantos:
In JsonDeserializer:
- when reading a property, check for the most common types first (String).
- make instances reusable by creating a new MemoryNodeBuilder for every call to deserialize
- Remove usages of Guava collections.
- Move the methods in the inner class DeserializationSupport to top level methods in JsonDeserializer, as these inner class was not accessed anywhere outside JsonDeserializer and did not contain any state.

### DIFF
--- a/oak-store-spi/src/main/java/org/apache/jackrabbit/oak/json/JsonDeserializer.java
+++ b/oak-store-spi/src/main/java/org/apache/jackrabbit/oak/json/JsonDeserializer.java
@@ -19,12 +19,12 @@
 
 package org.apache.jackrabbit.oak.json;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 
 import javax.jcr.PropertyType;
 
-import org.apache.jackrabbit.guava.common.collect.Lists;
 import org.apache.jackrabbit.oak.api.PropertyState;
 import org.apache.jackrabbit.oak.api.Type;
 import org.apache.jackrabbit.oak.commons.LongUtils;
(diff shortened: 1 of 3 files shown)


oak-store-spi/src/main/java/org/apache/jackrabbit/oak/json/JsonDeserializer.java
```diff
@@ -39,49 +39,67 @@
...
 
-    public JsonDeserializer(BlobDeserializer blobHandler) {
-        this(blobHandler, EMPTY_NODE.builder(), DeserializationSupport.INSTANCE);
+    private static boolean hasOrderableChildren(NodeBuilder builder) {
+        PropertyState primaryType = builder.getProperty(JCR_PRIMARYTYPE);
+        return primaryType != null && ORDERABLE_TYPE.equals(primaryType.getValue(Type.NAME));
+    }
+
+    private static boolean hasSingleColon(String jsonString) {
+        // In case the primaryType was encoded then it would be like
```
thomasmueller, line 69: What about using two indexOf operations instead? E.g. one to get the first colon, another to get the last, and then check if they are equal. I think that would be simpler and faster. Ah I see it's just code that is moved around...

thomasmueller (approved):

oak-store-spi/src/main/java/org/apache/jackrabbit/oak/json/JsonDeserializer.java
nfsantos: Yes, I moved it from the bottom of the class to be at the top, as is the convention for static code.
Your proposed solution might indeed be faster, given that `indexOf()` is intrinsified, so it is executed with native code, likely using vector instructions. This should be faster than a loop on `charAt()`. Might be worth to explore, although the gains might not be large because property names are usually small.

PR #1619, opened 2024-08-01
OAK-10803 - Fix memory consumption of uncompress properties

ionutzpi:
In [OAK-10803](https://issues.apache.org/jira/browse/OAK-10803) we introduce the possibility of compressing property values. During its development we encountered performance issues, namely with repeated decompression calls, even just during getNodeAtRevision. Before we can confidentially enable and use compression, we need to revisit performance first. Likely we'll have to make further tweaks before we can use it.

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
@@ -51,7 +51,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * PropertyState implementation with lazy parsing of the JSOP encoded value.
+ * PropertyState compression implementation with lazy parsing of the JSOP encoded value.
  */
 public final class CompressedDocumentPropertyState implements PropertyState {
 
@@ -341,10 +341,3 @@ public final class CompressedDocumentPropertyState implements PropertyState {
         return createProperty(name, values, Type.fromTag(type, true));
     }
 }
-
-
-
-
-
-
-
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
@@ -227,52 +227,7 @@ public final class CompressedDocumentPropertyState implements PropertyState {
      * @return new property state
      */
     PropertyState readProperty(String name, JsopReader reader) {
-        return readProperty(name, store, reader);
-    }
-
-    /**
-     * Read a {@code PropertyState} from a {@link JsopReader}.
-     *
-     * @param name the name of the property state
-     * @param store the store
-     * @param reader the reader
-     * @return new property state
-     */
-    static PropertyState readProperty(String name, DocumentNodeStore store, JsopReader reader) {
-        if (reader.matches(JsopReader.NUMBER)) {
-            String number = reader.getToken();
-            Long maybeLong = LongUtils.tryParse(number);
-            if (maybeLong == null) {
-                return new DoublePropertyState(name, Double.parseDouble(number));
-            } else {
-                return new LongPropertyState(name, maybeLong);
-            }
-        } else if (reader.matches(JsopReader.TRUE)) {
-            return BooleanPropertyState.booleanProperty(name, true);
-        } else if (reader.matches(JsopReader.FALSE)) {
-            return BooleanPropertyState.booleanProperty(name, false);
-        } else if (reader.matches(JsopReader.STRING)) {
-            String jsonString = reader.getToken();
-            if (jsonString.startsWith(TypeCodes.EMPTY_ARRAY)) {
-                int type = PropertyType.valueFromName(jsonString.substring(TypeCodes.EMPTY_ARRAY.length()));
-                return PropertyStates.createProperty(name, emptyList(), Type.fromTag(type, true));
-            }
-            int split = TypeCodes.split(jsonString);
-            if (split != -1) {
-                int type = TypeCodes.decodeType(split, jsonString);
-                String value = TypeCodes.decodeName(split, jsonString);
-                if (type == PropertyType.BINARY) {
-
-                    return  BinaryPropertyState.binaryProperty(name, store.getBlobFromBlobId(value));
-                } else {
-                    return createProperty(name, StringCache.get(value), type);
-                }
-            } else {
-                return StringPropertyState.stringProperty(name, StringCache.get(jsonString));
-            }
-        } else {
-            throw new IllegalArgumentException("Unexpected token: " + reader.getToken());
-        }
+        return DocumentPropertyState.readProperty(name, store, reader);
     }
 
     /**
@@ -283,61 +238,6 @@ public final class CompressedDocumentPropertyState implements PropertyState {
      * @return new property state
      */
     PropertyState readArrayProperty(String name, JsopReader reader) {
-        return readArrayProperty(name, store, reader);
-    }
-
-    /**
-     * Read a multi valued {@code PropertyState} from a {@link JsopReader}.
-     *
-     * @param name the name of the property state
-     * @param store the store
-     * @param reader the reader
-     * @return new property state
-     */
-    static PropertyState readArrayProperty(String name, DocumentNodeStore store, JsopReader reader) {
-        int type = PropertyType.STRING;
-        List<Object> values = new ArrayList<>();
-        while (!reader.matches(']')) {
-            if (reader.matches(JsopReader.NUMBER)) {
-                String number = reader.getToken();
-                Long maybeLong = LongUtils.tryParse(number);
-                if (maybeLong == null) {
-                    type = PropertyType.DOUBLE;
-                    values.add(Double.parseDouble(number));
-                } else {
-                    type = PropertyType.LONG;
-                    values.add(maybeLong);
-                }
-            } else if (reader.matches(JsopReader.TRUE)) {
-                type = PropertyType.BOOLEAN;
-                values.add(true);
-            } else if (reader.matches(JsopReader.FALSE)) {
-                type = PropertyType.BOOLEAN;
-                values.add(false);
-            } else if (reader.matches(JsopReader.STRING)) {
-                String jsonString = reader.getToken();
-                int split = TypeCodes.split(jsonString);
-                if (split != -1) {
-                    type = TypeCodes.decodeType(split, jsonString);
-                    String value = TypeCodes.decodeName(split, jsonString);
-                    if (type == PropertyType.BINARY) {
-                        values.add(store.getBlobFromBlobId(value));
-                    } else if (type == PropertyType.DOUBLE) {
-                        values.add(Conversions.convert(value).toDouble());
-                    } else if (type == PropertyType.DECIMAL) {
-                        values.add(Conversions.convert(value).toDecimal());
-                    } else {
-                        values.add(StringCache.get(value));
-                    }
-                } else {
-                    type = PropertyType.STRING;
-                    values.add(StringCache.get(jsonString));
-                }
-            } else {
-                throw new IllegalArgumentException("Unexpected token: " + reader.getToken());
-            }
-            reader.matches(',');
-        }
-        return createProperty(name, values, Type.fromTag(type, true));
+        return DocumentPropertyState.readArrayProperty(name, store, reader);
     }
 }
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
@@ -16,36 +16,20 @@
  */
 package org.apache.jackrabbit.oak.plugins.document;
 
-import static java.util.Collections.emptyList;
-import static org.apache.jackrabbit.oak.plugins.memory.PropertyStates.createProperty;
-
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.List;
-
-import javax.jcr.PropertyType;
 
 import org.apache.jackrabbit.oak.api.PropertyState;
 import org.apache.jackrabbit.oak.api.Type;
 import org.apache.jackrabbit.oak.commons.Compression;
-import org.apache.jackrabbit.oak.commons.LongUtils;
 import org.apache.jackrabbit.oak.commons.json.JsopReader;
 import org.apache.jackrabbit.oak.commons.json.JsopTokenizer;
 import org.apache.jackrabbit.oak.commons.properties.SystemPropertySupplier;
-import org.apache.jackrabbit.oak.json.TypeCodes;
 import org.apache.jackrabbit.oak.plugins.memory.AbstractPropertyState;
-import org.apache.jackrabbit.oak.plugins.memory.BinaryPropertyState;
-import org.apache.jackrabbit.oak.plugins.memory.BooleanPropertyState;
-import org.apache.jackrabbit.oak.plugins.memory.DoublePropertyState;
-import org.apache.jackrabbit.oak.plugins.memory.LongPropertyState;
-import org.apache.jackrabbit.oak.plugins.memory.PropertyStates;
-import org.apache.jackrabbit.oak.plugins.memory.StringPropertyState;
-import org.apache.jackrabbit.oak.plugins.value.Conversions;
 import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
@@ -24,12 +24,12 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
 import javax.jcr.PropertyType;
 
-import org.apache.jackrabbit.guava.common.collect.Lists;
 import org.apache.jackrabbit.oak.api.PropertyState;
 import org.apache.jackrabbit.oak.api.Type;
 import org.apache.jackrabbit.oak.commons.Compression;
@@ -296,7 +296,7 @@ public final class CompressedDocumentPropertyState implements PropertyState {
      */
     static PropertyState readArrayProperty(String name, DocumentNodeStore store, JsopReader reader) {
         int type = PropertyType.STRING;
-        List<Object> values = Lists.newArrayList();
+        List<Object> values = new ArrayList<>();
         while (!reader.matches(']')) {
             if (reader.matches(JsopReader.NUMBER)) {
                 String number = reader.getToken();
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
@@ -66,7 +66,6 @@ public final class CompressedDocumentPropertyState implements PropertyState {
     private PropertyState parsed;
     private final byte[] compressedValue;
     private final Compression compression;
-    private String decompressedValue;
 
     private static int COMPRESSION_THRESHOLD = SystemPropertySupplier
             .create("oak.documentMK.stringCompressionThreshold ", -1).loggingTo(LOG).get();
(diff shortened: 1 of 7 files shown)


reschke (commented):
This saves space, but instead will make every read access slower, right?

ionutzpi:
Done. Refactored it to make read access faster.

stefan-egli:
I don't see the value in this. This will restrict usability of compression to the time between creation and first access of the value. Thereafter it uses more space.

ionutzpi:
Revert it to first commit. Yes, the read access is slower.

stefan-egli:
`decompressedValue` is never set?

ionutzpi:
Done.

stefan-egli:
IIUC this PR has now addressed the [suggestion](https://issues.apache.org/jira/browse/OAK-10803?focusedCommentId=17869342&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17869342) on OAK-10803. That I think we should follow-up with indeed.

What about doing it in 2 steps though:
* first address those suggestions (bring memory consumption back to pre-compression)
* then look into performance improvements

Currently this PR seems to mix both concerns, which makes review discussion a bit more complex.

Having said that, reg the performance improvements: I still think we need to address the performance aspect differently. As it stands now, the first call to `decompress()` will expand the value again to its original state (I would then have perhaps set `value` instead of introducing/duplicating `decompressedValue`) - which means it will use up again the original amount of memory (at which point the gain done by compression is over). The issue is that `decompress()` will be call pretty much immediately after a property was created - namely when it needs to be put into the cache and when its memory consumption is estimated. Hence there would be zero gain of compression if decompression was done as it stands now. (we could for example verify how the memory consumption calculation could be fixed - after all it's probably broken now with compression anyway - and perhaps that could make the compression state live longer)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateFactory.java
```diff
@@ -0,0 +1,15 @@
+package org.apache.jackrabbit.oak.plugins.document;
```
stefan-egli, line 17: license header is missing

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateFactory.java
```diff
@@ -0,0 +1,15 @@
...
+
+import org.apache.jackrabbit.oak.api.PropertyState;
+import org.apache.jackrabbit.oak.commons.Compression;
+
+public class DocumentPropertyStateFactory {
+
+    public static PropertyState createPropertyState(DocumentNodeStore store, String name, String value, Compression compression) {
```
stefan-egli, line 28: what about adding a variant that doesn't ask for the `Compression`? in that case it would perhaps hard code to GZIP, or later that could be configured..

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
```diff
@@ -0,0 +1,354 @@
...
+    private static int COMPRESSION_THRESHOLD = SystemPropertySupplier
+            .create("oak.documentMK.stringCompressionThreshold ", -1).loggingTo(LOG).get();
+
+    CompressedDocumentPropertyState(DocumentNodeStore store, String name, String value, Compression compression) {
+        this.store = store;
+        this.name = name;
+        this.compression = compression;
+        int size = value.length();
+        byte[] compressedValue = null;
+        if (size > COMPRESSION_THRESHOLD) {
```
stefan-egli: I would suggest to move that out of the constructor and check the size in the factory. Would also simplify the code of this class

stefan-egli (commented):
Is the plan to re-purpose this PR to now address OAK-10803? Then we might want to change the subject of the PR (that usually also becomes the commit message later).

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateFactory.java
ionutzpi: Done.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateFactory.java
ionutzpi: Done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
ionutzpi: Done

ionutzpi:
I separated construction of DocumentPropertyState from CompressedDocumentPropertyState through a factory method for step 1: first address those suggestions (bring memory consumption back to pre-compression)
We can change the subject

stefan-egli (commented):
What's missing is the wiring. With this change, the compression code is actually never triggered since it doesn't go through the factory yet.

ionutzpi:
Done. Call DocumentPropertyState through factory method.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateFactory.java
```diff
@@ -0,0 +1,46 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DocumentPropertyStateFactory {
+
+    private static final Logger LOG = LoggerFactory.getLogger(DocumentPropertyStateFactory.class);
+
+    public static PropertyState createPropertyState(DocumentNodeStore store, String name, String value, Compression compression) {
+
+        if (compression != null && !compression.equals(Compression.NONE) && value.length() > CompressedDocumentPropertyState.getCompressionThreshold()) {
```
stefan-egli: the length check here would result to true for `-1` - shouldn't there be a test for that, did that not fail for some reason?

ionutzpi: The tests are checking default threshold exceeded or not, it didnt failed because of passing Compression.NONE as argument

stefan-egli: There should be a test that should fail now though. The condition is currently wrong. 

Probably we should also use GZIP by default, otherwise the feature cannot be enabled at all?

ionutzpi: Done. We can enable the feature by calling differently factory method.

stefan-egli: > We can enable the feature by calling differently factory method.

But that would require a code change. What I was referring to is enabling the feature by configuration - eg via the system property. We should have that - otherwise the code is not useable and disable in a hard-coded way. I do think we should use GZIP in the default, not NONE.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
```diff
@@ -0,0 +1,350 @@
...
+import org.apache.jackrabbit.oak.plugins.memory.LongPropertyState;
+import org.apache.jackrabbit.oak.plugins.memory.PropertyStates;
+import org.apache.jackrabbit.oak.plugins.memory.StringPropertyState;
+import org.apache.jackrabbit.oak.plugins.value.Conversions;
+import org.jetbrains.annotations.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * PropertyState implementation with lazy parsing of the JSOP encoded value.
```
stefan-egli: Would suggest to add some short mention of the fact that this class uses compression.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
```diff
@@ -0,0 +1,350 @@
...
+        return createProperty(name, values, Type.fromTag(type, true));
+    }
+}
+
+
+
+
+
+
+
```
stefan-egli: detail but we probably should remove those lines

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -47,64 +38,24 @@
...
-    private final Compression compression;
-
-    private static int COMPRESSION_THRESHOLD = SystemPropertySupplier
-            .create("oak.documentMK.stringCompressionThreshold ", -1).loggingTo(LOG).get();
 
     DocumentPropertyState(DocumentNodeStore store, String name, String value) {
-        this(store, name, value, Compression.GZIP);
-    }
-
-    DocumentPropertyState(DocumentNodeStore store, String name, String value, Compression compression) {
```
stefan-egli, line 77: Presumably this class should be reverted to the state before compression? In doing that it looks like it missed a commit that came in between : b04de7ea9dce0859de09f90964fa650a7bf5a978

stefan-egli (commented):
Haven't yet fully checked but it looks like some tests got lost in translation? eg multiValuedAboveThresholdSize. Also, compressValueThrowsException : now expects an exception while previously it didn't?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyStateTest.java
```diff
@@ -0,0 +1,168 @@
...
+public class CompressedDocumentPropertyStateTest {
+
+    private static final String STRING_HUGEVALUE = RandomStringUtils.random(10050, "dummytest");
+    private static final int DEFAULT_COMPRESSION_THRESHOLD = 1024;
+    private static final int DISABLED_COMPRESSION = -1;
+
+    @Rule
+    public DocumentMKBuilderProvider builderProvider = new DocumentMKBuilderProvider();
+
+    private Set<String> reads = newHashSet();
```
reschke: please now new use of Guava.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateFactory.java
ionutzpi: Done.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
ionutzpi: Done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
ionutzpi: Done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Manually added that commit

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyStateTest.java
ionutzpi: Done

ionutzpi:
The constructor in CompressedDocumentPropertyState has changed from the previous version of DocumentPropertyState

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateFactory.java
stefan-egli: The test still doesn't fail consistently. I'm referring to `DocumentPropertyStateFactoryTest.createPropertyStateWithDefaultCompression`. That is one test at least that should fail with the current code. The reason it doesn't fail is that it depends on test execution order. Some other test method in that class is executed first (eg `createPropertyStateWithCompressionThresholdNotExceeded`) and that sets the threshold, so that `createPropertyStateWithDefaultCompression` then runs on false assumptions.

Things that require fixing:
* createPropertyStateWithDefaultCompression should check that it actually uses the default - i.e. it should have an assertEquals before even executing the actual property creation (and that would currently fail, depending on test execution order)
* in addition to such asserts in each method, there must also be a `@After` and a `@Before` that resets the threshold
* plus the actual code fix

ionutzpi: Done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
```diff
@@ -0,0 +1,343 @@
...
+     * Read a multi valued {@code PropertyState} from a {@link JsopReader}.
+     *
+     * @param name the name of the property state
+     * @param store the store
+     * @param reader the reader
+     * @return new property state
+     */
+    static PropertyState readArrayProperty(String name, DocumentNodeStore store, JsopReader reader) {
+        int type = PropertyType.STRING;
+        List<Object> values = Lists.newArrayList();
```
Joscorbe: I think this can be refactored too, and then entirely avoid the import of the guava dependency. I think that's what @reschke wanted to point out.

ionutzpi: Done. It wasn't very clear from the comment.

stefan-egli:
> The constructor in CompressedDocumentPropertyState has changed from the previous version of DocumentPropertyState

What question or comment does this refer to?

ionutzpi:
> Also, compressValueThrowsException : now expects an exception while previously it didn't?

This question

stefan-egli:
(Adding a general comment here in the main discussion as it otherwise might be getting a bit complicated by now)

What I was trying to say: the fact that the `DocumentPropertyStateTest` got refactored into 3 different classes lost a few aspects and thus results in a regression.
* as mentioned I believe some test methods got lost (I will double-check and can add more details)
* in addition though, some of the `@Before` and `@After` logic got lost - which actually is also a regression. For example, the broken surrogate tests no longer work as they were intended - they now inherit the threshold from whatever test/test-method was executed before, which is breaking change from the previous behavior (I think previously they tested with `-1`)
* (and the `-1` bug mentioned several times is thus still undetected by tests and is still existing in the code)

Plus what is also a regression vs the previous state is what I mentioned previously:
> Also, compressValueThrowsException : now expects an exception while previously it didn't?

Previously the constructor was swallowing those exceptions, which I think is a good thing. Now that has been changed. Why?

ionutzpi:
Moved broken surrogate tests on CompressedDocumentPropertyStateTest(before and after are in place here).
In CompressDocumentPropertyState constructor the logic is only for compression, if it fails then throw new IllegalArgumentException.

stefan-egli:
> In CompressDocumentPropertyState constructor the logic is only for compression, if it fails then throw new IllegalArgumentException.

That's a change to the previous behavior though. We previously discussed that it shouldn't throw an exception for compressing a property but just fall back to uncompressed. What is the reason for this change?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyStateTest.java
```diff
@@ -113,287 +80,4 @@ public void multiValuedBinarySize() throws Exception {
...
-        DocumentNodeStore store = mock(DocumentNodeStore.class);
-
-        DocumentPropertyState.setCompressionThreshold(-1);
-        DocumentPropertyState state = new DocumentPropertyState(store, "propertyNameNew", "\"" + STRING_HUGEVALUE + "\"", Compression.GZIP);
-
-        byte[] result = state.getCompressedValue();
-
-        assertNull(result);
-        assertEquals(state.getValue(Type.STRING), STRING_HUGEVALUE);
-    }
```
stefan-egli, line 211: This test seems removed. Is there a replacement for it?

ionutzpi: Done. I added a replacement

ionutzpi: In CompressedDocumentPropertyState constructor if it throws an exception, it will be caught in factory method and call DocumentPropertyState constructor.

stefan-egli:
> In CompressedDocumentPropertyState constructor if it throws an exception, it will be caught in factory method and call DocumentPropertyState constructor.

(Moving this back to main thread)

Could the test (`compressValueThrowsException`) then be adjusted to use the factory and thus avoid `(expected = IllegalArgumentException.class)` ?

stefan-egli:
On the test methods that seem to have gotten lost, here's a list of them - those existed prior to this PR and now seem gone (unless renamed) :
```
uncompressValueThrowsException
multiValuedAboveThresholdSize
stringAboveThresholdSizeNoCompression
testEqualsWithoutCompression
testInterestingStringsWithoutCompression
testOneCompressOtherUncompressInEquals
uncompressValueThrowsException
```

Also noticed that `stringBelowThresholdSize` actually tests _below_ threshold - so either the test name or the test code is wrong.

Also, as a general comment : I think it would be useful to go via the factory method whenever possible. That would extend the test coverage of the factory method - which is a very key element. (I think this would be the lost prio comment though)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
```diff
@@ -19,24 +19,16 @@
...
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.OutputStream;
-import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
 import java.util.List;
 
 import javax.jcr.PropertyType;
 
 import org.apache.jackrabbit.guava.common.collect.Lists;
```
stefan-egli: we might want to remove this guava usage as well, while we're effectively changing this class entirely.

ionutzpi:
All the tests mention above were refactored to use new context. The test suite created covered all scenarios previous done.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
```diff
@@ -0,0 +1,343 @@
...
+import org.apache.jackrabbit.oak.plugins.memory.StringPropertyState;
+import org.apache.jackrabbit.oak.plugins.value.Conversions;
+import org.jetbrains.annotations.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * PropertyState compression implementation with lazy parsing of the JSOP encoded value.
+ */
+public final class CompressedDocumentPropertyState implements PropertyState {
```
stefan-egli, line 40: I think this class could deserve some refactoring. The way it is now it basically duplicates a lot of code from DocumentPropertyState, which is not ideal.

Perhaps the reason for not directly extending DocumentPropertyState is that would inherit `value` and thus use more memory?

Then again I'm not liking the current code duplication a lot neither.

I think at least we should try to reuse what's easily possible - namely the static methods. Those we don't need in this class as well and we can use the DocumentPropertyState one.

wdyt?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentPropertyState.java
ionutzpi: Done.

stefan-egli:
> All the tests mention above were refactored to use new context. The test suite created covered all scenarios previous done.

where is uncompressValueThrowsException covered? where testInterestingStringsWithoutCompression ?

edit: plus where is testOneCompressOtherUncompressInEquals covered?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
ionutzpi: Done. Remove duplicate static methods. 
The 2 classes have different logic for calculating value, and also use more memory if subclassing, it should be separated.

ionutzpi:
Added required test methods in CompressedDocumentPropertyStateTest

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyState.java
```diff
@@ -0,0 +1,243 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.jackrabbit.oak.plugins.document;
+
+import static java.util.Collections.emptyList;
+import static org.apache.jackrabbit.oak.plugins.memory.PropertyStates.createProperty;
```
stefan-egli: looking good now - except I'd suggest to remove the unused imports here (in this class), then we should be good to go

ionutzpi: Done. Thanks

stefan-egli (approved):
+1, looks good now. Will let the test run finish (but I think 2 tests also fail in trunk, so those should anyway be ignored), and then (squash) merge depending on results.

rishabhdaim (approved):

stefan-egli:
ok, test failed, but the failures looks like unrelated flaky ones. going to merge anyway.

PR #1621, opened 2024-08-02
OAK-10999 - indexing-job: report nodes that take the longest to index

nfsantos:
- Log any document that takes more than 1 second to index.
- Log the top 20 slowest documents at the end of indexing.


reschke (commented):
Do we really need "nanoTime" here?

nfsantos:
> Do we really need "nanoTime" here?

`System.currentTimeMillis()` is not guaranteed to be monotonically increasing or to match the real elapsed time, so for tracking times `nanoTime()` is the only one that will provide accurate results in all conditions.

reschke:
> System.currentTimeMillis() is not guaranteed to be monotonically increasing or to match the real elapsed time

But that is only true when somebody changed the system clock; if this bothers us, there are probably many places in the code base we need to adjust...

nfsantos:
> > System.currentTimeMillis() is not guaranteed to be monotonically increasing or to match the real elapsed time
> 
> But that is only true when somebody changed the system clock; if this bothers us, there are probably many places in the code base we need to adjust...

Or when NTP adjusts the clock. Or in another situations that we may not be considering. For me the important point is that the specification of `System.currentTimeMillis()` does not give enough guarantees to be sure that the elapsed time is accurate, so I will not use it for that purpose. I don't think we should make assumptions about the situations where the results will be inaccurate and say that it will never happen in our case. It may even be true now, but change in the future. So I prefer to program to the specification and `System.nanoTime()` is what guarantees accurate calculation of elapsed time.

And I do think we should not make use of `System.currentTimeMillis()` to measure times, but this is up to the owners of those modules to decide. The chances of this causing us problems are small, especially that in most cases the results are only used for human consumption.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -107,6 +107,8 @@ public abstract class DocumentStoreIndexerBase implements Closeable {
     protected final IndexerSupport indexerSupport;
     private static final int MAX_DOWNLOAD_ATTEMPTS = Integer.parseInt(System.getProperty("oak.indexer.maxDownloadRetries", "5")) + 1;
 
+    private static final int TOP_SLOWEST_PATHS_TO_LOG = 20;
```
nit0906: should we make this configurable ?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -350,15 +352,29 @@ public void reindex() throws CommitFailedException, IOException {
...
+                        // inaccuracy will not significantly change the results.
+                        long entryEnd = System.nanoTime();
+                        long elapsedMillis = (entryEnd - entryStart) / 1_000_000;
+                        entryStart = entryEnd;
+                        slowestTopKElements.add(entry.getPath(), elapsedMillis);
+                        if (elapsedMillis > 1000) {
+                            log.info("Indexing {} took {} ms", entry.getPath(), elapsedMillis);
+                        }
                     }
+                    log.info("Top slowest nodes to index (ms): {}", slowestTopKElements);
```
nit0906, line 377: Would it make sense to log the content of the nodes as well ?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/TopKSlowestPaths.java
```diff
@@ -0,0 +1,70 @@
...
+            this.timeMillis = timeMillis;
+        }
+
+        @Override
+        public int compareTo(PathAndTime o) {
+            return Long.compare(timeMillis, o.timeMillis);
+        }
+
+        @Override
+        public String toString() {
```
nit0906, line 45: Would it make sense to log the node content as well ?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -350,15 +352,29 @@ public void reindex() throws CommitFailedException, IOException {
             INDEXING_PHASE_LOGGER.info("[TASK:INDEXING:START] Starting indexing");
             Stopwatch indexerWatch = Stopwatch.createStarted();
             try {
-
                 if (flatFileStores.size() > 1) {
                     indexParallel(flatFileStores, indexer, progressReporter);
                 } else if (flatFileStores.size() == 1) {
                     FlatFileStore flatFileStore = flatFileStores.get(0);
+                    TopKSlowestPaths slowestTopKElements = new TopKSlowestPaths(TOP_SLOWEST_PATHS_TO_LOG);
```
Amoratinos, line 360: Why not accommodate the `IndexingProgressReporter` to provide this information?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
nfsantos: Done.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/TopKSlowestPaths.java
nfsantos: I don't think it is necessary or advisable. Once we know what are the paths of the nodes, we can easily retrieve the node from the backing store and investigate further. And as we can't log the full contents of the node, we would have to select which parts to log, and we risk not to include the relevant information. But I think the more important reason not to log the contents is that we would be exposing more customer information on the logs, which we should avoid as much as possible.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
nfsantos: It seems to me that adapting `IndexingProgressReporter` to expose this information would require some refactoring. I'd rather keep this simple and orthogonal to existing mechanisms, at least until being more sure of the value of logging the slowest elements.

thomasmueller (approved):

nit0906 (approved):

PR #1622, opened 2024-08-02
OAK-11000: Test coverage for common JCR operations related to importing content

reschke:


oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/ImportOperationsTest.java
```diff
@@ -0,0 +1,222 @@
...
+import org.junit.Test;
+
+/**
+ * Test coverage for common JCR operations related to importing content.
+ * <p>
+ * Note that the purpose of these tests is not to check conformance with the JCR
+ * specification, but to observe the actual behavior of the implementation
+ * (which may be hard to change).
+ */
+public class ImportOperationsTest extends AbstractRepositoryTest {
```
kwin: The class name is confusing to me. What about ProtectedPropertyTest because IIUC this is mainly about jcr:uuid and jcr:created for cases when both become protected.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/ImportOperationsTest.java
```diff
@@ -0,0 +1,222 @@
...
+    public void jcrMixinCreatedOnNtUnstructured() throws RepositoryException {
+        Node test = testNode.addNode(TEST_NODE_NAME, NodeType.NT_UNSTRUCTURED);
+        try {
+            test.setProperty("jcr:created", false);
+            session.save();
+            test.addMixin(NodeType.MIX_CREATED);
+            session.save();
+            // surprise: setting the mixin succeeds, and the subsequent state of
+            // the node is inconsistent with the mixin's type definition
+            assertEquals(false, test.getProperty("jcr:created").getBoolean());
```
kwin: https://developer.adobe.com/experience-manager/reference-materials/spec/jcr/2.0/3_Repository_Model.html#3.7.11.7%20mix:created says:

> In cases where this mixin is added to an already existing node the semantics of these properties are implementation specific (see §10.10.3 Assigning Mixin Node Types).

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/ImportOperationsTest.java
```diff
@@ -0,0 +1,222 @@
...
+        Node test = testNode.addNode(TEST_NODE_NAME, NodeType.NT_UNSTRUCTURED);
+        try {
+            String testUuid = UUID.randomUUID().toString();
+            test.setProperty("jcr:uuid", testUuid);
+            session.save();
+            test.addMixin(NodeType.MIX_REFERENCEABLE);
+            session.save();
+            // surprise: setting the mixin succeeds, and the subsequent state of
+            // the node doesn't have an autocreated UUID
+            assertEquals(testUuid, test.getProperty("jcr:uuid").getString());
```
kwin: why should it get a new id assigned if the previous one does not have any conflicts? I think a test where you assign an already taken value and later on add the mixin is more interesting.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/ImportOperationsTest.java
```diff
@@ -0,0 +1,222 @@
...
+        Node test = testNode.addNode(TEST_NODE_NAME, NodeType.NT_UNSTRUCTURED);
+        test.addMixin(NodeType.MIX_REFERENCEABLE);
+        session.save();
+        try {
+            String testUuid = UUID.randomUUID().toString();
+            test.setProperty("jcr:uuid", testUuid);
+            session.save();
+            fail("Setting jcr:uuid after adding mixin:referenceable should fail");
+        } catch (ConstraintViolationException ex) {
+            // expected
```
kwin: This is IMHO mandated by the JCR spec as mix:referenceable defines jcr:uuid as protected (so this test rather belongs to the JCR TCK)

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/ImportOperationsTest.java
```diff
@@ -0,0 +1,222 @@
...
+            Node check = session.getNodeByIdentifier(testUuid);
+            assertTrue(test.isSame(check));
+        } finally {
+            test.remove();
+            session.save();
+        }
+    }
+
+    @Test
+    public void jcrMixinReferenceableOnNtUnstructuredAfterRemovingMixinButDanglingReference() throws RepositoryException {
```
kwin: JCR TCK?

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/ImportOperationsTest.java
kwin: Maybe even split up general protected properties from jcr:uuid into dedicated test classes make sense because the latter imposes additional restrictions.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/ImportOperationsTest.java
reschke: ok

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/ImportOperationsTest.java
reschke: If it's there we could drop it.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/ImportOperationsTest.java
reschke: Done the remaining; let's discuss splitting later on.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/ImportOperationsTest.java
reschke: well, the spec requires the jcr:uuid to be "autocreated", which IMHO does not mean "supplied by the caller". I *do* agree that another test would be helpful here.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/ImportOperationsTest.java
reschke: Maybe; if you can find it. For now I'll leave it here, because having related tests in different places doesn't seem to be helpful.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/ImportOperationsTest.java
reschke: Done (and confirmed that this fails with exception, altough it happes early - before mixin:ref was added).

stefan-egli:
Just to mention one further, somewhat brutal, option : Equivalent to temporarily removing the `mix:referenceable` (with the goal to allow changing the `jcr:uuid`) would be to temporarily change the primaryType to an unprotected one that would allow changing the `jcr:uuid`. That could be achieved either by using a hard-coded raw type (eg `nt:unstructured`, `oak:Unstructured` depending on orderable or not) - or by temporarily adding an node type that extends the existing one, and in the temporary one make `jcr:uuid` not protected, eg:
```
            NodeTypeTemplate unprotectedNTT = ntMgr.createNodeTypeTemplate();
            unprotectedNTT.setName(tmpNodeTypeName);
            unprotectedNTT.setDeclaredSuperTypeNames(new String[] {primaryType.getName()});
            PropertyDefinitionTemplate pdt = ntMgr.createPropertyDefinitionTemplate();
            pdt.setName("jcr:uuid");
            pdt.setProtected(false);
            unprotectedNTT.getPropertyDefinitionTemplates().add(pdt);
            session.getWorkspace().getNodeTypeManager().registerNodeType(unprotectedNTT, true);
            target.setPrimaryType(unprotectedNTT.getName());
            target.setProperty("jcr:uuid", newUUID);
            target.setPrimaryType(primaryType.getName());
            session.getWorkspace().getNodeTypeManager().unregisterNodeType(unprotectedNTT.getName());
            session.save();
```

reschke:
The first two options would be problematic if the parent node requires a certain type (ack @kwin :-). The latter would work if extending a type would indeed allow "unprotecting" a property...

stefan-egli:
unprotecting does work - I tested with the snipped added previously - it's just a bit ... brutal

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/ProtectedPropertyTest.java
```diff
@@ -0,0 +1,376 @@
...
+        try {
+            String testUuid = UUID.randomUUID().toString();
+            test.setProperty("jcr:uuid", testUuid);
+            session.save();
+            test.addMixin(NodeType.MIX_REFERENCEABLE);
+            session.save();
+            // JCR spec
+            // (https://developer.adobe.com/experience-manager/reference-materials/spec/jcr/2.0/3_Repository_Model.html#3.8%20Referenceable%20Nodes)
+            // requests an "auto-created" property, so it might be a surprise
+            // that Oak actually keeps the application-assigned previous value.
```
mbaedke, line 108: The algorithm used to create the property value is free to use the existing one and I think that is also the most sane approach, so is it really that surprising?

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/ProtectedPropertyTest.java
```diff
@@ -0,0 +1,376 @@
...
+            session.save();
+        }
+    }
+
+    @Test
+    public void jcrMixinReferenceableOnNtUnstructuredAfterRemovingMixinButDanglingReference() throws RepositoryException {
+        Node test = testNode.addNode(TEST_NODE_NAME, NodeType.NT_UNSTRUCTURED);
+        test.addMixin(NodeType.MIX_REFERENCEABLE);
+        session.save();
+        Node ref = testNode.addNode(TEST_NODE_NAME_REF, NodeType.NT_UNSTRUCTURED);
```
mbaedke, line 236: Nitpick: Earlier the name ref was used for the referenced node, not the referencing one. So we might want to change that to avoid confusion.

mbaedke (approved):

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/ProtectedPropertyTest.java
reschke: the spec says "autocreated" :-)

mbaedke (approved):

PR #1627, opened 2024-08-02
OAK-11004: Remove usage of Guava Supplier interface

reschke:


reschke:
(as usual I'll let that sit for ~1 week)

nfsantos (approved):

PR #1633, opened 2024-08-09
OAK-11008 - AzureBlobStore: do not create a new CloudBlobContainer instance for every request to the backend

nfsantos:
Currently, whenever AzureBlobStoreBackend needs to make a call to Azure, it creates a new instance of CloudBlobContainer. But instances of this class can be reused and the configuration does not change after an instance of AzureBlobStoreBackend is created.

When making many calls to the Azure blob store, reusing the connection will likely speedup the calls to Azure.


amit-jain (approved):
lgtm!!

smiroslav (approved):

PR #1644, opened 2024-08-14
OAK-10792 : rename enum value DGC to FGC

rishabhdaim:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java
```diff
@@ -428,6 +428,18 @@ public GCResult(boolean ignoreGC, TimeInterval gcScope) {
...
 
     enum GcType {
-        RGC, DGC
+        RGC("Revision GC"), // revision GC
+        FGC("Full GC"); // full GC
+
+        private final String name;
+
+        GcType(String name) {
+            this.name = name;
```
stefan-egli, line 437: noticed that this changes a [log.warn](https://github.com/apache/jackrabbit-oak/blob/19ee59c6f484c272ffbc04e58c8ba254d7c6da9e/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGCRecommendations.java#L408) so could influence down-stream automation that greps for the existing string (RGC vs "Revision GC")

stefan-egli: if we want this, perhaps there is a way to keep that log.warn intact, eg append this new `name` at the end and keep the original one at `RGC` .. not sure ...

stefan-egli: but maybe it's fine :)

rishabhdaim: the main reason for this PR is to remove `DGC` with `FGC`. I added names to the enum constructor to give them meaning names for better clarity.

stefan-egli (approved):

PR #1645, opened 2024-08-14
OAK-11025 Silence more warnings for ordered properties

thomasmueller:


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMaker.java
```diff
@@ -300,18 +300,42 @@ protected boolean indexTypeOrderedFields(Document doc, String pname, int tag, Pr
...
-            log.warn(
-                    "[{}] Ignoring ordered property. Could not convert property {} of type {} to type {} for path {}",
-                    getIndexName(), pname,
-                    Type.fromTag(property.getType().tag(), false),
-                    Type.fromTag(tag, false), path, e);
+            boolean unknownWarning = true;
+            String message = e.getMessage();
+            if (message.startsWith("Not a date string") ||
+                    message.startsWith("Unable to parse the provided date field") ||
+                    message.startsWith("For input string")) {
```
nfsantos: I think we should treat each of the messages separately. As it is, this implementation may totally suppress some classes of warnings. A message like "Not a date string" might never be shown if it just happens to always occur when the logging is being suppressed. Maybe we can use this class, which already does something like what we want:

https://github.com/apache/jackrabbit-oak/blob/trunk/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/LogSilencer.java

thomasmueller: > I think we should treat each of the messages separately. 

At first, I also thought about doing something similar: having a separate delay for each property. But that would complicate the code, and in my view unnecessarily: We don't have any evidence that someone (except us) is looking at these warnings. If someone _is_ looking at the warnings, then he would take action, no matter what message.

I had a look at the LogSilencer, but it seems it is not protected against multi-threaded access to the cache. I don't want to risk running out of memory.

So I think the patch should be kept as-is.

nfsantos: I don't agree that we should suppress altogether some types of errors. Either the error message is important and should always be logged (although removing duplicates), or else it is not important and it should not be logged at all. But a probabilistic approach where the message is logged in some runs but not in other runs, depending only on chance does not seem to be correct.

From what I saw, the LogSilencer is thread-safe and supports bounding the memory usage, so there is no risk of OOME.
https://github.com/apache/jackrabbit-oak/blob/396b3c4626cec75d35638146ee9aab335e8a387d/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/LogSilencer.java#L34-L45

The only thing I don't like so much about the LogSilencer is that in this case it would be more useful to count the number of times a message is suppressed and then print a warning like:

> <ErrorMessage> (this message was repeated n times).

So that we would know how widespread is the particular issue (if a few dozen times in a run, then ignore, if it appears millions of times, may be worth to investigate).

thomasmueller: > I don't agree that we should suppress altogether some types of errors

Well, the patch doesn't. It just avoid filling the log file.

> Either the error message is important and should always be logged (although removing duplicates), or else it is not important and it should not be logged at all.

I don't see this as a binary thing. We do log a warning if we find a problem, but we need to avoid logging too much. That is what the patch does.

> From what I saw, the LogSilencer is thread-safe

Ah I didn't see Collections.synchronizedMap sorry.

OK I'll try to use it.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMaker.java
```diff
@@ -299,19 +301,43 @@ protected boolean indexTypeOrderedFields(Document doc, String pname, int tag, Pr
...
+            // - NumberFormatException: For input string
+            // For these we do not log a stack trace, and we only log once every 10 seconds
+            // (the location of the code can be found if needed, as it's in Oak)
+            if (message.startsWith("Not a date string")) {
+                key = LOG_KEY_NOT_A_DATE_STRING;
+            } else if (message.startsWith("Unable to parse the provided date field")) {
+                key = LOG_KEY_UNABLE_TO_PARSE;
+            } else if (message.startsWith("For input string")) {
+                key = LOG_KEY_FOR_INPUT_STRING;
+            }
```
nfsantos, line 324: This could be simplified a bit, something like:
```
List<String> errorMessagePrefixes = List.of("Not a date string", "Unable to parse the provided date field", "For input string");
...
String key = errorMessagePrefixes.stream().find(t -> message.startsWith(t))...
```

nfsantos (approved):

PR #1655, opened 2024-08-19
OAK-11039: Flaky MissingLastRevSeekerTest with RDB

sii74826:


stefan-egli (approved):

reschke (commented):
I'd prefer to actually find the root cause; there seems to be another test that breaks the test contract for RDB DocumentStore tests.

sii74826:
I fully agree, but for now we need to apply a workaround (similar to [OAK-11019](https://issues.apache.org/jira/browse/OAK-11019) ) to recover pipelines that have been failing for several weeks.
If we want to merge it, could you please merge it (with squash option)? No permissions.

Joscorbe (approved):
I agree we should find the root cause and fix it, but could we merge this temporarily to unblock the pipeline meanwhile? @reschke 

I would vote to merge this, but with the aim of fixing it properly _in future_

reschke (changes requested):
I believe this has been overtaken by events; see [OAK-11113](https://issues.apache.org/jira/browse/OAK-11113).

PR #1656, opened 2024-08-20
OAK-11035: Remove usage of Guava Strings.nullToEmpty

reschke:


rishabhdaim:
This PR title is not in line with the changes. In the PR, we are also changing guava's `Objects.hashCode` with java native and the same for guava's `Objects.equal`.

I would suggest that either we change the title of PR (and jira as well) OR we create separate PRs to remove usage of guava's `Objects.hashcode` & `Objects.equal`.

reschke:
@rishabhdaim - that's because this change was on top of OAK-11033. Now that that is merged, it should be much smaller.

That said, something is indeed strange here. Let me check this.

nfsantos (approved):

PR #1664, opened 2024-08-22
OAK-11054 Oak AsyncCheckpointCreatorTest sometimes fails

thomasmueller:
Wait 1 ms

### DIFF
--- a/oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreatorTest.java
+++ b/oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreatorTest.java
@@ -32,10 +32,12 @@ public class AsyncCheckpointCreatorTest {
         NodeStore store = new MemoryNodeStore();
         int minConcurrentCheckpoints = 3;
         int maxConcurrentCheckpoints = 5;
-        AsyncCheckpointCreator task = new AsyncCheckpointCreator(store, "test", 1000, minConcurrentCheckpoints, maxConcurrentCheckpoints);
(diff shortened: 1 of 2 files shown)


oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/AsyncCheckpointCreatorTest.java
```diff
@@ -47,14 +49,36 @@ public void testAsync() {
...
+                Thread.sleep(1);
+            } catch (InterruptedException e) {
+                // ignore
+            }
+            long time = System.currentTimeMillis();
+            if (time - start >= 1) {
+                break;
+            }
+        }
+    }
```
nfsantos, line 83: Why wait in a loop instead of doing only `Thread.sleep(1)`? Is there any reason why Thread.sleep() might return earlier unless it is interrupted? If the thread was interrupted, then we do not need to ensure 1ms delay because the test is being aborted. Furthermore, catching the InterruptException without reseting the interrupt flag will suppress the interrupt, which can block an attempt of terminating the thread/JVM.

thomasmueller: There are two cases: the thread could be interrupted, and the milliseconds could change (adjusting the system clock). The second case is probably more likely. In which case, the test still fails, because the millisecond is still the same. (The real code uses the System.currentTimeMillis).

> which can block an attempt of terminating the thread/JVM.

Yes. But this is a test case. I'm not worried about this.

I'm fixing a broken test case here.

fabriziofortino (approved):

PR #1667, opened 2024-08-22
OAK-11059: Remove usage of Guava BiMap

reschke:


thomasmueller (commented):
Can't we just use HashMap? I don't see (in the patch) that the reverse lookup is made at all... Possibly it was used in the past, and then at some point removed, and now it only uses one-way? (I didn't do an investigation..)

reschke:
It also guarentees that values are unique, right?

thomasmueller:
Yes, it guarantees entries to be unique, by removing keys if the same value was added.

```
        BidiMap<String, String> map = new DualHashBidiMap<>();
        map.put("key1", "value1");
        map.put("key2", "value2");
        map.put("key3", "valueA");
        map.put("key4", "valueA");
        assertEquals("value1", map.get("key1"));
        assertEquals("value2", map.get("key2"));
        assertEquals(null, map.get("key3"));
        assertEquals("valueA", map.get("key4"));

```

But at least in the BlobThroughPutTest.java this behavior doesn't seem to be used. And inverseBidiMap doesn't seem to be used. So it seems a regular hash map is sufficient there.

I didn't check the other usage.

thomasmueller (commented):
At least in the test case, a regular hash map seems sufficient.

reschke:
Well.

This is a case where only two classes were affected (one of which is a test).

Doing this kind of analysis when removng 100s of other cases of Guava usage will not scale. I very much prefer to change things systematically in a uniform way.

thomasmueller:
> Doing this kind of analysis when removng 100s of other cases of Guava usage will not scale.

Well but changes always have some risk. If we simply replace without understanding the implications, then we risk introducing bugs, which is worse.

I think you remember a case where we replaced (without review) the Guava Hash Map with a non-Guava one, and then this broke many customer environments? If not let me know, I can point you to the issue.

So I would prefer to review the changed, work on them, and then work on the next one, etc. I know it is slow, but it's better to be slow than to introduce bugs.

reschke:
Yes, I do remember that case. At the end of the day, every issue can be tracked to lack of tests.

In this particular case, we could of course decide not to change any test classes, because it really does not matter whether they use Guava or not.

The main strategy that I used for this activity is to identify the Guava usages that are simple and mechanical to do, because we have equivalents in the JDK or common-collections. Where this is the case, we should just do the obvious translation, and be done with it. Potential optimizations (like using a simple Map instead of a BidiMap) can of course be done if we feel it's important, but it should *not* be combined with the actual Guava replacment, unless the change *really* is trivial. Anything that requires some kind of analysis is not, because we have *hundreds* of lines of code to change.

WRT your example: this happened because (a) it was less than obvious that the replacement was not fully compatible, and (b), of course, lack of test coverage. I do agree that we need to be careful, and the best way to do so is to resist the temptation to micro-manage every change.

So, in summary: I disagree. I'd like to hear what others think.

thomasmueller:
> unless the change really is trivial

Well I think using HashMap is a trivial change, but OK, I don't block this change. It is important that we progress.

thomasmueller (approved):
Even thought I think it would be possible to use HashMap, I think it's OK.

PR #1669, opened 2024-08-23
OAK-11061 - Indexing job: during indexing phase, download blobs ahead of time in separate thread pool

nfsantos:
During indexing it is often necessary to download non-inlined blobs from the Blob store. This can significantly slow down indexing when done by the main indexing thread.

This PR adds an asynchronous blob downloader that scans the Flat file store independently of the indexer, searching for blobs that match a given pattern and downloading them ahead of the indexer. This will populate the local caches (in-memory blob store cache and in-disk data store cache), so that the indexing thread will find them locally when it needs to retrieve them.

The AOT downloader has a throttling mechanism to prevent it from going too far ahead of the indexer. This is to avoid filling up the local caches and causing evictions of blobs that were not yet processed by the indexer. In this case, the indexer would have to download again the blob. This throttling mechanism requires that the indexer periodically inform the AOT downloader of its position. The AOT downloader keeps internally a prefetch window and will download a blob only if there is space in the prefetch window, otherwise it blocks until there is space again. This window is bounded by two parameters:
- Number of blobs in the window
- Cumulative size of the blobs in the window 

With these settings, it is possible to configure the prefetch window to be just a little smaller than the in-memory blob store cache, so that the blobs are still in memory when they are needed by the indexer. Otherwise, a prefetch window larger than the blob store cache but smaller than the data store cache will still be effective, as the blobs will be found on the data store cache (with a slightly higher latency than if they were in the in-memory cache). The prefect window should not be bigger than the blob store cache, otherwise there is the risk of overflowing the data store cache and causing evictions.

New configuration properties:

- `oak.indexer.blobPrefetch.enableForIndexesPrefixes` - Name of indexes for which ahead of time blob downloading is enabled. Default: `""`.
- `oak.indexer.blobPrefetch.binaryNodesSuffix` - The suffix of the Oak nodes whose binary property `jcr:data` will be downloaded. Empty disables the AOT downloader. Default: `""`.
- `oak.indexer.blobPrefetch.downloadThreads` - Number of parallel download threads. Default: 4.
- `oak.indexer.blobPrefetch.downloadAheadWindowMB` - Maximum total size of the blobs in the download window. Default: 4096 
- `oak.indexer.blobPrefetch.downloadAheadWindowSize` - Maximum number of blobs in the download window. 32MB


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -357,26 +369,37 @@ public void reindex() throws CommitFailedException, IOException {
...
-                        // the measured times will also include the bookkeeping at the end of indexing each entry, but
-                        // we are only interested in entries that take a significant time to index, so this extra
-                        // inaccuracy will not significantly change the results.
-                        long entryEnd = System.nanoTime();
-                        long elapsedMillis = (entryEnd - entryStart) / 1_000_000;
-                        entryStart = entryEnd;
-                        slowestTopKElements.add(entry.getPath(), elapsedMillis);
-                        if (elapsedMillis > 1000) {
-                            log.info("Indexing {} took {} ms", entry.getPath(), elapsedMillis);
+                    try (AheadOfTimeBlobDownloader aheadOfTimeBlobDownloader = getAheadOfTimeBlobDownloaderInterface(flatFileStore, indexer)) {
```
thomasmueller: I would prefer if we wrap the index store, instead of adding the logic in the DocumentStoreIndexerBase.

This is because the tree store doesn't require (this type of) the AheadOfTimeBlobDownloader.

We already have the IndexStore API, which can be wrapped.

thomasmueller (changes requested):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
nfsantos: Can you elaborate? I'm not sure I understand the suggestion. 

The AOT downloader needs to receive callbacks with the position of the indexer, in order to not go to far ahead of the indexer. That is, it requires the callback this callback: `aheadOfTimeBlobDownloader.updateIndexed(entriesRead);`. 
which is done in the code below. Where could this code be moved to?

thomasmueller: The AheadOfTimeBlobDownloader is an interface and the only "real" implementation requires a FlatFileStore. So another IndexStore is not supported (which is fine because only the FlatFileStore needs it). But it means all IndexStores need to support somehow the AheadOfTimeBlobDownloader which is not true: the TreeStore doesn't need it, so the AheadOfTimeBlobDownloader should not appear here.

I would instead extend the FlatFileStore such that it can be wrapped by the AheadOfTimeBlobDownloader and then the AheadOfTimeBlobDownloader could be an IndexStore.

That way there is no hard dependency on the AheadOfTimeBlobDownloader here. And so the TreeStore doesn't require a dependency on the AheadOfTimeBlobDownloader or implement any of it's methods. Instead, we can use the IndexStore here. 

I'm against "forcing" a dependency on the AheadOfTimeBlobDownloader on other IndexStores that do not need it.

We can discuss in person tomorrow.

nfsantos: As discussed in person, could you suggest a refactoring of this PR to remove the dependency from the DocumentStoreIndexerBase to the AheadOfTimeBlobDownloader? I gave it a try but I did not find any obvious and simple way to do it.

thomasmueller (approved):

nfsantos:
This PR requires reworking. The logic to decide if a blob should be downloaded is too generic, because it assumes that if a node name matches a given suffix, then it should download the `jcr:data` property. But in fact, we should download the `jcr:data` property if the node matches the suffix and if the `jcr:data` property is actually required for the index being indexes. As a first approach, we can add an additional configuration property that enables AOT blob download only if the index name is in a list given as a configuration property. This would required that we configure Oak with the indexes that require this field. 

A better approach would be to somehow introspect the index definitions and figure out if they access the jcr:data field.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/AheadOfTimeBlobDownloadingFlatFileStore.java
```diff
@@ -75,17 +80,35 @@ private AheadOfTimeBlobDownloadingFlatFileStore(FlatFileStore ffs, CompositeInde
                     maxPrefetchWindowSize,
                     maxPrefetchWindowMB);
         }
-    }    
+    }
+
+    static boolean isEnabledForIndexes(String indexesEnabledPrefix, List<String> indexPaths) {
```
thomasmueller, line 85: Assuming indexesEnabledPrefix is not set, we have an empty string here...

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/AheadOfTimeBlobDownloadingFlatFileStore.java
```diff
@@ -75,17 +80,35 @@ private AheadOfTimeBlobDownloadingFlatFileStore(FlatFileStore ffs, CompositeInde
...
+        List<String> enableForIndexes = splitAndTrim(indexesEnabledPrefix);
+        for (String indexPath : indexPaths) {
+            if (enableForIndexes.stream().anyMatch(indexPath::startsWith)) {
+                return true;
```
thomasmueller, line 89: ... in which case we have "/oak:index/...".startsWith("")... that would always be true.

I think if enableForIndexes is empty, then this methods should always return false.

nfsantos: There is a test case for when `indexesEnabledPrefix` is empty, the function correctly returns false. See the first case in the companion test class below. The loop won't be executed if the string is empty and it will fall through to the `return false` at the end.

PR #1673, opened 2024-08-28
OAK-10532 Cost estimation for not(@x) calculates cost for @x='value' 

thomasmueller:
Half of the change is wiring in the feature toggle.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
@@ -584,6 +584,10 @@ public class Oak {
             LOG.info("Registered Prefetch feature: " + QueryEngineSettings.FT_NAME_PREFETCH_FOR_QUERIES);
             closer.register(prefetchFeature);
             queryEngineSettings.setPrefetchFeature(prefetchFeature);
+            Feature improvedIsNullCostFeature = newFeature(QueryEngineSettings.FT_NAME_IMPROVED_IS_NULL_COST, whiteboard);
+            LOG.info("Registered improved cost feature: " + QueryEngineSettings.FT_NAME_IMPROVED_IS_NULL_COST);
+            closer.register(improvedIsNullCostFeature);
+            queryEngineSettings.setImprovedIsNullCostFeature(improvedIsNullCostFeature);
         }
 
         return this;
@@ -977,6 +981,10 @@ public class Oak {
 
         public void setPrefetchFeature(@Nullable Feature prefetch) {
             settings.setPrefetchFeature(prefetch);
+        }
+
+        public void setImprovedIsNullCostFeature(@Nullable Feature feature) {
+            settings.setImprovedIsNullCostFeature(feature);
         }
 
         @Override
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
@@ -218,7 +218,8 @@ public class QueryEngineSettings implements QueryEngineSettingsMBean, QueryLimit
 
     @Override
     public boolean getImprovedIsNullCost() {
-        return improvedIsNullCostFeature != null && improvedIsNullCostFeature.isEnabled();
+        // enabled if the feature toggle is not used
+        return improvedIsNullCostFeature == null || improvedIsNullCostFeature.isEnabled();
     }
 
     public String getStrictPathRestriction() {
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
@@ -63,6 +63,8 @@ public class QueryEngineSettings implements QueryEngineSettingsMBean, QueryLimit
 
     public static final String FT_NAME_PREFETCH_FOR_QUERIES = "FT_OAK-10490";
 
+    public static final String FT_NAME_IMPROVED_IS_NULL_COST = "FT_OAK-10532";
+
     public static final int DEFAULT_PREFETCH_COUNT = Integer.getInteger(OAK_QUERY_PREFETCH_COUNT, -1);
 
     public static final String OAK_QUERY_FAIL_TRAVERSAL = "oak.queryFailTraversal";
@@ -111,6 +113,7 @@ public class QueryEngineSettings implements QueryEngineSettingsMBean, QueryLimit
     private final long queryLengthErrorLimit = Long.getLong(OAK_QUERY_LENGTH_ERROR_LIMIT, 100 * 1024 * 1024); //100MB
 
     private Feature prefetchFeature;
+    private Feature improvedIsNullCostFeature;
 
     private String autoOptionsMappingJson = "{}";
     private QueryOptions.AutomaticQueryOptionsMapping autoOptionsMapping = new QueryOptions.AutomaticQueryOptionsMapping(autoOptionsMappingJson);
@@ -133,6 +136,10 @@ public class QueryEngineSettings implements QueryEngineSettingsMBean, QueryLimit
 
     public void setPrefetchFeature(@Nullable Feature prefetch) {
         this.prefetchFeature = prefetch;
+    }
+
+    public void setImproveCostFeature(@Nullable Feature improvedCost) {
+        this.improvedIsNullCostFeature = improvedCost;
     }
 
     @Override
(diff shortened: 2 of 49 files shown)


oak-query-spi/src/main/java/org/apache/jackrabbit/oak/spi/query/package-info.java
```diff
@@ -18,7 +18,7 @@
 /**
  * This package contains oak query index related classes.
  */
-@Version("2.0.0")
+@Version("3.0.0")
```
nit0906: why a major version bump ? There doesn't seem to be any breaking change included in this PR.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
```diff
@@ -135,6 +138,10 @@ public void setPrefetchFeature(@Nullable Feature prefetch) {
         this.prefetchFeature = prefetch;
     }
 
+    public void setImproveCostFeature(@Nullable Feature improvedCost) {
```
nit0906: This is not getting used anywhere

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
```diff
@@ -584,6 +584,10 @@ public Oak with(@NotNull Whiteboard whiteboard) {
             LOG.info("Registered Prefetch feature: " + QueryEngineSettings.FT_NAME_PREFETCH_FOR_QUERIES);
             closer.register(prefetchFeature);
             queryEngineSettings.setPrefetchFeature(prefetchFeature);
+            Feature improvedIsNullCostFeature = newFeature(QueryEngineSettings.FT_NAME_IMPROVED_IS_NULL_COST, whiteboard);
```
nit0906, line 587: I would argue this being behind a "feature" toggle - it's more of fixing an incorrect behaviour for me.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndexPlanner.java
```diff
@@ -838,17 +841,29 @@ private int getMaxPossibleNumDocs(Map<String, PropertyDefinition> propDefns, Fil
...
             int weight = propDef.getValue().weight;
 
-            PropertyRestriction pr = filter.getPropertyRestriction(key);
             if (pr != null) {
                 if (pr.isNotNullRestriction()) {
                     // don't use weight for "is not null" restrictions
+                    // as all documents with this field can match;
+                    weight = 1;
+                } else if (improvedIsNullCost && pr.isNullRestriction()) {
+                    // don't use the weight for "is not null" restrictions
```
nit0906: did you mean 

// don't use the weight for "is null" restrictions

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndexPlanner.java
```diff
@@ -838,17 +841,29 @@ private int getMaxPossibleNumDocs(Map<String, PropertyDefinition> propDefns, Fil
...
 
             int weight = propDef.getValue().weight;
 
-            PropertyRestriction pr = filter.getPropertyRestriction(key);
             if (pr != null) {
                 if (pr.isNotNullRestriction()) {
                     // don't use weight for "is not null" restrictions
+                    // as all documents with this field can match;
+                    weight = 1;
+                } else if (improvedIsNullCost && pr.isNullRestriction()) {
```
nit0906, line 864: we could possibly combine this else if with the above if since both assign the same weight.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
thomasmueller: Oh!

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
thomasmueller: Yes, it is fixing an incorrect behavior. The problem is, someone might depend on the bug! That is the reason to have a feature toggle. See also https://www.hyrumslaw.com/ and https://xkcd.com/1172/

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndexPlanner.java
thomasmueller: Yes, of course it's possible. I kept it separate on purpose: so that the conditions and the comments are easier to read. I think in this case having the more verbose "duplication" of one line is better than to combine the cases and have a complex condition and comment. But it's just a personal preference, I assume.

oak-query-spi/src/main/java/org/apache/jackrabbit/oak/spi/query/package-info.java
reschke: I believe this is a non-internal API, so we need to be extra careful here.

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
thomasmueller: BTW if no feature toggle is set, then it is enabled. So the feature toggle is more about the ability to (temporarily) switch to the old behavior, in case of problems.

oak-query-spi/src/main/java/org/apache/jackrabbit/oak/spi/query/package-info.java
thomasmueller: The "QueryLimits" has a new method "getImprovedIsNullCost", I think that's why: the build broke without this change. I can try to add a default implementation in for the new method if you think the version bump is a problem?

thomasmueller: > I believe this is a non-internal API, so we need to be extra careful here.

It is supposed to be an internal API. I'm not aware of any external users.

We (relatively) recently added getIgnoredClassNamesInCallTrace which also caused a version bump, if I remember correctly. This didn't cause any issues.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
thomasmueller: Fixed.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndexPlanner.java
thomasmueller: Fixed.

oak-query-spi/src/main/java/org/apache/jackrabbit/oak/spi/query/package-info.java
reschke: If it is internal, it should be marked as that ("@Internal") annotation.

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
```diff
@@ -584,6 +584,10 @@ public Oak with(@NotNull Whiteboard whiteboard) {
...
+            LOG.info("Registered improved cost feature: " + QueryEngineSettings.FT_NAME_IMPROVED_IS_NULL_COST);
```
nfsantos, line 588: To be consistent, would be better to use string interpolation for all log statements.

nfsantos (approved):

nit0906 (approved):

oak-query-spi/src/main/java/org/apache/jackrabbit/oak/spi/query/package-info.java
reschke: You can't do that. It is not internal. Once it's public it stays public.

reschke (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
thomasmueller: Actually it would make it inconsistent with the above line, which is:

      LOG.info("Registered Prefetch feature: " + QueryEngineSettings.FT_NAME_PREFETCH_FOR_QUERIES);

reschke (approved):

PR #1675, opened 2024-08-29
OAK-11069 - Include RuntimeExceptions so we get logs for all of these.

ahanikel:
AzureRepositoryLock should handle other types of exceptions in refreshLease.
https://github.com/apache/jackrabbit-oak/blob/53b51f29e40cd0f0f9e30999751b63de64038737/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureRepositoryLock.java#L140-L140

This PR does not solve the problem entirely: the refresher thread is still giving up, but at least the situation is discoverable in the logs and can be acted upon.

I am tempted to add a `Runtime.getRuntime().exit(1)` to AzurePersistence:106 (replacing the "TODO"), but I think this needs further examination. The situation is non-recoverable, however.

### DIFF
--- a/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureRepositoryLock.java
+++ b/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureRepositoryLock.java
@@ -96,7 +96,7 @@ public class AzureRepositoryLock implements RepositoryLock {
                 leaseId = blob.acquireLease(leaseDuration, null);
                 writeAccessController.enableWriting();
                 log.info("Acquired lease {}", leaseId);
-            } catch (StorageException | IOException e) {
+            } catch (Exception e) {
                 if (ex == null) {
                     log.info("Can't acquire the lease. Retrying every 1s. Timeout is set to {}s.", timeoutSec);
                 }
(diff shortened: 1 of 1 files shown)


oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureRepositoryLock.java
```diff
@@ -137,17 +137,23 @@ private void refreshLease() {
...
                     writeAccessController.disableWriting();
                 }
 
-                if (Set.of(StorageErrorCodeStrings.OPERATION_TIMED_OUT, StorageErrorCode.SERVICE_INTERNAL_ERROR, StorageErrorCodeStrings.SERVER_BUSY, StorageErrorCodeStrings.INTERNAL_ERROR).contains(e.getErrorCode())) {
-                    log.warn("Could not renew the lease due to the operation timeout or service unavailability. Retry in progress ...", e);
-                } else if (e.getHttpStatusCode() == Constants.HeaderConstants.HTTP_UNUSED_306) {
-                    log.warn("Client side error. Retry in progress ...", e);
+                StorageException storageException = e instanceof StorageException ? (StorageException) e : null;
+
+                if (storageException != null
```
t-rana: can we wrap the entire block in  a single storage storageException != null. Something like this

```
                if (e instanceof StorageException) {
                    StorageException storageException = (StorageException) e;
                    if (Set.of(StorageErrorCodeStrings.OPERATION_TIMED_OUT,
                            StorageErrorCode.SERVICE_INTERNAL_ERROR,
                            StorageErrorCodeStrings.SERVER_BUSY,
                            StorageErrorCodeStrings.INTERNAL_ERROR).contains(storageException.getErrorCode())) {
                        log.warn("Could not renew the lease due to the operation timeout or service unavailability. Retry in progress ...", e);
                    } else if (storageException.getHttpStatusCode() == Constants.HeaderConstants.HTTP_UNUSED_306) {
                        log.warn("Client side error. Retry in progress ...", e);
                    } else {
                        log.warn("Could not renew lease due to storage exception. Retry in progress ... ", e);
                    }
                } else {
                    log.error("Can't renew the lease", e);
                    shutdownHook.run();
                    doUpdate = false;
                    return;
                }
```
               
               
Looks a bit cleaner to me. 
The PR looks good to me.

ahanikel: Your solution is much better because it also keeps retrying if the StorageException is not of the kind that are being checked for. In the current state, we could also be giving up on a StorageException that falls through the ifs. Perfect!

smiroslav (approved):

ahanikel:
Thanks @smiroslav ! Could you also merge it? (I don't have permission)

PR #1687, opened 2024-09-04
OAK-11078: Remove usage of Guava checkArgument

reschke:
Before proceeding with this, I'd love to get some feedback.

JDK lacks this feature. We can either use something identical from commons-lang (which we use anyway, and which I did here), or roll our own.

Opinions?

(note that I also took the freedom to fix "{}" in format strings, and to avoid string concatenation)

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
@@ -18,7 +18,6 @@
  */
 package org.apache.jackrabbit.oak.plugins.index;
 
-import static org.apache.jackrabbit.guava.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
 import static org.apache.jackrabbit.guava.common.base.Throwables.getStackTraceAsString;
 import static org.apache.jackrabbit.guava.common.collect.Sets.newHashSet;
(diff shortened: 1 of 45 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -244,8 +244,8 @@ public static String checkValidName(String asyncName){
         if (IndexConstants.ASYNC_REINDEX_VALUE.equals(asyncName)){
             return asyncName;
         }
-        checkArgument(asyncName.endsWith("async"), "async name [%s] does not confirm to " +
-                "naming pattern of ending with 'async'", asyncName);
+        Validate.isTrue(asyncName.endsWith("async"), "async name [%s] does not confirm to naming pattern of ending with 'async'",
```
mbaedke, line 247: s/confirm/conform

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/JournalEntry.java
```diff
@@ -700,7 +700,7 @@ private static final class TreeNode {
         }
 
         TreeNode(MapFactory mapFactory, TreeNode parent, String name) {
-            checkArgument(!name.contains("/"),
+            Validate.isTrue(!name.contains("/"),
```
mbaedke, line 703: Does "{}" even work in a format string?

mbaedke (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/JournalEntry.java
reschke: No, according to the documentation (even in checkArgument)

so yes, this needs to be fixed.

rishabhdaim (approved):

PR #1688, opened 2024-09-05
OAK-10741 : updated logs for fullGc

rishabhdaim:


Joscorbe:
Not sure if enabling this in general would be a good option, I worry it could be very verbose when running any other oak-run command. Wouldn't be better if we add a new argument like `--debug` to fullGC commands that enable the logging?

rishabhdaim:
@Joscorbe I have updated the log to `INFO` for fullGc in https://github.com/apache/jackrabbit-oak/pull/1688/commits/b39c13f75a84a38b4f5b4949e1682b1d23677cf3

Joscorbe (approved):

PR #1690, opened 2024-09-05
OAK-11073: add Set.of(varargs) and stream support

reschke:


oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/collections/CollectionUtilsTest.java
```diff
@@ -64,4 +64,10 @@ public void iteratorToSet() {
 
         Assert.assertEquals(s, CollectionUtils.toSet(iterable.iterator()));
     }
+
+    @Test
+    public void arrayToSet() {
+        final Set<String> s = CollectionUtils.toSet(data);
+        Assert.assertEquals(s, CollectionUtils.toSet(data.toArray()));
```
thomasmueller, line 75: Doesn't this test case basically call the same method twice? If we change it to return an empty set, then the test won't fail

thomasmueller (changes requested):

thomasmueller (commented):
Ah I think I misread... there are two "toSet" methods.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/collections/CollectionUtils.java
```diff
@@ -81,4 +81,17 @@ public static <T> Set<T> toSet(final Iterator<T> iterator) {
...
     }
 
+    /**
+     * Convert a vararg list of items to a set.  The returning set is mutable and supports all optional operations.
+     * @param elements elements to convert
+     * @return the set
+     * @param <T> the type of the elements
+     */
+    public static <T> Set<T> toSet(@SuppressWarnings("unchecked") final T... elements) {
+        final Set<T> result = new HashSet<>();
```
reschke: Should we specify the size for the HashSet?

reschke:
https://github.com/apache/jackrabbit-oak/pull/1690/files#r1747561226 -> @thomasmueller , @rishabhdaim

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/collections/CollectionUtils.java
rishabhdaim: We should also handle cases where `elements` are passed a `null`.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/collections/CollectionUtils.java
```diff
@@ -81,4 +81,17 @@ public static <T> Set<T> toSet(final Iterator<T> iterator) {
         return result;
     }
 
+    /**
+     * Convert a vararg list of items to a set.  The returning set is mutable and supports all optional operations.
+     * @param elements elements to convert
+     * @return the set
+     * @param <T> the type of the elements
+     */
+    public static <T> Set<T> toSet(@SuppressWarnings("unchecked") final T... elements) {
```
rishabhdaim: We remove the redundant suppression for `unchecked` and should add `@SafeVarargs` to method.

oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/collections/CollectionUtilsTest.java
rishabhdaim: Add a test case for the `null` element.

This throws a surprise result.
If we pass `null` to `toSet(T... elements)`, it gives a compilation error due to an ambiguous method call since we have 2 other methods with the same name.

On passing `null` after type casting to `Object` as `toSet((Object)null)`, it returns a `set` with one `null` element.

`@Test
    public void nullToSet() {
        Assert.assertEquals(new HashSet<>(), CollectionUtils.toSet((Object) null));
    }`

The above fails whereas (below) if we typecast it to `Object[]` passes. 
It returns the correct value of an empty set.

`@Test
    public void nullToSet() {
        Assert.assertEquals(new HashSet<>(), CollectionUtils.toSet((Object[]) null));
    }`

rishabhdaim (changes requested):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/collections/CollectionUtils.java
reschke: Well, it'll be a NPE. I actually do not like the additional checks; they provide zero additional information, unless they come with a message.

oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/collections/CollectionUtilsTest.java
reschke: is an empty set really desirable? Shouldn't we throw an NPE in both cases?

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/collections/CollectionUtils.java
```diff
@@ -81,4 +83,69 @@ public static <T> Set<T> toSet(final Iterator<T> iterator) {
...
-}
+    /**
+     * Convert a vararg list of items to a set.  The returning set is mutable and supports all optional operations.
+     * @param elements elements to convert
+     * @return the set
+     * @param <T> the type of the elements
+     */
+    @SafeVarargs
+    public static <T> Set<T> toSet(final T... elements) {
+        final Set<T> result = new HashSet<>(elements.length);
```
nfsantos: Here you are specifying the `capacity` of the set, which in case of HashSet and HashMap works in a different way than arrays. The HashSet will ensure that `numberOfElements < capacity*loadFactor`. So setting the capacity to the number of elements that will be inserted here, guarantees that there will be a resizing operation. So it is probably better to set capacity to something like `elements.length * 1.25` (the default load factor is 0.75)

reschke: Let's just get rid of the argument.

nfsantos: I'm not sure about the performance implications of using load factor of 1, I guess it will increase collisions which may degrade performance, of both inserts and lookups. As the returned map is mutable, we can expect that it will modified, so I think it's safer to keep the default load factor, which I guess was carefully chosen by very clever people to provide good performance. If we create the map with load factor of 1 and then it is used for a performance critical operation, we risk that the map will perform more poorly than if it had been created with the default load factor.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/collections/CollectionUtils.java
```diff
@@ -75,10 +86,87 @@ public static <T> Set<T> toSet(final Iterable<T> iterable) {
...
+     * @return the set
+     * @param <T> the type of the elements
+     */
+    @SafeVarargs
+    @NotNull
+    public static <T> Set<T> toSet(@NotNull final T... elements) {
+        Objects.requireNonNull(elements);
+        // make sure the set does not need to be resized given the initial content
+        float loadFactor = (float) 0.75; // HashSet default
+        int initialCapacity = 1+ (int) (elements.length / loadFactor);
```
nfsantos:
```suggestion
        int initialCapacity = 1 + (int) (elements.length / loadFactor);
```

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/collections/CollectionUtils.java
```diff
@@ -75,10 +86,87 @@ public static <T> Set<T> toSet(final Iterable<T> iterable) {
...
+            @Override
+            public Iterator<T> iterator() {
+                if (consumed) {
+                    throw new IllegalStateException("Iterator already returned once");
+                } else {
+                    consumed = true;
+                    return iterator;
+                }
+            }
+        };
```
nfsantos, line 147: This is not thread safe, it can be consumed two or more times under concurrent access. If we don't want to make it thread safe, would be better to mention it in the Java docs.

reschke: will do

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/collections/CollectionUtils.java
```diff
@@ -75,10 +86,91 @@ public static <T> Set<T> toSet(final Iterable<T> iterable) {
...
+        for (T element : elements) {
+            result.add(element);
+        }
+        return result;
+    }
+
+    /**
+     * Convert an {@code Iterator} to an {@code Iterable}.
+     * <p>
+     * This method is not thread-safe
```
nfsantos, line 120: This is not accurate. The method is thread-safe, it can be called concurrently without any problems. But the Iterable returned is not thread safe. It's a really minor issue.

nfsantos (approved):

PR #1691, opened 2024-09-06
OAK-11084 - Improve error handling and logging in mongo downloader

nfsantos:
When one of the mongo download threads finishes and cancels the other one, in many cases the other thread would log an ERROR level log message. This is wrong because in this case the cancellation is a normal process, so it should not generate ERROR or even WARN log messages. Changes the logic to only log an INFO level message when the download thread is cancelled. 

Additional change: fix tracking of maximum size of Persisted Linked list buffers. With the previous logic, the maximum values could go be set to lower values.

### DIFF
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreIterator.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreIterator.java
@@ -121,6 +121,16 @@ class FlatFileStoreIterator extends AbstractIterator<NodeStateEntry> implements 
     private NodeStateEntry computeNextEntry() {
         if (!buffer.isEmpty()) {
             if (buffer.size() > maxBufferSize || buffer.estimatedMemoryUsage() > maxBufferSizeBytes) {
+                // We need to take the max of the current max and of the current value, because the two metrics above
+                // can and will increase and decrease independently. For instance, consider the following
+                //buffer.size() = 10, buffer.estimatedMemoryUsage() = 1000
+                //10 elements of size 1000 in total.
+                //
+                //And then
+                //buffer.size() = 1, buffer.estimatedMemoryUsage() = 2000
+                //1 element of size 2000.
+                //
+                //So in this case, we want to update the max memory estimate but leave the max size at 10.
                 maxBufferSize = Math.max(buffer.size(), maxBufferSize);
                 maxBufferSizeBytes = Math.max(buffer.estimatedMemoryUsage(), maxBufferSizeBytes);
                 LOG.info("Max buffer size changed {} (estimated memory usage: {} bytes) for path {}",
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreIterator.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreIterator.java
@@ -123,14 +123,11 @@ class FlatFileStoreIterator extends AbstractIterator<NodeStateEntry> implements 
             if (buffer.size() > maxBufferSize || buffer.estimatedMemoryUsage() > maxBufferSizeBytes) {
                 // We need to take the max of the current max and of the current value, because the two metrics above
                 // can and will increase and decrease independently. For instance, consider the following
-                //buffer.size() = 10, buffer.estimatedMemoryUsage() = 1000
-                //10 elements of size 1000 in total.
                 //
-                //And then
-                //buffer.size() = 1, buffer.estimatedMemoryUsage() = 2000
-                //1 element of size 2000.
+                // buffer.size() = 10, buffer.estimatedMemoryUsage() = 1000  - 10 elements using 1000 bytes of memory.
+                // buffer.size() = 1, buffer.estimatedMemoryUsage() = 2000   - 1 element using 2000 bytes of memory.
                 //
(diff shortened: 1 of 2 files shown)


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreIterator.java
```diff
@@ -121,8 +121,8 @@ protected NodeStateEntry computeNext() {
     private NodeStateEntry computeNextEntry() {
         if (!buffer.isEmpty()) {
             if (buffer.size() > maxBufferSize || buffer.estimatedMemoryUsage() > maxBufferSizeBytes) {
-                maxBufferSize = buffer.size();
-                maxBufferSizeBytes = buffer.estimatedMemoryUsage();
+                maxBufferSize = Math.max(buffer.size(), maxBufferSize);
```
steffenvan, line 131: Im not against this change, but I'm wondering if it strictly necessary? Seems like the `bufferSize` will be larger than `maxBufferSize` if we enter this condition either way.

nfsantos: We enter this condition if either `buffer.size() > maxBufferSize` or `buffer.estimatedMemoryUsage() > maxBufferSizeBytes`. And these two conditions are independent. For instance, we can have the following sequence:

`buffer.size() = 10, buffer.estimatedMemoryUsage() = 1000`
10 elements of size 1000 in total.

And then 
`buffer.size() = 1, buffer.estimatedMemoryUsage() = 2000`
1 element of size 2000.

So in this case, we want to update the max memory estimate but leave the max size at 10.

thomasmueller: Yes the change makes sense.

We could also have to "if" statements and two updates methods (without max), but then we would need two log statements. Your change is a bit shorter (even thought, harder to understand).

But the old behavior looks incorrect.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -547,15 +547,9 @@ private Future<?> submitDownloadTask(ExecutorCompletionService<Void> executor, D
             try {
                 downloadTask.download(mongoFilter);
                 downloadTask.reportFinalResults();
-            } catch (InterruptedException | MongoInterruptedException e) {
-                LOG.info("Thread interrupted: {}", e.toString());
+            } catch (Throwable e) {
+                LOG.warn("Error during download: {}", e.toString());
```
thomasmueller, line 551: Technically, with InterruptedException we should set the interrupt flag?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -682,8 +677,15 @@ private void download(Bson mongoQueryFilter) throws InterruptedException, Timeou
                             downloadRange(nextRange, mongoQueryFilter, downloadOrder);
                         }
                         downloadCompleted = true;
+                    } catch (IllegalStateException | InterruptedException | MongoInterruptedException e) {
+                        if (cancelled) {
```
thomasmueller, line 681: Same here with the interrupt flag, https://www.baeldung.com/java-interrupted-exception

thomasmueller (approved):
I'm not blocking the PR, just wondering if we should would need to set the interrupt flag... I _guess_ it's not needed.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileStoreIterator.java
nfsantos: With the two if statements, to avoid logging twice for the same entry, I would need to add a boolean flag to keep track if we already logged or not. So I opted to the current logic with a single if. It is indeed harder to understand. I will add a comment.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
nfsantos: My understanding is that we only need to set the interrupt flag again after catching an InterruptedException if we do not propagate an exception that signals that the thread should stop. In this particular case, the code throws a RuntimeException, wrapping the InterruptedException, which will stop the task. So the interrupt already fulfilled its task, which was to stop the thread.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
nfsantos: Here we have handled the interrupt by stopping the task, so I don't think we need to propagate the interrupt. Can you think of any case where it may be necessary to propagate? The task will terminate after the return below.

steffenvan (approved):

PR #1694, opened 2024-09-06
OAK-11087: Remove usage of Guava Predicates.in()

reschke:


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/aggregate/AggregationCursor.java
```diff
@@ -28,7 +28,6 @@
 import org.apache.jackrabbit.oak.spi.query.IndexRow;
 import org.apache.jackrabbit.oak.spi.query.QueryIndex;
 import org.apache.jackrabbit.oak.spi.state.NodeState;
-
```
thomasmueller, line 31: So the patch doesn't remove the import just yet...

thomasmueller (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/aggregate/AggregationCursor.java
reschke: it might be used for a diffent method (such as "not()"), but I'll double-check

PR #1699, opened 2024-09-06
OAK-11091: Make Segment#readByte/readBytes public

Nicolapps:
These methods are currently package-private. Since they can be helpful when writing unit tests that depend on the Segment class, I’m proposing to make them public instead.

Closes OAK-11091


mreutegg (approved):
Looks good to me.

CI failure in oak-run-elastic looks unrelated.

PR #1700, opened 2024-09-06
OAK-11094: Allow creating Segment with a provided SegmentData

Nicolapps:
All currently existing constructors of `Segment` instantiate a `SegmentData` instance from the provided data buffer (either `byte[]` or `org.apache.jackrabbit.oak.commons.Buffer`)

In some cases where we would like to use a custom implementation of `SegmentData`, it could be helpful to have a constructor that takes a `SegmentData` instance instead.

This PR also makes `SegmentReferences` and `RecordNumber` public so that the constructor I’m adding can also be public.

Closes OAK-11094

### DIFF
--- a/oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/Segment.java
+++ b/oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/Segment.java
@@ -217,6 +217,21 @@ public class Segment {
         }
     }
 
+    public Segment(
+            @NotNull SegmentReader reader,
+            @NotNull SegmentId id,
+            @NotNull SegmentData data,
+            @NotNull RecordNumbers recordNumbers,
+            @NotNull SegmentReferences segmentReferences
+    ) {
+        this.reader = requireNonNull(reader);
+        this.id = requireNonNull(id);
+        this.data = requireNonNull(data);
+        this.recordNumbers = requireNonNull(recordNumbers);
+        this.segmentReferences = requireNonNull(segmentReferences);
+        this.version = LATEST_VERSION;
+    }
+
     private static String toHex(byte[] bytes) {
         ByteArrayOutputStream out = new ByteArrayOutputStream();
         try {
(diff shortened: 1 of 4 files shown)


oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/Segment.java
```diff
@@ -217,6 +217,21 @@ public String toString() {
...
+            @NotNull SegmentData data,
+            @NotNull RecordNumbers recordNumbers,
+            @NotNull SegmentReferences segmentReferences
+    ) {
+        this.reader = requireNonNull(reader);
+        this.id = requireNonNull(id);
+        this.data = requireNonNull(data);
+        this.recordNumbers = requireNonNull(recordNumbers);
+        this.segmentReferences = requireNonNull(segmentReferences);
+        this.version = LATEST_VERSION;
```
mreutegg: I think the version should be read from SegmentData.
```suggestion
        this.version = SegmentVersion.fromByte(data.getVersion());
```
Or is there a specific reason why LATEST_VERSION must be used here? Can you also please add a test for this constructor. E.g. verifying the version is as expected.

Nicolapps: I agree, I updated my PR. I also added unit tests for the constructor.

oak-segment-tar/src/test/java/org/apache/jackrabbit/oak/segment/SegmentTest.java
```diff
@@ -0,0 +1,55 @@
...
+        var segmentReader = mock(SegmentReader.class);
+        var segmentId = mock(SegmentId.class);
+        var segmentData = mock(SegmentData.class);
+        when(segmentData.getVersion()).thenReturn((byte) 12);
+        var recordNumbers = mock(RecordNumbers.class);
+        var segmentReferences = mock(SegmentReferences.class);
+
+        var segment = new Segment(segmentReader, segmentId, segmentData, recordNumbers, segmentReferences);
+
+        assertEquals(segment.getSegmentVersion(), SegmentVersion.V_12);
```
mreutegg: Please swap the arguments. The first argument is the expected value.

mreutegg (changes requested):

oak-segment-tar/src/test/java/org/apache/jackrabbit/oak/segment/SegmentTest.java
Nicolapps: Fixed it, thanks!

mreutegg (approved):
Looks good to me.

CI failures seem unrelated to this change.

PR #1704, opened 2024-09-09
OAK-11097: remove remaining use of Guava predicates

reschke:


oak-it/src/test/java/org/apache/jackrabbit/oak/composite/CompositeNodeStoreTest.java
```diff
@@ -652,12 +650,14 @@ public void duplicatedChildren() throws CommitFailedException {
         deepMountBuilder.child("new").setProperty("store", "deepMounted", Type.STRING);
         deepMountedStore.merge(deepMountBuilder, EmptyHook.INSTANCE, CommitInfo.EMPTY);
 
-        List<ChildNodeEntry> children = newArrayList(filter(store.getRoot().getChildNodeEntries(), compose(Predicates.equalTo("new"), GET_NAME::apply)));
+        List<ChildNodeEntry> children = newArrayList(filter(store.getRoot().getChildNodeEntries(),
+                x -> Objects.equals(x == null ? null : x.getName(), "new")));
```
nfsantos, line 654: I find this confusing. Suggested alternative:
```
x -> x != null && Objects.equals(x.getName(), "new")
```

nfsantos (approved):

oak-it/src/test/java/org/apache/jackrabbit/oak/composite/CompositeNodeStoreTest.java
reschke: I *believe* we need "equals(null, null) -> true".

nfsantos: I think in this case the right-hand side is always "new", so if x == null, it will be false. Anyway, not too important, no need to change anything, but the fact that we are arguing over this line of code is a good indication that it is confusing.

PR #1705, opened 2024-09-11
OAK-11107 Index statistics support for multi-threaded indexing

thomasmueller:


nfsantos (approved):
> The use of AtomicLong will add some overhead compared to non atomic variables. For indexes that require little work per entry, this overhead might be significant. 

According to my test, System.nanoTime() takes twice the time of AtomicLong.addAndGet(): 17 ns vs 7 ns per call. (Both on Mac OS / Apple M1.) Both overheads, compared to what indexing is doing, seem insignificant to me.

PR #1706, opened 2024-09-11
OAK-11101 : removed usage of guava ImmutableSet.of()

rishabhdaim:


reschke (approved):
Wow, this is huge.

I assume you replace with "Set.of()" throughout, except in places where null can occur?

If this passes tests (which I'll take granted) please proceed.

rishabhdaim:
Thanks for review @reschke. Both `ImmutableSet` & `Set.of()` don't allow `null` values.
But the Set created after `Set.of()` doesn't even allow `null` inside `Set.contains()` method whereas `ImmutableSet` allows.

For that, I added `null` checks before `contains` where we have the possibility of `null` values.

rishabhdaim:
@reschke I have made some changes to add `null` checks before calling `Set.contains()` where there is a potential risk of NPE.

reschke:
Maybe then we shouldn't use "Set.of()" in the first place?

rishabhdaim:
The potential risk has been taken care of and I have added more unit cases to cover additional cases.

All unit cases are passing including the new ones.

We could decide to use something else if we are not confident enough.

PR #1707, opened 2024-09-11
OAK-11108 Tree store: support parallel indexing

thomasmueller:


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/tree/ParallelTreeStore.java
```diff
@@ -0,0 +1,103 @@
...
+                }
+                if (currentIterator == null) {
+                    return false;
+                }
+                return true;
+            }
+
+            @Override
+            public NodeStateEntry next() {
+                hasNext();
```
fabriziofortino: should we check the result of `hasNext` here?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/tree/ParallelTreeStore.java
```diff
@@ -0,0 +1,103 @@
...
+
+            @Override
+            public boolean hasNext() {
+                while (currentIterator != null && !currentIterator.hasNext()) {
+                    currentIterator = backend.nextSubsetIterator();
+                }
+                if (currentIterator == null) {
+                    return false;
+                }
+                return true;
```
fabriziofortino: this can be simplified with

```suggestion
                return currentIterator != null
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/tree/Prefetcher.java
```diff
@@ -93,12 +96,22 @@ public void setNodeReadAheadCount(long nodeReadAheadCount) {
...
 
-    public boolean shutdown() throws InterruptedException {
+    public boolean shutdown() {
+        closed.set(true);
         executorService.shutdown();
-        return executorService.awaitTermination(1, TimeUnit.SECONDS);
+        try {
+            return executorService.awaitTermination(10, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            LOG.info("Prefetcher termination was interrupted: {}", e.toString());
```
fabriziofortino, line 106: should we set the interrupt flag here by calling:

```
Thread.currentThread().interrupt();
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/tree/ParallelTreeStore.java
thomasmueller: You are right this is not obvious.
Update: yes actually this is needed!

fabriziofortino (approved):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/tree/ParallelTreeStore.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.apache.jackrabbit.oak.index.indexer.document.NodeStateEntry;
+import org.apache.jackrabbit.oak.index.indexer.document.indexstore.IndexStore;
+
+/**
+ * A wrapper around the tree store that only iterates over a subset of the
+ * nodes. Each parallel tree store for reads the next block (a block is a tiny
```
nfsantos:
```suggestion
 * nodes. Each parallel tree store reads the next block (a block is a tiny
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/tree/ParallelTreeStore.java
```diff
@@ -0,0 +1,105 @@
...
+ * subset of the nodes). Once it has finished iterating over the block, it asks
+ * the backend for the next block. The result is that the whole range is
+ * covered, but it doesn't matter much which thread is how fast.
```
nfsantos, line 34:
```suggestion
 * covered, but it doesn't matter how fast each thread is.
```

nfsantos (approved):
I think it's fine to explore parallelization as done here.

However, shared-state parallelization as done here is dangerous and tricky to get right. We have tried it before for the download phase, with very poor results (code was very complex, buggy and performed poorly). My recommendation would be to first try to split it into stages, as we did for the download. One stage creating the Lucene document, and another to write it to Lucene. This is safer, as most of the logic remains single-threaded, there would be just a queue in between the stages. From what I have seen so far in the metrics from the indexing phase, these two tasks (build the Lucene document and writing it to disk) take approximately the same time, so separating them in two separate stages of a pipeline would roughly cut in half the indexing phase, while avoiding all the difficulties of parallelization.
Once that is deployed and we collect new metrics of this functional split, I would look into optimizing the index generation, there may be opportunities to speed it up.

I would only consider parallelization as done here once we run out of other less complex and dangerous options and if the performance is still not good enough. I think having shared state is very dangerous and should be avoided as much as possible.

thomasmueller:
> I think it's fine to explore parallelization

Yes, I think the approach here has quite some limitations. I also think that using a queue would be better. 

On the other hand, "IndexWriter instances are completely thread safe" according to the documentation. Using multiple threads was tested (up to some point) before, and _not_ trying to do it that way seems like a missed opportunity, because it is relatively easy to implement. That's why I did this. But the benchmark results are not very promising: this only resulted in a speedup of about 50 when using 3 threads instead of one. I'm not quite sure yet why.

PR #1713, opened 2024-09-12
OAK-11112 : replaced Maps.newHashMapWithExpectedSize() with new HashM…

rishabhdaim:
…ap<>(initialCapacity)

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/GroupImpl.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/GroupImpl.java
@@ -18,7 +18,6 @@ package org.apache.jackrabbit.oak.security.user;
 
 import org.apache.jackrabbit.guava.common.base.Stopwatch;
 import org.apache.jackrabbit.guava.common.base.Strings;
-import org.apache.jackrabbit.guava.common.collect.Maps;
 import org.apache.jackrabbit.api.security.user.Authorizable;
 import org.apache.jackrabbit.api.security.user.Group;
 import org.apache.jackrabbit.api.security.user.UserManager;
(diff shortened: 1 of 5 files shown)


reschke (changes requested):
I believe for the size calculation we should java common utility method, or even implement newHash* directly in CollectionUtils. @nfsantos ?

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/collections/CollectionUtils.java
```diff
@@ -34,6 +36,8 @@
  */
 public class CollectionUtils {
 
+    private static final int MAX_CAPACITY = 1 << 30;
```
reschke: why do we need this?

rishabhdaim: This is the maximum capacity of `hashmap`. Any value above this would be reverted to this value.

Also, we might hit negative values if we don't specify an upper limit.

See test below:

`@Test
    public void ensureCapacityWithMaxValue() {
        int capacity = CollectionUtils.ensureCapacity(1908874369);
    }`

it returns `-2147483648`, which is wrong.

reschke (changes requested):
Could you please separate the changes for CollectionUtils from the cases where we *use* them?

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/collections/CollectionUtils.java
reschke: and maybe also add a comment with a short explanation?

rishabhdaim:
@reschke Created https://github.com/apache/jackrabbit-oak/pull/1722 for `ensureCapacity` method changes.

mbaedke (approved):
If we don't care that the maps will resize, it looks fine. Otherwise we might want to increase the initial capacity.

rishabhdaim:
@mbaedke we are using `CollectionUtils.newHashMap` API which creates a hashmap with the expected size to accommodate the expected number of elements without resizing.

mbaedke:
> @mbaedke we are using `CollectionUtils.newHashMap` API which creates a hashmap with the expected size to accommodate the expected number of elements without resizing.

Yes, of course. I was confused, sorry for the noise.

PR #1715, opened 2024-09-13
OAK-11111 Check if principal cache property exists

Amoratinos:
Added a condition to check if the principal cache value it's already present

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/CachedPrincipalMembershipReader.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/CachedPrincipalMembershipReader.java
@@ -131,17 +131,18 @@ class CachedPrincipalMembershipReader extends PrincipalMembershipReader {
                 cacheGroups(authorizableTree, groups);
             } else {
                 // another thread is already updating the cache, which leaves two options
-                // 1. serve a stale cache if allowed
+                // 1. serve a stale cache if allowed and if the cache values exist
                 // 2. read membership from membership-provider without caching the result
-                if (canServeStaleCache(expirationTime, now)) {
+                if (canServeStaleCache(expirationTime, now) && hasCacheValues(principalCache)) {
                     // the cache cannot be updated by the current thread but a stale cache can be returned and reading
                     // membership again can be avoided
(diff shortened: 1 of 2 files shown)


anchela (approved):
the fix looks good to me.
there is one thing, i would suggest to take another look at (maybe in a separate PR): there is a threshold for the number of membership required to have the cache written on line 209:

`groupPrincipals.size() <= MEMBERSHIP_THRESHOLD`

the default is 0 but with the current code, this means that 'no group membership' would also result in no cache being written.... i am wondering if it would be better to change this to 
`groupPrincipals.size() < MEMBERSHIP_THRESHOLD`
and make sure that in the default setup we always get the cache property written.

regarding the test:
the newly added verification
`verify(systemRoot, times(1)).commit(CacheValidatorProvider.asCommitAttributes());`
seems to indicates that we have no test coverage for the situation when the cache is not present and cannot be written. if my reading is correct, that might be valuable to cover as well. 

wdyt?
otherwise we are good.

Amoratinos:
>and make sure that in the default setup we always get the cache property written.

The `MEMBERSHIP_THRESHOLD` cannot be changed by configuration so changing to `< MEMBERSHIP_THRESHOLD`  will have the same effect like remove the condition, which is also fine for me.  

> we have no test coverage for the situation when the cache is not present and cannot be written

There's a test when the cache cannot be written or an exception is triggered. 


https://github.com/apache/jackrabbit-oak/blob/fd78e686ffce2369766519964a3c42d9ff3e03cf/oak-core/src/test/java/org/apache/jackrabbit/oak/security/user/CachedPrincipalMembershipReaderTest.java#L138-L150

anchela:
@Amoratinos , thanks. i think we are good.not sure why the checks failed. i will build your branch locally and merge it if there are no issues. hope that's ok.

anchela:
@Amoratinos , i get a build failure due to lack of branch coverage:

[WARNING] Rule violated for package org.apache.jackrabbit.oak.security.user: branches covered ratio is 0.98, but expected minimum is 0.99

the new method hasCacheValues(@NotNull Tree principalCache) only has 50% branch coverage and the condition if (canServeStaleCache(expirationTime, now) && hasCacheValues(principalCache)) also doesn't have full coverage. i guess this the reason for the drop of coverage.

once the build passes i will merge it.

PR #1716, opened 2024-09-13
OAK-11114 - Filter downloaded Mongo documents by path suffix

nfsantos:
Adds support to the indexing job to filter documents downloaded from Mongo early in the processing pipeline, more precisely at the NodeDocumentCodec, when they are being decoded from a binary stream of tokens. The motivation is to reduce the work that the indexing job has to do and the size of the intermediate flat file store, by not processing the documents that are not necessary for indexing, but that cannot be easily filtered using Mongo regex filtering.

This PR adds two system configuration properties:
- `oak.indexer.pipelined.nodeDocument.filter.filteredPath` - The path where the filter should be applied. Documents outside this path are not considered for filtering.  
- `oak.indexer.pipelined.nodeDocument.filter.suffixesToSkip` - The suffix of the documents that should be excluded.


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/ConfigHelper.java
```diff
@@ -48,4 +52,11 @@ public static boolean getSystemPropertyAsBoolean(String name, boolean defaultVal
         LOG.info("Config {}={}", name, value);
         return value;
     }
+
+    public static List<String> getSystemPropertyAsStringList(String name, String defaultValue, String separator) {
+        String result = System.getProperty(name, defaultValue);
+        List<String> parts = result.isBlank() ? List.of() : Arrays.stream(result.split(separator)).map(String::trim).collect(Collectors.toList());
```
thomasmueller: What about to methods: `getSystemProperty` and `splitIgnoreSpace`? I see the code is quite lenient in what it accepts (isBlank vs isEmpty, trim). For user-facing configuration, I think it's the right thing to do; we ran into problems because people added spaces eg. "type" = "lucene " (trailing space). 

Having a separate method `splitIgnoreSpace` would be quite useful I think... it would also simplify adding a test case just for this part.

(I think it is not necessary in this case because this is not end-user facing but only for internal use. For internal use, I think non-lenient behavior is better because it results in shorter, easier to read code. But it's not that bad here, so I think it's fine.)

But more importantly, what if spaces have meaning? The method name `getSystemPropertyAsStringList` doesn't convey what it does, and there is no Javadoc.

What about renaming "separator" -> "separatorRegex", or then quote the separator (using `Pattern.quote` I think)?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeDocumentCodec.java
```diff
@@ -49,44 +56,107 @@
...
         this.stringCoded = (Codec<String>) bsonTypeCodecMap.get(BsonType.STRING);
         this.longCoded = (Codec<Long>) bsonTypeCodecMap.get(BsonType.INT64);
         this.booleanCoded = (Codec<Boolean>) bsonTypeCodecMap.get(BsonType.BOOLEAN);
     }
 
+    /**
+     * Skipping over values in the BSON file is faster than reading them. Skipping is done by advancing a pointer in
+     * an internal buffer, while reading requires converting them to a Java data type (typically String).
+     */
+    private void skipUntilEndOfDocument(BsonReader reader) {
```
thomasmueller, line 111: To avoid the duplicate line:
```suggestion
    private void skipUntilEndOfDocument(BsonReader reader) {
        while (reader.readBsonType() != BsonType.END_OF_DOCUMENT) {
            reader.skipName();
            reader.skipValue();
        }
        reader.readEndDocument();
    }
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeDocumentFilter.java
```diff
@@ -0,0 +1,115 @@
...
+        if (filteringDisabled) {
+            return false;
+        }
+        // Check if the NodeDocument should be considered for filtering, that is, if it starts with includePath.
+        // If the value is for an _id, then we must find the start of the path section, that is, the position of the first
+        // slash (3:/foo/bar/baz). If the value given is for a path, then it already contains only the path. In any case,
+        // we look up for the first occurrence of /
+        int idxOfFirstForwardSlash = idOrPathValue.indexOf('/');
+        if (idxOfFirstForwardSlash < 0) {
+            LOG.info("Invalid field. {} = {}", fieldName, idOrPathValue);
```
thomasmueller: Using info level and the message "invalid" are somewhat conflicting... Did you see such a case already? If not, I would suggest to use "warn". If you did, then what field / what was the reason? If it can happen and we know in which (rare) cases, then I wonder why a log message at all...

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -786,6 +786,10 @@ void download(FindIterable<NodeDocument> mongoIterable) throws InterruptedExcept
                 try {
                     while (cursor.hasNext()) {
                         NodeDocument next = cursor.next();
+                        // If the id is not set, then the document was filtered by NodeDocumentFilter and should be ignored
+                        if (next.getId() == null) {
+                            continue;
+                        }
                         String id = next.getId();
```
thomasmueller: I would
```suggestion
                        String id = next.getId();
                        if (id == null) {
                            continue;
                        }                        
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeDocumentFilter.java
nfsantos: I increased the log level to warn. It should never happen, but in case it ever does, we should log it. I thought about throwing an exception, but that can be dangerous, as it could block a job from succeeding because of a single bad path.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeDocumentCodec.java
nfsantos: Done.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/ConfigHelper.java
nfsantos: To reduce the chance of errors, I changed separator to be a `char` and use a method from Commons Lang that splits using char, instead of regex. I created a separate method as private. Maybe in the future could be made made into an utility method, in case it does not already exist.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
nfsantos: Done

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeDocumentFilter.java
thomasmueller: Yes a warning makes sense.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/ConfigHelper.java
```diff
@@ -48,4 +53,18 @@ public static boolean getSystemPropertyAsBoolean(String name, boolean defaultVal
         LOG.info("Config {}={}", name, value);
         return value;
     }
+
+    /**
+     * white space at the start/end of the string or at the start/end of the parts delimited by separator are trimmed
```
thomasmueller, line 58: Yes this Javadoc helps here.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/ConfigHelper.java
```diff
@@ -48,4 +53,18 @@ public static boolean getSystemPropertyAsBoolean(String name, boolean defaultVal
...
+     */
+    public static List<String> getSystemPropertyAsStringList(String name, String defaultValue, char separator) {
+        String result = System.getProperty(name, defaultValue);
+        List<String> parts = splitString(result, separator);
+        LOG.info("Config {}={}", name, parts);
+        return parts;
+    }
+
+    private static List<String> splitString(String str, char separator) {
```
thomasmueller, line 67: Hm, I would have hoped to make it public as well, add a Javadoc, and specially a unit test... But OK, it's a short method, and the risk is low...

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeDocumentCodec.java
```diff
@@ -49,44 +56,109 @@
...
+        while (true) {
+            BsonType bsonType = reader.readBsonType();
+            if (bsonType == BsonType.END_OF_DOCUMENT) {
+                break;
+            }
+            reader.skipName();
+            reader.skipValue();
+        }
```
thomasmueller: Sorry, even simpler:

```suggestion
        while (reader.readBsonType() != BsonType.END_OF_DOCUMENT) {
            reader.skipName();
            reader.skipValue();
        }
```

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeDocumentCodec.java
```diff
@@ -49,44 +56,109 @@
...
     private final Codec<String> stringCoded;
     private final Codec<Long> longCoded;
     private final Codec<Boolean> booleanCoded;
 
+    private final NodeDocumentFilter fieldFilter = new NodeDocumentFilter(filteredPath, suffixesToSkip);
+
+    // Statistics
+    private final AtomicLong totalDocsDecoded = new AtomicLong(0);
+    private final AtomicLong totalDataDownloaded = new AtomicLong(0);
+    private final ThreadLocal<NodeDocumentDecoderContext> perThreadContext = ThreadLocal.withInitial(NodeDocumentDecoderContext::new);
```
thomasmueller, line 94: I'm not a very big fan of using ThreadLocal because of the problems they can cause. One is out-of-memory (in an non-obvious way), another is that some web servers, and possibly also OSGi, can try to reset them (using things like "setAccessible(true)")... Because if they don't, then the class (NodeDocumentDecoderContext here) can't be unloaded, and so the classloader can't be unloaded, and so you get out-of-memory that way.

I think here it is OK. But I wouldn't want to use ThreadLocal in other parts of Oak. Just FYI. See the comments here: https://stackoverflow.com/questions/60487341/are-there-any-general-drawbacks-from-using-java-lang-threadlocal

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/ConfigHelper.java
nfsantos: There are unit tests of `getSystemPropertyAsStringList`, which should be enough to cover this method as well.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeDocumentCodec.java
nfsantos: Yes, this version is better, simple and clear.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/ConfigHelper.java
thomasmueller: I didn't see those sorry. Great!

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeDocumentCodec.java
nfsantos: I see, it can be a memory leak if the thread is long running, right. In this case, the thread pool from where these threads are taken is shutdown at the end of the download phase, so I think it won't be a problem. Also this logic will not run inside OSGI or a servlet container. 
But I understand your concerns, in fact it is not a good practice in general. Here I used it in an attempt to decrease contention and to have a single long-lived per-thread context object. I did not measure if it has an impact on performance, but it seemed cleaner that each thread has its own context.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeDocumentCodec.java
```diff
@@ -49,44 +56,106 @@
...
+        NodeDocumentDecoderContext threadLocalContext = perThreadContext.get();
+        threadLocalContext.estimatedSizeOfCurrentObject = 0;
         reader.readStartDocument();
         while (reader.readBsonType() != BsonType.END_OF_DOCUMENT) {
             String fieldName = reader.readName();
-            Object value = readValue(reader, fieldName, estimatedSizeOfCurrentObject);
+            Object value = readValue(reader, fieldName, threadLocalContext);
+            // Once we read the _id or the _path, apply the filter
+            if (fieldName.equals(NodeDocument.ID) || fieldName.equals(NodeDocument.PATH)) {
+                if (fieldFilter.shouldSkip(fieldName, (String) value)) {
```
thomasmueller: It seems there is no risk here that value is null or not a String.

thomasmueller (approved):

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/ConfigHelperTest.java
alagodasii, line 1: It's non-functional change, but could improve structure, readability and extensibility of the test. I suggest parameterising it. JUnit 4 doesn't support it well, but still we have capabilities of doing so. Here's the proposed solution:

_package org.apache.jackrabbit.oak.index.indexer.document.flatfile.pipelined;

import static java.util.Collections.emptyList;
import static org.junit.Assert.assertEquals;

import org.junit.Rule;
import org.junit.Test;
import org.junit.contrib.java.lang.system.RestoreSystemProperties;
import org.junit.runner.RunWith;
import org.junit.runners.Parameterized;
import org.junit.runners.Parameterized.Parameters;
import org.junit.runners.Suite;

import java.util.Collection;
import java.util.List;

@RunWith(Suite.class)
@Suite.SuiteClasses({ConfigHelperTest.SystemPropertyExistsTest.class, ConfigHelperTest.MissingSystemPropertyTest.class})
public class ConfigHelperTest {

  @Rule
  public final RestoreSystemProperties restoreSystemProperties = new RestoreSystemProperties();


  @RunWith(Parameterized.class)
  public static class SystemPropertyExistsTest {

    private final String key;
    private final String value;
    private final List<String> expected;

    public SystemPropertyExistsTest(String key, String value, List<String> expected) {
      this.key = key;
      this.value = value;
      this.expected = expected;
    }

    @Parameters
    public static Collection<Object[]> data() {
      return List.of(new Object[][]{
          {"key1", "value1", singletonList("value1")},
          {"key2", " ", emptyList()},
          {"key3", "v1;v2", List.of("v1", "v2")},
          {"key4", "v1; v2", List.of("v1", "v2")}
      });
    }

    @Test
    public void givenDefaultValue_whenSystemPropertyExists_thenGetSystemPropertyAsString() {
      // given
      var defaultValue = "default";
      var separator = ';';
      System.setProperty(key, value);

      // when
      final var result = ConfigHelper.getSystemPropertyAsStringList(key, defaultValue, separator);

      // then
      assertEquals(expected, result);
    }

  }

  @RunWith(Parameterized.class)
  public static class MissingSystemPropertyTest {

    private final String defaultValue;
    private final List<String> expected;

    public MissingSystemPropertyTest(String defaultValue, List<String> expected) {
      this.defaultValue = defaultValue;
      this.expected = expected;
    }

    @Parameters
    public static Collection<Object[]> data() {
      return List.of(new Object[][]{
          {"", emptyList()},
          {"default", singletonList("default")},
          {"default1;default2", List.of("default1", "default2")}
      });
    }

    @Test
    public void givenDefaultValue_whenSystemPropertyNotExists_thenReturnDefaultValueAsList() {
      // given
      final var key = "not.defined";
      final var separator = ';';

      // when
      final var result = ConfigHelper.getSystemPropertyAsStringList(key, defaultValue, separator);

      // then
      assertEquals(expected, result);
    }
  }
}_

Also could you provide the scenario where separator differs from the one used in system property? I.e. `default1;defalut2` and `,`.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
```diff
@@ -796,7 +800,7 @@ void download(FindIterable<NodeDocument> mongoIterable) throws InterruptedExcept
                         this.lastIdDownloaded = id;
                         this.documentsDownloadedTotal++;
                         downloadStatics.incrementDocumentsDownloadedTotal();
-                        if (this.documentsDownloadedTotal % 20_000 == 0) {
+                        if (this.documentsDownloadedTotal % 50_000 == 0) {
```
alagodasii, line 803: Can you extract `50_000` to constant?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeDocumentCodec.java
nfsantos: There shouldn't be, but it's a good point. The filtering is a best-effort performance optimization, not needed for correctness, so it should never cause the download to fail. I added a type check to value (which also covers the null case) before calling the filter method.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/ConfigHelperTest.java
nfsantos: The proposed solution is much more verbose, I don't think it is justified in this case. I have instead extracted into a separate method the core of the test logic, which reduces repetition and makes the test more readable.

I added a few test cases where the separator in the value of the system property does not match the expected separator.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/NodeDocumentCodec.java
thomasmueller: Yes I think it is fine in this case.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedMongoDownloadTask.java
nfsantos: We are inlining similar constants in other places. Maybe it is not a good practise, but it's ok for the time being, I don't want to expose this as a constant.

Additionally, the whole approach to logging progress should be reconsidered. I no longer think that logging based on the number of items processed is the best approach, as download speeds vary a lot from run to run, so we risk either not logging often enough or logging way too frequently. A better approach is to log periodically, so at a certain point in the future I would like to revise the whole approach to logging progress in the indexing job.

thomasmueller: No, I think this is not needed. Actually I'm against making it a constant: that would just complicate the code without giving any meaningful value, in my view. (I understand others might have a different opinion, that's fine.)

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/ConfigHelperTest.java
```diff
@@ -0,0 +1,52 @@
...
+        test(List.of(), " ", "default", ';');
+        test(List.of("v1", "v2"), "v1;v2", "default", ';');
+        test(List.of("v1", "v2"), "v1; v2", "default", ';');
+        test(List.of("v1", "v2"), "v1,v2", "default", ',');
+        test(List.of("v1,v2"), "v1,v2", "default", ';');
+    }
+
+    private void test(List<String> expected, String sysPropertyValue, String defaultValue, char separator) {
+        System.setProperty("key", sysPropertyValue);
+        assertEquals(expected, ConfigHelper.getSystemPropertyAsStringList("key", defaultValue, separator));
```
fabriziofortino: minor:  I would cleanup the system property

```
System.clearProperty("key");
```

fabriziofortino (approved):

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/ConfigHelperTest.java
nfsantos: Done. I have also changed the key to a value that is less likely to be used by any other code.

PR #1722, opened 2024-09-17
OAK-11073 : added ensureCapacity method to create hash based collecti…

rishabhdaim:
…ons with expected size


oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/collections/CollectionUtils.java
```diff
@@ -34,6 +36,8 @@
  */
 public class CollectionUtils {
 
+    private static final int MAX_CAPACITY = 1 << 30;
```
reschke, line 41: maybe add a comment here so it's clear why it's needed

reschke (approved):
...but it might be good to explain the capacity limit in a comment

PR #1723, opened 2024-09-17
OAK-11100: remove use of Guava transform/filter

reschke:


nfsantos (approved):
Ok for oak-run-common and oak-run modules. I did not look at the other modules.

oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/DynamicSyncContext.java
```diff
@@ -234,7 +235,8 @@ private Map<ExternalIdentityRef, SyncEntry> collectSyncEntries(@NotNull Iterable
      */
     private void collectSyncEntries(@NotNull Iterable<ExternalIdentityRef> declaredGroupRefs, long depth, @NotNull Map<ExternalIdentityRef, SyncEntry> map) throws ExternalIdentityException, RepositoryException {
         boolean shortcut = shortcut(depth);
-        for (ExternalIdentityRef ref : Iterables.filter(declaredGroupRefs, this::isSameIDP)) {
+        Iterable<ExternalIdentityRef> filtered = () -> CollectionUtils.toStream(declaredGroupRefs).filter(this::isSameIDP).iterator();
```
rishabhdaim, line 238: How about having a util for this transformation as well?

> @NotNull
>     public static <T> Iterable<T> toIterable(@NotNull Iterable<T> iterable, @NotNull Predicate<? super T> predicate) {
>         Objects.requireNonNull(iterable);
>         Objects.requireNonNull(predicate);
>         return () -> StreamSupport.stream(iterable.spliterator(), false).filter(predicate).iterator();
>     }

rishabhdaim (changes requested):

rishabhdaim:
Could you please revert changes to `ImmutableSet/List.copyOf()` from this PR.

oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/DynamicSyncContext.java
rishabhdaim: Another option could be to use `commons-collections` `IterableUtils.filteredIterable` and same goes for `Iterator`.

reschke: Using the first proposal wouldn't work well when mutiple stream manipulations are changed.

PR #1724, opened 2024-09-18
OAK-11130 - indexing-job: Improvements to logging of AOT Blob downloader

nfsantos:
Reduce frequency of logging messages.
Log text extraction cache statistics only once.

### DIFF
--- a/oak-run-elastic/src/main/java/org/apache/jackrabbit/oak/index/ElasticDocumentStoreIndexer.java
+++ b/oak-run-elastic/src/main/java/org/apache/jackrabbit/oak/index/ElasticDocumentStoreIndexer.java
@@ -18,8 +18,11 @@
  */
 package org.apache.jackrabbit.oak.index;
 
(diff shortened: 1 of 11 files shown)


steffenvan (approved):

oak-run-elastic/src/main/java/org/apache/jackrabbit/oak/index/ElasticDocumentStoreIndexer.java
```diff
@@ -18,8 +18,11 @@
  */
 package org.apache.jackrabbit.oak.index;
 
-import org.apache.jackrabbit.guava.common.collect.ImmutableList;
-import org.apache.jackrabbit.oak.index.indexer.document.*;
```
thomasmueller, line 22: It's good that we no longer need this... Star imports are problematic in my view.

thomasmueller (approved):

PR #1727, opened 2024-09-19
OAK-11113 -- fix flaky test methods

ionutzpi:
fix flaky tests

### DIFF
--- a/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
+++ b/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
@@ -230,7 +230,6 @@ public class MissingLastRevSeekerTest extends AbstractDocumentStoreTest {
     
     @Test
     public void getNonSplitDocs() throws Exception {
-        //assertTrue();
         String nodeName = this.getClass().getName() + "-foo";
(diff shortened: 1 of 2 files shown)


oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
```diff
@@ -248,6 +248,7 @@ public void dispose() {
             dns.merge(b1, EmptyHook.INSTANCE, CommitInfo.EMPTY);
         }
         dns.runBackgroundOperations();
+        Thread.sleep(1001);
```
fabriziofortino: This might still fail when running on nodes with low resources or add unneeded delays when the background operation completes quickly. In oak-search, we make extensive use of the following utility method that takes care of both issues:

https://github.com/apache/jackrabbit-oak/blob/3e315733e3932b77de9606271165190c2b106fa8/oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/TestUtil.java#L247

ionutzpi: Adding code with assertEventually(..) still fails on local, even if I add 100s as timeout. With no timeout it still fails.

rishabhdaim (approved):

fabriziofortino (approved):

PR #1731, opened 2024-09-23
OAK-10741 : added logging for skipping non-gap orphaned node

rishabhdaim:


reschke (approved):
...but the check for the LOG level may be unneeded.

PR #1734, opened 2024-09-23
OAK-11139 Allow downloading only recently changed nodes from MongoDB

thomasmueller:


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/MergeIncrementalTreeStore.java
```diff
@@ -85,14 +99,63 @@ public String getStrategyName() {
...
+                                line.path, line.value);
+                    }
+                    treeStore.putNode(line.path, line.value);
+                    break;
+                case UPSERT:
+                    upserted++;
+                    // upsert = insert or update
+                    treeStore.putNode(line.path, line.value);
+                    break;
+                case DELETE:
```
amit-jain, line 136: I am not really clear why are there two similar ops DELETE/REMOVE?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/MergeIncrementalTreeStore.java
```diff
@@ -67,16 +68,29 @@ public MergeIncrementalTreeStore(File baseFile, File incrementalFile, File merge
 
     @Override
     public void doMerge() throws IOException {
```
amit-jain, line 70: Test missing for this case where the TOPUP_FILE is added.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedTreeStoreIT.java
```diff
@@ -675,6 +675,7 @@ private PipelinedTreeStoreStrategy createStrategy(MongoTestBackend backend, Pred
                 pathPredicate,
```
amit-jain, line 746: The tests miss the case of modified date which I get is difficult to test. But can we add UT/ITs for cases like removeNode and the procedure using updateIndexStore above?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/MergeIncrementalTreeStore.java
thomasmueller: I'll document the difference!

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedTreeStoreIT.java
thomasmueller: Yes I'll try to add more tests!

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/IndexerSupport.java
```diff
@@ -76,12 +76,21 @@ public class IndexerSupport {
     private File indexDefinitions;
     private final String checkpoint;
     private File existingDataDumpDir;
+    private long minModified;
```
nit0906, line 84: this field is not self descriptive - maybe a comment here would help to point out that this is a timestamp ?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/FlatFileNodeStoreBuilder.java
```diff
@@ -191,6 +192,11 @@ public FlatFileNodeStoreBuilder withCheckpoint(String checkpoint) {
         return this;
     }
 
+    public FlatFileNodeStoreBuilder withMinModified(long minModified) {
```
nit0906, line 202: I would add a small comment here as well to describe what is the use for this field.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/IncrementalStoreOperand.java
```diff
@@ -18,10 +18,44 @@
...
+
+    // modify an existing node;
+    // log a warning if it doesn't exist
+    // (this is used by the 'diff' command)
     MODIFY("M"),
-    DELETE("D");
+
+    // remove a node that may or may not exist
+    // (this operation may be used by the 'top up' command)
+    REMOVE_IF_EXISTS("R"),
```
nit0906, line 53: The enum is REMOVE_IF_EXISTS but the description says remove a node that may or may not exist - should it not be - 

Remove a node if exists otherwise NOOP ? 

Also, I am not sure if it's clear to me how does it differ from DELETE ? That is also deleting and if the node is not present - then logging a warning.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/IndexerSupport.java
```diff
@@ -76,12 +76,21 @@ public class IndexerSupport {
...
         this.indexHelper = indexHelper;
         this.checkpoint = checkpoint;
     }
 
+    public long getMinModified() {
+        return minModified;
+    }
+
+    public void setMinModified(long minModified) {
+        this.minModified = minModified;
```
tihom88, line 96: [oak documentation](https://jackrabbit.apache.org/oak/docs/nodestore/documentmk.html) mentions:
`The _modified field contains an indexed low-resolution timestamp when the node was last modified. The time resolution is five seconds. This field is also updated when a branch commit modifies a node.
`
 If I am understanding it correctly anything updated till 4 sec will be set to 0 as per time resolution.
i.e  for a  14th second update  _modified property will be set to 10 sec.
For a  29th second update  _modified property will be set to 25 sec.

So to make sure that we don't miss any update 
I think minModified should be
`this.minModified = minModified - minModified/5`

@stefan-egli , @mreutegg am I correct?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/IndexerSupport.java
tihom88, line 84: Should we use lastModifed  instead of minModified as we are using _modified property of mongMK which denotes lastModified timestamp?

[Oak document](https://jackrabbit.apache.org/oak/docs/nodestore/documentmk.html):
`The _modified field contains an indexed low-resolution timestamp when the node was last modified.`

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/MergeIncrementalTreeStore.java
```diff
@@ -85,14 +99,67 @@ public String getStrategyName() {
...
+                case DELETE:
+                    deleted++;
+                    if (old == null) {
+                        LOG.warn(
+                                "DELETE: node {} doesn't exist",
+                                line.path);
+                    }
+                    treeStore.removeNode(line.path);
+                    break;
+                case REMOVE_IF_EXISTS:
```
amit-jain, line 145: The only difference from a behaviour point of view between the 2 modes is that there is warning logged for DELETE, which doesn't seem to warrant a new mode. But do you forsee different implementations going forward? In that case maybe it's ok to have these different

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/IndexerSupport.java
thomasmueller: "Last modified" refers to the last time a node was modified. It refers to the property "jcr:lastModified", which may or may not exist.

"Min modified" means the minimum value (lower bound) of the "_modified" property.

>  _modified property of mongMK which denotes lastModified timestamp

The "_modified" property does not denote the "jcr:lastModified" timestamp.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/MergeIncrementalTreeStore.java
thomasmueller: Well, that's the point: "REMOVE_IF_EXISTS" should not log a warning, but "DELETE" should.

I think both implementations are needed: one for the "diff" command (DELETE), and one for the "top-up" command (REMOVE_IF_EXISTS).

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/IndexerSupport.java
thomasmueller: minModified is the minimum (lower bound) of the "_modified" property.

thomasmueller: > anything updated till 4 sec will be set to 0

Anything between on January 1st, 1970 midnight (up to ~5 seconds later), will be set to 0.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/IncrementalStoreOperand.java
thomasmueller: Fixed.

> how does it differ from DELETE

It differs: DELETE logs a warning if the node doesn't exist, REMOVE_IF_EXISTS does not.

amit-jain (approved):
lgtm!

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/IndexerSupport.java
tihom88: yes, _modified is independent of jcr:lastModified. My point was to just add info about what this property maps to.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/incrementalstore/IncrementalStoreOperand.java
nit0906: But do we really need REMOVE_IF_EXISTS then - Is it just to avoid the warning ?

nit0906:
I have one query on the overall top up implementation - when getting the topup directly from mongo using the _modified filter - we would miss the deleted nodes - right ? 

Eventually this would lead to larger and larger index stores ? 

We filter the non-existant nodes in queries, so even if this won't effect query results - but would it still impact the performance ? Let's say a bulk delete on a large content repository just before the top up.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/IndexerSupport.java
thomasmueller: Yes, I understand now! Thanks!

thomasmueller:
> we would miss the deleted nodes - right ?

Only for those deleted nodes that the "diff" doesn't see. (My plan is to run _both_ the diff _and_ the top-up.)

> Eventually this would lead to larger and larger index stores ?

Until the bugs in the "diff" parts are fixed, yes. But the verification would still notice it, and so we could re-download everything once we see that it's too many.

> Let's say a bulk delete on a large content repository just before the top up.

If the bugs in the "diff" logic are not fixed, then I think the behavior is the same as it is now. This is because "diff" is also used for async index update.

If it turns out that the "diff" logic can not be fixed easily, then we could also change the download logic (pipelined strategy) to emit "removed" entries. But I think we don't need to implement this logic currently.

PR #1737, opened 2024-09-24
OAK-11142: org.apache.jackrabbit.oak.plugins.memory is for internal use only and should be marked as such

reschke:


stefan-egli:
What is the plan with these markings? I see we have like a dozen packages marked as internal - where I suspect that we probably have many more. IOW how does this one stand out from the others?

reschke:
If we find more, we should mark them as well.

The point here is to clarify what APIs that are used in Oak internally can be used from applications outside Oak.

For instance, we give no guarantee about API stability at all. We don't have to, as Oak is released as a monolithic thing.

stefan-egli (approved):

PR #1738, opened 2024-09-25
OAK-11131 - indexing-job: AOT Blob downloader may download blobs that are not needed for the indexes

nfsantos:
The AOT Blob downloader makes several checks on each line/node in the FFS before downloading an non-inlined blob. One of these checks is if the line/node is needed by the **indexers for which the feature is enabled**. The current implementation is **checking if any of the indexers requires the line**. This is incorrect, because it may download blobs for lines that are not required by the indexers with AOT blob download enabled. In the worst case, this will slow down significantly the downloader, which may make it fall behind the indexer, thereby preventing it from downloading the blobs for the lines that are actually needed by the indexers.

This PR changes the logic so that the AOT blob downloader to only check against the indexes for which AOT blob download is enabled when deciding if a line/node is to be considered for blob download or not.

This PR also adds additional tests for the AOT blob downloader.

### DIFF
--- a/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/CompositeIndexer.java
+++ b/oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/CompositeIndexer.java
@@ -30,6 +30,10 @@ import org.apache.jackrabbit.oak.plugins.document.NodeDocument;
 
 import static java.util.Objects.requireNonNull;
 
+/**
+ * Many methods in this class call themselves recursively, and are susceptible to infinite recursion if a composite
+ * indexer contains itself, directly or indirectly. In this case, the methods will throw a StackOverflowException.
+ */
 public class CompositeIndexer implements NodeStateIndexer {
 
     private final List<NodeStateIndexer> indexers;
@@ -89,8 +93,6 @@ public class CompositeIndexer implements NodeStateIndexer {
 
     @Override
     public String getIndexName() {
-        // This is susceptible to infinite recursion if a composite indexer contains itself, directly or indirectly.
-        // The same is true for several other methods in this class.
         return indexers.stream().map(NodeStateIndexer::getIndexName).collect(Collectors.joining(",", "CompositeIndexer[", "]"));
     }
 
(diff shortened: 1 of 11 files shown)


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/CompositeIndexer.java
```diff
@@ -86,6 +87,13 @@ public Set<String> getRelativeIndexedNodeNames() {
         return result;
     }
 
+    @Override
+    public String getIndexName() {
+        // This is susceptible to infinite recursion if a composite indexer contains itself, directly or indirectly.
```
thomasmueller: I like the comment (I wasn't aware of that), however in my view the comment is in the wrong location... Could you move it to the top of the class please (just before "public class CompositeIndexer")? Let me try to make a proposal...

thomasmueller:
```suggestion
        // Many methods are susceptible to infinite recursion if a composite indexer contains itself, directly or indirectly.
        // In this case, the methods would throw a StackOverflowException
```

oak-run-elastic/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/ElasticIndexer.java
```diff
@@ -126,6 +126,11 @@ public Set<String> getRelativeIndexedNodeNames() {
         return definition.getRelativeNodeNames();
     }
 
+    @Override
+    public String getIndexName() {
+        return definition.getIndexName();
```
thomasmueller, line 131: I think the usage of getIndexName() is fine. However, just FYI:

For on-prem and managed services, we support the same node name in multiple locations. For example it's possible to have 

* /oak:index/damAssetLucene
* /content/dam/oak:index/damAssetLucene

both in the same repository.

It is unlikely that this causes issues for this case here however (in my view).

thomasmueller (approved):

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/CompositeIndexer.java
nfsantos: Yes, I agree, it should be a warning in the public documentation of the class. The users of this class have to be aware to avoid this issue. I updated the comment.

oak-run-elastic/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/ElasticIndexer.java
nfsantos: `definition.getIndexName()` returns the full path of the index, so there is no risk of collisions for indexes that are outside `/oak:index`.

PR #1729, opened 2024-09-21
OAK-11073 : added new api in CollectionUtils to create HashSet with expected…

rishabhdaim:
… no. of elements

PR #1744, opened 2024-09-26
OAK-11153 : added utils to create LinkedHashSet/hashSet with expected…

rishabhdaim:
… capacity and linkedhashset from iterables


Joscorbe:
This seems identical to #1729 right?

rishabhdaim:
Yes, it is indeed identical to https://github.com/apache/jackrabbit-oak/pull/1729 but with different jira. Previous one was closed and released as part of 1.70.0 release.

reschke (approved):

PR #1745, opened 2024-09-27
Oak 11113 - MissingLastRevSeekerTest fix flaky test

ionutzpi:
MissingLastRevSeekerTest fix flaky test getNonSplitDocs

### DIFF
--- a/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyStateTest.java
+++ b/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyStateTest.java
@@ -238,6 +238,11 @@ public class CompressedDocumentPropertyStateTest {
             if (nodeStore != null) {
                 nodeStore.dispose();
             }
+            try {
+                fixture.dispose();
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
         }
 
     }
--- a/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyStateTest.java
+++ b/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyStateTest.java
@@ -47,7 +47,7 @@ import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 
-public class CompressedDocumentPropertyStateTest {
+public class CompressedDocumentPropertyStateTest extends AbstractDocumentStoreTest {
 
     private static final String STRING_HUGEVALUE = RandomStringUtils.random(10050, "dummytest");
     private static final int DEFAULT_COMPRESSION_THRESHOLD = 1024;
@@ -68,6 +68,10 @@ public class CompressedDocumentPropertyStateTest {
     };
 
     private DocumentNodeStore ns;
+
+    public CompressedDocumentPropertyStateTest(DocumentStoreFixture dsf) {
+        super(dsf);
+    }
 
     @Before
     public void before() throws Exception {
@@ -253,6 +257,7 @@ public class CompressedDocumentPropertyStateTest {
 
         PropertyState p = nodeStore.getRoot().getChildNode("test").getProperty("p");
         assertEquals(Objects.requireNonNull(p).getValue(Type.STRING), test);
+        removeMe.add("1:/test");
     }
 
 
--- a/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyStateTest.java
+++ b/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyStateTest.java
@@ -238,11 +238,6 @@ public class CompressedDocumentPropertyStateTest {
             if (nodeStore != null) {
                 nodeStore.dispose();
             }
-            try {
-                fixture.dispose();
-            } catch (Exception e) {
-                throw new RuntimeException(e);
-            }
         }
 
     }
--- a/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
+++ b/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
@@ -33,8 +33,6 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
-import java.util.Objects;
-
 import static org.apache.jackrabbit.oak.plugins.document.ClusterNodeInfo.DEFAULT_LEASE_DURATION_MILLIS;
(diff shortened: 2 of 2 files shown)


oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyStateTest.java
```diff
@@ -238,6 +238,11 @@ private void getBrokenSurrogateAndInitializeDifferentStores(DocumentStoreFixture
             if (nodeStore != null) {
                 nodeStore.dispose();
             }
+            try {
```
reschke: In theory, disposing the node store should take care of this. We should find out why it doesn't.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
```diff
@@ -249,7 +251,7 @@ public void dispose() {
         }
         dns.runBackgroundOperations();
         //seeker should return only non split documents
-        int docs = Iterables.size(seeker.getCandidates(0));
+        int docs = Iterables.size(seeker.getCandidates(Objects.requireNonNull(dns.getRoot().getLastRevision().getRevision(dns.getClusterId())).getTimestamp()));
```
reschke: Is this related to the flakyness?

ionutzpi: Yes. I removed 0 and replaced it with proper revision version timestamp.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyStateTest.java
ionutzpi: It is just a precocious adding. No truly impact.

reschke: Ok, then please remove that change, because a) not needed, b) not related, and c) potentially confusing.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
reschke: Ok.

a) Could you put a short description of what was wrong into the ticket?

b) I would recommend that @egli, @rishabhdaim or @Joscorbe confirm that his is the correct change.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyStateTest.java
ionutzpi: Done.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/MissingLastRevSeekerTest.java
rishabhdaim: `0` was used to fetch all the documents. Using the timestamp of `lastRevision` would result in skipping some documents.

If there is no proper reason, I will revert the change.

rishabhdaim: I was able to reproduce this test failure on my local. The root cause the is presence of document with id `1:/test` in the `RDB-H2`.

Once I did `mvn clean install` this problem went away.
IMO, it is because of some earlier test(s) didn't cleanup properly rather than test.

ionutzpi: Yes. That is the reason that I modified the timestamp to return only the documents needed, not all documents.
@reschke we can have inconsistencies retrieving all docs from RDB, but retrieving only the ones that this test added should solve the issue and future issues that could appear regarding this test. What do you think?

rishabhdaim: I think I found the Root cause.

These tests are failing due to the `CompressedDocumentPropertyStateTest` test not cleaning up properly.
On my local, it is causing either Mongo or RDB-H2 to fail.

`CompressedDocumentPropertyStateTest` creates the problematic node `1:/test` causing test failures.

`CompressedDocumentPropertyStateTest` has been added recently which explains the recent failures of `MissingLastRevSeekerTest`.

@ionutzpi could you please look into this?
cc @Joscorbe @reschke

ionutzpi: Yes. I added CompressedDocumentPropertyStateTest changes to solve this issue. Julian told that documentfixturestore should close when closing nodestore(seems that this is not true).

rishabhdaim: Calling `fixture.dispose();` after disposing of `nodeStore` [1] would fix this problem for `Mongo`.

[1] https://github.com/apache/jackrabbit-oak/blob/trunk/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/CompressedDocumentPropertyStateTest.java#L240

ionutzpi: It is not removing for RDB, I dont know the reason yet.

ionutzpi: @rishabhdaim if I add removeMe.add("1:/test"); after assert it will fix the issue. Do you agree with this change?

rishabhdaim:
Created PR https://github.com/apache/jackrabbit-oak/pull/1749 with proposed changes.

reschke:
@rishabhdaim - thanks for the analysis. This also explains why the test hasn't been flaky for a long time.

Note that since a few weeks, AbstractDocumentStoreTest will check the NODES table before/after. So if a test fails like that, a check of the unit test log might be interesting.

@ionutzpi - trust me; the RDB based tests assume that the test nodes are cleaned up (unless the test has been designed to purge tables). This is different from MongoDB because back then, I didn't want tests to nuke tables they did not create. This may have been the wrong approach, and we can discuss to change this. I'm just not sure it's worth the effort.

reschke:
Once it's clear which test is flaky, please adjust the issue title and make sure that the PR commit message matches.

rishabhdaim:
@reschke I have created PR to fix the flacky tests here https://github.com/apache/jackrabbit-oak/pull/1749.
Could you please review the same. Also the build passes after the proposed fix.

PR #1752, opened 2024-10-01
OAK-11026: Adding cache to external principals

Amoratinos:
## Summary

Added cached membership reader to external group principal reader allowing to cache principals read using UserConfiguration cache values. 

## Changes

- Changed CachedPrincipalMembershipReader to allow external calls
- Created CacheConfiguration class to define where to store the cached principals
- Adapted unit tests to the new class structure and created new unit tests to meet coverage threshold


oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/ExternalGroupPrincipalProvider.java
```diff
@@ -101,6 +102,7 @@
 class ExternalGroupPrincipalProvider implements PrincipalProvider, ExternalIdentityConstants, DynamicMembershipProvider {
 
     private static final Logger log = LoggerFactory.getLogger(ExternalGroupPrincipalProvider.class);
+    static final String CACHE_PRINCIPAL_NAMES = "externalPrincipalNames";
```
anchela: i would recommend to use "rep:" prefix for the property to mark it as oak-system maintained property

anchela (changes requested):
hi @Amoratinos , the overall approach looks very good to me. great stuff. just marked 1 thing that i spotted about the property name.
as discussed in slack i would try to avoid adding dependencies to oak-core in the oak-auth-external module. by extracting a cache-reader interface and extend userconfiguration.
if that turns out to be not feasible (or really ugly api-contract wise), we can live it with the dependency to oak-core... but lets give it a try.

one thing that i would like to verify explicitly (maybe you already do that): the CachedMembershipReader has that static final CACHE_UPDATES map to prevent concurrent updates from different threads. we need to make sure that a single repository login that is being handled by 2 login-modules and thus writing 2 cache properties actually succeeds... i think it should because i would expect this to covered by the same thread. but if you have not verified that explicitly that might be worth an extra test... from the top of my head something like
a) ext-user is member of ext-group which is member of local groups -> cache prop 1 (the new one)
b) ext-user is member of local group -> cache prop 2 (the original one)
both should be written if i am not mistaken

oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/ExternalGroupPrincipalProvider.java
Amoratinos: Done, instead of adding `rep:` in front I renamed it as there are some restrictions to that property definition.

anchela: @Amoratinos , how do you mean? sorry you lost me here.... why not adding the rep: prefix?

anchela: `rep:externalPrincipalNames` is already taken (used) for the real external principal names, right? so, i would avoid re-using the same name for the caching prop.... specially since it is not totally accurate. the cache will also contain local group names, which external groups are member of, no?
let's think about another name and add rep prefix. this is important to mark it as oak-internal property.
e.g.
- rep:groupPrincipalNamesForExternal
- rep:groupPrincipalNamesExternal
- ....
wdyt?

Amoratinos: I've renamed to `rep:externalCachedPrincipalNames` if that works

anchela (approved):
hi @Amoratinos , looks all good to me! thanks a lot for all the effort you put into this fix.

PR #1760, opened 2024-10-04
OAK-11062: document-store: refine skipping of MongoDB tests

mbaedke:
All tests against MongoDB will now be skipped if the system property oak.skipMongo is true.

### DIFF
--- a/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/prefetch/CacheWarmingTest.java
+++ b/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/prefetch/CacheWarmingTest.java
@@ -22,6 +22,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assume.assumeNotNull;
 
 import java.util.ArrayList;
 import java.util.HashSet;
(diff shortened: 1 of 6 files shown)


oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/prefetch/CacheWarmingTest.java
```diff
@@ -107,7 +108,7 @@ protected Clock getTestClock() throws InterruptedException {
 
     private DocumentStore newMongoDocumentStore() throws InterruptedException {
         mongoConnection = connectionFactory.getConnection();
-        assertNotNull(mongoConnection);
+        assumeNotNull(mongoConnection);
```
reschke: can we leave out that change? - it seems to be unrelated

reschke (commented):
The skipping is working fine; for instance, I don't see a docker image loaded.

Also:

With oak.skipMongo:

dockered mongo: 490/110 13:39
local mongo: 490/110 13:35
none: 490/110 14:06

(so the same, no matter whether docker is there, local mongo is there, or none of these)

Without:

dockered mongo: 1097/141 43:58
local mongo: 1097/142 29:48

reschke (approved):

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/prefetch/CacheWarmingTest.java
reschke: Ok; now I see why we needed it.

PR #1766, opened 2024-10-07
OAK-11169: MongoVersionGCSupport (oak-document-store) uses incorrect syntax for sort mongodb function

reschke:
The change in the test case is not really needed, but it would detect the case where sorting by ID does not happen (as opposed to be happening as default fallback).

@rgambelli - thanks for finding this.


Joscorbe (approved):
Looks good to me, has this been tested also with newer java driver? It seems it makes stricter validations, so maybe is a good test to check if the syntax is now correct.

reschke:
Actually, the test hasn't been run with the new driver, but I believe it's ok (it fails if we take out the broken sort statement *and* the exists(MODIFIED) condition, in which case the query returns docs in "insertion" order).

PR #1768, opened 2024-10-08
OAK-11174: VersionGCSupportTest should use standard fixtures

reschke:


oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGCSupportTest.java
```diff
@@ -111,12 +89,6 @@ public VersionGCSupportTest(DocumentStoreFixture fixture) {
         }
     }
 
-    @After
-    public void after() throws Exception {
-        store.remove(Collection.NODES, ids);
```
stefan-egli, line 116: the only thing I'm wondering is why this was done in the first place - as it seems `fixture.dispose()` will drop the collections anyway...

reschke: I believe we have several tests that ignore/bypass parts of out testing framework...

PR #1789, opened 2024-10-14
OAK-9447: Upgrade Mongo java driver to 5.2

reschke:


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -508,7 +507,7 @@ private void configureEstimators(IndexingProgressReporter progressReporter) {
     private long getEstimatedDocumentCount() {
         MongoConnection mongoConnection = indexHelper.getService(MongoConnection.class);
         if (mongoConnection != null) {
-            return mongoConnection.getDatabase().getCollection("nodes").count();
+            return mongoConnection.getDatabase().getCollection("nodes").countDocuments();
```
rishabhdaim, line 510:
```suggestion
            return mongoConnection.getDatabase().getCollection("nodes").estimatedDocumentCount();
```

`countDcuments` is very slow. I would use `estimatedDocumentCount` for faster results.

See https://www.mongodb.com/community/forums/t/countdocuments-is-extremly-slow/207357

oak-store-document/pom.xml
```diff
@@ -43,8 +43,8 @@
         <configuration>
           <instructions>
             <Import-Package>
-              com.mongodb*;version="[3.8, 4)";resolution:=optional,
-              org.bson*;version="[3.8, 4)";resolution:=optional,
+              com.mongodb*;version="[5.0, 5.2)";resolution:=optional,
```
rishabhdaim, line 46: Since we are switching to `5.2.0`, shouldn't it be

```suggestion
              com.mongodb*;version="[5.2, 6)";resolution:=optional,
```

oak-store-document/pom.xml
```diff
@@ -43,8 +43,8 @@
...
+              org.bson*;version="[5.0, 5.2)";resolution:=optional,
```
rishabhdaim, line 47: Same as above.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -2345,11 +2339,6 @@ private boolean withClientSession() {
         return connection.getStatus().isClientSessionSupported() && useClientSession;
     }
 
-    private boolean secondariesWithinAcceptableLag() {
-        return getClient().getReplicaSetStatus() == null
-                || connection.getStatus().getReplicaSetLagEstimate() < acceptableLagMillis;
```
rishabhdaim, line 2350: `acceptableLagMillis` should be marked as deprecated since it won't be used anymore.

rishabhdaim (changes requested):

oak-store-document/pom.xml
Joscorbe: Yes, definitely, or `5.3)` may be better, as sometimes they also introduce breaking changes or deprecated methods in minor versions. Example in 5.1: https://www.mongodb.com/docs/drivers/java/sync/v5.2/whats-new/#what-s-new-in-5.1 
And in 4.x it happened quite more frequently: https://www.mongodb.com/docs/drivers/java/sync/v5.2/upgrade/#version-4.8-breaking-changes

PR #1790, opened 2024-10-14
OAK-11195 : added api to create LinkedList from iterables and create IdentityHashSet

rishabhdaim:


oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/collections/CollectionUtils.java
```diff
@@ -180,6 +202,14 @@ public static <K> Set<K> newLinkedHashSet(final int capacity) {
...
     }
 
+    /**
+     * Creates a new, empty IdentityHashSet with default size.
+     */
+    @NotNull
+    public static <E> Set<E> newIdentityHashSet() {
+        return Collections.newSetFromMap(Maps.newIdentityHashMap());
+    }
+
```
mbaedke, line 212: Why not just return Sets.newIdentityHashSet() ?

rishabhdaim: We are trying to remove the guava usage. So returning `Sets.newIdentityHashSet() ` doesn't make sense.

reschke (approved):

PR #1793, opened 2024-10-14
OAK-11198: commons: add helper class for argument checking

reschke:
For now, exactly for those variants used in Oak.

(Bikeshed: this is in a new package because it is internal and no other package seemed to fit; suggestions for better package/class names are welcome)


oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/conditions/Checks.java
```diff
@@ -0,0 +1,118 @@
+package org.apache.jackrabbit.oak.commons.conditions;
```
rishabhdaim: missing license header.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/conditions/Checks.java
```diff
@@ -0,0 +1,118 @@
...
+            if (!CHECKMESSAGETEMPLATE) {
+                checkTemplate(messageTemplate, messageArgs);
+            }
+
+            String message = String.format(messageTemplate, messageArgs);
+            throw new IllegalArgumentException(message);
+        }
+    }
+
+    protected static boolean checkTemplate(@NotNull String messageTemplate, @Nullable Object... messageArgs) {
```
rishabhdaim: the scope should be `default`.
Since it is final and cannot be extended, it doesn't make sense to use a `protected` scope.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/conditions/Checks.java
```diff
@@ -0,0 +1,118 @@
...
+        int argsSpecified = messageArgs.length;
+        int argsInTemplate = countArguments(messageTemplate);
+        boolean result = argsSpecified == argsInTemplate;
+        if (argsSpecified != argsInTemplate) {
+            LOG.error("Invalid message format: template '{}', argument count {}", messageTemplate, argsSpecified);
+        }
+        return result;
+    }
+
+    protected static int countArguments(String template) {
```
rishabhdaim: same as above.

rishabhdaim (changes requested):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/conditions/Checks.java
reschke: Ups; had a conflict due to line ends, and that was lost.

rishabhdaim:
@reschke I would recommend using the name `Preconditions` or `Validate` over `Checks`.
@stefan-egli wdyt?

mbaedke (approved):

mbaedke:
Not "preconditions", because why "pre"?

stefan-egli (approved):
+1 for the code in general. 

regarding naming of class : I guess `Preconditions` could indeed be a good name - also considering guava has chosen that name. But the latter might be perhaps an argument to use a different name : if we use `Preconditions` it would not immediately be clear whether that is guava or not (unless the code is read in an IDEA which can clarify that).

rishabhdaim (approved):

reschke:
Let's use `Validate` for now.

rishabhdaim (approved):

PR #1795, opened 2024-10-15
OAK-10745 : added osgi configs for fullGc delayFactor & batch size

rishabhdaim:


stefan-egli (commented):
Few things:

* I'm wondering if we shouldn't also (or primarily) make `PROGRESS_BATCH_SIZE` configurable. Currently it's hardcoded to 10000 and we'd only be able to fiddle with the batch size, not the overall one. The overall one is however the one which controls how long fullGC occupies the scheduler thread. And in extremis we could set it to a ridiculously low value (say 10, then batch size also 10) and thus have near zero realistic performance impact on an otherwise very busy system. (Not suggesting 10 is a good value - but currently we wouldn't be able to configure that at all).
* One question is then perhaps: what are good default values for progress-/batchSize. Are the current ones OK or should we reduce them somewhat. That's something we could still find out - but it might become useful if we could tune both values.
* The default for classic delayFactor is 0 - which I think is not ideal - zero means it does not delay at all. In production system that should or is rather set to 0.5 or 1 etc. I'm not suggesting to change the default for classic GC. But for the new full GC one, we could perhaps indeed use a different default. Perhaps even as high as 5 (which would mean max 20% time usage)?
* Another aspect is that `delayOnModifications` (where the magic happens) is only invoked when deletion happens. But deletion isn't the only load on DocumentStore. There's also the reading and for fullGC the embedded verification that does many reads. For classic GC that probably applies as well to some extend - but the difference is : classic GC only ever works on deltas (as it existed since early oak days), it hardly has to scan a large amount. But full GC will do a one-off scan of large repositories thus do loads of reads. I'd suggest we introduce something like `delayOnBatch` - in addition to or perhaps actually in replacement of `delayOnModification` use.
* for `collectGarbageOnDocument` (which is used by oak-run) it doesn't seem useful for it to sleep, as it only ever runs on 1 document

rishabhdaim:
> I'm wondering if we shouldn't also (or primarily) make PROGRESS_BATCH_SIZE configurable. Currently it's hardcoded to 10000 and we'd only be able to fiddle with the batch size, not the overall one. The overall one is however the one which controls how long fullGC occupies the scheduler thread. And in extremis we could set it to a ridiculously low value (say 10, then batch size also 10) and thus have near zero realistic performance impact on an otherwise very busy system. (Not suggesting 10 is a good value - but currently we wouldn't be able to configure that at all).

yes, let us add a config for PROGRESS_BATCH_SIZE as well.

>The default for classic delayFactor is 0 - which I think is not ideal - zero means it does not delay at all. In production system that should or is rather set to 0.5 or 1 etc. I'm not suggesting to change the default for classic GC. But for the new full GC one, we could perhaps indeed use a different default. Perhaps even as high as 5 (which would mean max 20% time usage)?

I don't have any strong opinion here. Using default value 0 allows us to run it fast locally while testing. 
If we use a different default value, even local testing would be slower. But for local, we can always change the code.
So I am fine with other default values.

>Another aspect is that delayOnModifications (where the magic happens) is only invoked when deletion happens. But deletion isn't the only load on DocumentStore. There's also the reading and for fullGC the embedded verification that does many reads. For classic GC that probably applies as well to some extend - but the difference is : classic GC only ever works on deltas (as it existed since early oak days), it hardly has to scan a large amount. But full GC will do a one-off scan of large repositories thus do loads of reads. I'd suggest we introduce something like delayOnBatch - in addition to or perhaps actually in replacement of delayOnModification use.

Yes, we should add `dealyOnModifications` or `delayOnBatch` (or any other name) while collecting garbage as well.
Regarding `embeddedVerifiacation`, it is called before removing garbage, thus is already covered in the current code.

> for collectGarbageOnDocument (which is used by [oak-run](https://issues.apache.org/jira/browse/OAK-run)) it doesn't seem useful for it to sleep, as it only ever runs on 1 document

For oak run, we could set the default value to 0. It would allow us to reuse the existing code.

stefan-egli:
> I don't have any strong opinion here. Using default value 0 allows us to run it fast locally while testing.
If we use a different default value, even local testing would be slower. But for local, we can always change the code.
So I am fine with other default values.

Good point. I guess we have to choose between production and development unfortunately...

>  Regarding embeddedVerifiacation, it is called before removing garbage, thus is already covered in the current code.

the `timer` only measure removing garbage though - so that would have to be adjusted...

rishabhdaim:
>the timer only measure removing garbage though - so that would have to be adjusted...

No, it measures `embeddedVerification` as well.

rishabhdaim:
I have addressed your review comments here https://github.com/apache/jackrabbit-oak/pull/1795/commits/f6f8ce792f6f08c76fb03667be49993b6c6ef6e3

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1160,75 +1172,82 @@ public void collectGarbage(final NodeDocument doc, final GCPhases phases) {
...
+                    AUDIT_LOG.info("<Collected> [{}] garbage count in [{}]", op.getChanges().size(), doc.getId());
+                    updateOpList.add(op);
+                }
+                if (log.isTraceEnabled() && op.hasChanges()) {
+                    // only log in case of changes & debug level enabled
+                    log.trace("UpdateOp for [{}] is [{}]", doc.getId(), op);
+                }
+            } finally {
+                // pause the fullGC thread to avoid excessive load on the system
+                delayOnModifications(timer.stop().elapsed(MILLISECONDS), cancel, fullGCDelayFactor);
```
stefan-egli: My initial thought was to delay once per batch (the smaller one, not progressBatch). Agree, doing it on every collectGarbage would also work, but it might be a bit excessive? I'd have done it in `collectFullGC` upon completion of every batch - that would also include the time spent in the querying? Wdyt?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreHelper.java
```diff
@@ -72,7 +76,7 @@ public static VersionGarbageCollector createVersionGC(DocumentNodeStore nodeStor
                                                           boolean fullGCEnabled, boolean isFullGCDryRun,
                                                           boolean embeddedVerification, int fullGCMode) {
         return new VersionGarbageCollector(nodeStore, gcSupport, fullGCEnabled, isFullGCDryRun, embeddedVerification,
-                fullGCMode);
+                fullGCMode, DEFAULT_FGC_DELAY_FACTOR, DEFAULT_FGC_BATCH_SIZE, DEFAULT_FGC_PROGRESS_SIZE);
```
stefan-egli: fine with this, we could use `0` for delay factor here though, if we want (as this is oak-run, so more development or production-intervention)?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
rishabhdaim: I think it makes more sense to do it in a smaller batch, which would also avoid peaks on Mongo. Will make the change.

rishabhdaim: addressed in https://github.com/apache/jackrabbit-oak/pull/1795/commits/a43bd81a1c9421c6aa9427c75eef472c168a0c3b

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreHelper.java
rishabhdaim: addressed in https://github.com/apache/jackrabbit-oak/pull/1795/commits/a43bd81a1c9421c6aa9427c75eef472c168a0c3b

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
```diff
@@ -353,4 +353,24 @@
...
+                    "It allows the FullGC thread to stop by (fullGC run time * delayFactor) period after each cycle." +
+                    "The default value is " + DocumentNodeStoreService.DEFAULT_FGC_DELAY_FACTOR)
+    double fullGCDelayFactor() default DocumentNodeStoreService.DEFAULT_FGC_DELAY_FACTOR;
+
+    @AttributeDefinition(
+            name = "Batch Size to fetch data for each FullGC cycle",
+            description = "Integer value indicating the number of documents to fetch from database in a single query to check for Full GC." +
+                    "It should be a factor of fullGCProgressSize for better performance " +
+                    "The default value is " + DocumentNodeStoreService.DEFAULT_FGC_BATCH_SIZE)
+    int fullGCBatchSize() default DocumentNodeStoreService.DEFAULT_FGC_BATCH_SIZE;
```
shodaaan, line 369: Suggested change: for more clarity this could be renamed to "fullGCFetchBatchSize" or "fullGCFetchDocsBatchSize"

rishabhdaim: I believe the description mentions this very clearly. Not sure whether adding this in the config name would make it any clearer.

stefan-egli (approved):
+1, looks good now

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -255,26 +259,33 @@ static void setFullGcMode(int fullGcMode) {
...
         this.embeddedVerification = embeddedVerification;
+        this.fullGCDelayFactor = fullGCDelayFactor;
+        this.fullGCBatchSize = Math.min(fullGCBatchSize, fullGCProgressSize);
+        this.fullGCProgressSize = fullGCProgressSize;
         this.options = new VersionGCOptions();
 
         setFullGcMode(fullGCMode);
-        AUDIT_LOG.info("<init> VersionGarbageCollector created with fullGcMode = {}", fullGcMode);
+        AUDIT_LOG.info("<init> VersionGarbageCollector created with fullGcMode: {}, batchSize: {}, progressSize: {}",
+                fullGcMode, fullGCBatchSize, fullGCProgressSize);
```
stefan-egli, line 288: minor : could also log `delayFactor` (but would suggest to merge the PR anyway)

rishabhdaim: sure will log in next PR.

PR #1797, opened 2024-10-15
OAK-11200 use the improved API of the JackrabbitNode if possible

joerghoh:


oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/JackrabbitSession.java
```diff
@@ -268,6 +268,9 @@ public interface JackrabbitSession extends Session {
      */
     @Nullable 
     default Node getParentOrNull(@NotNull Item item) throws RepositoryException {
+        if (item instanceof JackrabbitNode) {
+            return ((JackrabbitNode) item).getParentOrNull();
+        }
```
reschke, line 273: I'd put an "else" here (I don't like unbalanced "return"s, but that's probably just a matter of taste.

anchela (approved):
lgtm

PR #1805, opened 2024-10-17
OAK-11182: Create and test general mechanism for creating MongoDocumentStore fullGC garbage for testing

shodaaan:
- created GenerateFullGCCommand oak-run class with methods for generating garbage of type GAP_ORPHANS and EMPTY_PROPERTIES. All garbage is generated under a specific main node. Also created a cleanup method for deleting all generated garbage
- added call to GenerateFullGCCommand to AvailableModes with specific oak-run input variable
- created GenerateFullGCCommandTest class with unit test methods for generating and cleaning up garbage
- made VersionGarbageCollector.fullGCMode public, so it can be used Command class


oak-run/src/main/java/org/apache/jackrabbit/oak/run/AvailableModes.java
```diff
@@ -75,5 +75,6 @@ public final class AvailableModes {
             .put("segment-copy", new SegmentCopyCommand())
             .put("server", new ServerCommand())
             .put("purge-index-versions", new LucenePurgeOldIndexVersionCommand())
+            .put("generateFullGC", new GenerateFullGCCommand())
```
rishabhdaim:
```suggestion
            .put("createGarbage", new CreateGarbageCommand())
```

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
```diff
@@ -0,0 +1,416 @@
...
+            cleanGarbage(options, closer);
+        } else {
+            System.err.println("unknown revisions command: " + subCmd);
+        }
+
+        return generateBasePaths;
+    }
+
+    private List<String> generateGarbageContinuously(GenerateFullGCOptions options, Closer closer) throws IOException, Exception {
+        ScheduledExecutorService executor = Executors.newScheduledThreadPool(1);
```
rishabhdaim: Could you please close the executor once finished.
See https://github.com/shodaaan/jackrabbit-oak/blob/trunk/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStoreThrottlingMetricsUpdater.java#L58 for reference.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/AvailableModes.java
stefan-egli: +1 for renaming the command

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
```diff
@@ -0,0 +1,416 @@
...
+public class GenerateFullGCCommand implements Command {
+    private static final Logger LOG = LoggerFactory.getLogger(GenerateFullGCCommand.class);
+
+    private static final String USAGE = "generateFullGC {<jdbc-uri> | <mongodb-uri>} [options]";
+
+    /**
+     * Root node for fullGC garbage generation.
+     * Necessary in order to allow cleanup of all generated garbage nodes by simply removing the root node.
+     */
+    public static String FULLGC_GEN_ROOT_PATH = "fullGCGenRoot";
```
stefan-egli: Would recommend not to add content to root and use something like "/tmp" instead, so eg "/tmp/fullGcTestGarbage" or something like that.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
```diff
@@ -0,0 +1,416 @@
...
+
+    /**
+     * Base path for fullGC garbage generation. The timestamp of the run will be appended to this path,
+     * which is necessary in order for each garbage generation run to be unique and not overwrite previous ones.
+     * If continuous generation is enabled, the index of the run will also be appended to this path.
+     */
+    public static String FULLGC_GEN_BASE_PATH = "fullGCGenTest_";
```
stefan-egli: once you are under `FULLGC_GEN_ROOT_PATH` and its name indicates that it is for testing and is garbage, do we still need all path elements to contain `fullGc`?

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
```diff
@@ -0,0 +1,416 @@
...
+     * of this class.
+     * @param collection
+     * @param key
+     * @param maxCacheAge
+     * @return
+     * @param <T>
+     * @throws DocumentStoreException
+     */
+    @Nullable
+    public <T extends Document> T getDocument(Collection<T> collection, String key, int maxCacheAge)
```
stefan-egli: as this is only used by test code I'd move it over there (and if needed create an accessor to `documentNodeStore` for example)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -168,7 +168,7 @@ public class VersionGarbageCollector {
      * Ultimately the goal is to clean up all possible garbage. After hardening these modes
      * might no longer be supported.
      */
-    enum FullGCMode {
+    public enum FullGCMode {
```
stefan-egli: not sure if we want this public? what about adding a helper under `org.apache.jackrabbit.oak.plugins.document` in oak-run instead?

that helper could perhaps also have some `includesEmptyProps` / `includesGapOrphans` etc methods - as fullgc modes is going to be a combination of different modes (and the current check is hard coded to only modes 1 and 2)

oak-run/src/test/java/org/apache/jackrabbit/oak/plugins/document/GenerateFullGCCommandTest.java
```diff
@@ -0,0 +1,293 @@
...
+        List<String> garbageRootNodeChildNames = getChildNodeNames(garbageRootNodeState);
+
+        assertEquals(garbageRootNodeChildNames.size(), 1);
+        for (String generateBasePath : generatedBasePaths) {
+            assertTrue(garbageRootNodeChildNames.contains(generateBasePath));
+        }
+
+        String generateBasePath = generatedBasePaths.get(0);
+        for (int i = 0; i < garbageNodesParentCount; i ++) {
+            NodeDocument missingDocument = cmd.getCommand().getDocument(Collection.NODES, "3:/"
```
stefan-egli: would suggest to use `Utils.getIdFromPath` etc rather than hard coding the depth - eg changing the base path to be `/tmp/xy` would now require this depth to be fixed
(applies to a few different occasions)

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
```diff
@@ -0,0 +1,416 @@
...
+        VersionGarbageCollector.FullGCMode fullGCMode = getFullGCMode(options);
+        if (fullGCMode == VersionGarbageCollector.FullGCMode.NONE) {
+            LOG.error("Invalid garbageType specified. Must be one of the following: 1 - EMPTYPROPS, 2 - GAP_ORPHANS, 3 - GAP_ORPHANS_EMPTYPROPS");
+            System.exit(1);
+        }
+
+        //1. Create nodes with properties
+        NodeBuilder rootNode = documentNodeStore.getRoot().builder();
+        NodeBuilder garbageRootNode = rootNode.child(FULLGC_GEN_ROOT_PATH);
+        garbageRootNode.child(generationBasePath).setProperty("jcr:primaryType", "nt:file", NAME);
```
stefan-egli: would use `org.apache.jackrabbit.JcrConstants.JCR_PRIMARYTYPE` instead

also, would not use `nt:file` as that requires eg `jcr:content` child etc (which wasn't noticed in the test as the merge happens without commithook - something that's a bit risky, but this is probably an exception as it is primarily a test command - but this aspect isn't ideal and might lead to surprises). we could use `org.apache.jackrabbit.oak.spi.nodetype.NT_OAK_UNSTRUCTURED` instead (as that is unorderable, and unstructured, it would be a good fix as it doesn't have any special requirements otherwise).

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
```diff
@@ -0,0 +1,416 @@
...
+        int nodesCountUnderParent = options.getCreateGarbageNodesCount() / options.getGarbageNodesParentCount();
+        for(int i = 0; i < options.getGarbageNodesParentCount(); i ++) {
+            garbageRootNode.child(generationBasePath).child(FULLGC_GEN_PARENT_NODE_PREFIX + i).setProperty("jcr:primaryType", "nt:folder", NAME);
+
+            for(int j = 0; j < nodesCountUnderParent; j ++) {
+                garbageRootNode.child(generationBasePath).child(FULLGC_GEN_PARENT_NODE_PREFIX + i).child(FULLGC_GEN_NODE_PREFIX + j).
+                        setProperty("jcr:primaryType", "nt:file", NAME);
+
+                if (fullGCMode == VersionGarbageCollector.FullGCMode.EMPTYPROPS || fullGCMode == VersionGarbageCollector.FullGCMode.GAP_ORPHANS_EMPTYPROPS) {
+                    garbageRootNode.child(generationBasePath).child(FULLGC_GEN_PARENT_NODE_PREFIX + i).child(FULLGC_GEN_NODE_PREFIX + j).
```
stefan-egli: what might be interesting is adding different types of "child subtrees". IIUC then the it will delete the nodes at level `FULLGC_GEN_PARENT_NODE_PREFIX` (which then leaves its children gap orphans). currently it only creates a 1-level deep child under that node. what would be interesting is to have different cases:
* (later perhaps) 0-level deep but just split docs (we actually probably will miss that type of garbage, that's going to be a new TODO probably, but this test could trigger it)
* 1-level deep (which you already have)
* 2-level deep
* 3-level deep

reason being that it might cause different behavior in fullGC code later on - and thus having slight variations of that garbage could be useful to have

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
```diff
@@ -0,0 +1,416 @@
...
+            System.exit(1);
+        }
+
+        String generationBasePath = FULLGC_GEN_BASE_PATH + timestamp + "_" + runIndex;
+        System.out.println("Generating fullGC on the document: " + generationBasePath);
+        documentNodeStore = builder.build();
+
+        VersionGarbageCollector.FullGCMode fullGCMode = getFullGCMode(options);
+        if (fullGCMode == VersionGarbageCollector.FullGCMode.NONE) {
+            LOG.error("Invalid garbageType specified. Must be one of the following: 1 - EMPTYPROPS, 2 - GAP_ORPHANS, 3 - GAP_ORPHANS_EMPTYPROPS");
```
stefan-egli: assume this log message will be outdated soon with more fullGC modes ready for rollout. eg we will support ALL_ORPHANS at some point etc... Perhaps just mention to use a valid fullGc mode (and perhaps add the classname)?

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
```diff
@@ -0,0 +1,416 @@
...
+        documentNodeStore.runBackgroundOperations();
+
+        //3.1. Generate garbage nodes - GAP_ORPHANS - remove parent nodes
+        if (fullGCMode == VersionGarbageCollector.FullGCMode.GAP_ORPHANS) {
+            StringBuilder sbNodePath = new StringBuilder();
+            List<String> deleteNodePaths = new ArrayList<>();
+            for (int i = 0; i < options.getGarbageNodesParentCount(); i++) {
+
+                sbNodePath.setLength(0);
+                sbNodePath.append("3:/").append(FULLGC_GEN_ROOT_PATH).append("/").append(generationBasePath).append("/").
```
stefan-egli: (same suggestion here to use Utils for path composition and id conversion)

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
```diff
@@ -0,0 +1,416 @@
...
+        public int getNumberOfRuns() {
+            return numberOfRuns.value(options);
+        }
+
+        public int getGenerateIntervalSeconds() {
+            return generateIntervalSeconds.value(options);
+        }
+    }
+
+    public void execute(String... args) throws Exception {
```
dare-gartners: throwing Exception is not nice, can't you narrow it down? Also I think it's actually throwing throwable, in which case you dont need to specify throws at all

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
```diff
@@ -0,0 +1,416 @@
...
+        Closer closer = Closer.create();
+        try {
```
dare-gartners: if you use try with resources, you wouldn't need to write finally any more. this will be done automatically

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
```diff
@@ -0,0 +1,416 @@
...
+        } else if (GenerateFullGCOptions.CMD_CLEAN.equals(subCmd)) {
+            cleanGarbage(options, closer);
+        } else {
+            System.err.println("unknown revisions command: " + subCmd);
+        }
+
+        return generateBasePaths;
+    }
+
+    private List<String> generateGarbageContinuously(GenerateFullGCOptions options, Closer closer) throws IOException, Exception {
```
dare-gartners: IoException is a child class of Exception, so you don't need to specify both in the throws. And again I'd suggest not throwing Exception and narrowing down to specific exceptions that the function might throw

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
```diff
@@ -0,0 +1,416 @@
...
+    private String generateGarbage(GenerateFullGCOptions options, Closer closer, int runIndex) throws IOException, Exception {
+
+        DocumentNodeStoreBuilder<?> builder = createDocumentMKBuilder(options, closer);
+        long generationTimestamp = System.currentTimeMillis();
+
+        return generateGarbage(options, closer, runIndex, builder, generationTimestamp);
+    }
+
+    private String generateGarbage(GenerateFullGCOptions options, Closer closer, int runIndex,
+                                  DocumentNodeStoreBuilder<?> builder, long timestamp) throws IOException, Exception {
```
dare-gartners: why dont you generate/create timestamp in this method instead of passing it down? It seems to be called fro only 1 another private function that does nothing other then calling this function

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
```diff
@@ -0,0 +1,416 @@
...
+            LOG.error("Invalid garbageType specified. Must be one of the following: 1 - EMPTYPROPS, 2 - GAP_ORPHANS, 3 - GAP_ORPHANS_EMPTYPROPS");
+            System.exit(1);
+        }
+
+        //1. Create nodes with properties
+        NodeBuilder rootNode = documentNodeStore.getRoot().builder();
+        NodeBuilder garbageRootNode = rootNode.child(FULLGC_GEN_ROOT_PATH);
+        garbageRootNode.child(generationBasePath).setProperty("jcr:primaryType", "nt:file", NAME);
+
+        int nodesCountUnderParent = options.getCreateGarbageNodesCount() / options.getGarbageNodesParentCount();
```
dare-gartners: can options.getGarbageNodesParentCount() be 0 ?

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
```diff
@@ -0,0 +1,416 @@
...
+                garbageRootNode.child(generationBasePath).child(FULLGC_GEN_PARENT_NODE_PREFIX + i).child(FULLGC_GEN_NODE_PREFIX + j).
+                        setProperty("jcr:primaryType", "nt:file", NAME);
+
+                if (fullGCMode == VersionGarbageCollector.FullGCMode.EMPTYPROPS || fullGCMode == VersionGarbageCollector.FullGCMode.GAP_ORPHANS_EMPTYPROPS) {
+                    garbageRootNode.child(generationBasePath).child(FULLGC_GEN_PARENT_NODE_PREFIX + i).child(FULLGC_GEN_NODE_PREFIX + j).
+                            setProperty(EMPTY_PROPERTY_NAME, "bar", STRING);
+                }
+            }
+        }
+        documentNodeStore.merge(rootNode, EmptyHook.INSTANCE, CommitInfo.EMPTY);
```
dare-gartners: these 2 operations you do again on Line 305&306 and Line 321/322. If fullGCMode can only have 1 value, wouldn't it be enough to do it just once at the end of all If:s ? 
`documentNodeStore.merge(rootNode, EmptyHook.INSTANCE, CommitInfo.EMPTY);
        documentNodeStore.runBackgroundOperations();`

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
```diff
@@ -0,0 +1,416 @@
...
+import static org.apache.jackrabbit.oak.api.Type.STRING;
+import static org.apache.jackrabbit.oak.run.Utils.createDocumentMKBuilder;
+
+/**
+ * GenerateFullGCCommand generates garbage nodes in the repository in order to allow for testing fullGC functionality.
+ */
+public class GenerateFullGCCommand implements Command {
+    private static final Logger LOG = LoggerFactory.getLogger(GenerateFullGCCommand.class);
+
+    private static final String USAGE = "generateFullGC {<jdbc-uri> | <mongodb-uri>} [options]";
```
rishabhdaim:
```suggestion
    private static final String USAGE = "createGarbage {<jdbc-uri> | <mongodb-uri>} [options]";
```

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
rishabhdaim: how about we use `garbage` to make it clear than we are creating garbage under this folder.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
```diff
@@ -0,0 +1,416 @@
...
+     * If continuous generation is enabled, the index of the run will also be appended to this path.
+     */
+    public static String FULLGC_GEN_BASE_PATH = "fullGCGenTest_";
+
+    /**
+     * Prefix for parent nodes under which garbage nodes will be created.
+     * The index of the parent node will be appended to this prefix.
+     */
+    public static String FULLGC_GEN_PARENT_NODE_PREFIX = "fullGCParent_";
+    public static String FULLGC_GEN_NODE_PREFIX = "fullGCNode_";
```
rishabhdaim: I would avoid using FULLGC in variable names.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
rishabhdaim: @dare-gartners we are overriding the `execute` method from `Command` interface here. We have to follow the syntax.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
rishabhdaim: Yes, this is entirely possible. Good catch @dare-gartners.

rishabhdaim (changes requested):

oak-run/src/main/java/org/apache/jackrabbit/oak/run/AvailableModes.java
shodaaan: Changed. Thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
shodaaan: Changed. Thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
shodaaan: Made both changes. Thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
shodaaan: Changed node names. Thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
shodaaan: Done. Thank you for the suggestion.

oak-run/src/test/java/org/apache/jackrabbit/oak/plugins/document/GenerateFullGCCommandTest.java
shodaaan: Changed. Thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
shodaaan: Changed. Thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
shodaaan: Changed. Thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
shodaaan: Done. Thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
shodaaan: As @rishabhdaim said, we have to follow the Command interface. No change made.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
shodaaan: Removed IOException, just Exception is thrown now.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
shodaaan: Added check for garbage nodes count and parent nodes count from options. Thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
shodaaan: The merge and runBackgroundOperations calls are necessary after each batch of operations in order for the changes to be applied to the MongoDB database.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
shodaaan: Done, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
shodaaan: Done, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
shodaaan: The timestamp is necessary for the flow of calling generateGarbage from generateGarbageContinuously, where all of the generated garbage must be under the parent with the initial timestamp.
No change made.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
shodaaan: Done, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
```diff
@@ -0,0 +1,427 @@
...
+
+        public GenerateFullGCOptions(String usage) {
+            super(usage);
+            createGarbageNodesCount = parser
+                    .accepts("garbageNodesCount", "the total number of garbage nodes to create").withRequiredArg()
+                    .ofType(Integer.class).defaultsTo(0);
+            garbageNodesParentCount = parser
+                    .accepts("garbageNodesParentCount", "total number of parent nodes under which to create garbage nodes").withRequiredArg()
+                    .ofType(Integer.class).defaultsTo(1);
+            garbageType = parser
```
rishabhdaim: We should allow it to create multiple types of garbage.

@stefan-egli wdyt?

Here is a sample code to enable delimited values in parameters: https://github.com/apache/jackrabbit-oak/blob/trunk/oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCheckCommand.java#L148-L150

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
```diff
@@ -0,0 +1,427 @@
...
+import static org.apache.jackrabbit.oak.plugins.document.util.Utils.getIdFromPath;
+import static org.apache.jackrabbit.oak.run.Utils.createDocumentMKBuilder;
+
+/**
+ * GenerateFullGCCommand generates garbage nodes in the repository in order to allow for testing fullGC functionality.
+ */
+public class GenerateGarbageCommand implements Command, Closeable {
+    private static final Logger LOG = LoggerFactory.getLogger(GenerateGarbageCommand.class);
+
+    private final ScheduledExecutorService continuousRunExecutor = Executors.newScheduledThreadPool(1);
```
rishabhdaim: We need to register this executor with `Closer` to close this.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
```diff
@@ -0,0 +1,427 @@
...
+     */
+    private String generateGarbage(GenerateFullGCOptions options, Closer closer, int runIndex) throws IOException, Exception {
+
+        DocumentNodeStoreBuilder<?> builder = createDocumentMKBuilder(options, closer);
+        long generationTimestamp = System.currentTimeMillis();
+
+        return generateGarbage(options, closer, runIndex, builder, generationTimestamp);
+    }
+
+    private String generateGarbage(GenerateFullGCOptions options, Closer closer, int runIndex,
```
rishabhdaim: `closer` is not used in this method.

oak-run/src/main/java/org/apache/jackrabbit/oak/plugins/document/GenerateGarbageHelper.java
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Helper class to generate garbage for testing purposes.
+ */
+public class GenerateGarbageHelper {
+
+    public static boolean isInvalidGarbageGenerationMode(int fullGCMode) {
+        return fullGCMode == 0;
+    }
+
+    public static boolean includesEmptyProps(int fullGCMode) {
```
rishabhdaim:
```suggestion
    public static boolean isEmptyProps(int fullGCMode) {
```

oak-run/src/main/java/org/apache/jackrabbit/oak/plugins/document/GenerateGarbageHelper.java
```diff
@@ -0,0 +1,19 @@
...
+        return fullGCMode == 1;
+    }
+
+    public static boolean includesGapOrphans(int fullGCMode) {
```
rishabhdaim:
```suggestion
    public static boolean isGapOrphans(int fullGCMode) {
```

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
```diff
@@ -0,0 +1,427 @@
...
+                    garbageRootNode.child(generationBasePath).child(GEN_PARENT_NODE_PREFIX + i).child(GEN_NODE_PREFIX + j).
+                            removeProperty(EMPTY_PROPERTY_NAME);
+                }
+            }
+        }
+        documentNodeStore.merge(rootNode, EmptyHook.INSTANCE, CommitInfo.EMPTY);
+        documentNodeStore.runBackgroundOperations();
+
+        //3.1. Generate garbage nodes - GAP_ORPHANS - remove parent nodes
+        if (GenerateGarbageHelper.includesGapOrphans(options.getGarbageType())) {
```
rishabhdaim: One problem with this approach is that the more garbage types we add in the future, the more if/else statements we need to add.

This would make the code more complex, less readable, and very hard to understand.

How about we create `enums` for fullGcMode with `createGarbage` method and each mode would create its type of garbage?

This would ensure that for adding a new type of garbage, we don't need to update the base class, add a new implementation of garbage type, and plug in the base class.

cc @stefan-egli @Joscorbe

rishabhdaim: A sample implementation could be like this.

`private interface GarbageType {
        void createGarbage();
    }`

    `private class EmptyProps implements GarbageType {

        @Override
        public void createGarbage() {
            for (int i = 0; i < options.getGarbageNodesParentCount(); i++) {
                for (int j = 0; j < nodesCountUnderParent; j++) {
                    garbageRootNode.child(generationBasePath).child(GEN_PARENT_NODE_PREFIX + i).child(GEN_NODE_PREFIX + j).
                            removeProperty(EMPTY_PROPERTY_NAME);
                }
            }

            documentNodeStore.merge(rootNode, EmptyHook.INSTANCE, CommitInfo.EMPTY);
            documentNodeStore.runBackgroundOperations();
        }
    }`

rishabhdaim (changes requested):

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
rishabhdaim: All these implementations could be saved inside a `Map` of `fullGcMode --> Implementation` and we can simply iterate over garbageType and create garbage of different types in one simple loop.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
shodaaan: Done, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateFullGCCommand.java
shodaaan: Implemented, thank you for the suggestion.
Added orphansDepth and orphansLevelGap parameters for running the command, which allow for generating gap orphans up to 15 levels deep and also allows for deleting nodes in their hierarchy up to the specified orphansLevelGap level.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
shodaaan: In the interest of keeping the functionality as simple as possible, we can do this in a future update of the GenerateGarbageCommand, if necessary.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
shodaaan: Fixed, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/plugins/document/GenerateGarbageHelper.java
shodaaan: Fixed, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
shodaaan: Thank you for the suggestion, this is a good idea for the future, when we will have multiple garbageTypes generated by the GenerateGarbageCommand.
For now, since we only have 2 types, one if/else is short and quite readable.

As suggested, I have completely separated the methods for generating EMPTY_PROPS / GAP_ORPHANS garbage.

Will definitely revisit the rest of the suggestion when more garbageType generation methods are implemented.

oak-run/src/main/java/org/apache/jackrabbit/oak/plugins/document/GenerateGarbageHelper.java
shodaaan: Fixed, thank you for the suggestion.

oak-run/src/test/java/org/apache/jackrabbit/oak/plugins/document/GenerateFullGCCommandTest.java
```diff
@@ -0,0 +1,293 @@
...
+    public void cleanupAllGarbage() {
+
+        // generate garbage first
+        GenerateFullGCCmd cmdGenerateGarbage = new GenerateFullGCCmd(OPTION_GARBAGE_NODES_COUNT, "10", OPTION_GARBAGE_NODES_PARENT_COUNT, "1",
+                OPTION_GARBAGE_TYPE, "2");
+        cmdGenerateGarbage.run();
+        Closer cmd1Closer = cmdGenerateGarbage.getCloser();
+
+        // cleanup garbage
+        String mongoDBConnString = "mongodb://cm-p107053-e251100-username:3fUEP9UHPD9yP4H@cmp107053e251100r355e01-shard-00-00.wxmqq.mongodb.net:27017,cmp107053e251100r355e01-shard-00-01.wxmqq.mongodb.net:27017,cmp107053e251100r355e01-shard-00-02.wxmqq.mongodb.net:27017/cm-p107053-e251100-database?ssl=true&authSource=admin&replicaSet=atlas-4x602o-shard-0&retryWrites=true&readPreference=primaryPreferred&readConcernLevel=majority&w=majority";
```
Joscorbe: Check this please!

Joscorbe (changes requested):

oak-run/src/main/java/org/apache/jackrabbit/oak/plugins/document/GenerateGarbageHelper.java
```diff
@@ -0,0 +1,19 @@
+package org.apache.jackrabbit.oak.plugins.document;
```
stefan-egli: license header is missing

oak-run/src/main/java/org/apache/jackrabbit/oak/run/AvailableModes.java
```diff
@@ -75,5 +75,6 @@ public final class AvailableModes {
             .put("segment-copy", new SegmentCopyCommand())
             .put("server", new ServerCommand())
             .put("purge-index-versions", new LucenePurgeOldIndexVersionCommand())
+            .put("generateGarbage", new GenerateGarbageCommand())
```
stefan-egli: oak-run commands need to be lower case. for readability it might be easier to use something like `"generate-garbage"`?

Or .. on second thought .. what about expressing more explicitly that this new command is meant for testing only by .. eg calling it `generate-test-garbage` ?

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
```diff
@@ -0,0 +1,553 @@
...
+
+/**
+ * GenerateFullGCCommand generates garbage nodes in the repository in order to allow for testing fullGC functionality.
+ */
+public class GenerateGarbageCommand implements Command, Closeable {
+    private static final Logger LOG = LoggerFactory.getLogger(GenerateGarbageCommand.class);
+
+    private final ScheduledExecutorService continuousRunExecutor = Executors.newScheduledThreadPool(1);
+
+    private static final String USAGE = "createGarbage {<jdbc-uri> | <mongodb-uri>} [options]";
```
stefan-egli: should use the actual command instead of `createGarbage`.

Also, somewhere, not sure where, mandatory options should be mentioned, such as IIUC `garbageNodesCount` at least.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
```diff
@@ -0,0 +1,553 @@
...
+
+    private int continuousRunIndex = 0;
+
+    private DocumentNodeStore documentNodeStore;
+
+    public DocumentNodeStore getDocumentNodeStore() {
+        return documentNodeStore;
+    }
+
+    private static class GenerateFullGCOptions extends Utils.NodeStoreOptions {
```
stefan-egli: I think this shouldn't have "FullGC" in the name anymore?

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
```diff
@@ -0,0 +1,553 @@
...
+
+    /**
+     * FullGC garbage should be generated under tmp node.
+     */
+    public static String GARBAGE_GEN_ROOT_PATH_BASE = "tmp";
+
+    /**
+     * The root node name for fullGC garbage generation, one level under tmp.
+     */
+    public static String GARBAGE_GEN_ROOT_NODE_NAME = "fullGC_GarbageRoot";
```
stefan-egli: not sure about the name here, and whether it should have `fullGC` in it...
what about `oak-run-generated-test-garbage` ?

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
```diff
@@ -0,0 +1,553 @@
...
+    /**
+     * Recursively get all paths of the tree nodes under the given root node.
+     * @param rootNode
+     * @param basePath
+     * @param treeNodePaths
+     * @param nodeLevel
+     */
+    private void getTreeNodePaths(NodeBuilder rootNode, String basePath,
+                                  List<String> treeNodePaths, int nodeLevel) {
+        String childBasePath = basePath.replaceFirst(nodeLevel + ":/", (nodeLevel + 1) + ":/");
```
stefan-egli: This seems like a good candidate to make use of  `org.apache.jackrabbit.oak.plugins.document.util.Utils.getIdFromPath` or `org.apache.jackrabbit.oak.plugins.document.Path.fromString`. Working on the Path object would also allow to verify that anything added to `treeNodePaths` is indeed at least 2 levels deep and is under the expected main generated-test-garbage path.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
```diff
@@ -0,0 +1,553 @@
...
+        String garbageRootNodePath = "2:/"+ GARBAGE_GEN_ROOT_PATH;
+        List<String> childNodePaths = new ArrayList<>();
+        childNodePaths.add(garbageRootNodePath);
+
+        // get all paths of the tree nodes under the garbage root node
+        getTreeNodePaths(generatedGarbageRootBuilder, garbageRootNodePath, childNodePaths, 2);
+
+        documentNodeStore.getDocumentStore().remove(org.apache.jackrabbit.oak.plugins.document.Collection.NODES, childNodePaths);
+        documentNodeStore.merge(rootBuilder, EmptyHook.INSTANCE, CommitInfo.EMPTY);
+        documentNodeStore.runBackgroundOperations();
```
stefan-egli: (same as above, shouldn't be necessary)

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
```diff
@@ -0,0 +1,553 @@
...
+        //2. Generate garbage nodes - EMPTY_PROPERTIES
+        for (int i = 0; i < options.getGarbageNodesParentCount(); i++) {
+            for (int j = 0; j < nodesCountUnderParent; j++) {
+                garbageRootNode.child(generationBasePath).child(GEN_PARENT_NODE_PREFIX + i).child(GEN_NODE_PREFIX + j).
+                        removeProperty(EMPTY_PROPERTY_NAME);
+            }
+        }
+
+        documentNodeStore.merge(rootNode, EmptyHook.INSTANCE, CommitInfo.EMPTY);
+        documentNodeStore.runBackgroundOperations();
```
stefan-egli: this shouldn't be necessary - automatically happens in the background and at dispose time.

stefan-egli (commented):
Added a few more points.

One general question is : as this command is clearly a test command not to be used against a production environment : how could that be emphasized. For example: we could add an explicit: "are you really sure you want to generate test garbage against ...". That would make it clear that the target repository is going to be written to. Wdyt?

oak-run/src/main/java/org/apache/jackrabbit/oak/plugins/document/GenerateGarbageHelper.java
shodaaan: Fixed, thank you for the catch.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/AvailableModes.java
shodaaan: Changed, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
shodaaan: Changed, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
shodaaan: Changed, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
shodaaan: Changed, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
shodaaan: Changed, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
shodaaan: Changed, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
shodaaan: Changed, thank you for the suggestion.

shodaaan:
> Added a few more points.
> 
> One general question is : as this command is clearly a test command not to be used against a production environment : how could that be emphasized. For example: we could add an explicit: "are you really sure you want to generate test garbage against ...". That would make it clear that the target repository is going to be written to. Wdyt?

I resolved all of the points, thank you for the suggestions.

Regarding the explicit check, I don't think this is necessary because oak-run is a tool that by design makes important / deep changes in the MongoDB it is run on, so users should always be careful and know the exact consequences of the commands they run. 
The USAGE text clearly states what the command and associated subcommands do.
Also, since oak-run commands are designed to be used "with intent", and not "by accident", having an additional confirmation step for each garbage generation command run would just annoy the user IMHO. I didn't see this check for any other command in oak-run.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/AvailableModes.java
```diff
@@ -75,5 +75,6 @@ public final class AvailableModes {
             .put("segment-copy", new SegmentCopyCommand())
             .put("server", new ServerCommand())
             .put("purge-index-versions", new LucenePurgeOldIndexVersionCommand())
+            .put("generate-test-garbage", new GenerateGarbageCommand())
```
rishabhdaim: I would suggest changing `generate` to `create`.
Generate is used with non-tangible things whereas `create` is used with tangible ones.

`create` seems more accurate grammatically.

shodaaan: Fixed, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
stefan-egli: any particular reason you would like to keep these ? I think they apply to tests but not in this case.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
stefan-egli: same here

stefan-egli:
> Regarding the explicit check, I don't think this is necessary

Fair enough.

Two unrelated additional comments:
* the DocumentNodeStore is not properly disposed. You can see this by running the oak-run command consecutively : the second start will have to wait 2min for the lease timeout of the first non-gracefully shut down run. We should ensure a graceful disposal.
* what might be useful is to add the base path where the garbage is generated to - such that the user immediately knows (and doesn't eg have to consult the source code)

shodaaan:
> > Regarding the explicit check, I don't think this is necessary
> 
> Fair enough.
> 
> Two unrelated additional comments:
> 
> * the DocumentNodeStore is not properly disposed. You can see this by running the [oak-run](https://issues.apache.org/jira/browse/OAK-run) command consecutively : the second start will have to wait 2min for the lease timeout of the first non-gracefully shut down run. We should ensure a graceful disposal.
> * what might be useful is to add the base path where the garbage is generated to - such that the user immediately knows (and doesn't eg have to consult the source code)

I added documentNodeStore.dispose() calls to CreateGarbageCommand. Thank you for the suggestion

The root path where garbage is generated (/tmp/oak-run-generated-test-garbage) is specified in the USAGE string of CreateGarbageCommand, so users will know what the path is where garbage is generated.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
shodaaan: Fixed.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/GenerateGarbageCommand.java
shodaaan: Fixed.

shodaaan:
> > > Regarding the explicit check, I don't think this is necessary
> > 
> > 
> > Fair enough.
> > Two unrelated additional comments:
> > 
> > * the DocumentNodeStore is not properly disposed. You can see this by running the [oak-run](https://issues.apache.org/jira/browse/OAK-run) command consecutively : the second start will have to wait 2min for the lease timeout of the first non-gracefully shut down run. We should ensure a graceful disposal.
> > * what might be useful is to add the base path where the garbage is generated to - such that the user immediately knows (and doesn't eg have to consult the source code)
> 
> I added documentNodeStore.dispose() calls to CreateGarbageCommand. Thank you for the suggestion
> 
> The root path where garbage is generated (/tmp/[oak-run-generated-test-garbage](https://issues.apache.org/jira/browse/OAK-run-generated-test-garbage)) is specified in the USAGE string of CreateGarbageCommand, so users will know what the path is where garbage is generated.


I added proper disposal of DocumentNodeStore for both createGarbage and cleanupGarbage commands. Tested oak-run with repeated runs and disposal works correctly.

I also added some println messages with the base path of the generated garbage, as well as the paths of the documents deleted by the "clean" sub-command.

oak-run/src/test/java/org/apache/jackrabbit/oak/plugins/document/CreateGarbageCommandTest.java
```diff
@@ -0,0 +1,437 @@
...
+
+        private final ImmutableList<String> args;
+        private CreateGarbageCommand command;
+        private Closer closer;
+        private List<String> generatedBasePaths;
+
+        public CreateGarbageCmd(String... args) {
+            // append the default mongodb connection string if one was not provided
+            if (args[0].startsWith("mongodb://")) {
+                this.args = ImmutableList.copyOf(args);
```
rishabhdaim: please avoid the usage of guava collections. We have this OAK-10685 epic to remove guava from Oak.

You can use `List.copyOf` with only the caveat that it doesn't allow `null` values.

oak-run/src/test/java/org/apache/jackrabbit/oak/plugins/document/CreateGarbageCommandTest.java
```diff
@@ -0,0 +1,437 @@
...
+            } else {
+                this.args = ImmutableList.<String>builder().add(MongoUtils.URL)
```
rishabhdaim: same as above.

oak-run/src/main/java/org/apache/jackrabbit/oak/plugins/document/GenerateGarbageHelper.java
```diff
@@ -0,0 +1,37 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.jackrabbit.oak.plugins.document;
+
+/**
+ * Helper class to generate garbage for testing purposes.
+ */
+public class GenerateGarbageHelper {
```
rishabhdaim:
```suggestion
public class CreateGarbageHelper {
```

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
```diff
@@ -0,0 +1,574 @@
...
+import java.util.concurrent.TimeUnit;
+
+import static org.apache.jackrabbit.oak.api.Type.NAME;
+import static org.apache.jackrabbit.oak.api.Type.STRING;
+import static org.apache.jackrabbit.oak.plugins.document.util.Utils.getIdFromPath;
+import static org.apache.jackrabbit.oak.run.Utils.asCloseable;
+import static org.apache.jackrabbit.oak.run.Utils.createDocumentMKBuilder;
+
+/**
+ * GenerateGarbageCommand generates garbage nodes in the repository in order to allow for testing fullGC functionality.
```
rishabhdaim:
```suggestion
 * CreateGarbageCommand creates garbage nodes in the repository to allow for testing fullGC functionality.
```

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
```diff
@@ -0,0 +1,574 @@
...
+public class CreateGarbageCommand implements Command, Closeable {
+    private static final Logger LOG = LoggerFactory.getLogger(CreateGarbageCommand.class);
+
+    private final ScheduledExecutorService continuousRunExecutor = Executors.newScheduledThreadPool(1);
+
+    private static final String USAGE = Joiner.on(System.lineSeparator()).join(
+            "create-test-garbage {<jdbc-uri> | <mongodb-uri>} <sub-command> [options] ",
+            "where sub-command is one of: ",
+            "  clean - clean up all generated garbage under the specified root node",
+            "  generate - generate garbage nodes in the repository, under the root node tmp/oak-run-generated-test-garbage. ",
```
rishabhdaim:
```suggestion
            "  create - create garbage nodes in the repository, under the root node tmp/oak-run-created-test-garbage. ",
```

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
```diff
@@ -0,0 +1,574 @@
...
+            "           Use the --garbageType (required) option to specify the type of garbage to generate",
```
rishabhdaim:
```suggestion
            "           Use the --garbageType (required) option to specify the type of garbage to create",
```

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
```diff
@@ -0,0 +1,574 @@
...
+        return documentNodeStore;
+    }
+
+    private static class CreateGarbageOptions extends Utils.NodeStoreOptions {
+
+        /**
+         * Sub-command for generating garbage.
+         * This is the default sub-command to run if none is specified.
+         */
+        static final String CMD_GENERATE = "generate";
```
rishabhdaim:
```suggestion
        static final String CMD_CREATE = "create";
```

rishabhdaim (changes requested):
I would replace the usage of `generate` with `create`

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
```diff
@@ -0,0 +1,574 @@
...
+        public int getNumberOfRuns() {
+            return numberOfRuns.value(options);
+        }
+
+        public int getGenerateIntervalSeconds() {
+            return generateIntervalSeconds.value(options);
+        }
+    }
+
+    public void execute(String... args) throws Exception {
```
rishabhdaim, line 206: missing `@Override` annotation.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
```diff
@@ -0,0 +1,574 @@
...
+            cleanGarbage(options, closer);
+        } else {
+            System.err.println("Unknown revisions command: " + subCmd);
+        }
+
+        return generateBasePaths;
+    }
+
+    @Override
+    public void close() throws IOException {
```
rishabhdaim: We can safely remove this since we are registering `continuousRunExecutor` with a `closer` instance along with the `Closeable` interface.

oak-run/src/test/java/org/apache/jackrabbit/oak/plugins/document/CreateGarbageCommandTest.java
shodaaan: Done, thank you for the suggestion.

oak-run/src/test/java/org/apache/jackrabbit/oak/plugins/document/CreateGarbageCommandTest.java
shodaaan: Done, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/plugins/document/GenerateGarbageHelper.java
shodaaan: Done, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
shodaaan: Done, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
shodaaan: Done, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
shodaaan: Done, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
shodaaan: Done, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
shodaaan: Done, thank you for the suggestion.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
shodaaan: Done, thank you for the suggestion.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/FullGCHelper.java
```diff
@@ -70,6 +70,14 @@ public static RevisionVector unmergedBranchCommit(final NodeStore store, final C
...
     }
 
+    public static boolean includesEmptyProps(int fullGCMode) {
+        return fullGCMode == 1;
+    }
+
+    public static boolean includesGapOrphans(int fullGCMode) {
+        return fullGCMode == 2;
+    }
+
```
stefan-egli: these methods are not used. would suggest to either remove them or consolidate them with the same in CreateGarbageHelper

oak-run/src/main/java/org/apache/jackrabbit/oak/plugins/document/CreateGarbageHelper.java
```diff
@@ -0,0 +1,37 @@
...
+        return fullGCMode == 0;
+    }
+
+    public static boolean isEmptyProps(int fullGCMode) {
+        return fullGCMode == 1;
+    }
+
+    public static boolean isGapOrphans(int fullGCMode) {
+        return fullGCMode == 2;
+    }
```
stefan-egli, line 36: I think it's ok to have these for now (except they are duplicated by FullGCHelper currently).

I do think though that we could move them to oak-store-document's VersionGarbageCollector - or actually make the FullGCMode in there a bit more than just an enum (or some other reasonable refactoring).

but I think the refactoring can be done later and doesn't have to block this PR

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
```diff
@@ -0,0 +1,568 @@
...
+
+    /**
+     * Recursively get all paths of the tree nodes under the given root node.
+     * If there are any gaps in the tree, the paths of the gap nodes will NOT be included.
+     * @param rootNode
+     * @param basePath
+     * @param treeNodePaths
+     */
+    private void getTreeNodePaths(NodeBuilder rootNode, String basePath,
+                                  List<String> treeNodePaths) {
```
stefan-egli: not sure this is a `get` method? what about a `buildTreeNodePaths`?

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
```diff
@@ -0,0 +1,568 @@
...
+                    if (!continuousRunExecutor.awaitTermination(5, TimeUnit.SECONDS)) {
+                        continuousRunExecutor.shutdownNow();
+                    }
+                } catch (InterruptedException e) {
+                    throw new RuntimeException(e);
+                }
+            });
+
+            execute(closer, args);
+
```
stefan-egli: when used in continuous mode (eg by setting `--numberOfRuns 42`) then the code here terminates early and closes the Closer - preventing the continuous mode from practically ever running

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
```diff
@@ -0,0 +1,568 @@
...
+        if (builder == null) {
+            System.err.println("generateGarbage mode only available for DocumentNodeStore");
+            System.exit(1);
+        }
+
+        String generationBasePath = GEN_BASE_PATH + timestamp + "_" + runIndex;
+        documentNodeStore = builder.build();
+        closer.register(asCloseable(documentNodeStore));
+
+        NodeBuilder rootNode = documentNodeStore.getRoot().builder();
```
stefan-egli: nitpick : it could be somewhat confusing if a `NodeBuilder` is called `somethingNode`. Usually we call it `somethingBuilder` or `something` (as `somethingNode` is typically used for a `NodeState`)
(elsewhere I see you're also using the classic terminology, so this applies only to part of the code)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -168,7 +168,7 @@ public class VersionGarbageCollector {
      * Ultimately the goal is to clean up all possible garbage. After hardening these modes
      * might no longer be supported.
      */
-    enum FullGCMode {
+     enum FullGCMode {
```
stefan-egli: nitpick, detail, but I would remove this, it is an unnecessary change

stefan-egli (commented):
Had troubles getting the continuous mode running and I believe it's broken because of the new closer logic closing it early.

(other than that added a few detail remarks as well)

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
shodaaan: Fixed, thank you for the comment.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
shodaaan: Fixed, thank you for the comment.

oak-run/src/main/java/org/apache/jackrabbit/oak/plugins/document/CreateGarbageHelper.java
shodaaan: Fixed, thank you for the comment.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/CreateGarbageCommand.java
shodaaan: Fixed, thank you for the comment.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
shodaaan: Fixed, thank you for the comment.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/FullGCHelper.java
shodaaan: Fixed, thank you for the comment.

shodaaan:
> Had troubles getting the continuous mode running and I believe it's broken because of the new closer logic closing it early.
> 
> (other than that added a few detail remarks as well)

Tested oak-run with all modes (including continuous generation), everything works fine now.

stefan-egli (approved):
+1 lgtm, started the jenkins run, would suggest to go ahead with merge unless we find a regression (unlikely), and then we can iterate from there

stefan-egli:
> would suggest to go ahead with merge unless we find a regression

**squash merge** obviously!

PR #1808, opened 2024-10-18
OAK-11210: add util to convert iterable to arrayDeque

t-rana:


rishabhdaim (approved):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/collections/package-info.java
```diff
@@ -21,7 +21,7 @@
  * Utilities for Java collections and streams.
  */
 @Internal(since = "1.0.0")
-@Version("1.2.0")
+@Version("1.3.0")
```
rishabhdaim: This shouldn't be required.
Could you please create PR without updating the baseline version?

t-rana: Done.

PR #1811, opened 2024-10-18
OAK-11171:  Added cached external membership benchmarks

Amoratinos:
## Summary

Created new benchmark to test external membership login with external groups and user added to local groups to compare the performance using cache.


anchela (commented):
hi @Amoratinos , can you limit this PR to changes to benchmarks only? otherwise i would rather merge it into the general fix PR. wdyt?

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/PrincipalMembershipReaderImpl.java
anchela, line 1: this class should not be part of the benchmark PR, should it?

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/principal/PrincipalMembershipReaderFactory.java
anchela, line 1: also the interface in security-spi should not be needed for adding/changing benchmarks. is suspect that's just a left-over from some other refactoring. no?

anchela (changes requested):
hi @Amoratinos , there are 2 new classes (one in oak-core and one in security-spi) that imho should not be part of this PR, which just adds/modifies benchmarks. i suspect that's only a left over from some other refactoring. can you check?

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/PrincipalMembershipReaderImpl.java
Amoratinos: Yes, as I rebased on trunk this was a left over

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/principal/PrincipalMembershipReaderFactory.java
Amoratinos: Same here, as you mentioned a left-over after the rebase

anchela (approved):
lgtm! thanks a lot.

PR #1812, opened 2024-10-18
OAK-11104 : removed usage of Preconditions.checkState

rishabhdaim:


oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/conditions/Validate.java
```diff
@@ -119,9 +119,9 @@ public static void checkState(final boolean expression) {
      * @param errorMessage message to use in exception
      * @throws IllegalStateException if expression is false
      */
-    public static void checkState(final boolean expression, @NotNull final String errorMessage) {
+    public static void checkState(final boolean expression, @NotNull final Object errorMessage) {
```
rishabhdaim, line 122: This is required because in some places the string we pass is not initialized until the expression has run.

reschke (approved):
I trust this change was a) purely mechanical, and b) tested...

PR #1814, opened 2024-10-21
OAK-11203 : added support for delayFactor, progress/batch size overri…

rishabhdaim:
…de from oak-run for full GC

### DIFF
--- a/oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
+++ b/oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
@@ -73,8 +73,6 @@ import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreHelper
 import static org.apache.jackrabbit.oak.plugins.document.FormatVersion.versionOf;
 import static org.apache.jackrabbit.oak.plugins.document.util.Utils.getIdFromPath;
 import static org.apache.jackrabbit.oak.plugins.document.util.Utils.getRootDocument;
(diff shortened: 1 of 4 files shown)


oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -146,6 +147,10 @@ private static class RevisionsOptions extends Utils.NodeStoreOptions {
             delay = parser
                     .accepts("delay", "introduce delays to reduce impact on system").withRequiredArg()
                     .ofType(Double.class).defaultsTo(0.0);
+            fullGcDelayFactor = parser
+                    .accepts("fullGcDelayFactor", "introduce delays after each fullGc batch to reduce load " +
+                            "on system")
+                    .withRequiredArg().ofType(Double.class).defaultsTo(0.0);
```
stefan-egli: as this will now be a command line parameter, we could use a more conservative value here. this would protect against unknowingly overloading a production system.

rishabhdaim: let's use the same value as in OSGi configurations.

rishabhdaim: addressed in https://github.com/apache/jackrabbit-oak/pull/1814/commits/8cb54bf67b1a9ffa24e5017fb43322887f4aa287

stefan-egli (approved):

stefan-egli:
merging as discussed offline

PR #1815, opened 2024-10-21
OAK-11218: Array containing empty String should parse as empty Array

Joscorbe:
Otherwise, it entirely disables FullGC.

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
@@ -326,7 +326,7 @@ public class DocumentNodeStoreBuilder<T extends DocumentNodeStoreBuilder<T>> {
     }
 
     public T setFullGCExcludePaths(@Nullable String[] excludePaths) {
-        if (isNull(excludePaths) || excludePaths.length == 0) {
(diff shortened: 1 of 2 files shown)


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -326,7 +326,7 @@ public Set<String> getFullGCIncludePaths() {
     }
 
     public T setFullGCExcludePaths(@Nullable String[] excludePaths) {
-        if (isNull(excludePaths) || excludePaths.length == 0) {
+        if (isNull(excludePaths) || excludePaths.length == 0 || Arrays.equals(excludePaths, new String[]{""})) {
```
daniancu, line 329: I would extract this `new String[]{""}` as a constant field, e.g EMPTY_STRING_ARRAY

daniancu (approved):

rishabhdaim (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
Joscorbe: I think we don't gain much by doing that. I would definitely agree if we could reuse it with the `IncludePaths` method, but the array value is different.

PR #1820, opened 2024-10-24
OAK-11222: Remove usage of Guava newArrayList (no param variant)

reschke:


reschke:
As this is a purely mechanical edit which does affect a ton of files, I'll merge this sonnish.

PR #1824, opened 2024-10-27
OAK-11231: Few test cases reproducing OAK-11231 bug

Joscorbe:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
```diff
@@ -253,9 +253,9 @@ public Iterable<NodeDocument> getModifiedDocs(final long fromModified, final lon
         // (_modified = fromModified && _id > fromId || _modified > fromModified && _modified < toModified)
         final Bson query = or(
                 withIncludeExcludes(includedPathPrefixes, excludedPathPrefixes,
-                        and(eq(MODIFIED_IN_SECS, getModifiedInSecs(fromModified)), gt(ID, fromId))),
+                        and(eq(MODIFIED_IN_SECS, TimeUnit.MILLISECONDS.toSeconds(fromModified)), gt(ID, fromId))),
```
rishabhdaim: I don't think this is a correct fix. In the `document store`, we round off `_modified` to the nearest 5 sec value.
@stefan-egli wdyt?

stefan-egli: Right, by default it is rounded. The question is whether there is a guarantee that all `_modified` are rounded. IIUC then in this case a document was injected with an unrounded `_modified` - in which case it resulted in a unnecessary large number of iterations over the same few documents.

So we could say : there are never any unrounded `_modified` and then not need the above. Or we could say : if the above is corrected irrespective of whether `_modified` is rounded or not, then it could be a useful fix to have in any case.

Joscorbe: I think the fix is ready, I changed some  I'm just adding some warnings to log some edge cases, but the core of the fix should remain the same. @rishabhdaim I found examples of documents not rounded, and in those cases they are iterated infinitely (up to 10k times of the while loop) because the second part of the query returns everything in the 5 second time window, regardless of any previous progress.

The fix essentially removes the rounding when we are using a fromId, that means, when the modifiedFrom is taken from an existing document in the repository. In that case, I think we shouldn't do any rounding and take the value directly from the document to ensure there is progress.

rishabhdaim (approved):

PR #1829, opened 2024-10-29
OAK-11234 - Use virtual clock on Pipelined IT tests.

nfsantos:
This is required to better test recovery from failures in the download phase of the Pipeline strategy. The tests should verify that the download does not miss any field, regardless of the `_modified` value. Without a virtual clock, it is not practical to create test content with different `_modified` values, as that would require waiting for more than 5 seconds between updates, as this is the granularity of the field. With a virtual clock, we can advance the clock by arbitrary amounts.

### DIFF
--- a/oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedTreeStoreIT.java
+++ b/oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedTreeStoreIT.java
@@ -68,6 +68,7 @@ import org.apache.jackrabbit.oak.spi.commit.CommitInfo;
 import org.apache.jackrabbit.oak.spi.commit.EmptyHook;
 import org.apache.jackrabbit.oak.spi.filter.PathFilter;
 import org.apache.jackrabbit.oak.spi.state.NodeBuilder;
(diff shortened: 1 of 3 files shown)


oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/PipelinedTreeStoreIT.java
```diff
@@ -364,11 +365,12 @@ public void getOnlyModified() throws Exception {
...
             contentDamBuilder.child("change.png").setProperty("test", 0);
             rwNodeStore.merge(rootBuilder, EmptyHook.INSTANCE, CommitInfo.EMPTY);
-            long oldMod = NodeDocument.getModifiedInSecs(System.currentTimeMillis());
+            Clock clock = rwNodeStore.getClock();
+            long oldMod = NodeDocument.getModifiedInSecs(clock.getTime());
             // wait until the modified time changes
             do {
-                Thread.sleep(10);
-                minModified = NodeDocument.getModifiedInSecs(System.currentTimeMillis());
+                clock.waitUntil(clock.getTime() + 1000);
```
thomasmueller, line 372: Great! I didn't think about using the virtual clock. I wasn't aware waitUntil increments the counter...

thomasmueller (approved):

fabriziofortino (approved):

PR #1836, opened 2024-10-31
OAK-11239: Remove usage of Guava newArrayList(vargargs) in test classes

reschke:


oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/basic/DefaultSyncContextTest.java
```diff
@@ -1459,7 +1459,7 @@ public void testCreateValuesEmptyCollection() throws Exception {
 
     @Test
     public void testCreateValuesSkipsNull() throws Exception {
-        List<String> strings = Lists.newArrayList("s", null, null, "t");
+        List<String> strings = new ArrayList<>(Arrays.asList("s", null, null, "t"));
```
mbaedke, line 1462: Why not just `List.of("s", null, null, "t")` ?

reschke, line 1462: Because "List.of" does not suppoet null. See https://issues.apache.org/jira/browse/OAK-11239

mbaedke (approved):
LGTM, though I think that at some places List.of() could be used instead of an ArrayList. Maybe I'm wrong, I don't think it matters much in test cases.

reschke:
> I think that at some places List.of() could be used instead of an ArrayList. 

I actually tested that.

PR #1837, opened 2024-10-31
OAK-11233: Don't apply MongoDockerRule if the Mongo Docker image isn't actually used.

mbaedke:
Implemented.


reschke:
I assume you tested the three cases?

- no local Mongo nor Docker?
- local Mongo available
- Docker available

?

reschke (approved):

mbaedke:
Yes, I tested that.

PR #1838, opened 2024-10-31
OAK-11242 - AzureRepositoryLock: Thread should not die silently on any kind of error

ahanikel:
This is a follow-up to [OAK-11069](https://issues.apache.org/jira/browse/OAK-11069): Apparently the refresher thread still dies without a trace. I'm therefore assuming some java.lang.Error occurs, which is hidden behind the executor. An approach could be to avoid the executor in favour of creating the thread ourselves, and catching all Throwables to make sure we handle them. We leave fatal errors such as OOME to pop up in other parts of the code.


smiroslav (approved):

smiroslav:
There is a build failure in oak-store-document but it is unrelated to the PR change.

smiroslav:
Created dedicated issue 
https://issues.apache.org/jira/browse/OAK-11249

PR #1851, opened 2024-11-06
OAK-11247: Migrated groovy test files to Java

Amoratinos:
Migrated Groovy test files in oak-pojosr package to Java.

Asserts were changed to use junit assert methods imported statically.


anchela (approved):
hi @Amoratinos , i just looked at the SecurityProviderRegistrationTest and that one looks good if we conclude that we want to keep the existing module and the way it's running tests... which is probably what we can achieve with limited effort.
an alternative approach would be to rewrite the intention of the tests within the oak modules they belong to, get rid of the mocking and use PaxExam to build ITs that accurately reflect running oak in a OSGi setup like we have it in Adobe AEM. but maybe that's too much of an effort compared to the gain.

Amoratinos:
> use PaxExam to build ITs that accurately reflect running oak in a OSGi setup like we have it in Adobe AEM. but maybe that's too much of an effort compared to the gain.

I'm fine to take that task with me but I think that's a broader scope of the task. Specially from reviewer's point of view it would need to invest more time and effort to go over the changes. If it's fine for you I'll create a new ticket asking to change that test to use PaxExam.

reschke (approved):
I really can't review.

It the Groovy dependency is gone, and integration tests are passing ("-PintegrationTesting"), this should be good.

reschke:
@anchela  - paxExam is abandoned; we should not use it in "new" code (see https://mvnrepository.com/artifact/org.ops4j.pax.exam/pax-exam)

Amoratinos:
@reschke I suppose you wanted to mention @anchela  😅 

At least the pax exam repo seems somehow mantained https://github.com/ops4j/org.ops4j.pax.exam/activity 

I ran the integrationTesting locally and it went through without issues

reschke:
@anchela - time to merge?

anchela:
yep

PR #1858, opened 2024-11-08
OAK-11260 Speed up Lucene cost estimation

thomasmueller:


oak-lucene/src/test/java/org/apache/jackrabbit/oak/jcr/query/ManyIndexesAndUnionPlanTest.java
```diff
@@ -0,0 +1,94 @@
...
+import javax.jcr.query.Row;
+import javax.jcr.query.RowIterator;
+
+import org.apache.jackrabbit.core.query.AbstractQueryTest;
+import org.apache.jackrabbit.oak.plugins.index.search.FulltextIndexConstants;
+import org.apache.jackrabbit.oak.plugins.index.search.IndexFormatVersion;
+
+public class ManyIndexesAndUnionPlanTest extends AbstractQueryTest {
+
+    public void testResultSize() throws Exception {
```
fabriziofortino, line 36: should we add a `@Test` annotation here?

thomasmueller: On Eclipse, the test is run even without annotation... but I'm not sure about Maven. I'll add it!

fabriziofortino (approved):

PR #1863, opened 2024-11-13
OAK-11271 Execute Sonar analysis in a dedicated job

kwin:


.github/workflows/build.yml
```diff
@@ -24,34 +24,61 @@ on:
...
       - name: Cache Maven packages
-        uses: actions/cache@v3.2.6
+        uses: actions/cache@v4
         with:
           path: ~/.m2
           key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-m2
-      - name: Build and analyze
+      - name: Build
+        run: mvn -B verify -Pcoverage
```
kwin: In order to fully replace ASF Jenkins we probably need to add profile `integrationTesting` (https://github.com/apache/jackrabbit-oak/blob/4a03d67dffd5619c41f5b54ccacfa49e1eb57daf/Jenkinsfile#L29C5-L43C6), however I would refrain from building each module individually but rather build the whole reactor with that profile.

Also `-DtrimStackTrace=false -Dnsfixtures=SEGMENT_TAR,DOCUMENT_NS -Dmongo.db=MongoMKDB-${MONGODB_SUFFIX}` would need to be added, right?

PR #1865, opened 2024-11-14
OAK-11269 add dedicated workflow for checking commit messages

kwin:
OAK-11269 my test

### DIFF
--- a/.commit-check.yml
+++ b/.commit-check.yml
@@ -1,0 +1,17 @@
+# format as outlined in https://github.com/commit-check/commit-check/blob/main/.commit-check.yml
+# further information in https://github.com/commit-check/commit-check/blob/main/README.rst
+checks:
+  - check: message
+    regex: '^OAK-\d+\s\S+.*'
+    error: "The commit message must start with 'OAK-<ID> ' followed by some descriptive text"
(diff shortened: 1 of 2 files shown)


Joscorbe:
This is a good idea, but got a question, will this check enforce the OAK-<ID> also in commits created by bots? I think that will fail in PRs like #1860

kwin:
For now it is just a check which fails (on every PR which is not compliant). However there is nothing enforced right now, i.e. a failed check will not prevent merging via UI.

Joscorbe:
> For now it is just a check which fails (on every PR which is not compliant). However there is nothing enforced right now, i.e. a failed check will not prevent merging via UI.

Sounds good, thanks!

reschke:
@Joscorbe - re automatically generated PRs (such as from dependabot): these are useful, but IMHO we still should open a ticket (release notes!) and test things properly, so I would still prefer *us* creating the PR. In your example, the PR seems simple enough, but in fact breaks baseline, see https://issues.apache.org/jira/browse/OAK-11286?focusedCommentId=17901680&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17901680

reschke (approved):

PR #1878, opened 2024-11-26
OAK-11285 - Check if paths used in included/excluded Paths are absolute and fail if they are not

nfsantos:
The `includedPaths` and `excludedPaths` in an index definition must be absolute. If they are not or if they start with a space, then the indexing job may fail or hang. This PR adds a check to the `PathFilter` constructor to check that all paths are absolute and fail immediately if they are not.

### DIFF
--- a/oak-store-spi/src/main/java/org/apache/jackrabbit/oak/spi/filter/PathFilter.java
+++ b/oak-store-spi/src/main/java/org/apache/jackrabbit/oak/spi/filter/PathFilter.java
@@ -122,7 +122,9 @@ public class PathFilter {
      * @param includes list of paths which should be included
      * @param excludes list of paths which should not be included
      */
-    public PathFilter(Iterable<String> includes, Iterable<String> excludes) {
+    public PathFilter(@NotNull Iterable<String> includes, @NotNull Iterable<String> excludes) {
+        checkPathsAreAbsolute(includes, "included");
+        checkPathsAreAbsolute(excludes, "excluded");
         Set<String> includeCopy = CollectionUtils.toSet(includes);
         Set<String> excludeCopy = CollectionUtils.toSet(excludes);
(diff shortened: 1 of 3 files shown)


oak-store-spi/src/main/java/org/apache/jackrabbit/oak/spi/filter/PathFilter.java
```diff
@@ -195,4 +197,12 @@ public boolean areAllDescendantsIncluded(String path) {
         }
         return false;
     }
+
+    private static void checkPathsAreAbsolute(Iterable<String> paths, String pathType) {
+        for (String path : paths) {
+            if (!PathUtils.isAbsolute(path)) {
+                throw new IllegalStateException("Invalid path in " + pathType + " paths list: " + path + ". Paths must be absolute.");
```
nit0906, line 204: would it be better to throw an IllegalArgumentException instead ?

steffenvan:
Should we add a test case where one of the paths start with a space? On another note, I thought we trimmed those string values?

oak-store-spi/src/main/java/org/apache/jackrabbit/oak/spi/filter/PathFilter.java
nfsantos: I agree it would be better, but the PathFilter constructor is already throwing `IllegalStateException` in a check it does in the constructor, so I prefer to keep it consistent and use the same exception.

nfsantos:
> Should we add a test case where one of the paths start with a space? On another note, I thought we trimmed those string values?

One of the tests is already testing paths that start with space. Let me know if you think we should add more tests.

steffenvan:
Oh right my eyes must have deceived me. That one space character can be quite subtle, I can understand why someone might miss it

fabriziofortino (approved):

PR #1882, opened 2024-11-28
OAK-10466: Allow to configure anonymous user disablement

Amoratinos:
Created a configuration value that allows to prevent anonymous user to be disabled. 

To support backward compatibility be default anonymous user can still be disabled

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserImpl.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserImpl.java
@@ -25,6 +25,7 @@ import org.apache.jackrabbit.api.security.user.User;
 import org.apache.jackrabbit.oak.api.PropertyState;
 import org.apache.jackrabbit.oak.api.Tree;
 import org.apache.jackrabbit.oak.namepath.NamePathMapper;
+import org.apache.jackrabbit.oak.spi.security.ConfigurationParameters;
 import org.apache.jackrabbit.oak.spi.security.user.AuthorizableType;
 import org.apache.jackrabbit.oak.spi.security.user.UserConstants;
 import org.apache.jackrabbit.oak.spi.security.user.UserIdCredentials;
@@ -44,15 +45,15 @@ class UserImpl extends AuthorizableImpl implements User {
     private final boolean isAdmin;
     private final boolean isAnonymous;
     private final PasswordHistory pwHistory;
-    private final boolean allowDisableAnonymous;
+    private final ConfigurationParameters configurationParameters;
 
     UserImpl(String id, Tree tree, UserManagerImpl userManager) throws RepositoryException {
         super(id, tree, userManager);
 
-        isAdmin = UserUtil.isAdmin(userManager.getConfig(), id);
-        isAnonymous = UserUtil.getAnonymousId(userManager.getConfig()).equals(id);
-        allowDisableAnonymous = userManager.getConfig().getConfigValue(UserConstants.PARAM_ALLOW_DISABLE_ANONYMOUS, true);
-        pwHistory = new PasswordHistory(userManager.getConfig());
+        configurationParameters = userManager.getConfig();
+        isAdmin = UserUtil.isAdmin(configurationParameters, id);
+        isAnonymous = UserUtil.getAnonymousId(configurationParameters).equals(id);
+        pwHistory = new PasswordHistory(configurationParameters);
     }
 
     //---------------------------------------------------< AuthorizableImpl >---
@@ -135,7 +136,7 @@ class UserImpl extends AuthorizableImpl implements User {
 
     @Override
     public void disable(@Nullable String reason) throws RepositoryException {
-        canDisableUser();
+        validateDisableUser();
 
         getUserManager().onDisable(this, reason);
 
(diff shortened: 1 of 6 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserImpl.java
```diff
@@ -42,12 +42,16 @@
...
+    private final boolean isAnonymous;
     private final PasswordHistory pwHistory;
+    private final boolean allowDisableAnonymous;
 
     UserImpl(String id, Tree tree, UserManagerImpl userManager) throws RepositoryException {
         super(id, tree, userManager);
 
         isAdmin = UserUtil.isAdmin(userManager.getConfig(), id);
+        isAnonymous = UserUtil.getAnonymousId(userManager.getConfig()).equals(id);
+        allowDisableAnonymous = userManager.getConfig().getConfigValue(UserConstants.PARAM_ALLOW_DISABLE_ANONYMOUS, true);
```
joerghoh: Having this in the constructor will impose a small performance penalty on every construction of a user object. Is it possible to move this code down into the ``canDisableUser`` method, so it is executed only when it is needed?

oak-core/src/test/java/org/apache/jackrabbit/oak/security/user/UserImplTest.java
```diff
@@ -165,6 +178,36 @@ public void testDisableNullReason() throws Exception {
...
+
+        cleanUserManager();
+        before();
+
+        User anonymous = getAnonymousUser();
+        assertFalse(anonymous.isDisabled());
+
+        anonymous.disable("Test anonymous disable");
+
+        fail("Shouldn'td have reached this point");
```
joerghoh:
```suggestion
        fail("Shouldn't have reached this point");
```

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/user/UserConstants.java
```diff
@@ -265,4 +265,12 @@ public interface UserConstants {
      * @since Oak 1.3.3
      */
     int PASSWORD_HISTORY_DISABLED_SIZE = 0;
+
+    /**
+     * Optional configuration parameter indicating if the anonymous user can be disabled or not.
+     * By default, the anonymous user can be disabled.
+     *
+     * @since Oak 1.73.0
```
joerghoh: there is no Oak 1.73.0; Oak uses the odd-even versioning scheme, so there is an official 1.70.0, 1.72.0, 1.74.0 etc.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserImpl.java
```diff
@@ -148,6 +150,16 @@ public void disable(@Nullable String reason) throws RepositoryException {
         }
     }
 
+    private void canDisableUser() throws RepositoryException {
```
joerghoh: nit: I am not super-happy with the name of this method, as it makes me think that it returns a ``boolean``. But I don't have a good name for it either.
(Feel free to ignore this, if you disagree with it.)

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserImpl.java
Amoratinos: I also considered it but in that case I'd have to keep a reference to the configuration parameters on each UserImpl instance. Not sure if this could have a performance impact as well.

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/user/UserConstants.java
Amoratinos: Ok, I wasn't aware of this. I took it based from the pom.xml. <s>Will change it to 1.74 then</s> Not sure if it's required or not or if could be removed

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserImpl.java
Amoratinos: I'm also not happy with it. Initially I call it `validateCanDisableUser` but again I dislike it.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserImpl.java
anchela: @Amoratinos , it should be possible to retrieve the configuration params on the fly.... the usermanager field stored with the base class is an instance of UserManagerImpl, which comes with a method getConfig(), which are the config params obtained from UserConfiguration afaik.

so, i believe it's possible to move the evaluation to the validateDisableUser method and avoiding story isAnonymous _and_ avoiding storing the config params again as they are already attached to the user manager

Amoratinos: You're right, changed to use getUserManager and removed those fields in favor of online evaluation.

anchela (approved):
hi @Amoratinos , lgtm. thank you

anchela:
@joerghoh , can you take another look to see if your concerns are addressed? ty.

PR #1888, opened 2024-12-04
OAK-11298 Fix test IncrementalStoreTest

thomasmueller:
* Ensure there is at most one property per node.
* Fix compiler warnings about generics.


amit-jain (approved):
lgtm

PR #1891, opened 2024-12-09
OAK-11199 getSubject is supported only if a security manager is allowed

mbaedke:


reschke (changes requested):
a) Compat*i*bility
b) The new package in commons would need a package-info

Otherwise, looks good.

reschke:
Recommendation:

- new package commons.jdkcompat (non-public, "internal" in package-info), export in pom.xml
- name of util class more specific, like Java23Security

mbaedke:
Don't know about the name Java23Security. It's certainly not more specific :)

reschke (approved):
The changes in commons now look good.

The other parts of the PR are minimal, which is good.

I haven't tested, I assume you did, @mbaedke 

The one thing that came to mind is that if we change Java23Security to "Subject", changes would be even less intrusive. Or even "Java23Subject"?  (this is bikeshedding, ignore if you want to)

reschke (commented):
And maybe we should deal with

https://sonarcloud.io/component_measures?id=org.apache.jackrabbit%3Ajackrabbit-oak&pullRequest=1891&metric=new_coverage&view=list

mbaedke:
I like Java23Subject. I wouldn't call it Subject at least for the reason that the method signatures are slightly different, which might confuse readers. Of course I tested, but will run integration tests again and look into the coverage issue before merging.

oak-security-spi/pom.xml
```diff
@@ -109,6 +109,11 @@
       <artifactId>oak-jackrabbit-api</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
```
reschke, line 112: Duplicated dependency.

PR #1892, opened 2024-12-10
OAK-11299: Missing Segments With Oak Run Segment Copy

lweitzendorf:


oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/tool/SegmentStoreMigrator.java
```diff
@@ -162,9 +162,13 @@ private void migrateArchives() throws IOException, ExecutionException, Interrupt
         List<String> targetArchives = targetManager.listArchives();
 
         if (appendMode && !targetArchives.isEmpty()) {
-            //last archive can be updated since last copy and needs to be recopied
-            String lastArchive = targetArchives.get(targetArchives.size() - 1);
-            targetArchives.remove(lastArchive);
+            // last archive could have been updated since last copy and needs to be recopied
+            try {
+                targetArchives.sort(String::compareTo);
```
mreutegg: Sorting lexicographically only work with up to 5 digits in the archive name.
```
data00001a.tar
data33985b.tar
data99999a.tar
```
But the archive name permits more digits up to an archive number of `Integer.MAX_VALUE`.

https://github.com/apache/jackrabbit-oak/blob/5b13b559dc9d31a20676353782abff7c96e59f13/oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/tar/TarFiles.java#L261

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/tool/SegmentStoreMigrator.java
```diff
@@ -162,9 +162,13 @@ private void migrateArchives() throws IOException, ExecutionException, Interrupt
...
+                targetArchives.remove(targetArchives.size() - 1);
+            } catch (UnsupportedOperationException e) {
```
mreutegg: I assume `UnsupportedOperationException` is caught here because the `List` returned by `targetManager.listArchives()` may be immutable. I would rather copy the returned archive names in a new `ArrayList`.

mreutegg (changes requested):
I would also like to see unit tests that validate the changes in this PR.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/tool/SegmentStoreMigrator.java
lweitzendorf: It's a bit odd. It returns an ArrayList, but `java.util.Arrays.ArrayList` instead of `java.util.ArrayList` and the former doesn't support any resizing operation.

lweitzendorf: But more generally, yes the list may be fully immutable or some other type that doesn't allow resizing. Why copy it into an ArrayList instead of just always using subList for example?

mreutegg: > But more generally, yes the list may be fully immutable or some other type that doesn't allow resizing. Why copy it into an ArrayList instead of just always using subList for example?

Getting a sub list from an immutable list should work, but SegmentStoreMigrator also sorts the list and that is not possible when the list is immutable.

mreutegg: There is also similar code in `AwsSegmentStoreMigrator` that needs the same fix: https://github.com/apache/jackrabbit-oak/blob/69ee5b5034275b981943b8eec871c27def94cb24/oak-segment-aws/src/main/java/org/apache/jackrabbit/oak/segment/aws/tool/AwsSegmentStoreMigrator.java#L157-L161

mreutegg: So, probably makes sense to move this logic into a utility method. At least sorting the archive names.

lweitzendorf: They share a lot of code in general. Pulling the common logic out into a base class for both might be a good option.

mreutegg: This PR is now pretty big and includes quite some refactoring. Can you please split the changes into what is necessary to fix OAK-11299 and a new issue that covers the refactoring that is not essential for OAK-11299?

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/RemoteUtilities.java
```diff
@@ -50,4 +51,17 @@ public static UUID getSegmentUUID(@NotNull String segmentFileName) {
         }
         return UUID.fromString(m.group(2));
     }
+
+    public static class ArchiveIndexComparator implements Comparator<String> {
```
mreutegg: Looks good. Though, I would not expose the inner class. You could make it private and available as a public static field of type `Comparator<String>`.
```
    public static final Comparator<String> ARCHIVE_INDEX_COMPARATOR = new ArchiveIndexComparator();
```

mreutegg (approved):

PR #1903, opened 2024-12-13
OAK-11217 : set existing object meta data for s3 only

t-rana:


oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
```diff
@@ -343,7 +343,10 @@ public void write(DataIdentifier identifier, File file)
                     objectMetaData.getLastModified().getTime());
                 CopyObjectRequest copReq = new CopyObjectRequest(bucket, key,
                     bucket, key);
-                copReq.setNewObjectMetadata(objectMetaData);
+                LOG.warn("Object MetaData before copy: {}", objectMetaData.getRawMetadata());
+                if (properties.getOrDefault(S3Constants.MODE, RemoteStorageMode.S3) == RemoteStorageMode.S3) {
```
rishabhdaim:
```suggestion
                if (Objects.equals(RemoteStorageMode.S3, properties.get(S3Constants.MODE))) {
```

rishabhdaim (approved):

PR #1914, opened 2024-12-20
OAK-11317 : removed usage of ImmutableSet from guava

rishabhdaim:


reschke (commented):
I think it would be good to analyze the context about whether null can be present; in that case we could avoid the use of filter(). Here, this seems to be the case.

This probably applies to more places.

oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/ExternalGroupPrincipalProviderTest.java
```diff
@@ -90,13 +90,13 @@ void sync(@NotNull ExternalUser externalUser) throws Exception {
...
-            return ImmutableSet.copyOf(idp.getUser(userId).getDeclaredGroups()).stream().map(externalIdentityRef -> {
+            return CollectionUtils.toSet(idp.getUser(userId).getDeclaredGroups()).stream().map(externalIdentityRef -> {
                 try {
                     return new PrincipalImpl(idp.getIdentity(externalIdentityRef).getPrincipalName());
                 } catch (ExternalIdentityException e) {
                     throw new RuntimeException(e);
                 }
-            }).collect(Collectors.toSet());
+            }).collect(Collectors.toUnmodifiableSet());
         } else {
```
reschke: There seem to be two conversions to Set here (inherited from earlier).

oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/SyncHandlerMappingTrackerTest.java
```diff
@@ -72,23 +73,23 @@ public void testAddingServiceWithProperties() {
...
+        assertEquals(Set.of("testIDP-2"), Collections.unmodifiableSet(CollectionUtils.toSet(tracker.getIdpNames("testSH-2"))));
 
         ServiceReference ref3 = mock(ServiceReference.class);
         when(ref3.getProperty(PARAM_SYNC_HANDLER_NAME)).thenReturn("testSH");
         when(ref3.getProperty(PARAM_IDP_NAME)).thenReturn("testIDP-3");
         tracker.addingService(ref3);
 
-        assertEquals(Set.of("testIDP", "testIDP-3"), ImmutableSet.copyOf(tracker.getIdpNames("testSH")));
-        assertEquals(Set.of("testIDP-2"), ImmutableSet.copyOf(tracker.getIdpNames("testSH-2")));
+        assertEquals(Set.of("testIDP", "testIDP-3"), Collections.unmodifiableSet(CollectionUtils.toSet(tracker.getIdpNames("testSH"))));
```
reschke: Do we need conversions to immutable when the set id only used in a subsequent use of "assert"? Maybe we can check the diff for uses of unmodifiableSet in the context of asserts?

reschke (approved):
I have reviewed a small part and left a few comments where I can assume that things coulf be simplified; maybe even after a somehat mechanical check (like looking for uses of "unmodifiable" close to an assert).

Other than that, if test are passing,,,

oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/SyncHandlerMappingTrackerTest.java
rishabhdaim: good catch @reschke. I would fix it.

rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/1914/commits/412cad0bd43ffc420df2ecde81ddbfadebeaeade along with other uses of `Collections.unmodifiableSet` in test cases.

oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/ExternalGroupPrincipalProviderTest.java
rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/1914/commits/412cad0bd43ffc420df2ecde81ddbfadebeaeade

rishabhdaim:
> I think it would be good to analyze the context about whether null can be present; in that case we could avoid the use of filter(). Here, this seems to be the case.
> 
> This probably applies to more places.

I tried to use `filter` only where we get data from external sources or where I can't confirm whether null is possible. In case I have missed something, could you please help in pointing that out.

reschke:
> I tried to use filter only where we get data from external sources or where I can't confirm whether null is possible. In case I have missed something, could you please help in pointing that out.

That of course will be a lot of work... Maybe we could check for all uses of "::nonNull" in test cases first; but only later on.

oak-security-spi/src/test/java/org/apache/jackrabbit/oak/spi/security/principal/TestPrincipalProvider.java
```diff
@@ -17,6 +17,8 @@
 package org.apache.jackrabbit.oak.spi.security.principal;
 
 import java.security.Principal;
+import java.util.ArrayList;
```
reschke: https://sonarcloud.io/project/issues?id=org.apache.jackrabbit%3Ajackrabbit-oak&pullRequest=1914&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

rishabhdaim: thanks, fixed in https://github.com/apache/jackrabbit-oak/pull/1914/commits/47073558a422e1b46f1e6a47f101e2e90f20143d

rishabhdaim:
>That of course will be a lot of work... Maybe we could check for all uses of "::nonNull" in test cases first; but only later on.

There is no usage in test files, only 3 usages in code files where we ingest data from outside.

PR #1916, opened 2024-12-20
OAK-11320 Separate JaCoCo execution data files for UTs and ITs

kwin:
Late evaluation of property "test.opts.coverage" as being calculated by jacoco-m-p

Follow example from
https://www.eclemma.org/jacoco/trunk/doc/examples/build/pom-it.xml


kwin:
Not necessary as outlined in https://issues.apache.org/jira/browse/OAK-11320?focusedCommentId=17907472&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17907472

PR #1935, opened 2025-01-02
OAK-9500 Deploy Snapshot artifacts from trunk to Apache Nexus

kwin:


reschke (commented):
Q: does the github workflow run sufficient integration tests so that we are sure that the SNAPSHOT is good enough?

kwin:
I guess so: https://github.com/apache/jackrabbit-oak/blob/a8a14e9ea2faf60b9b2e5934e4cf6898b22f1c6c/.github/workflows/build.yml#L40. Any other IT you can think of?

kwin:
If there are no further concerns, I am gonna merge end of this week.

PR #1955, opened 2025-01-07
 OAK-11357: Create StreamUtils

reschke:
...proof of concept. If we're ok with this, we can replay that with the other types of methods (Map, List, etc.)


rishabhdaim:
I would suggest doing the refactor of `CollectionUtils` after `1.74.0` release.

mbaedke (approved):

PR #1958, opened 2025-01-08
OAK-11360: remove usage of Guava Ints.contains()

reschke:


Amoratinos (approved):
LGTM

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/FieldFactory.java
```diff
@@ -70,12 +72,10 @@ public final class FieldFactory {
         OAK_TYPE_NOT_STORED.setIndexOptions(DOCS_AND_FREQS_AND_POSITIONS);
         OAK_TYPE_NOT_STORED.setTokenized(true);
         OAK_TYPE_NOT_STORED.freeze();
-
-        Arrays.sort(TYPABLE_TAGS);
     }
 
     public static boolean canCreateTypedField(Type<?> type) {
-        return Ints.contains(TYPABLE_TAGS, type.tag());
+        return TYPABLE_TAGS.contains(type.tag());
```
rishabhdaim, line 75:
```suggestion
        return Arrays.stream(TYPABLE_TAGS).anyMatch(tag -> tag == type.tag());
```
and we keep the type of `TYPABLE_TAGS` as `int[]`.

rishabhdaim: Or we could change `TYPABLE_TAGS` to `Set` and take advantage of `Set.contains` API.

@reschke wdyt ?

reschke: I wanted to avoid to convert ints to Integers every time we do the check.

I also assumed that the original author believed that sorting the array/list would help improve performance.

If this is not the case we of course could make it a Set.

rishabhdaim: AFAIK `List.contains` always run in linear time irrespective of whether `List` is sorted or not.

reschke: Yes, sorting only helps if the lower values need to be looked up more frequently.

We could check the code what is really needed here. Or we can just keep things the way they are :-)

rishabhdaim (approved):

mbaedke (approved):

PR #1963, opened 2025-01-08
OAK-11366: Remove usage of Guava UnsignedBytes.compare()

reschke:


oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/SegmentVersion.java
```diff
@@ -56,7 +53,7 @@ public enum SegmentVersion {
      * Latest segment version
      */
     public static final SegmentVersion LATEST_VERSION = max(allOf(SegmentVersion.class),
-        (v1, v2) -> UnsignedBytes.compare(v1.version, v2.version));
+        (v1, v2) -> Integer.compare(v1.version, v2.version));
```
rishabhdaim:
```suggestion
        (v1, v2) -> Integer.compareUnsigned(Byte.toUnsignedInt(v1.version), Byte.toUnsignedInt(v2.version));
```

reschke: But now we don't need "compareUnsigned" anymore, right?

rishabhdaim: I think we should use `compareUnsigned`.

rishabhdaim (approved):

rishabhdaim (approved):

PR #1964, opened 2025-01-08
OAK-11367: Remove usage of Guava Chars.contains()

reschke:


oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndex.java
```diff
@@ -307,7 +307,7 @@ public static String rewriteQueryText(String textsearch) {
                     escaped = false;
                 }
                 rewritten.append(c);
-            } else if (Chars.contains(QUERY_OPERATORS, c)) {
+            } else if (QUERY_OPERATORS.indexOf(c) >= 0) {
```
rishabhdaim, line 310: I would convert `QUERY_OPERATORS` to `Set<Character>` and then use `Set.contains`.

nfsantos: That would be more expensive. A `Set<Character>` requires boxing the elements. And a call to `Set.contains` will box the argument and calculate an hash. A linear search on a character array, as String.indexOf does, uses much less logic per call and will certainly be faster for an array of just 6 elements (all contiguous in memory).

rishabhdaim (approved):

reschke:
trunk: [59912f4fad|https://github.com/apache/jackrabbit-oak/commit/59912f4fad14e2fa920ae7def6b52354ae80b2d1]

PR #1968, opened 2025-01-09
OAK-11370 : add utils in oak-commons to check equality of 2 iterables

t-rana:


oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/collections/CollectionUtils.java
```diff
@@ -520,6 +521,34 @@ public static <T> Stream<T> toStream(@NotNull Iterator<T> iterator) {
...
+        Objects.requireNonNull(iterable2);
+
+        if (iterable1 instanceof Collection && iterable2 instanceof Collection) {
+            Collection<?> collection1 = (Collection<?>) iterable1;
+            Collection<?> collection2 = (Collection<?>) iterable2;
+            if (collection1.size() != collection2.size()) {
+                return false;
+            }
+        }
+
```
reschke, line 545: I would put that into an "else" (return values from the same nesting level).

t-rana: Changes done.

PR #1969, opened 2025-01-09
OAK-11374 : removed usage of Guava's Maps.immutableEntry

rishabhdaim:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/MissingBcSweeper2.java
```diff
@@ -158,7 +157,7 @@ public Map.Entry<Path, UpdateOp> apply(NodeDocument doc) {
                     lastYield = context.getClock().getTime();
                     yieldCnt = 0;
                 }
-                return immutableEntry(doc.getPath(), sweepOne(doc));
+                return new SimpleImmutableEntry<>(doc.getPath(), sweepOne(doc));
```
rishabhdaim, line 160: since `sweepOne` could return a `null` value, we couldn't use `Map.entry` here.

reschke (approved):

PR #1970, opened 2025-01-09
OAK-11375 : removed usage of Guava's Maps.transformValues

rishabhdaim:


reschke (commented):
Not convinced; this has obly two usages, one of which is a test.

reschke (approved):

PR #1992, opened 2025-01-16
OAK-11399: Remove usage of Guava io.BaseEncoding

reschke:


oak-blob/src/main/java/org/apache/jackrabbit/oak/spi/blob/AbstractBlobStore.java
```diff
@@ -119,6 +120,11 @@ public abstract class AbstractBlobStore implements GarbageCollectableBlobStore,
      */
     private int blockSize = 2 * 1024 * 1024;
 
+    /**
+     * Encode in Base 32, hex encoding, no line breaks, padding with "="
+     */
+    private static Base32 BASE32ENCODER = Base32.builder().setLineSeparator(new byte[0]).setPadding((byte)'=').get();
+
```
mbaedke: Isn't there `.setEncodeTable(...)` missing?

rishabhdaim (approved):

oak-blob/src/main/java/org/apache/jackrabbit/oak/spi/blob/AbstractBlobStore.java
reschke: It was.

I now changed to the old-style constructor, because we otherwise would have to speficy the actual values (the constants in the Base32 class are private).

mbaedke: Yes, that's a design flaw in the builder API.

oak-blob/src/main/java/org/apache/jackrabbit/oak/spi/blob/AbstractBlobStore.java
```diff
@@ -237,7 +244,7 @@ public String getReference(@NotNull String blobId) {
             Mac mac = Mac.getInstance(ALGORITHM);
             mac.init(new SecretKeySpec(getReferenceKey(), ALGORITHM));
             byte[] hash = mac.doFinal(blobId.getBytes("UTF-8"));
-            return blobId + ':' + BaseEncoding.base32Hex().encode(hash);
+            return ':' + BASE32ENCODER.encodeToString(hash);
```
mbaedke: It used to be: return blobId + ':' +...

reschke: oh, so much for opening PRs while tests are still running, and making last-second changes.

PR #2005, opened 2025-01-20
OAK-11409: Allow multiple cache expiration properties

Amoratinos:
Changed CachePrincipalMembershipReader to allow to define an alternative cache expiration property.

### DIFF
--- a/oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/ExternalGroupPrincipalProviderWithCacheTest.java
+++ b/oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/ExternalGroupPrincipalProviderWithCacheTest.java
@@ -153,7 +153,6 @@ public class ExternalGroupPrincipalProviderWithCacheTest extends AbstractPrincip
         Authorizable user = getUserManager(root).getAuthorizable(USER_ID);
         assertNotNull(user);
 
-        // same as in test before even if the principal is not a tree-based-principal
         Set<Principal> expected = getExternalGroupPrincipals(USER_ID);
         expected.add(testGroup.getPrincipal());
         Set<? extends Principal> principals = principalProvider.getMembershipPrincipals(user.getPrincipal());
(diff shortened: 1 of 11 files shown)


anchela (approved):
hi @Amoratinos , sorry for the long delay. the fix looks very good to me.

if you really wanted me to nitpick
- there are some formatting/whitespace changes and improvements that are not directly related to the fix at hand. in general, i recommend to push those separately to keep the actually changes minimal
- there is an extra whitespace on line 84 in CacheConfiguration. i can fix that myself now that i noticed

Amoratinos:
Sorry about the blank spaces, I have the editor configured to remove trailing blank spaces by default and that's why it appears like that :(

reschke:
Right; please do not reformat code that does not need to change.

See https://jackrabbit.apache.org/oak/docs/participating.html#pull-requests-prs

reschke (changes requested):
Right; please do not reformat code that does not need to change.

See https://jackrabbit.apache.org/oak/docs/participating.html#pull-requests-prs

Amoratinos:
Sorry about that, I've removed the extra formatting in the classes. It's a pity that there's no formatting agreed so we could have a common ground so all these issues could be avoided

oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/ExternalGroupPrincipalProviderWithCacheTest.java
```diff
@@ -169,42 +168,4 @@ public void testGetGroupMembershipExternalUserAndLocal() throws Exception {
         Set<Principal> readFromCache = principalProvider.getMembershipPrincipals(user.getPrincipal());
         assertEquals(expected, readFromCache);
     }
-
-//    @Test
```
reschke, line 173: While I don't understand why we have a commented-out test case, wouldn't it be better to leave it in?

reschke (approved):
Thanks for the changes.

I can't comment on the actual technical change, and leave that to Angela.

I would propose not to remove the commented-out test, unless it's evident that nobody will look at it again.

oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/ExternalGroupPrincipalProviderWithCacheTest.java
Amoratinos: I originally copied this class from another test class and left out this test commented out. Now while revisiting the tests I removed it.

anchela:
i am running the checks now and will merge it if it doesn't surface anything odd

anchela:
the build fails with
`Error:  org.apache.jackrabbit.oak.spi.security.user: Version increase required; detected 2.8.0, suggested 3.0.0`

hmmmm... that should not be the case and we can't have major version changes.

-> sorted out in a quick call :-)

anchela (approved):
lgtm

anchela:
i don't understand why checks except for maven build fails.... i am merging it now.
ping me if i have to revert

PR #2009, opened 2025-01-21
OAK-11415: TypeEditor: constraint exception for node type violations should contain that node's effective type

reschke:


mbaedke (approved):

rombert (commented):
It would be useful IMO to also surface the node's path, if available. I noticed this exception as part of a FileVault content package install and it's not easy to understand which node definition caused this.

reschke:
> It would be useful IMO to also surface the node's path, if available. I noticed this exception as part of a FileVault content package install and it's not easy to understand which node definition caused this.

I agree, but the path does not seem to be available here.

PR #2010, opened 2025-01-21
OAK-11416: Docker support for oak-jcr tests with RdbDocumentStore nee…

mbaedke:
…ds a way to specify environment variables for the container

Implemented.


oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/RdbConnectionUtils.java
```diff
@@ -44,6 +49,7 @@ public class RdbConnectionUtils {
     public static final String USERNAME = System.getProperty("rdb.jdbc-user", "sa");
```
reschke, line 47: at some point, we should use SystemPropertySupplier

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/RdbConnectionUtils.java
```diff
@@ -148,4 +155,22 @@ private static boolean checkImageAvailability() throws TimeoutException {
     private static boolean checkDockerAvailability() {
         return DockerClientFactory.instance().isDockerAvailable();
     }
+
+    private static Map<String, String> parseDockerEnv(String raw) {
+        if (StringUtils.isEmpty(raw)) {
```
reschke: do we need to special-case this? Won't the tokenizer just return no tokens on an empty string?

mbaedke: The only difference in behavior would be that we would get a warning on something like -Drdb.docker-env='some whitespace', but that's an improvement. I'll change it.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/RdbConnectionUtils.java
mbaedke: Separate issue, I feel.

reschke (approved):

PR #2027, opened 2025-01-25
 OAK-11431: Remove usage of Guava Files.createTempDir() 

reschke:
(also assign prefixes to temp files to help debugging when they are left open)

### DIFF
--- a/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/tool/AzureCheck.java
+++ b/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/tool/AzureCheck.java
@@ -16,7 +16,6 @@
  */
 package org.apache.jackrabbit.oak.segment.azure.tool;
 
-import com.google.common.io.Files;
 import com.microsoft.azure.storage.blob.CloudBlobDirectory;
 import org.apache.jackrabbit.oak.segment.azure.AzurePersistence;
 import org.apache.jackrabbit.oak.segment.azure.AzureStorageCredentialManager;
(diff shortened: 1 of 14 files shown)


oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/tool/AzureCheck.java
```diff
@@ -388,7 +389,13 @@ public int run() {
...
         }
 
-        FileStoreBuilder builder = fileStoreBuilder(Files.createTempDir()).withCustomPersistence(persistence);
+        FileStoreBuilder builder = null;
+        try {
+            builder = fileStoreBuilder(Files.createTempDirectory(getClass().getSimpleName() + "-").toFile()).withCustomPersistence(persistence);
+        } catch (IOException e) {
+            e.printStackTrace(err);
+            return 1;
+        }
```
mbaedke, line 398: Maybe throw an IllegalStateException instead of returning 1 to keep closest to the Guava behavior?

mbaedke (changes requested):

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/tool/AzureCheck.java
reschke: This is part of a commend line tool.

Look at the end of the method; this is exactly what it's already doing; the only new thing is the creation of the temp directory can fail.

mbaedke: I don't understand, the Guava version may also fail, just with a RuntimeException.

mbaedke (approved):

PR #2031, opened 2025-01-28
OAK-11438 - Add support for parallelization in the Lucene index writer backend.

nfsantos:
Indexing generation consists of the following tasks:

```
for each node in the NodeStore {
  for each index being generated {
    if node is indexed by index {
      d = make Lucene/Elastic Document
      write document d to Lucene/Elastc indexer
    }
  }
}
```

Currently, these phases are done sequentially, by a single thread. This PR adds support to execute the writes to Lucene index in parallel, by introducing a `IndexWriterPool`, a pool of writer threads, and a new implementation of `LuceneIndexWriter`, named `PooledLuceneIndexWriter`, which delegates the actual execution of operations to an instance of `IndexWriterPool`. This improvement works with all NodeStores, because the Lucene writing stage is agnostic to the node store being used for indexing. For instance, this works both with the Mongo Document and segment Tar node stores. The only limitation is that the producer side of the `IndexWriterPool` must be single-threaded (i.e., there must be only one thread producing Lucene documents).

This improvement provides a 40 to 50% speedup to the indexing phase of the indexing job in the common cases, because writing a document to the Lucene index is CPU and I/O intensive. The following shows the speed-up in 4 runs of the indexing job, in 3 different instances of MongoDocumentStore:

document-store-reindex

segment-tar-speedup

New configuration parameters:
- `oak.indexer.document.parallelWriter.enabled` - Whether to enable parallel writes when indexing the Mongo document store. Default is `false`.
- `oak.indexer.editor.parallelWriter.enabled` - Whether to enable parallel writes when indexing using the Out Of Band indexer (Segment tar, Editor based indexing). Default is `false`.
- `oak.indexer.parallelWriter.numberThreads` - Number of threads in writer pool. Default is 4.
- `oak.indexer.parallelWriter.maxBatchSize` - Maximum size of the batches operations on the Lucene index passed from the producer (the thread generating Lucene documents) to the writers. Default is 256.
- `oak.indexer.parallelWriter.queueSize` - Size of the queue between producer and writers. Default is 64.

Note:
- This PR only adds parallelization to the Lucene writer. The Elastic writers do little work on the client side, they just do remote calls to the Elastic servers. And the current Elastic client already supports batching and asynchronous calls, so there would not be much gain in parallelizing the Elastic writers.

Follow up work: [Support parallel Lucene writers in incremental indexer](https://issues.apache.org/jira/browse/OAK-11492)


oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/IndexWriterPoolTest.java
```diff
@@ -0,0 +1,170 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import static org.junit.Assert.*;
```
thomasmueller: I would avoid star imports.

nfsantos: Removed the wildcard import

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -413,12 +413,13 @@ public void reindex() throws CommitFailedException, IOException {
...
                     log.info("Top slowest nodes to index (ms): {}", slowestTopKElements);
+                }
 
-                    indexerProviders.forEach(indexProvider -> {
-                        ExtractedTextCache extractedTextCache = indexProvider.getTextCache();
-                        CacheStats cacheStats = extractedTextCache == null ? null : extractedTextCache.getCacheStats();
-                        log.info("Text extraction cache statistics: {}", cacheStats == null ? "N/A" : cacheStats.cacheInfoAsString());
-                    });
+                for (NodeStateIndexerProvider indexerProvider : indexerProviders) {
+                    indexerProvider.close();
```
thomasmueller: Hm, shouldn't we close it after logging?
(I guess currently "indexerProvider.getTextCache()" will work after closing, but it seems wrong to me to close it first).

thomasmueller (commented):
It looks good! it's a large patch, I think someone else should review as well.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/IndexWriterPool.java
```diff
@@ -0,0 +1,407 @@
...
+        public void execute() {
+            try {
+                try {
+                    boolean closeResult = delegate.close(timestamp);
+                    sync.put(new CloseResult(closeResult));
+                } catch (IOException e) {
+                    sync.put(new CloseResult(e));
+                }
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
```
thomasmueller, line 189: I wonder if this is the right thing to do... Why not 
```suggestion
                LOG.warn("[{}] Interrupted while closing", timestamp, e);
                sync.put(new CloseResult(new IOException(e)));
```
or something like that?

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/IndexWriterPool.java
```diff
@@ -0,0 +1,407 @@
...
+                                id, op.sequenceNumber, op.operations.length, durationMillis);
+                    }
+                    synchronized (pendingBatchesLock) {
+                        pendingBatches.remove(op.sequenceNumber);
+                        pendingBatchesLock.notifyAll();
+                    }
+                }
+            } catch (InterruptedException | IOException e) {
+                LOG.warn("[{}] Interrupted while waiting to take an update operation from the queue", id, e);
+                throw new RuntimeException(e);
```
thomasmueller, line 242: Here, I assume it makes sense to throw a RuntimeException

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/IndexWriterPool.java
```diff
@@ -0,0 +1,407 @@
...
+                    if (durationMillis > 1000) {
+                        LOG.info("[{}] Processing batch {} of size {} took {} millis.",
+                                id, op.sequenceNumber, op.operations.length, durationMillis);
+                    }
+                    synchronized (pendingBatchesLock) {
+                        pendingBatches.remove(op.sequenceNumber);
+                        pendingBatchesLock.notifyAll();
+                    }
+                }
+            } catch (InterruptedException | IOException e) {
```
thomasmueller: Should we have a different log message for IOException?

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/IndexWriterPool.java
nfsantos: `sync.put()` throws an `InterrupedException`, so we cannot use it on the outermost exception handler. If `sync.put()` fails, then I don't think we can do anything else other than quitting with a `RuntimeException`.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/IndexWriterPool.java
nfsantos: Yes, here we are not trying to return a value to the thread that is waiting for these tasks to complete, the only information that needs to be sent back is success vs failure.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/IndexWriterPool.java
nfsantos: Yes, the message was not correct for the case of an IOException. I removed the IOException from the handler of the InterruptedException and left only the catch Throwable afterwards to deal with all other errors.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
nfsantos: Good point, it's a bad practise to access an object after closing it. I moved the close to the end of the block.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/DefaultIndexWriterFactory.java
```diff
@@ -50,14 +64,30 @@ public LuceneIndexWriter newInstance(IndexDefinition def, NodeBuilder definition
...
+        return wrapWithPipelinedIndexWriter(writer, definition.getIndexName());
+    }
+
+    public void close() {
+        LOG.info("Closing LuceneIndexWriterFactory");
+        if (indexWriterPool == null) {
+            LOG.info("Not using an Index writer pool");
+        } else {
+            indexWriterPool.close();
+        }
```
fabriziofortino: I am not sure if incremental indexing opens/closes a writer every time it kicks in.
In that case, we will see a lot of these logs. Should we lower the debug level for the log messages here?

nfsantos: I have lowered the level to DEBUG.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/DefaultIndexWriter.java
```diff
@@ -146,6 +146,7 @@ public boolean close(long timestamp) throws IOException {
                 PERF_LOGGER.end(start, -1, "Completed suggester for directory {}", definition);
             }
 
+            writer.forceMergeDeletes();
```
thomasmueller: What is the reason for this change?

thomasmueller (approved):

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/writer/DefaultIndexWriter.java
nfsantos: It went in by mistake, I reverted it.

fabriziofortino (approved):

PR #2037, opened 2025-01-30
OAK-11441 : removed usage of Guava Iterables.addAll

rishabhdaim:


reschke (commented):
hmmm... If the replacement is a one liner, maybe use that?

rishabhdaim:
@reschke the replacement code is of one line but we need those `null` checks as well.

Doing `null` checks everywhere in the code would make it less readable and more cluttered.

That's why added the util.

reschke:
Do we *really* need the null checks?

If we have them, we get an NPE. If we don't, we get an NPE as well.

Or am I missing something?

The only difference is that pre-Java 17 the NPE would have more information.

see https://dev.to/harsvnc/the-new-nullpointer-exception-in-java-17-4bcb

rishabhdaim:
@reschke addressed the changes in https://github.com/apache/jackrabbit-oak/pull/2037/commits/4783eb2337bcac4166d8a447043ebc9ed6b08e6c.

mbaedke (approved):

PR #2039, opened 2025-01-30
OAK-11445: Remove usage of Guava Files.toString()

reschke:
See

 https://github.com/greenbytes/guavaVsJDKTests/blob/main/src/test/java/org/greenbytes/java/src/test/java/guavaVsJDKTests/io/FilesTest.java
 
 for why this replacement variant was used.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/datastore/DataStoreTextWriter.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/datastore/DataStoreTextWriter.java
@@ -16,7 +16,6 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-
 package org.apache.jackrabbit.oak.plugins.index.datastore;
 
 import java.io.BufferedWriter;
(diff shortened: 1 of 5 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/datastore/DataStoreTextWriter.java
```diff
@@ -131,7 +130,7 @@ public void write(@NotNull String blobId,@NotNull String text) throws IOExceptio
         File textFile = getFile(stripLength(blobId));
         ensureParentExists(textFile);
         //TODO should we compress
-        Files.write(text, textFile, StandardCharsets.UTF_8);
+        org.apache.jackrabbit.guava.common.io.Files.write(text, textFile, StandardCharsets.UTF_8);
```
rishabhdaim, line 133: shouldn't this be removed ?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/datastore/DataStoreTextWriter.java
```diff
@@ -233,7 +232,7 @@ public String toString() {
     private Set<String> loadFromFile(File file) throws IOException {
         Set<String> result = new HashSet<>();
         if (file.exists()) {
-            result.addAll(Files.readLines(file, StandardCharsets.UTF_8));
+            result.addAll(org.apache.jackrabbit.guava.common.io.Files.readLines(file, StandardCharsets.UTF_8));
```
rishabhdaim, line 235: same as above.

rishabhdaim (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/datastore/DataStoreTextWriter.java
reschke: That's no functional change. We need a FQN here because JDK Files replaces Guava Files just for `Files.toString()`. Other methods will be replaced later on.

rishabhdaim (approved):

PR #2041, opened 2025-01-31
OAK-11443 - Use a single DefaultIndexWriterFactory in the LuceneIndexEditorProvider

nfsantos:


oak-store-spi/src/main/java/org/apache/jackrabbit/oak/plugins/memory/ModifiedNodeState.java
```diff
@@ -165,7 +166,7 @@ static Iterable<? extends PropertyState> getProperties(
                     x -> !keys.contains(x == null ? null : x.getName());
             return concat(
                     filter(base.getProperties(), predicate::test),
-                    filter(properties.values(), x -> x != null));
+                    filter(properties.values(), Objects::nonNull));
```
thomasmueller, line 169: This change seems unrelated...

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexEditorProvider.java
```diff
@@ -257,6 +261,22 @@ public Editor getIndexEditor(
         return null;
     }
 
+    private void initDefaultWriterFactory() {
+        defaultWriterFactoryInitLock.lock();
+        try {
+            if (defaultIndexWriterFactory == null) {
+                LOG.info("Initializing DefaultIndexWriterFactory");
+                cowDirectoryCleanupCallback = new COWDirectoryCleanupCallback();
+                BlobDeletionCallback blobDeletionCallback = activeDeletedBlobCollector.getBlobDeletionCallback();
```
thomasmueller, line 270: I wonder if it is possible to add a test case to show the problem of the current code

nfsantos:
On hold until OAK-11438 is merged. The main motivation for this PR was to make it easy for OAK-11438 for close the thread pool `IndexWriterPool` that would be added inside `DefaultIndexWriterFactory`. But I have refactored OAK-11438 to move the ownership of the `IndexWriterPool` from the instances of `DefaultIndexWriterFactory` to the `LuceneIndexWriterProvider` that creates them. The provider can be easily closed at the end of the indexing. Therefore, the individual instances of `DefaultIndexWriterFactories` will no longer own a thread pool, so they can be left for the garbage collector to handle as it is done currently. There would still be an efficiency gain of reusing a single DefaultIndexWriterFactory but the current implementation makes this quite complex and would need a significant refactoring.

PR #2050, opened 2025-02-02
OAK-11450 Correctly detect expanded names using empty namespace or

kwin:
internal namespace

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/namepath/impl/GlobalNameMapper.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/namepath/impl/GlobalNameMapper.java
@@ -16,21 +16,21 @@
  */
 package org.apache.jackrabbit.oak.namepath.impl;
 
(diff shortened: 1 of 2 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/namepath/impl/GlobalNameMapper.java
```diff
@@ -67,10 +68,15 @@ protected static boolean isHiddenName(String name) {
     protected static boolean isExpandedName(String name) {
         if (name.startsWith("{")) {
             int brace = name.indexOf('}', 1);
-            return brace != -1 && name.substring(1, brace).indexOf(':') != -1;
-        } else {
-            return false;
+            if (brace != -1) {
+                String namespace = name.substring(1, brace);
+                // the empty namespace and "internal" are valid as well, otherwise it always contains a colon (as it is a URI)
```
reschke, line 73: As most people do not understand the difference between URI and "relative reference" we may want to refer to Section 3 of RFC 3986.

reschke (approved):

PR #2055, opened 2025-02-06
OAK-11458 : added utils class for replacing Guava's Iterables

rishabhdaim:


reschke:
What do we do when a class needs both our IterarbleUtils and those from commons-collections?

(I'd say implement the methods and delegate)

rishabhdaim:
This is why I created the **IterableUtils** inside **oak-commons** module so we can implement the method with the same name and delegate it to Apache **commons-collections4**.

reschke:
Maybe I'm mssing something. Aren't we already using commons-collection4's IterableUtils in other places?

rishabhdaim:
Yes, we are using commons-collection4's IterableUtils in other places. But for this **Iterables.concat** use case, we hit a roadblock and had to implement our replacement for Iterables<Iterables<E>> [1].

To avoid using different IterableUtils in each class (and polluting it with a complete class name), I have added other methods (to replace Iterables.concat) in oak-commons IterableUtils.

Now, to avoid using both IterableUtils in the same class for other methods, When we hit this case, we will implement the method in oak-common's IterableUtils and replace it.

So, gradually we would be replacing the commons-collection4's IterableUtils with our delegating methods.

[1] https://guava.dev/releases/20.0/api/docs/com/google/common/collect/Iterables.html#concat-java.lang.Iterable-

reschke:
Please see https://sonarcloud.io/project/issues?id=org.apache.jackrabbit%3Ajackrabbit-oak&pullRequest=2055&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

mbaedke (dismissed):

rishabhdaim:
@mbaedke the sonar complaints regarding [Remove this array creation and simply pass the elements.](https://sonarcloud.io/project/issues?pullRequest=2055&open=AZTa0NS6q1ntFmbx0i9k&id=org.apache.jackrabbit%3Ajackrabbit-oak) are due to fact that I am creating an array beforehand passing the variables to an method that accepts varargs.

```
public static <E> Iterable<E> chainedIterable(final Iterable<? extends E> a,
                                                  final Iterable<? extends E> b) {
        return chainedIterable(new Iterable[] {a, b});
    }
```
is calling 
```
@SafeVarargs
    public static <E> Iterable<E> chainedIterable(final Iterable<? extends E>... iterables) {
        Objects.requireNonNull(iterables);
        return org.apache.commons.collections4.IterableUtils.chainedIterable(iterables);
    }
```
from same class i.e. **IterableUtils** in oak-commons.

In case I don't create an array, I would be calling the caller method thus creating an infinite loop.

```
public static <E> Iterable<E> chainedIterable(final Iterable<? extends E> a,
                                                  final Iterable<? extends E> b) {
        return chainedIterable(a, b);
    }
```

this is the same pattern used inside **_commons-collections4_**.

One possible way to avoid is to call _**commons-collections4**_ api's from each method rather than delegating the call to _commons-collections4_ to a single method.

```
public static <E> Iterable<E> chainedIterable(final Iterable<? extends E> a,
                                                  final Iterable<? extends E> b) {
        return org.apache.commons.collections4.IterableUtils.chainedIterable(a, b);
    }
```

rishabhdaim:
@mbaedke @reschke changes addressed in https://github.com/apache/jackrabbit-oak/pull/2055/commits/16134e53afd6db71cf9913648fe3b2bfebc96ec4

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/collections/IterableUtils.java
```diff
@@ -0,0 +1,145 @@
...
+     * <p>
+     * The returned {@code Iterable} will iterate over all elements of the first {@code Iterable},
+     * then all elements of the second, and so on.
+     *
+     * @param <E> the type of elements returned by the iterator
+     * @param iterables an {@code Iterable} of {@code Iterable} instances to be chained
+     * @return an {@code Iterable} that provides a single view of all elements in the input {@code Iterable} instances
+     * @throws NullPointerException if the input {@code Iterable} or any of its elements are null
+     */
+    public static <E> Iterable<E> chainedIterable(final Iterable<? extends Iterable<? extends E>> iterables) {
```
thomasmueller, line 135: Is this actually used anywhere in the code?

rishabhdaim: @thomasmueller yes. 

This method would be repalcing Guava's **Iterables.concat** [1] and that method is used at many places inside code.

[1] https://guava.dev/releases/20.0/api/docs/com/google/common/collect/Iterables.html#concat-java.lang.Iterable-

reschke:
> Now, to avoid using both IterableUtils in the same class for other methods, When we hit this case, we will implement the method in [oak-common](https://issues.apache.org/jira/browse/OAK-common)'s IterableUtils and replace it.

Is this planned to happen next? Otherwise I'd want to include this here, or even do things in the opposite order.

reschke:
That's why I was asking. I'd want this to be consistent throughout first.

If somebody else changes code and comes into this situation, there'll be confusion.

rishabhdaim:
Yes, we will make these changes in their respective PRs and then proceed to make changes in classes where these would be used.

So, first I would be adding methods in this Util class and then only proceeding further to other classes.

reschke (approved):

PR #2056, opened 2025-02-06
OAK-11459 - Code cleanups on indexing related packages

nfsantos:
Move ConfigHelper from oak-run-commons to oak-core
Minor refactorings and code cleanups to indexing modules.

Mostly cosmetic PR to reduce the changeset of OAK-11438.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/progress/IndexingProgressReporter.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/progress/IndexingProgressReporter.java
@@ -91,7 +91,7 @@ public class IndexingProgressReporter implements NodeTraversalCallback {
     }
 
     public void traversedNode(PathSource pathSource) throws CommitFailedException {
-        if (++traversalCount % 100_000 == 0) {
+        if (++traversalCount % 10_000 == 0) {
             double rate = traversalRateEstimator.getNodesTraversedPerSecond();
             String formattedRate = String.format("%1.2f nodes/s, %1.2f nodes/hr", rate, rate * 3600);
             String estimate = estimatePendingTraversal(rate);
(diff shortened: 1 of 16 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/progress/IndexingProgressReporter.java
```diff
@@ -91,7 +91,7 @@ public void setMessagePrefix(String messagePrefix) {
     }
 
     public void traversedNode(PathSource pathSource) throws CommitFailedException {
-        if (++traversalCount % 10000 == 0) {
+        if (++traversalCount % 100_000 == 0) {
```
fabriziofortino: is this intended? We are printing logs every 100K. It was 10K before.

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/flatfile/pipelined/ConfigHelperTest.java
fabriziofortino, line 1: would it make sense to move the test in oak-run?

nfsantos: I missed the test. I have now moved it to the same package as `ConfigHelper`, `oak-core`. I assume the `oak-run` mentioned above was a typo, and you meant `oak-core`, right?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/progress/IndexingProgressReporter.java
nfsantos: The increase to 100K is intentional. In my tests, I found that printing every 10K is too much. But now I'm wondering, this logic is also used for the AsyncIndexing, which is  much slower than the OutOfBandIndexer (segment tar indexer), so maybe I should leave it at 10K. I will revert to 10K for the time being.

nfsantos: A better approach would be to print the messages at a fixed time interval, but that is a bigger change, so better leave it for another PR.

fabriziofortino (approved):

PR #2032, opened 2025-01-29
OAK-11433 : added OSGi config for setting max revision age for full gc

rishabhdaim:

PR #2057, opened 2025-02-06
OAK-11461 : DocumentNodeStoreService -  disable discovery 

niekraaijmakers:
Make it configurable to disable discovery for the document node store service

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
@@ -398,6 +398,6 @@ import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreServic
 
     @AttributeDefinition(
             name = "Invisible for discovery",
-            description = "Boolean value indicating that the instance should not be discoverable for the cluster. The default value is " + DocumentNodeStoreService.DEFAULT_INVISIBLE_FOR_DISCOVERY)
+            description = ""Boolean value indicating whether the instance should be discoverable by the cluster. The default value is " + DocumentNodeStoreService.DEFAULT_INVISIBLE_FOR_DISCOVERY)
     boolean invisibleForDiscovery() default DocumentNodeStoreService.DEFAULT_INVISIBLE_FOR_DISCOVERY;
 }
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
@@ -395,4 +395,9 @@ import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreServic
             description = "Integer value indicating the number of documents to check for garbage in each Full GC cycle." +
                     "The default value is " + DocumentNodeStoreService.DEFAULT_FGC_PROGRESS_SIZE)
     int fullGCProgressSize() default DocumentNodeStoreService.DEFAULT_FGC_PROGRESS_SIZE;
+
+    @AttributeDefinition(
+            name = "Invisible for discovery",
+            description = "Boolean value indicating that the instance should not be discoverable for the cluster. The default value is " + DocumentNodeStoreService.DEFAULT_INVISIBLE_FOR_DISCOVERY)
+    boolean invisibleForDiscovery() default DocumentNodeStoreService.DEFAULT_INVISIBLE_FOR_DISCOVERY;
 }
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
@@ -398,6 +398,6 @@ import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreServic
 
     @AttributeDefinition(
             name = "Invisible for discovery",
-            description = ""Boolean value indicating whether the instance should be discoverable by the cluster. The default value is " + DocumentNodeStoreService.DEFAULT_INVISIBLE_FOR_DISCOVERY)
+            description = "Boolean value indicating whether the instance should be discoverable by the cluster. The default value is " + DocumentNodeStoreService.DEFAULT_INVISIBLE_FOR_DISCOVERY)
     boolean invisibleForDiscovery() default DocumentNodeStoreService.DEFAULT_INVISIBLE_FOR_DISCOVERY;
 }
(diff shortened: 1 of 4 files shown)


Joscorbe:
Re-running the pipeline because it failed for:
`[ERROR] org.apache.maven.surefire.booter.SurefireBooterForkException: The forked VM terminated without properly saying goodbye. VM crash or System.exit called?`

So far no tests failed:
`[WARNING] Tests run: 1685, Failures: 0, Errors: 0, Skipped: 13`

rishabhdaim:
Please add unit cases in https://github.com/apache/jackrabbit-oak/blob/trunk/oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceConfigurationTest.java also.

For reference you can see https://github.com/apache/jackrabbit-oak/pull/2032 PR.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
```diff
@@ -395,4 +395,9 @@
             description = "Integer value indicating the number of documents to check for garbage in each Full GC cycle." +
                     "The default value is " + DocumentNodeStoreService.DEFAULT_FGC_PROGRESS_SIZE)
     int fullGCProgressSize() default DocumentNodeStoreService.DEFAULT_FGC_PROGRESS_SIZE;
+
+    @AttributeDefinition(
+            name = "Invisible for discovery",
+            description = "Boolean value indicating that the instance should not be discoverable for the cluster. The default value is " + DocumentNodeStoreService.DEFAULT_INVISIBLE_FOR_DISCOVERY)
```
rishabhdaim:
```suggestion
            description = ""Boolean value indicating whether the instance should be discoverable by the cluster. The default value is " + DocumentNodeStoreService.DEFAULT_INVISIBLE_FOR_DISCOVERY)
```

rishabhdaim (changes requested):

Joscorbe:
I'm just a bit concerned if there are enough tests covering all the cases under `DocumentDiscoveryLiteInvisibleServiceCrashTest` class, but this PR is just adding an extra parameter to configure the property and not changing functionality.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
Joscorbe: You added an extra double quote, probably by mistake.

niekraaijmakers:
@rishabhdaim please re-review :D thanks!

rishabhdaim:
@niekraaijmakers could you please add test cases as mentioned in this comment: https://github.com/apache/jackrabbit-oak/pull/2057#issuecomment-2640071711

Also, I agree with @Joscorbe (see https://github.com/apache/jackrabbit-oak/pull/2057#issuecomment-2640170061) that though the change looks simple enough, it might introduce some regression since it affects a core functionality.

To avoid that, I suggest you parametrizing the test cases for `BaseDocumentDiscoveryLiteServiceTest` and its subclasses to cover both visible & invisible scenarios.

You can look at the `NodeDocumentTest` class for parametrized tests with boolean values.

cc @stefan-egli @reschke

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
```diff
@@ -395,4 +395,9 @@
             description = "Integer value indicating the number of documents to check for garbage in each Full GC cycle." +
                     "The default value is " + DocumentNodeStoreService.DEFAULT_FGC_PROGRESS_SIZE)
     int fullGCProgressSize() default DocumentNodeStoreService.DEFAULT_FGC_PROGRESS_SIZE;
+
+    @AttributeDefinition(
+            name = "Invisible for discovery",
+            description = ""Boolean value indicating whether the instance should be discoverable by the cluster. The default value is " + DocumentNodeStoreService.DEFAULT_INVISIBLE_FOR_DISCOVERY)
```
Joscorbe: This double quote symbol is breaking the build now:

```suggestion
            description = "Boolean value indicating whether the instance should be discoverable by the cluster. The default value is " + DocumentNodeStoreService.DEFAULT_INVISIBLE_FOR_DISCOVERY)

Joscorbe (changes requested):

niekraaijmakers:
Great. Push failed ;_; and I wasted everybody's time. Sorry guys.

niekraaijmakers:
@rishabhdaim sorry for the initial comment. my push was rejected. i added the test you requested.

Regarding the second point, from a policy standpoint I unfortunately cannot make myself responsible for this part, so the answer is I won't do this. This is for the team that wrote this functionality to write. If this is required for this PR, then i'll need to create a blocker ticket and push on that.

I would love to help out, but I need to limit myself on what I take on, especially in domains that are not mine, and avoid getting trapped in rabbit holes. 

Nothing personal, and you're totally right that it should be tested. Just a question of who is gonna do it and if this PR / effort needs to be halted until the test is done.

rishabhdaim:
@niekraaijmakers Thanks for adding the unit cases.

Earlier only from test cases or oak-run, we were allowed to make the node invisible.

However, this config makes sure that now in the production code, we would be making the cluster invisible for discovery which is not the case currently.
Since, this is a major change and unfortunately, I don't have the full knowledge of this piece of code, I would suggest we add unit cases for invisible mode in `BaseDocumentDiscoveryLiteServiceTest` and its subclasses.

I am also fine with you creating a blocker ticket to address that.

P.S. I would wait for @stefan-egli to have a look into this PR.

niekraaijmakers:
@rishabhdaim  i understand. 
Just making clear: this will be only for specific deployments / pods in the cluster, not all. 

If the rest agrees this is a requirement il create a blocker ticket and basically block all other work behind this.

stefan-egli (approved):
I'm +1 for this change : by default (DEFAULT_INVISIBLE_FOR_DISCOVERY=false) this change is a no-op - if the config is added, then it does what it is intended for : namely to hide a DocumentNodeStore from discovery. Setting that flag should be done very consciously of course. But if say only 1 out of n DocumentNodeStores in the cluster does that, discovery is still satisfied and running fine.
Ok to add more tests of course, it should be noted though that the tests could be complex and perhaps it wouldn't be feasible to reuse many existing tests (as if they are discovery dependent, then hiding would exactly break that aspect - while if they are not discovery dependent it wouldn't add test coverage).

rishabhdaim (approved):

niekraaijmakers:
@rishabhdaim  @stefan-egli can someone merge this? the error's in the PR seem to be not specific to the PR?

niekraaijmakers:
@Joscorbe

Joscorbe (approved):

Joscorbe:
I'm rebuilding the pipelines and see why they are failing, but could be not related with the PR.

PR #2058, opened 2025-02-06
OAK-11460 Allow configuring the index.mapping.total_fields.limit

thomasmueller:


tihom88:
@thomasmueller Should we also set index.mapping.total_fields.ignore_dynamic_beyond_limit to true.

Here is the case let's say a document comes in with two fields, one is fixed and other is dynamic and lets say the mapping limit is already breached for dynamic property. So should we throw error for this document as a whole or just ignore the dynamic field and index rest of the document.

fabriziofortino (approved):

PR #2065, opened 2025-02-07
OAK-11469 - Speedup resolution of mount paths by SimpleMountInfoProvider

nfsantos:
The method `DefaultMountInfoProvider.getMountByPath(String)` is called once for every node indexed by the out-of-band indexer, which can be on the order of hundreds of millions. The current implementation of `getMountByPath(String)` method is iterating over the values of a Map. In a JFR profile run, the instances of this iterator over the values of a collection showed up as the second largest source of object allocation.

This PR optimizes this logic by keeping the list of mounts in a separate immutable list, which allows iterating directly over these values. The iteration over such a list is more direct. I measured a small but significant improvement and speed (around 5% in a micro benchmark), but more importantly once I rerun the indexing job with the new logic, the allocation of iterators by the `getMountByPath(String)` disappeared from the profile results. My guess for the drastic reduction in object allocation is that the JVM can more easily optimize a direct iteration over an immutable list than over the values of a map.


oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/mount/SimpleMountInfoProvider.java
```diff
@@ -18,47 +18,50 @@
...
 import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * A simple and inefficient implementation to manage mount points
  */
 final class SimpleMountInfoProvider implements MountInfoProvider {
 
-    private final Map<String, Mount> mounts;
+    private final Map<String, Mount> nameToMounts;
```
thomasmueller: For me, "nameToMounts" would mean the key is name, and the value is an _array_ of mounts... So I would use "nameToMount". 

```suggestion
    private final Map<String, Mount> nameToMount;
```

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/mount/SimpleMountInfoProvider.java
```diff
@@ -67,9 +70,9 @@ public boolean hasNonDefaultMounts() {
     }
 
     @Override
-    public Collection<Mount> getMountsPlacedUnder(String path) {
-        Collection<Mount> mounts = new ArrayList<>();
-        for ( Mount mount : this.mounts.values()) {
+    public @NotNull Collection<Mount> getMountsPlacedUnder(String path) {
+        Collection<Mount> mounts = new ArrayList<>(this.mounts.size());
+        for ( Mount mount : this.mounts) {
```
thomasmueller:
```suggestion
        for (Mount mount : this.mounts) {
```

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/mount/SimpleMountInfoProvider.java
```diff
@@ -78,9 +81,9 @@ public Collection<Mount> getMountsPlacedUnder(String path) {
     }
 
     @Override
-    public Collection<Mount> getMountsPlacedDirectlyUnder(String path) {
-        Collection<Mount> mounts = new ArrayList<>();
-        for ( Mount mount : this.mounts.values()) {
+    public @NotNull Collection<Mount> getMountsPlacedDirectlyUnder(String path) {
+        Collection<Mount> mounts = new ArrayList<>(this.mounts.size());
+        for ( Mount mount : this.mounts) {
             if ( mount.isDirectlyUnder(path) ) {
```
thomasmueller:
```suggestion
        for (Mount mount : this.mounts) {
            if (mount.isDirectlyUnder(path)) {
```

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/mount/SimpleMountInfoProvider.java
```diff
@@ -67,9 +70,9 @@ public boolean hasNonDefaultMounts() {
     }
 
     @Override
-    public Collection<Mount> getMountsPlacedUnder(String path) {
-        Collection<Mount> mounts = new ArrayList<>();
-        for ( Mount mount : this.mounts.values()) {
+    public @NotNull Collection<Mount> getMountsPlacedUnder(String path) {
+        Collection<Mount> mounts = new ArrayList<>(this.mounts.size());
```
thomasmueller: If I'm not mistaken, the average size of the result is 1, so actually `new ArrayList<>(1);` would be better.
```suggestion
        Collection<Mount> mounts = new ArrayList<>(1);
```

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/mount/SimpleMountInfoProvider.java
```diff
@@ -78,9 +81,9 @@ public Collection<Mount> getMountsPlacedUnder(String path) {
     }
 
     @Override
-    public Collection<Mount> getMountsPlacedDirectlyUnder(String path) {
-        Collection<Mount> mounts = new ArrayList<>();
-        for ( Mount mount : this.mounts.values()) {
+    public @NotNull Collection<Mount> getMountsPlacedDirectlyUnder(String path) {
+        Collection<Mount> mounts = new ArrayList<>(this.mounts.size());
+        for ( Mount mount : this.mounts) {
```
thomasmueller:
```suggestion
        for (Mount mount : this.mounts) {
```

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/mount/SimpleMountInfoProvider.java
nfsantos: Done

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/mount/SimpleMountInfoProvider.java
nfsantos: I see, in the common case, only one mount will pass the test. I also considered creating the array lazily, only when one match is found, but I don't think this method is called frequently enough to be worth the extra complexity. So I changed it to create an array of fixed size 1, as suggested.

thomasmueller (approved):

PR #2068, opened 2025-02-10
OAK-11472 - Better error message when Tika configuration is missing.

nfsantos:


mbaedke:
Maybe a stupid question: what if someone wants to use Oak without full-text search. Isn't it perfectly fine for the Tika config to be missing?

nfsantos:
> Maybe a stupid question: what if someone wants to use Oak without full-text search. Isn't it perfectly fine for the Tika config to be missing?

I suppose so, but that's outside the scope of this PR. The goal here is only to improve the error message, to replace the generic NPE by a more informative message.

mbaedke:
Ah, I missed that it also threw an exception before. Sorry for the noise.

nfsantos:
> Ah, I missed that it also threw an exception before. Sorry for the noise.

No problem, it was a fair comment.

thomasmueller (approved):

PR #2076, opened 2025-02-13
OAK-8748: Run performance tests and compare with baseline

Joscorbe:
Script to run the performance tests in oak-benchmarks and compare the results with a base version.

This script allows to detect performance differences after a change in the codebase. It was reviewed years ago by @mreutegg but never committed. I have added a new argument to allow to define the output folder for all the generated and downloaded files, which defaults to `target` folder, as suggested in the ticket comments.


mreutegg (approved):
LGTM

PR #2078, opened 2025-02-14
OAK-11480 - Graceful handling when getErrorCode() returns null.

ahanikel:
Make sure Set.contains() does not throw an NPE if getErrorCode() returns null.


smiroslav (approved):

ahanikel:
@smiroslav Thanks for the approval Miro! Could you also merge the PR? I don't have permission to do that.

joerghoh (approved):

PR #2081, opened 2025-02-14
OAK-11486: Add failing test case for session-mapped non-registered

kwin:
namespace URIs

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/namepath/impl/LocalNameMapper.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/namepath/impl/LocalNameMapper.java
@@ -59,15 +59,12 @@ public class LocalNameMapper extends GlobalNameMapper {
         checkArgument(!isExpandedName(oakName), oakName); // expanded name
 
         int colon = oakName.indexOf(':');
-        boolean globallyMappedPrefix = false;
-        boolean locallyMappedPrefix = false;
         if (colon > 0) {
             String oakPrefix = oakName.substring(0, colon);
             String uri = getNamespacesProperty(oakPrefix);
 
-            globallyMappedPrefix = uri != null;
-            locallyMappedPrefix = local.containsValue(uri);
-            if (locallyMappedPrefix) {
+            if (local.containsValue(uri)) {
+                //local prefix replaces global prefix
                 for (Map.Entry<String, String> entry : local.entrySet()) {
                     if (uri.equals(entry.getValue())) {
                         String jcrPrefix = entry.getKey();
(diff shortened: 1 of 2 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/namepath/impl/LocalNameMapper.java
```diff
@@ -64,8 +64,8 @@ public synchronized String getJcrName(@NotNull String oakName) {
                 String oakPrefix = oakName.substring(0, colon);
                 String uri = getNamespacesProperty(oakPrefix);
                 if (uri == null) {
-                    throw new IllegalStateException(
-                            "No namespace mapping found for " + oakName);
+                    //the namespace has a local mapping, but no global mapping, which is fine
```
kwin: I don't think this comment is accurate, this only means that there is some local namespace mapping (irrespective for which prefix/uri combination) but the prefix is not registered globally.

mbaedke: Yes, that is true. I'll fix it.

oak-core/src/main/java/org/apache/jackrabbit/oak/namepath/impl/LocalNameMapper.java
```diff
@@ -91,9 +91,10 @@ public synchronized String getJcrName(@NotNull String oakName) {
                         }
                     }
                 }
+            } else {
+                throw new IllegalStateException("No namespace mapping found for " + oakName);
```
kwin, line 80: With that we are back at the same exception as before :-)

kwin: actually we are now throwing an exception even earlier (i.e. for the first getName() without any mapping)

mbaedke: That was an intermediate version. This current one should hopefully do, except for the problem I mentioned on Jira, which probably is something completely different.

kwin: I wouldn't understand why `Session.addNode("foo:bar").getName()` wouldn't end up here, because

a) uri is null (there is no namespace URI registered for "foo" neither globally nor locally
b) there is no local namespace mapping for "foo"

This previously didn't lead to an exception. Now it does.

kwin:
@mbaedke Do you plan to pick this up again?

PR #2082, opened 2025-02-16
OAK-8413 Reset property "segment.azure.v12.enabled" after test

kwin:
This fixes the "java.lang.ClassCastException: class org.apache.jackrabbit.oak.segment.azure.AzurePersistence cannot be cast to class org.apache.jackrabbit.oak.segment.azure.v8.AzurePersistenceV8" issue

### DIFF
--- a/oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/AzureSegmentStoreServiceTest.java
+++ b/oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/AzureSegmentStoreServiceTest.java
@@ -26,6 +26,7 @@ import org.apache.jackrabbit.oak.segment.azure.util.Environment;
 import org.apache.jackrabbit.oak.segment.spi.persistence.SegmentNodeStorePersistence;
 import org.apache.sling.testing.mock.osgi.junit.OsgiContext;
 import org.jetbrains.annotations.NotNull;
+import org.junit.AfterClass;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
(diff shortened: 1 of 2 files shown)


kwin:
And another unstable test: 
```
java.lang.AssertionError: batchWriteLines() should be significantly faster (>10x) than the naive implementation, but took 81ms while naive implementation took 790ms
```

oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/AzureSegmentStoreServiceTest.java
```diff
@@ -88,6 +89,11 @@ public void setup() throws Exception {
         }
     }
 
+    @AfterClass
+    public static void tearDownClass() {
+        System.clearProperty("segment.azure.v12.enabled");
```
thomasmueller, line 94: That assumes the system property was not set before running the test. This assumption might not be correct. Instead, the old value should be store in the setup, and restored in the tear down.

thomasmueller (changes requested):

oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/AzureSegmentStoreServiceTest.java
kwin: @smiroslav Please take over this PR/branch to improve further. My primary goal was to bring the GHA build back to green.

smiroslav: hi @kwin, I have addressed the issue  with the system property in 
https://github.com/apache/jackrabbit-oak/pull/2086/files

kwin:
Superseded by #2086

PR #2091, opened 2025-02-18
OAK-11497 Add support for Date fields in DocumentNodeStore

daniancu:
Allow values of type ISODate fields in Mongo documents
This type is needed when creating TTL indexes

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/UpdateOp.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/UpdateOp.java
@@ -17,6 +17,7 @@
 package org.apache.jackrabbit.oak.plugins.document;
 
 import java.util.Collections;
+import java.util.Date;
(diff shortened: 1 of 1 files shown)


shodaaan (approved):

reschke (changes requested):
1. Test coverage?
2. Do we expect this to round-trip? Would be strange if it could be set but not retrieved.
3. How does this work with other implementations of document store?

daniancu:
> * Do we expect this to round-trip? Would be strange if it could be set but not retrieved.
> * How does this work with other implementations of document store?

I didn't consider the retrieval part because we need this one-way. But  once the API allows the Date values to be written the [convertFromDBObject](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java#L1959) will crash when reading any document containing Date values. I'm closing the PR for now as it needs more analysis

PR #2092, opened 2025-02-18
OAK-11268: Remove Jenkinsfile and add GitHub Actions metadata to POM

kwin:


kwin:
Must no be merged until all Jobs having "oak" in their name below https://ci-builds.apache.org/job/Jackrabbit/ are disabled/removed.

PR #2093, opened 2025-02-18
OAK-11498: Expose Session-bound principals via JackrabbitSession

kwin:


oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionImpl.java
```diff
@@ -845,6 +846,12 @@ public UserManager getUserManager() throws RepositoryException {
         return sessionContext.getUserManager();
     }
 
+    @Override
+    @NotNull
+    public Set<Principal> getPrincipals() throws RepositoryException {
+        return sd.getAuthInfo().getPrincipals();
+    }
```
reschke, line 858: We may want to check for consistency with other APIs. Maybe Iterable would be better.

kwin: I already shared my view in https://issues.apache.org/jira/browse/OAK-11498?focusedCommentId=17928122&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17928122.

kwin:
I am gonna merge end of this week, if there is no other feedback.

anchela (changes requested):
without a default implementation for the new API call, this modification is a breaking change that will cause issues for our downstream project AEM.

as i said before the fact that principals can be exposed on the jackrabbit API is very much an implementation detail of oak and i am not really convinced that this should be exposed in jackrabbit API.

reschke:
@anchela - how would a default implementation help? (FWIW, I'd like to understand as well how we would implement this in Jackrabbit Classic).

kwin:
> as i said before the fact that principals can be exposed on the jackrabbit API is very much an implementation detail of oak and i am not really convinced that this should be exposed in jackrabbit API.

We do have several principal related API classes already in https://github.com/apache/jackrabbit-oak/tree/trunk/oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/security/principal. As JCR 2.0 spec relies on principals for access control (https://developer.adobe.com/experience-manager/reference-materials/spec/jcr/2.0/16_Access_Control_Management.html#16.5.1%20Access%20Control%20Entries) it is natural that somehow you have it bound to a session as each impl would need to somehow compute those to check for access.

For JR2 I think the subject.getPrincipals (https://github.com/apache/jackrabbit/blob/trunk/jackrabbit-core/src/main/java/org/apache/jackrabbit/core/SessionImpl.java#L384) should work in general. Maybe with some additional logic to get transitive group memberships like in https://github.com/apache/jackrabbit/blob/trunk/jackrabbit-core/src/main/java/org/apache/jackrabbit/core/security/authentication/AbstractLoginModule.java#L682C30-L682C43.

reschke:
@anchela - I do not understand how having a default impl would avoid bumping up the package version - or did you have something else in mind?

@kwin - that said, would it be possible to have a meaningful default impl?

kwin:
I don't think a default impl in the JR-API module without additional dependencies would work, rather one impl for Oak and one for JR2.

anchela (changes requested):
imho it should be fairly easy to come up with a best-effort default implementation.
and if that is not considered good enough the default implementation should at the very least throw an UnsupportedRepositoryOperationException to avoid runtime exceptions

best-effort default could e.g. be an attempt to compute a set of principals from existing public API and if that is not possible throw an exception.

kwin:
Having a default implementation which just throws an exception is actually worse than requiring any provider. to come up with a reasonable impl. Otherwise semantic versioning becomes useless

anchela:
@kwin , i was not thinking about an impl that throws but one that leverages public API. i will work with @Amoratinos to get this unblocked.

oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/JackrabbitSession.java
```diff
@@ -277,6 +279,15 @@ default Node getParentOrNull(@NotNull Item item) throws RepositoryException {
...
         }
     }
+   
+    /**
+     * Returns the set of principals associated with this session.
+     * @return the set of principals associated with this session.
+     * @throws RepositoryException in case principal information cannot be retrieved.
+     * @since 1.81
+     */
+    @NotNull Set<Principal> getPrincipals() throws RepositoryException;
```
Amoratinos: @kwin what do you think about: 

```java
    @NotNull
    default Set<Principal> getPrincipals() throws RepositoryException {
        String userId = getUserID();
        if (userId == null) {
            throw new UnsupportedRepositoryOperationException("No user ID associated with this session.");
        }

        Authorizable authorizable = getUserManager().getAuthorizable(userId);
        if (authorizable == null) {
            throw new UnsupportedRepositoryOperationException("No authorizable found for user ID: " + userId);
        }

        if (!authorizable.isGroup() && ((User) authorizable).isSystemUser()) {
            throw new UnsupportedRepositoryOperationException("Unable to calculate effective set of principals for system user " + userId);
            // alternatively:
            // return Collections.singleton(authorizable.getPrincipal));
        }

        Principal userPrincipal = authorizable.getPrincipal();
        Set<Principal> principals = new java.util.HashSet<>();
        principals.add(userPrincipal);
        PrincipalIterator iterator =  getPrincipalManager().getGroupMembership(userPrincipal);
        while (iterator.hasNext()) {
            principals.add(iterator.nextPrincipal());
        }
        return principals;
    }
```

I've aligned with @anchela about the implementation, we thought about returning an empty set instead of throwing an exception but this could be misleading.

EDIT: I've edit the code to include a check for system user

kwin: In general looks good, however I am not sure about throwing [`UnsupportedRepositoryOperationException`](https://developer.adobe.com/experience-manager/reference-materials/spec/javax.jcr/javadocs/jcr-2.0/javax/jcr/UnsupportedRepositoryOperationException.html?is-external=true)
> Thrown by methods that are not supported by a particular implementation.

In most case this is a particular issue with the current session only (and not due to a non-compliant implementation), therefore I consider IllegalStateException more suitable here (however this should be mentioned in the API) as the same request for another session may succeed. Feel free to directly add you commits to my branch or create a new one based on mine and raise a new PR.

Amoratinos: I've created a PR changing to IllegalStateException as I cannot do it directly on jackrabbit repo I did from my fork. 
https://github.com/apache/jackrabbit-oak/pull/2356

Amoratinos (approved):
LTGM

oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/JackrabbitSession.java
```diff
@@ -304,4 +309,37 @@ default Node getParentOrNull(@NotNull Item item) throws RepositoryException {
...
     @NotNull String getExpandedPath(@NotNull Item item) throws RepositoryException;
+
+    /**
+     * Returns the set of principals associated with this session.
+     * @return the set of principals associated with this session.
+     * @throws RepositoryException in case principal information cannot be retrieved.
+     * @throws IllegalStateException if user information is not available or if the user is a system user.
+     * @since 1.82
+     */
+    @NotNull default Set<Principal> getPrincipals() throws RepositoryException {
```
anchela: i wonder if the method name is self-explanatory.
what about 'getBoundPrincipals' instead? this would be aligned with the repository descriptor that was created for the same reason.
@kwin , @Amoratinos wdyt?

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/session/JackrabbitSessionTest.java
```diff
@@ -131,4 +134,23 @@ public void testGetExpandedPath() throws RepositoryException {
...
+        JackrabbitSession guest = (JackrabbitSession) getHelper().getRepository().login(new GuestCredentials());
+        try {
+            Set<Principal> principals = guest.getPrincipals();
+            assertNotNull(principals);
+            assertFalse("Admin principal not expected", principals.contains(s.getPrincipalManager().getPrincipal("admin")));
+            assertTrue("Everyone principal expected", principals.contains(s.getPrincipalManager().getEveryone()));
+        } finally {
+            guest.logout();
+        }
+    }
```
anchela, line 176: IMHO there should be more tests

1. for the default implementation and the different scenarios that throw exceptions
2. for the oak implementation
- verifying that really the principal set is returned
- verifying that the principal set cannot modify the set attached to the session
- verifying that there is no automatic membership resolution (e.g. when the subject does not contain groups)

also: it feels wrong to hard-code the 'admin' name and make assumptions whether an admin principal is part of the final set or not....

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionImpl.java
```diff
@@ -845,6 +846,12 @@ public UserManager getUserManager() throws RepositoryException {
         return sessionContext.getUserManager();
     }
 
+    @Override
+    @NotNull
+    public Set<Principal> getPrincipals() throws RepositoryException {
+        return sd.getAuthInfo().getPrincipals();
```
anchela, line 857: i would suggest to additionally make sure we don't have 2 implementations
- SessionImpl.getPrincipals
- SessionImpl.getAttribute

-> i would either refactor the code in getAttribute or the method here (or both) and make sure they are always sync even if one gets modified.
e.g. extracting a new method that is call by both or have getAttribute call getPrincipals or the otherway round.

anchela (changes requested):

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionImpl.java
kwin: Agree, will take it up.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/session/JackrabbitSessionTest.java
kwin: Regarding 2: We have that feature since quite a while in Oak (exposed as session attribute). This PR is only about exposing this via another API. You were fine back then without additional tests, but let me try to improve test coverage a bit.

oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/JackrabbitSession.java
kwin: `getBoundPrincipals` sounds good to me, will change accordingly.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/session/JackrabbitSessionTest.java
kwin: I extended test coverage, please have another look @anchela

oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/JackrabbitSession.java
mbaedke: Late and maybe stupid question: why do we need this if a call to #getAttribute() does the job already? I wasn't aware of that.

kwin: This was clearly stated in the description for https://issues.apache.org/jira/browse/OAK-11498. A proper API allows to depend on a specific version of Jackrabbit API instead of implementing a fallback in the custom code in case the attribute is not provided, also it greatly increases visibility of that feature so that 

>  I wasn't aware of that.

is less likely to happen :-)

kwin:
@anchela I am gonna merge beginning of next week if I don't receive any further comments here.

PR #2094, opened 2025-02-18
OAK-11499: Remove parameter "aggregate"

kwin:
It is no longer evaluated since m-javadoc-p 3.0
(https://issues.apache.org/jira/browse/MJAVADOC-457)


reschke (commented):
It might be deprecated but I believe we use it for the side javadoc:

https://jackrabbit.apache.org/oak/docs/apidocs/

Is there a replacement?

kwin:
For aggregates you need to use a dedicated mojo: https://maven.apache.org/plugins/maven-javadoc-plugin/aggregate-mojo.html

PR #2097, opened 2025-02-19
OAK-11503 : added utils method for replacing Guava's Iterables.isEmpty

rishabhdaim:


rishabhdaim:
With this PR, we are done with adding API's to IterableUtils in oak-commons.

mbaedke (approved):

PR #2098, opened 2025-02-19
OAK-11505 - Reduce object allocation in FullTextIndexEditor

nfsantos:
The methods `#getMatcherState(String name, NodeState after)` and `#checkAggregates(String name)` are called multiple times for every node traversed by the indexer, so they are critical for performance. This PR reduces the number of objects allocated by these methods.

Change the collection type of `MatcherState.affectedMatchers` from a `Set<Aggregate.Matcher>` to a more compact and efficient `BitSet`.

Replace all loops on matcher state that were using iterators with indexed loops, to avoid allocating iterator objects.

## Comparing performance of iterating over empty list vs pre-checking for empty

The program below compares iterating over a list using three approaches:
- unconditional iterator - enhanced for loop with iterator
- conditional iteration with empty check - check if collection is empty before iterating
- indexed traversal - using a classic for loop with an index

Results after 10 repetitions:

`List<String> l = List.of();`
- unconditional iterator: 654ms 
- conditional iteration with empty check: 0 ms
- indexed traversal: 0 ms

`List<String> l = List.of("a");`
- unconditional iterator: 1327 ms 
- conditional iteration with empty check: 1328 ms
- indexed traversal: 623 ms

`List<String> l = List.of("a", "b", "c", "d", "e", "f", "g", "h", "i", "j");`
- unconditional iterator: 8783 ms 
- conditional iteration with empty check: 8764 ms
- indexed traversal: 8302 ms

Using an indexed traversal is always faster, with the relative difference being greater for smaller lists.

```java
import java.util.List;

class Scratch {
    private static long count1 = 0;
    private static long count2 = 0;
    private static long count3 = 0;

    private static void test1(List<String> list) {
        for (int i = 0; i < 1_000_000_000; i++) {
            for (String s : list) {
                count1 += s.length();
            }
        }
        System.out.println("count: " + count1 + " test1: " + (System.nanoTime() - start) / 1_000_000);
    }

    private static void test2(List<String> list) {
        for (int i = 0; i < 1_000_000_000; i++) {
            if (!list.isEmpty()) {
                for (String s : list) {
                    count2 += s.length();
                }
            }
        }
        System.out.println("count: " + count2 + " test2: " + (System.nanoTime() - start) / 1_000_000);
    }

    private static void test3(List<String> list) {
        for (int i = 0; i < 1_000_000_000; i++) {
            for (int j = 0; j < list.size(); j++) {
                count3 += list.get(j).length();
            }
        }
        System.out.println("count: " + count3 + " test3: " + (System.nanoTime() - start) / 1_000_000);
    }


    public static void main(String[] args) {
        List<String> l = List.of("a", "b", "c", "d", "e", "f", "g", "h", "i", "j");
        for (int i = 0; i < 10; i++) {
            count1 = count2 = count2 = 0;
            test1(l);
            test2(l);
            test3(l);
        }
    }
}
```

### DIFF
--- a/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java
+++ b/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java
@@ -23,7 +23,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 
-import org.apache.jackrabbit.guava.common.collect.Iterables;
 import org.apache.jackrabbit.oak.api.CommitFailedException;
 import org.apache.jackrabbit.oak.api.PropertyState;
 import org.apache.jackrabbit.oak.commons.PathUtils;
@@ -173,9 +172,9 @@ public class FulltextIndexEditor<D> implements IndexEditor, Aggregate.AggregateR
     public void propertyAdded(PropertyState after) {
         if (isIndexable()) {
             markPropertyChanged(after.getName());
-            checkAggregates(after.getName());
             propertyUpdated(null, after);
         }
+        checkAggregates(after.getName());
     }
 
     @Override
@@ -183,9 +182,9 @@ public class FulltextIndexEditor<D> implements IndexEditor, Aggregate.AggregateR
         if (isIndexable()) {
             markPropertyChanged(before.getName());
             propertiesModified.add(before);
-            checkAggregates(before.getName());
             propertyUpdated(before, after);
         }
+        checkAggregates(before.getName());
     }
 
     @Override
@@ -193,9 +192,9 @@ public class FulltextIndexEditor<D> implements IndexEditor, Aggregate.AggregateR
         if (isIndexable()) {
             markPropertyChanged(before.getName());
             propertiesModified.add(before);
-            checkAggregates(before.getName());
             propertyUpdated(before, null);
         }
+        checkAggregates(before.getName());
     }
 
     @Override
@@ -297,14 +296,14 @@ public class FulltextIndexEditor<D> implements IndexEditor, Aggregate.AggregateR
             Aggregate.Matcher result = m.match(name, after);
             if (result.getStatus() == Aggregate.Matcher.Status.MATCH_FOUND) {
                 if (matched == EMPTY_AGGREGATE_MATCHER_LIST) {
-                    matched = new ArrayList<>(2);
+                    matched = new ArrayList<>();
                 }
                 matched.add(result);
             }
 
             if (result.getStatus() != Aggregate.Matcher.Status.FAIL) {
                 if (inherited == EMPTY_AGGREGATE_MATCHER_LIST) {
-                    inherited = new ArrayList<>(2);
+                    inherited = new ArrayList<>();
                 }
                 inherited.addAll(result.nextSet());
             }
--- a/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java
+++ b/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java
@@ -20,7 +20,6 @@ package org.apache.jackrabbit.oak.plugins.index.search.spi.editor;
 
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java
```diff
@@ -41,361 +40,370 @@
...
+            }
+        }
 
-      if (result.getStatus() != Aggregate.Matcher.Status.FAIL){
-        inherited.addAll(result.nextSet());
-      }
+        if (!matched.isEmpty() || !inherited.isEmpty()) {
+            return new MatcherState(matched, inherited);
+        }
+        return MatcherState.NONE;
```
thomasmueller: As an alternative, what about:

```suggestion
        if (matched.isEmpty() && inherited.isEmpty()) {
            return MatcherState.NONE;
        }
        return new MatcherState(matched, inherited);
```

nfsantos: Yes, cleaner to read. I don't think it will improve performance, it seems to be equivalent, but I will change it because it is nicer.

reschke: FWIW; my preference is to have all `return` statements at the same nesting level :-) ... unless it's really an _early_ exit, which is not the case here.

nfsantos: Yes, I also prefer to use if ... else ... in these cases. I was hesitating to change to that style to minimize the changeset and also because it is mainly a stylist option, but probably it's better to fix this. having the else clause is also safer and more clear.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java
```diff
@@ -285,61 +296,73 @@ public void markDirty() {
...
-                            List<Aggregate.Matcher> inherited) {
+        public MatcherState(List<Aggregate.Matcher> matched, List<Aggregate.Matcher> inherited) {
             this.matched = matched;
             this.inherited = inherited;
 
             //Affected matches would only be used when there are
             //some matched matchers
             if (matched.isEmpty()) {
-                affectedMatchers = Collections.emptySet();
+                affectedMatchers = Set.of();
```
thomasmueller: I find it interesting to see that there is, in fact, a difference between the two:
https://stackoverflow.com/questions/68759254/why-is-set-of-implemented-differently-from-collections-emptyset

```
Collections.emptySet().contains(null); // false
Set.of().contains(null); // NullPointerException
```

I don't think we rely on this specific difference, but it is still interesting.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java
```diff
@@ -285,61 +296,73 @@ public void markDirty() {
...
      *
      * @param name modified property name
      */
     private void checkAggregates(String name) {
-        for (Aggregate.Matcher m : matcherState.matched) {
-            if (!matcherState.affectedMatchers.contains(m)
-                    && m.aggregatesProperty(name)) {
-                matcherState.affectedMatchers.add(m);
+        // Avoid allocating the iterator over matcherState.matched in the common case of the list being empty
+        if (!matcherState.matched.isEmpty()) {
```
thomasmueller: I first thought there is no measurable performance impact in this additional "if", but then I did measure and found that yes, indeed the additional condition does help. For things that are called millions of times, this should help a bit.

Unfortunately, Java doesn't support macros (custom "for" loops) like [some other languages](https://github.com/thomasmueller/bau-lang?tab=readme-ov-file#macros-and-ternary-condition)...

thomasmueller (approved):

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java
nfsantos: I found this when looking at a JFR profile of an indexing job using the OutOfBandIndexer. The allocation of the iterators over `matcherState.matched` was the 3rd or 4th largest source of object allocations. I was also surprised, I thought the JIT would be able to optimize away this allocation. It made me wonder if in performance critical regions, it may be better to iterate with an old-fashioned for loop with an explicit index, `for (int i=0; i<c.length...)`.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java
```diff
@@ -144,8 +145,11 @@ public void leave(NodeState before, NodeState after)
             }
         }
 
-        for (Aggregate.Matcher m : matcherState.affectedMatchers) {
-            m.markRootDirty();
+        for (int i = 0; i < matcherState.matched.size(); i++) {
+            if (matcherState.affectedMatchers.get(i)) {
+                Aggregate.Matcher m = matcherState.matched.get(i);
+                m.markRootDirty();
+            }
```
thomasmueller: Something like this:
```suggestion
        BitSet bitSet = matcherState.affectedMatchers;
        for (int i = bitSet.nextSetBit(0); i != -1; i = bitSet.nextSetBit(i + 1)) {
            Aggregate.Matcher m = matcherState.matched.get(i);
            m.markRootDirty();
        }
```

fabriziofortino (approved):

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java
nfsantos: Good idea, I changed the code to iterate over the BitSet instead over all the matchers.

PR #2099, opened 2025-02-19
OAK-11504 Elasticsearch: support flattened fields

thomasmueller:
Notes:
* Support is optional, and only enabled when using "isFlattened: true" in the property definition (the one that has "isRegexp: true")
* Flattened is not the default, for backward compatibility.
* If "analyzed" is enabled, then "flattened" is disabled (silently - logging it might log a lot, repeatedly - but we could make it an Oak-Pal rule).
* The hidden properties such as ":nodeName" are included still, even thought it is not needed. This is because changing it might break existing (weird) use cases. But we could consider doing that in the future. Best would be to make this optional or together with another breaking change (maybe the one above).
* Assuming we have a index definition regex property with the node name "allProperties" with a regular expression, and a property "propa", the document looks like this: `{:depth=2, flat:allProperties.propa=foo, :ancestors=/test, flat:allProperties.:nodeName=a}`
* Notice the `:nodeName` is also included, which is weird. Fixing that would break backward compatilibity, so I left it. Changing this would also affect Lucene. However, there is now disabled code where this is described, and that could be changed.
* I added a toString() method, to help debugging (there is no test case for this, and that's why code coverage complains).
* I fixed the indentation of `ElasticPropertyDefinition`, but I can undo this, if needed, to make the patch smaller.
* The query is `[{"term":{"flat:allProperties.propa":{"value":"foo"}}}]`
* The index definition is below; the last line is added:
* There are two test cases, one with many fields and flattened (including ordering), and another test without (disabled explicitly).

```
{  
    :depth=Property: {"type":"integer","doc_values":false}, 
    :fulltext=Property: {"type":"text","analyzer":"oak_analyzer"}, 
    :ancestors=Property: {"type":"text","analyzer":"ancestor_analyzer","search_analyzer":"keyword","search_quote_analyzer":"keyword"}, 
    :dynamic-boost-ft=Property: {"type":"text","analyzer":"oak_analyzer"}, 
    :suggest=Property: {"type":"nested","properties":{"value":{"type":"text","analyzer":"oak_analyzer"}}}, 
    :path=Property: {"type":"keyword"}, 
    :path-random-value=Property: {"type":"integer","doc_values":true,"index":false}, 
    :dynamic-properties=Property: {"type":"nested","properties":{"name":{"type":"keyword"},"value":{"type":"text","analyzer":"oak_analyzer"}}}, 
    :spellcheck=Property: {"type":"text","analyzer":"trigram"}, 
    :lastUpdated=Property: {"type":"date"}, 
    flat:allProperties=Property: {"type":"flattened","index":true}
}
```

### DIFF
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticPropertyDefinition.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticPropertyDefinition.java
@@ -46,6 +46,11 @@ public class ElasticPropertyDefinition extends PropertyDefinition {
     public static final String PROP_IS_FLATTENED = "isFlattened";
     private final boolean isFlattened;
 
+    /**
+     * The default value for the "isFlattened" property.
+     */
+    public static final boolean PROP_IS_FLATTENED_DEFAULT = false;
+
     public ElasticPropertyDefinition(IndexDefinition.IndexingRule idxDefn, String nodeName, NodeState defn) {
         super(idxDefn, nodeName, defn);
         if (this.useInSimilarity) {
@@ -56,7 +61,7 @@ public class ElasticPropertyDefinition extends PropertyDefinition {
                     getOptionalValue(defn, PROP_CANDIDATES, DEFAULT_CANDIDATES));
         }
         this.useInFullTextQuery = this.dynamicBoost && getOptionalValue(defn, PROP_USE_IN_FULL_TEXT_QUERY, true);
-        boolean flattened = getOptionalValue(defn, PROP_IS_FLATTENED, false);
+        boolean flattened = getOptionalValue(defn, PROP_IS_FLATTENED, PROP_IS_FLATTENED_DEFAULT);
         if (analyzed) {
             // if analyzed is enabled, then flattened needs to be disabled,
             // because flattened types do not support fulltext queries
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticPropertyDefinition.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticPropertyDefinition.java
@@ -24,96 +24,119 @@ import org.apache.jackrabbit.oak.spi.state.NodeState;
 
 public class ElasticPropertyDefinition extends PropertyDefinition {
 
-  public static final String DEFAULT_SIMILARITY_METRIC = "l2_norm";
-  static final String PROP_SIMILARITY_METRIC = "similarityMetric";
-  private static final String PROP_SIMILARITY = "similarity";
-  private static final String PROP_K = "k";
-  private static final String PROP_CANDIDATES = "candidates";
-  private static final float DEFAULT_SIMILARITY = 0.95f;
-  private static final int DEFAULT_K = 10;
-  private static final int DEFAULT_CANDIDATES = 500;
-  private KnnSearchParameters knnSearchParameters;
+    public static final String DEFAULT_SIMILARITY_METRIC = "l2_norm";
+    static final String PROP_SIMILARITY_METRIC = "similarityMetric";
+    private static final String PROP_SIMILARITY = "similarity";
+    private static final String PROP_K = "k";
+    private static final String PROP_CANDIDATES = "candidates";
+    private static final float DEFAULT_SIMILARITY = 0.95f;
+    private static final int DEFAULT_K = 10;
+    private static final int DEFAULT_CANDIDATES = 500;
+    private KnnSearchParameters knnSearchParameters;
 
-  /**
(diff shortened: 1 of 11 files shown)


oak-search-elastic/src/test/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticRegexPropertyIndexTest.java
```diff
@@ -0,0 +1,62 @@
...
+        setIndex("test1", builder);
+        root.commit();
+
+        Tree test = root.getTree("/").addChild("test");
+        test.addChild("a").setProperty("propa", "foo");
+        test.addChild("b").setProperty("propa", "foo");
+        test.addChild("c").setProperty("propa", "foo2");
+        test.addChild("d").setProperty("propc", "foo");
+        test.addChild("e").setProperty("propd", "foo");
+        root.commit();
```
fabriziofortino, line 61: minor: I have slightly modified the test to simulate a scenario with high cardinality and it works well. It does not break and performance are okay. Feel free to include it if you think is useful.
```suggestion
       // create 10k nodes with different property names to have high cardinality
        for (int i = 0; i < 10000; i++) {
            test.addChild("node" + i).setProperty("prop" + i, "foo");
        }
        root.commit();
```

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticPropertyDefinition.java
```diff
@@ -50,6 +61,14 @@ public ElasticPropertyDefinition(IndexDefinition.IndexingRule idxDefn, String no
                     getOptionalValue(defn, PROP_CANDIDATES, DEFAULT_CANDIDATES));
         }
         this.useInFullTextQuery = this.dynamicBoost && getOptionalValue(defn, PROP_USE_IN_FULL_TEXT_QUERY, true);
+        boolean flattened = getOptionalValue(defn, PROP_IS_FLATTENED, PROP_IS_FLATTENED_DEFAULT);
+        if (analyzed) {
+            // if analyzed is enabled, then flattened needs to be disabled,
+            // because flattened types do not support fulltext queries
```
nit0906, line 67: should we also add `nodeScopeIndex` to this exception ? As per the oak docs 

```
nodeScopeIndex
Control whether the value of a property should be part of fulltext index. That is, you can do a jcr:contains(., 'foo') and it will return nodes that have a string property that contains the word foo. Example
/jcr:root/content//element(*, app:Asset)[jcr:contains(., 'image')]

```

nit0906 (approved):
lgtm - just one minor comment/question

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticPropertyDefinition.java
```diff
@@ -50,6 +61,14 @@ public ElasticPropertyDefinition(IndexDefinition.IndexingRule idxDefn, String no
                     getOptionalValue(defn, PROP_CANDIDATES, DEFAULT_CANDIDATES));
         }
         this.useInFullTextQuery = this.dynamicBoost && getOptionalValue(defn, PROP_USE_IN_FULL_TEXT_QUERY, true);
+        boolean flattened = getOptionalValue(defn, PROP_IS_FLATTENED, PROP_IS_FLATTENED_DEFAULT);
+        if (analyzed) {
```
fabriziofortino, line 65: I am wondering if we should actually fail in this case.
If we don't, the index won't be created as flattened and it might fail while indexing in case the number of fields exceeds the maximum.

fabriziofortino (commented):
Overall it looks good. Just wondering if we should fail immediately when a regex property is configured as flattened and analyzed.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticPropertyDefinition.java
thomasmueller: First I thought so, yes. But actually it's not needed: nodeScopeIndex adds the text to the ":fulltext" field, and that one is not flattened (doesn't need to be).

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticPropertyDefinition.java
thomasmueller: > I am wondering if we should actually fail in this case.

Sorry I saw the comment too late...

Yes, we could reject this combination. The main reason I didn't is: If we change the default, this would immediately fail many tests. I wanted to be able to test with the setting enabled by default... Sure, there are other options as well: for example, only fail if analyzed is enabled and flattened is explicitly set.

PR #2102, opened 2025-02-20
OAK-11509 Incremental flat file store: reduce logging

thomasmueller:
Most of the messages are actually in the document store:

    diffJournalChildren failed with ..., falling back to classic diff

I also saw that iteration count is 32 bit: it should be 64 bit everywhere.

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
@@ -3429,9 +3429,14 @@ public final class DocumentNodeStore
                 // see OAK-6016 and OAK-6011
                 if (e instanceof IllegalStateException &&
                         "Root document does not have a lastRev entry for local clusterId 0".equals(e.getMessage())) {
+<<<<<<< Updated upstream
                     LOG.debug("diffJournalChildren failed with " +
                             e.getClass().getSimpleName() +
                             ", falling back to classic diff : " + e.getMessage());
+=======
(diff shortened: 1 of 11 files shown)


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
```diff
@@ -3429,7 +3429,7 @@ private String diffImpl(AbstractDocumentNodeState from, AbstractDocumentNodeStat
                 // see OAK-6016 and OAK-6011
                 if (e instanceof IllegalStateException &&
                         "Root document does not have a lastRev entry for local clusterId 0".equals(e.getMessage())) {
-                    LOG.warn("diffJournalChildren failed with " +
+                    LOG.debug("diffJournalChildren failed with " +
```
nfsantos: It is more efficient to use the parametrized version of this method (with {} placeholders). As this log will rarely be printed, it is wasteful to create the String that is passed as an argument.

nfsantos (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
fabriziofortino: +1

fabriziofortino (approved):

thomasmueller:
I'm merging it now, bug @Joscorbe @rishabhdaim if you find it is not good then we can discuss the change for diffJournalChildren!

PR #2108, opened 2025-02-21
OAK-11510 - Performance improvements to IndexDefinition class

nfsantos:
- Replace a HashMap with the property configs that was being used with case-insensitive keys by always converting to lower case Strings used in get/put operations, by a TreeMap using a case-insensitive String comparator, thereby avoiding the need to convert to lower-case before put/get. This reduces significantly object allocation during indexing, because this map is queried for every node that is indexed.
- Add an additional field with the list of `PropertyDefinitions`, which is a copy of the values of the `PropertyConfig` map. It is faster to iterate directly over this list than to iterate over the values of the map. This should give a significant speedup to indexing operations, as they call very often these operations.
- Do not create intermediate array in `getApplicableIndexingRule()` methods and return from method as soon as first matching rule is found.

Other minor cleanups.

### DIFF
--- a/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/IndexDefinition.java
+++ b/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/IndexDefinition.java
@@ -53,7 +53,6 @@ import javax.jcr.RepositoryException;
 import javax.jcr.nodetype.NodeTypeIterator;
 import java.io.InputStream;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -61,6 +60,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.UUID;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
@@ -679,7 +679,7 @@ public class IndexDefinition implements Aggregate.AggregateMapper {
     }
 
     public String getTikaMappedMimeType(String type) {
-        return customTikaMimeTypeMappings.getOrDefault(type, type);
+        return type == null ? null : customTikaMimeTypeMappings.getOrDefault(type, type);
     }
 
     public String getIndexName() {
@@ -966,7 +966,7 @@ public class IndexDefinition implements Aggregate.AggregateMapper {
 
     private boolean evaluateSuggestionEnabled() {
         for (IndexingRule indexingRule : definedRules) {
-            for (PropertyDefinition propertyDefinition : indexingRule.propConfigs.values()) {
+            for (PropertyDefinition propertyDefinition : indexingRule.propDefinitions) {
                 if (propertyDefinition.useInSuggest) {
                     return true;
                 }
@@ -994,7 +994,7 @@ public class IndexDefinition implements Aggregate.AggregateMapper {
 
     private boolean evaluateSpellcheckEnabled() {
         for (IndexingRule indexingRule : definedRules) {
-            for (PropertyDefinition propertyDefinition : indexingRule.propConfigs.values()) {
+            for (PropertyDefinition propertyDefinition : indexingRule.propDefinitions) {
                 if (propertyDefinition.useInSpellcheck) {
                     return true;
                 }
@@ -1045,7 +1045,8 @@ public class IndexDefinition implements Aggregate.AggregateMapper {
         /**
          * Case-insensitive map of lower cased propertyName to propertyConfigs
          */
-        private final Map<String, PropertyDefinition> propConfigs;
+        private final TreeMap<String, PropertyDefinition> propConfigs;
+        private final List<PropertyDefinition> propDefinitions;
         /**
          * List of {@code NamePattern}s configured for this rule
          */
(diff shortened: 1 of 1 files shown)


oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/IndexDefinition.java
```diff
@@ -887,31 +884,36 @@ public IndexingRule getApplicableIndexingRule(String primaryNodeType) {
...
-        List<IndexingRule> r = indexRules.get(getPrimaryTypeName(state));
-        if (r != null) {
-            rules = new ArrayList<>(r);
+        {
+            List<IndexingRule> rules = indexRules.get(getPrimaryTypeName(state));
+            IndexingRule rule = getApplicableIndexingRule(state, rules);
+            if (rule != null) {
+                return rule;
+            }
         }
```
thomasmueller:
```suggestion
        List<IndexingRule> rules = indexRules.get(getPrimaryTypeName(state));
        IndexingRule rule = getApplicableIndexingRule(state, rules);
        if (rule != null) {
            return rule;
        }
```

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/IndexDefinition.java
```diff
@@ -887,31 +884,36 @@ public IndexingRule getApplicableIndexingRule(String primaryNodeType) {
...
 
         for (String name : getMixinTypeNames(state)) {
-            r = indexRules.get(name);
-            if (r != null) {
-                if (rules == null) {
-                    rules = new ArrayList<>();
-                }
-                rules.addAll(r);
+            List<IndexingRule> rules = indexRules.get(name);
+            IndexingRule rule = getApplicableIndexingRule(state, rules);
```
thomasmueller:
```suggestion
            rule = getApplicableIndexingRule(state, rules);
```

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/IndexDefinition.java
```diff
@@ -1302,16 +1307,16 @@ public boolean isBasedOnNtBase() {
...
+        private TreeMap<String, PropertyDefinition> collectPropConfigs(NodeState config,
+                                                                       List<NamePattern> patterns,
+                                                                       List<Aggregate.Include> propAggregate,
+                                                                       List<PropertyDefinition> nonExistentProperties,
+                                                                       List<PropertyDefinition> existentProperties,
+                                                                       List<PropertyDefinition> nodeScopeAnalyzedProps,
+                                                                       List<PropertyDefinition> functionRestrictions,
+                                                                       List<PropertyDefinition> syncProps,
+                                                                       List<PropertyDefinition> similarityProperties) {
+            TreeMap<String, PropertyDefinition> propDefns = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
thomasmueller, line 1317: I'm not sure if TreeMap with CASE_INSENSITIVE_ORDER is faster than HashMap... I know, toLowerCase is not needed, but then, aren't most property names already lowercase? So toLowerCase() will just return "this".

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/IndexDefinition.java
```diff
@@ -1395,27 +1400,27 @@ private Map<String, PropertyDefinition> collectPropConfigs(NodeState config,
                 }
             }
             ensureNodeTypeIndexingIsConsistent(propDefns, syncProps);
-            return Map.copyOf(propDefns);
```
thomasmueller, line 1398: Hm, we used to copy the map... I don't know why... But are you sure this is not needed?

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/IndexDefinition.java
nfsantos: done

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/IndexDefinition.java
nfsantos: I have noticed this optimization opportunity while profiling a run of the indexing job, where the calls to `toLowerCase()` were responsible for a significant part of object allocation in that run, and also showed up on CPU profiling. Property names are usually Camel case, so most of them will in fact have to be converted to lower case as most consist of two or more logical words.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/IndexDefinition.java
nfsantos: I don't see a reason why we must make a copy. Once this method finishes, there is no other reference to the map other than the one returned by the method, so there is no risk of the map being modified elsewhere. And the IndexDefinition class does not modify the map anywhere, just reads it. So I don't think it is necessary to make a copy or to  make the map immutable. Maybe it was done just as a best-practice to ensure immutability. Or as an attempt at performance optimization, as an immutable map created by `Map.copyOf()` may be faster than a HashMap due to a more efficient internal representation. I am not sure, but in this case, avoiding the creation of lower case strings is a big gain, easily offsets any additional overhead of searching for a key on a tree as compared to search on a hashmap. Searching on a HashMap also requires computing the hash of the String, which can be expensive.

thomasmueller (approved):

PR #2111, opened 2025-02-21
OAK-11518 : added utils method to replace Iterables.partition

rishabhdaim:


oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/collections/IterableUtils.java
```diff
@@ -212,4 +214,46 @@ public static <T> T[] toArray(final Iterable<T> itr, final Class<T> type) {
...
+     *
+     * @param <T> the type of elements in the itr
+     * @param itr the itr to split, may not be null
+     * @param size the size of each sub-iterator, must be greater than 0
+     * @return an iterator of sub-iterators, each of the specified size
+     * @throws NullPointerException if the itr is null
+     * @throws IllegalArgumentException if size is less than or equal to 0
+     */
+    public static <T> Iterable<List<T>> partition(final Iterable<T> itr, final int size) {
+        if (itr == null) {
```
reschke: Why can't we use "requireNonNull" here? (or even do not do it at all?)

rishabhdaim: IMO, we should fail fast.

rishabhdaim: addressed in https://github.com/apache/jackrabbit-oak/pull/2111/commits/8103c0116fc53d00436f65a5a0f2339b6c9b3ff0

reschke: Not having the null check would fail fast as well here.

My opinion is that requireNonNull is completely unnecessary noise in 99% of the cases where it's used.

reschke (approved):

mbaedke (changes requested):
The following test case fails:

`    public void testTmp() {
        Iterable<List<Integer>> partition = IterableUtils.partition(Collections.emptyList(), 1);
        Iterator<List<Integer>> iterator = partition.iterator();
        Assert.assertFalse(iterator.hasNext());
        Assert.assertThrows(NoSuchElementException.class, iterator::next);
    }
`

IMHO it should not.

rishabhdaim:
> The following test case fails:
> 
> ` public void testTmp() { Iterable<List<Integer>> partition = IterableUtils.partition(Collections.emptyList(), 1); Iterator<List<Integer>> iterator = partition.iterator(); Assert.assertFalse(iterator.hasNext()); Assert.assertThrows(NoSuchElementException.class, iterator::next); }`
> 
> IMHO it should not.

addressed in https://github.com/apache/jackrabbit-oak/pull/2111/commits/c2a3b4b2a4cc7d0cbb31ecf7f3ac8702bf5738af

mbaedke (approved):

PR #2113, opened 2025-02-24
OAK-11520 - Performance improvements to Aggregate class in indexing logic

nfsantos:
- `Aggregate.getElementNameIfNotAPattern()` is creating a String representation of the elements arrays every time it is called, even though it was only needed if the call to `checkArgument()` fails, which should never happen. Replace this call by a conditional that avoids this expensive conversion to String.
- In calls to collectAggregates, avoid the creation of ArrayList<Matcher> instances, instead use Matcher[]. Iteration over arrays with the enhanced for loop are converted to indexed traversals, which avoids the creation of an Iterator object while preserving the clean syntax of the enhanced for loop.
- Replace usages of static constructor methods  from Collections by corresponding methods in Set and Map.
- In collectAggregatesForDirectMatchers(), do not create temporary children map and do not call matchChildren() unless there is a match among the children.
- Fix white space.

### DIFF
--- a/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/Aggregate.java
+++ b/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/Aggregate.java
@@ -160,6 +160,7 @@ public class Aggregate {
 
     private static void matchChildren(Matcher[] matchers, ResultCollector collector,
                                       Iterable<? extends ChildNodeEntry> children) {
+        // Performance critical code: create nextSet lazily. And once created, reuse the same instance.
         List<Matcher> nextSet = null;
         for (ChildNodeEntry cne : children) {
             for (Matcher m : matchers) {
@@ -177,6 +178,7 @@ public class Aggregate {
             }
             if (nextSet !=null && !nextSet.isEmpty()) {
                 collectAggregates(cne.getNodeState(), nextSet.toArray(new Matcher[0]), collector);
+                // Clear the set so it can be reused. This reduces object allocation overhead.
                 nextSet.clear();
             }
         }
--- a/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/Aggregate.java
+++ b/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/Aggregate.java
@@ -60,7 +60,7 @@ public class Aggregate {
     private final String nodeTypeName;
     private final List<? extends Include> includes;
     public final int reAggregationLimit;
-    private final List<NodeInclude> relativeNodeIncludes;
+    private final NodeInclude[] relativeNodeIncludes;
     private final boolean nodeAggregates;
 
     Aggregate(String nodeTypeName) {
@@ -95,7 +95,8 @@ public class Aggregate {
         for (int i = 0; i < matchers.length; i++) {
             matchers[i] = new Matcher(this, includes.get(i), root);
         }
-        return List.of(matchers);
+        // Wrap the array in an ArrayList, this avoids copying the array
+        return Arrays.asList(matchers);
     }
 
     public boolean hasRelativeNodeInclude(String nodePath) {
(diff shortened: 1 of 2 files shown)


oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/Aggregate.java
```diff
@@ -588,37 +588,36 @@ public Matcher match(String name, NodeState nodeState) {
...
                     List<Matcher> result = new ArrayList<>(nextAgg.includes.size());
-                    for (Include i : nextAgg.includes){
-                        result.add(new Matcher(this,  i, currentPath));
+                    for (Include i : nextAgg.includes) {
+                        result.add(new Matcher(this, i, currentPath));
                     }
                     return result;
                 }
-                return Collections.emptyList();
+                return List.of();
```
alagodasii: Actually, what's the benefit of using `List.of()` in comparison to `emptyList()`? Both create immutable lists without elements.

nfsantos: Several reasons:
- `List.of()` is a more modern API and seems to be the preferred choice for new APIs.

- It reads better and is more maintainable when used with the other static factory constructors in the List class:
```
List.of()
List.of("a")
List.of(1,2,3)
```
If we want to change between empty and non-empty, it is easier. And is more consistent.

- Performance may also be better is using only List.of() for every list used in a particular code path. This is because List.of() will return only 2 different implementations of the List class, which will mean that any call site that uses these objects will be mono or bimorphic. The JIT compiler can optimize these call sites to have similar performance as static calls. However, if there are more than 2 types of lists, the call site becomes polimorphic and performance is worse. Usually, this will have no measurable difference in performance, but in performance critical loops like here, the impact can be significant. 
https://stackoverflow.com/questions/33317720/performance-of-collections-emptylist-and-empty-arraylist-with-jit-compiler

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/Aggregate.java
```diff
@@ -132,82 +126,90 @@ private static boolean matchingType(String nodeTypeName, NodeState nodeState) {
...
+                        nextSet = new ArrayList<>();
+                    }
+                    result.nextSet(nextSet);
                 }
             }
-            if (!nextSet.isEmpty()) {
-                collectAggregates(cne.getNodeState(), nextSet, collector);
+            if (nextSet !=null && !nextSet.isEmpty()) {
+                collectAggregates(cne.getNodeState(), nextSet.toArray(new Matcher[0]), collector);
+                nextSet.clear();
```
thomasmueller, line 182: I had some trouble understanding why nextSet.clear() is needed... Yes, it is correct and matches the previous logic: before, a new ArrayList was created inside the  the children loop. Now, this is outside the loop.
Yes, it is probably faster that way, but I'm worried that the logic is harder to understand, and in the future might cause some trouble... Not now... now the code is fine! But after some more changes.

thomasmueller (changes requested):

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/Aggregate.java
alagodasii: For the readability reasons, `emptyList()` vs `List.of()` is arguable. Some could say, that "list of nothing" is less readable than "empty list of whatever". More interesting is that performance impact. I'm curious what would be the gain of performance after switch. Since, there are no performance tests in place, it would be worth to monitor executions and compare to historical ones.

nfsantos: The readable improvement is mainly to use a uniform way of creating immutable lists.  I think this is nicer:
```
List.of()
List.of("a")
List.of(1,2,3)
```
than
```
Collections.emptyList()
List.of("a")
List.of(1,2,3)
```
The performance improvements of bimorphic vs polimorphic calls will likely not be measurable in this case. However, as using `List.of` everywhere arguably makes the code more consistent and simpler, any minor gains are just a welcome extra.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/Aggregate.java
nfsantos: Good point, I added some comments to explain the intent of this optimization. Let me know if it is clear.

PR #2115, opened 2025-02-25
OAK-11522 : removed usage of Guava's Iterables.filter

rishabhdaim:


reschke (approved):
This appears to be completely mechanical.

PR #2120, opened 2025-02-26
OAK-11529 RegExp recursion in AstElement - workaround

thomasmueller:
Workaround for - what seems to be - a JDK bug

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/AstElement.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/AstElement.java
@@ -41,12 +41,12 @@ abstract class AstElement {
         }
     }
 
-    protected String quote(String pathOrName) {
+    protected static String quote(String pathOrName) {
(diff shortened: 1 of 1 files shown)


nfsantos:
Can you explain what the bug is and how this PR avoids the issue?

nfsantos:
For consistency, I'd recommend to also make `static` the method `protect()` and to reorder the methods to follow the conventions of static methods appearing first in the class.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/AstElement.java
```diff
@@ -41,12 +41,12 @@ protected String protect(Object expression) {
         }
     }
 
-    protected String quote(String pathOrName) {
+    protected static String quote(String pathOrName) {
         pathOrName = pathOrName.replaceAll("]", "]]");
         return '[' + pathOrName + ']';
```
reschke, line 46: While we are at it, I would collapse that into a single line, avoiding the reassignment...

PR #2121, opened 2025-02-26
OAK-11530 Elasticsearch: 'field name cannot be an empty string' 

thomasmueller:
if fields start with a dot

### DIFF
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexWriter.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexWriter.java
@@ -204,9 +204,33 @@ class ElasticIndexWriter implements FulltextIndexWriter<ElasticDocument> {
             if (ese.status() == 400 && ese.getMessage().contains("resource_already_exists_exception")) {
                 LOG.warn("Index {} already exists. Ignoring error", indexName);
             } else {
+                LOG.warn("Failed to create index {}", indexName, ese);
+                StringBuilder sb = new StringBuilder();
+                int old = JsonpUtils.maxToStringLength();
+                try {
+                    JsonpUtils.maxToStringLength(16_000_000);
+                    JsonpUtils.toString(request, sb);
+                    String[] array = splitLargeString(sb.toString(), 1024);
+                    for (int i = 0; i < array.length; i++) {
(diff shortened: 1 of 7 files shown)


fabriziofortino (approved):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexWriter.java
```diff
@@ -204,11 +204,35 @@ private void provisionIndex() throws IOException {
...
+                try {
+                    JsonpUtils.maxToStringLength(16_000_000);
+                    JsonpUtils.toString(request, sb);
+                    String[] array = splitLargeString(sb.toString(), 1024);
+                    for (int i = 0; i < array.length; i++) {
+                        LOG.warn("request chunk[{}] = {}", i, array[i]);
+                    }
+                } finally {
+                    JsonpUtils.maxToStringLength(old);
+                }
```
nfsantos: This may print 16MB worth of data to the logs, which would likely overload the logger and any log management systems that may be capturing the logs, which could lead to us losing other logs which are more important. And all this data would likely not be useful. I suggest that we truncate the string to a reasonable size, like 1K or so.

nfsantos: And allocating 16MB of memory just to create a log message also sounds risky, as it can create memory pressure. For instance, if there are many writers, all hitting the same error, they would all in a short period of time allocate up to 32MB of memory (16MB for the full Json representation, then 16MB for the array split).

thomasmueller: Notice this is just the _maximum_ length. The default maximum length is 10'000 characters, and unfortunately, the interesting part is not logged at all at that length. The request for a "normal" index definition is around 13'000 characters. So we could limit to 100'000. However, possibly this is not enough: it is hard to tell. But I changed it to 1 MB now.

> allocating 16MB of memory

It doesn't allocate 16 MB of memory. Internally, JsonpUtils uses StringBuilder, which doesn't allocate 16 MB either, if the message is smaller.

> which could lead to us losing other logs which are more important

This message is important, because without it, it is very hard to understand why building the index failed. Truncating this message would be very problematic.

PR #2129, opened 2025-03-03
 OAK-11500: Remove usage of Guava io.Closeable 

reschke:


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/directory/ActiveDeletedBlobCollectorFactory.java
```diff
@@ -324,7 +323,11 @@ public void purgeBlobsDeleted(long before, @NotNull GarbageCollectableBlobStore
             long startBlobTrackerSyncTime = clock.getTime();
             // Synchronize deleted blob ids with the blob id tracker
             try {
-                Closeables.close(idTempDeleteWriter, true);
+                try {
```
rishabhdaim, line 326: Can't we create an Util for this. Instead of repeating these 4 lines, put them under an Utility class in oak-commons. wdyt @reschke ?

rishabhdaim (approved):

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/directory/ActiveDeletedBlobCollectorFactory.java
reschke: I really don't think we need this. In this case, inlining actually makes things clearer.

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/AbstractSharedCachingDataStore.java
```diff
@@ -335,7 +334,7 @@ public InputStream getStream() throws DataStoreException {
                         copyInputStreamToFile(in, tmpFile);
                         return new LazyFileInputStream(tmpFile);
                     } finally {
-                        Closeables.close(in, false);
+                        in.close();
```
thomasmueller, line 337: What if "in" is null?

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
```diff
@@ -380,7 +375,13 @@ protected void markAndSweep(boolean markOnly, boolean forceBlobRetrieve) throws
         } finally {
             statsCollector.updateDuration(sw.elapsed(TimeUnit.MILLISECONDS), TimeUnit.MILLISECONDS);
             if (!LOG.isTraceEnabled() && !traceOutput) {
-                Closeables.close(fs, threw);
+                try {
+                    fs.close();
```
thomasmueller, line 379: What if "fs" is null?

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
```diff
@@ -769,7 +770,13 @@ public long checkConsistency(boolean markOnly) throws Exception {
             }
         } finally {
             if (!traceOutput && (!LOG.isTraceEnabled() && candidates == 0)) {
-                Closeables.close(fs, threw);
+                try {
+                    fs.close();
```
thomasmueller, line 774: What if "fs" is null?

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
```diff
@@ -1091,7 +1098,7 @@ void retrieve(GarbageCollectableBlobStore blobStore,
             } finally {
                 if (idsIter instanceof Closeable) {
                     try {
-                        Closeables.close((Closeable) idsIter, false);
+                        ((Closeable)idsIter).close();
```
thomasmueller, line 1101: null check

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/datastore/DataStoreBlobStore.java
```diff
@@ -341,9 +340,15 @@ public String writeBlob(InputStream stream, BlobOptions options) throws IOExcept
             stats.uploadFailed();
             throw new IOException(e);
         } finally {
-            //DataStore does not closes the stream internally
+            //DataStore does not close the stream internally
             //So close the stream explicitly
-            Closeables.close(stream, threw);
+            try {
+                stream.close();
```
thomasmueller, line 346: null check

thomasmueller (commented):
This is a very dangerous commit.

PR #2130, opened 2025-03-04
OAK-11541: Reduce number of traversed documents on first iterations of FullGC

Joscorbe:


Joscorbe:
Waiting for now, let's try #2134 first.

PR #2134, opened 2025-03-04
OAK-11541: Reduce logQueryExplain to only happen if Trace logs are enabled

Joscorbe:
Apparently the query explain in Mongo can take `ages` to run. One of my local executions took more than 7 hours to run, on a database with a massive amount of documents.

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
@@ -197,10 +197,9 @@ public class MongoVersionGCSupport extends VersionGCSupport {
     }
 
     /**
-     * Logs an explain of a mongo query. If log level is INFO it does it one-lined,
-     * if log level is DEBUG it does it pretty print multi-lined.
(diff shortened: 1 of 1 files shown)


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
```diff
@@ -209,24 +208,19 @@ private Bson withIncludeExcludes(@NotNull Set<String> includes, @NotNull Set<Str
...
-        final BasicDBObject winningPlan = MongoUtils.getWinningPlan(explainResult);
-        final BasicDBObject result = winningPlan == null ? explainResult : winningPlan;
-        if (LOG.isDebugEnabled()) {
-            // if log level is DEBUG, let's do a pretty print
-            String prettyPrinted = JsopBuilder.prettyPrint(result.toJson());
-            LOG.debug(logMsg, hint, prettyPrinted);
-        } else {
-            // otherwise let's just do a compact print
-            LOG.info(logMsg, hint, result);
+        if (LOG.isTraceEnabled()) {
```
rishabhdaim: I think we should call `logQueryExplain` itself from `isTraceEnabled` block.

rishabhdaim (approved):

PR #2136, opened 2025-03-04
OAK-11500: Remove usage of Guava io.Closeable

reschke:


oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/datastore/OakFileDataStore.java
```diff
@@ -139,12 +138,13 @@ public void addMetadataRecord(InputStream input, String name)
...
         try {
             File file = new File(getPath(), name);
-            FileOutputStream os = new FileOutputStream(file);
-            try {
+            try (FileOutputStream os = new FileOutputStream(file)) {
                 IOUtils.copyLarge(input, os);
             } finally {
-                Closeables.close(os, true);
-                Closeables.close(input, true);
+                try {
```
rishabhdaim: Shouldn't we use `IOUtils.closeQuietly` here too, since we are not doing anything with exception?

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/datastore/FSBackend.java
```diff
@@ -189,12 +188,13 @@ public void addMetadataRecord(InputStream input, String name)
...
         try {
             File file = new File(fsPathDir, name);
-            FileOutputStream os = new FileOutputStream(file);
-            try {
+            try (FileOutputStream os = new FileOutputStream(file)) {
                 IOUtils.copyLarge(input, os);
             } finally {
-                Closeables.close(os, true);
-                Closeables.close(input, true);
+                try {
```
rishabhdaim: same here.

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/datastore/OakFileDataStore.java
reschke: yep

rishabhdaim (approved):

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
```diff
@@ -380,7 +375,13 @@ protected void markAndSweep(boolean markOnly, boolean forceBlobRetrieve) throws
         } finally {
             statsCollector.updateDuration(sw.elapsed(TimeUnit.MILLISECONDS), TimeUnit.MILLISECONDS);
             if (!LOG.isTraceEnabled() && !traceOutput) {
-                Closeables.close(fs, threw);
+                try {
+                    fs.close();
```
thomasmueller: This can throw a NullPointerException.

thomasmueller (changes requested):

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/MarkSweepGarbageCollector.java
reschke: How so?

thomasmueller (approved):

rishabhdaim (approved):

PR #2137, opened 2025-03-04
OAK-11267 - Upgrade Azure SDK V8 to V12 for oak-blob-azure

andreeastroe96:


oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
```diff
@@ -1264,34 +1132,41 @@ private boolean loadItems() {
...
                     return false;
                 }
-                firstCall = false;
-                ResultSegment<ListBlobItem> results = container.listBlobsSegmented(null, false, EnumSet.noneOf(BlobListingDetails.class), null, resultContinuation, null, null);
-                resultContinuation = results.getContinuationToken();
-                for (ListBlobItem item : results.getResults()) {
-                    if (item instanceof CloudBlob) {
-                        items.add(AzureBlobInfo.fromCloudBlob((CloudBlob)item));
-                    }
+
```
amit-jain: Is this needed? Can't we just return the blobItems.getIterator()? Pagination based on max results is already handled by the sdk, https://learn.microsoft.com/en-us/azure/developer/java/sdk/pagination

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/Constants.java
```diff
@@ -0,0 +1,37 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.jackrabbit.oak.blob.cloud.azure.blobstorage;
+
+public interface Constants {
```
amit-jain, line 21: Update to latest values

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/binary/fixtures/datastore/AzureDataStoreFixture.java
```diff
@@ -94,12 +97,22 @@ public DataStore createDataStore() {
 
         String connectionString = Utils.getConnectionStringFromProperties(azProps);
         try {
-            CloudBlobContainer container = Utils.getBlobContainer(connectionString, containerName);
-            container.createIfNotExists();
+            boolean useSDK12 = Boolean.parseBoolean(azProps.getProperty(AZURE_SDK_12_ENABLED, "false"));
```
amit-jain, line 100: What about tests/implementations dependent upon the AzureDataStoreFixture, we should have parametrized tests or some other mechanism that test those with both v12 & v8 versions. 
There are many command supproted in oak run utility which would have corresponding tests which should support above.

amit-jain (changes requested):

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureBlobStoreBackend.java
amit-jain: Consider just using the sdk without any decorator as well. Just returning blobItems.getIterator() directly in the getAll* methods

amit-jain (approved):
lgtm!

oak-blob-cloud-azure/oak-blob-cloud-azure/pom.xml
reschke, line 1: This seems to be misplaced (see folder name).

oak-blob-cloud-azure/pom.xml
```diff
@@ -68,6 +71,14 @@
...
                             azure-identity,
                             azure-json,
+                            azure-xml,
+                            azure-storage-blob,
+                            azure-storage-common,
+                            azure-storage-internal-avro,
+                            com.microsoft.aad,
+                            com.microsoft.aad.msal4jextensions,
+                            com.microsoft.aad.msal4jextensions.persistence,
+                            jackson-dataformat-xml,
```
reschke, line 81: Why do we need to embed all ot this? For instance, jackson???

oak-blob-cloud-azure/pom.xml
```diff
@@ -170,6 +181,11 @@
             <groupId>com.microsoft.azure</groupId>
             <artifactId>azure-storage</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.azure</groupId>
+            <artifactId>azure-storage-blob</artifactId>
+            <version>12.27.1</version>
+        </dependency>
```
reschke: I'd move this so it's not between two com.microsoft.azure dependencies.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/Utils.java
```diff
@@ -16,114 +16,88 @@
...
 import org.jetbrains.annotations.Nullable;
 
-public final class Utils {
-
+public class Utils {
+    public static final String DASH = "-";
     public static final String DEFAULT_CONFIG_FILE = "azure.properties";
 
-    public static final String DASH = "-";
+    public Utils() {}
```
reschke: Is making this public and non-final intentional?

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v8/UtilsV8.java
```diff
@@ -0,0 +1,167 @@
...
+import com.microsoft.azure.storage.RetryExponentialRetry;
+import com.microsoft.azure.storage.RetryNoRetry;
+import com.microsoft.azure.storage.RetryPolicy;
+import com.microsoft.azure.storage.StorageException;
+import com.microsoft.azure.storage.blob.BlobRequestOptions;
+import com.microsoft.azure.storage.blob.CloudBlobClient;
+import com.microsoft.azure.storage.blob.CloudBlobContainer;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.jackrabbit.core.data.DataStoreException;
+import com.google.common.base.Strings;
```
reschke, line 33: No use of Guava, please.

oak-run-elastic/pom.xml
```diff
@@ -42,7 +42,7 @@
         105 MB: Azure updates
         107 MB: RDB/Tomcat (OAK-10752)
         -->
-        <max.jar.size>115000000</max.jar.size>
+        <max.jar.size>115550000</max.jar.size>
```
reschke, line 46: nit: please add a comment explaining the increase

reschke (changes requested):

oak-blob-cloud-azure/pom.xml
andreeastroe96: Without all this, when testing the changes in a AEM build, it failed

reschke: Well, that information is insufficient.

It just hints that deploying this in an OSGi environment requires additional components. Find out which they are and deploy them.

Optimally, include the component in OsgiIT and see what happens (that's the reason why we have these tests).

PR #2139, opened 2025-03-05
OAK-11553 Elastic: index name is swallowed in case of IllegalArgument…

thomasmueller:
…Exception

### DIFF
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexWriter.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexWriter.java
@@ -189,7 +189,7 @@ class ElasticIndexWriter implements FulltextIndexWriter<ElasticDocument> {
         try {
             request = ElasticIndexHelper.createIndexRequest(indexName, indexDefinition);
         } catch (IllegalArgumentException e) {
-            LOG.warn("Failed to create index {}: {}", indexName, e.toString());
+            LOG.error("Failed to create index {}: {}", indexName, e.toString());
             throw e;
         }
         LOG.debug("Creating Index with request {}", request);
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexWriter.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexWriter.java
@@ -185,7 +185,13 @@ class ElasticIndexWriter implements FulltextIndexWriter<ElasticDocument> {
             return;
         }
 
-        final CreateIndexRequest request = ElasticIndexHelper.createIndexRequest(indexName, indexDefinition);
+        CreateIndexRequest request;
+        try {
+            request = ElasticIndexHelper.createIndexRequest(indexName, indexDefinition);
+        } catch (IllegalArgumentException e) {
+            LOG.warn("Failed to create index {}", indexName, e);
+            throw e;
+        }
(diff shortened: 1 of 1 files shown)


reschke (approved):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexWriter.java
```diff
@@ -185,7 +185,13 @@ private void provisionIndex() throws IOException {
             return;
         }
 
-        final CreateIndexRequest request = ElasticIndexHelper.createIndexRequest(indexName, indexDefinition);
+        CreateIndexRequest request;
+        try {
+            request = ElasticIndexHelper.createIndexRequest(indexName, indexDefinition);
+        } catch (IllegalArgumentException e) {
+            LOG.warn("Failed to create index {}", indexName, e);
+            throw e;
```
nfsantos: Won't the exception be logged twice? Maybe here we could just print the `.toString()` of the exception (in addition to the index name) and then let the top level handler that catches this exception log the stack trace, to avoid duplicate traces in the logs. Or create a new exception wrapping this exception and with a message that includes the index name. This could be nicer to have a single log entry with all the information, index name and stack trace.

nfsantos (approved):
Just have a minor comment.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexWriter.java
thomasmueller: > Won't the exception be logged twice?

Yes. I don't consider this a huge problem, but yes we can avoid it.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexWriter.java
```diff
@@ -185,7 +185,13 @@ private void provisionIndex() throws IOException {
             return;
         }
 
-        final CreateIndexRequest request = ElasticIndexHelper.createIndexRequest(indexName, indexDefinition);
+        CreateIndexRequest request;
+        try {
+            request = ElasticIndexHelper.createIndexRequest(indexName, indexDefinition);
+        } catch (IllegalArgumentException e) {
```
fabriziofortino: should we be more generic here and catch `Exception`?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexWriter.java
```diff
@@ -185,7 +185,13 @@ private void provisionIndex() throws IOException {
...
+            LOG.warn("Failed to create index {}: {}", indexName, e.toString());
```
fabriziofortino: this should be logged as `error`

thomasmueller:
```suggestion
            LOG.error("Failed to create index {}: {}", indexName, e.toString());
```

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexWriter.java
thomasmueller:
```suggestion
        } catch (Exception e) {
```

fabriziofortino (approved):

PR #2145, opened 2025-03-06
OAK-11555 Elastic: support dot in property and function names

thomasmueller:
This is longer than expected. The problem is that there was no conversion function from "JCR property name" to "Elasticsearch field name" so far, so I had to add it everywhere. That also affects tests etc.

Side node: ElasticIndexUtilsTest: I also added unit tests of the existing functions. I find it really weird that we convert randomly distributed bytes to a string like so: 

    new String(<byte array with completely random values>)

... and then use that as the String value for "_id" field in Elasticsearch. But... (a) changing it would introduce incompatibility, and (b) it seems to work.

How I tested: I changed the ElasticIndexUtils.fieldName method to return "X" + the actual result, and like this could find other places where the method is needed.

I added a method to undo (revert / unescape) the conversion, so that we can verify it is always possible to do so. Even thought, right now it is not needed.

### DIFF
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
@@ -31,6 +31,7 @@ import java.util.stream.Stream;
 
 import org.apache.jackrabbit.oak.api.Type;
 import org.apache.jackrabbit.oak.commons.collections.StreamUtils;
+import org.apache.jackrabbit.oak.plugins.index.elastic.util.ElasticIndexUtils;
 import org.apache.jackrabbit.oak.plugins.index.search.FieldNames;
 import org.apache.jackrabbit.oak.plugins.index.search.FulltextIndexConstants;
 import org.apache.jackrabbit.oak.plugins.index.search.IndexDefinition;
@@ -324,13 +325,17 @@ public class ElasticIndexDefinition extends IndexDefinition {
             ElasticPropertyDefinition pd = getMatchingRegexPropertyDefinition(propertyName);
             if (pd != null) {
                 if (pd.isFlattened()) {
-                    return FieldNames.FLATTENED_FIELD_PREFIX + pd.nodeName + "." + propertyName;
+                    String fieldName = ElasticIndexUtils.fieldName(propertyName);
+                    String flattenedFieldName = FieldNames.FLATTENED_FIELD_PREFIX +
+                            ElasticIndexUtils.fieldName(pd.nodeName) + "." + fieldName;
+                    return flattenedFieldName;
                 }
             }
-            return propertyName + ".keyword";
+            String fieldName = ElasticIndexUtils.fieldName(propertyName);
+            return fieldName + ".keyword";
         }
 
-        String field = propertyName;
+        String field = ElasticIndexUtils.fieldName(propertyName);
         // it's ok to look at the first property since we are sure they all have the same type
         int type = propertyDefinitions.get(0).getType();
         if (isAnalyzable.apply(type) && isAnalyzed(propertyDefinitions)) {
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
@@ -319,23 +319,18 @@ public class ElasticIndexDefinition extends IndexDefinition {
      */
     public String getElasticKeyword(String propertyName) {
         List<PropertyDefinition> propertyDefinitions = propertiesByName.get(propertyName);
+        String field = ElasticIndexUtils.fieldName(propertyName);
         if (propertyDefinitions == null) {
             // if there are no property definitions we return the default keyword name
             // this can happen for properties that were not explicitly defined (eg: created with a regex)
             ElasticPropertyDefinition pd = getMatchingRegexPropertyDefinition(propertyName);
-            if (pd != null) {
-                if (pd.isFlattened()) {
-                    String fieldName = ElasticIndexUtils.fieldName(propertyName);
-                    String flattenedFieldName = FieldNames.FLATTENED_FIELD_PREFIX +
-                            ElasticIndexUtils.fieldName(pd.nodeName) + "." + fieldName;
-                    return flattenedFieldName;
-                }
+            if (pd != null && pd.isFlattened()) {
+                return FieldNames.FLATTENED_FIELD_PREFIX +
+                        ElasticIndexUtils.fieldName(pd.nodeName) + "." + field;
+            } else {
+                return field + ".keyword";
             }
-            String fieldName = ElasticIndexUtils.fieldName(propertyName);
-            return fieldName + ".keyword";
         }
-
-        String field = ElasticIndexUtils.fieldName(propertyName);
         // it's ok to look at the first property since we are sure they all have the same type
         int type = propertyDefinitions.get(0).getType();
         if (isAnalyzable.apply(type) && isAnalyzed(propertyDefinitions)) {
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
@@ -649,16 +649,20 @@ public class ElasticRequestHandler {
     }
 
     private Stream<NestedQuery> dynamicScoreQueries(String text) {
-        return elasticIndexDefinition.getDynamicBoostProperties().stream().map(pd -> NestedQuery.of(n -> n
-                .path(ElasticIndexUtils.fieldName(pd.nodeName))
-                .query(q -> q.functionScore(s -> s
+        return elasticIndexDefinition.getDynamicBoostProperties().stream()
+            .map(pd -> {
+                String field = ElasticIndexUtils.fieldName(pd.nodeName);
+                return NestedQuery.of(n -> n
+                    .path(field)
+                    .query(q -> q.functionScore(s -> s
                         .boost(DYNAMIC_BOOST_WEIGHT)
                         .query(fq -> fq.match(m -> m.field(
-                                ElasticIndexUtils.fieldName(pd.nodeName) + ".value").
+                                field + ".value").
                                 query(FieldValue.of(text))))
                         .functions(f -> f.fieldValueFactor(fv -> fv.field(
-                                ElasticIndexUtils.fieldName(pd.nodeName) + ".boost")))))
-                .scoreMode(ChildScoreMode.Avg))
+                                field + ".boost")))))
+                    .scoreMode(ChildScoreMode.Avg));
+            }
         );
     }
 
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
@@ -263,7 +263,8 @@ public class ElasticRequestHandler {
                     continue;
                 }
 
-                String similarityPropFieldName = FieldNames.createSimilarityFieldName(pd.name);
+                String similarityPropFieldName = FieldNames.createSimilarityFieldName(
+                        ElasticIndexUtils.fieldName(pd.name));
                 KnnQuery knnQuery = baseKnnQueryBuilder(similarityPropFieldName, bytes, pd).build();
                 return Optional.of(knnQuery);
             }
@@ -649,11 +650,14 @@ public class ElasticRequestHandler {
 
     private Stream<NestedQuery> dynamicScoreQueries(String text) {
         return elasticIndexDefinition.getDynamicBoostProperties().stream().map(pd -> NestedQuery.of(n -> n
-                .path(pd.nodeName)
+                .path(ElasticIndexUtils.fieldName(pd.nodeName))
                 .query(q -> q.functionScore(s -> s
                         .boost(DYNAMIC_BOOST_WEIGHT)
-                        .query(fq -> fq.match(m -> m.field(pd.nodeName + ".value").query(FieldValue.of(text))))
-                        .functions(f -> f.fieldValueFactor(fv -> fv.field(pd.nodeName + ".boost")))))
+                        .query(fq -> fq.match(m -> m.field(
+                                ElasticIndexUtils.fieldName(pd.nodeName) + ".value").
+                                query(FieldValue.of(text))))
+                        .functions(f -> f.fieldValueFactor(fv -> fv.field(
+                                ElasticIndexUtils.fieldName(pd.nodeName) + ".boost")))))
                 .scoreMode(ChildScoreMode.Avg))
         );
     }
@@ -889,8 +893,8 @@ public class ElasticRequestHandler {
                 .type(TextQueryType.CrossFields)
                 .tieBreaker(0.5d);
         if (FieldNames.FULLTEXT.equals(fieldName)) {
-            for(PropertyDefinition pd: pr.indexingRule.getNodeScopeAnalyzedProps()) {
-                qsqBuilder.fields(pd.name + "^" + pd.boost);
+            for (PropertyDefinition pd : pr.indexingRule.getNodeScopeAnalyzedProps()) {
+                qsqBuilder.fields(ElasticIndexUtils.fieldName(pd.name) + "^" + pd.boost);
             }
             // dynamic boost is included only for :fulltext field
             if (includeDynamicBoostedValues) {
@@ -951,6 +955,11 @@ public class ElasticRequestHandler {
         if (planResult.isPathTransformed()) {
             propertyName = PathUtils.getName(propertyName);
         }
-        return propertyName;
+        if ("*".equals(propertyName)) {
+            // elasticsearch does support the pseudo-field "*" meaning all fields,
+            // but (arguably) what we really want is the field ":fulltext".
+            return FieldNames.FULLTEXT;
+        }
+        return ElasticIndexUtils.fieldName(propertyName);
     }
 }
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/facets/ElasticSecureFacetAsyncProvider.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/facets/ElasticSecureFacetAsyncProvider.java
@@ -22,6 +22,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import org.apache.jackrabbit.oak.plugins.index.elastic.query.ElasticRequestHandler;
 import org.apache.jackrabbit.oak.plugins.index.elastic.query.ElasticResponseHandler;
 import org.apache.jackrabbit.oak.plugins.index.elastic.query.async.ElasticResponseListener;
+import org.apache.jackrabbit.oak.plugins.index.elastic.util.ElasticIndexUtils;
 import org.apache.jackrabbit.oak.plugins.index.search.spi.query.FulltextIndex;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -44,6 +45,7 @@ class ElasticSecureFacetAsyncProvider implements ElasticFacetProvider, ElasticRe
     private static final Logger LOG = LoggerFactory.getLogger(ElasticSecureFacetAsyncProvider.class);
 
     private final Set<String> facetFields;
+    private final Set<String> elasticFieldNames;
     private final Map<String, Map<String, Integer>> accessibleFacetCounts = new ConcurrentHashMap<>();
     private final ElasticResponseHandler elasticResponseHandler;
     private final Predicate<String> isAccessible;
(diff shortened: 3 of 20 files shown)


oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
```diff
@@ -324,13 +325,17 @@ public String getElasticKeyword(String propertyName) {
...
                 if (pd.isFlattened()) {
-                    return FieldNames.FLATTENED_FIELD_PREFIX + pd.nodeName + "." + propertyName;
+                    String fieldName = ElasticIndexUtils.fieldName(propertyName);
+                    String flattenedFieldName = FieldNames.FLATTENED_FIELD_PREFIX +
+                            ElasticIndexUtils.fieldName(pd.nodeName) + "." + fieldName;
+                    return flattenedFieldName;
                 }
             }
-            return propertyName + ".keyword";
+            String fieldName = ElasticIndexUtils.fieldName(propertyName);
```
nfsantos: The line `String fieldName = ElasticIndexUtils.fieldName(propertyName);` is executed in both cases of the conditional, so it would be more clear if it was moved to before the if condition. 

As a personal preference, I like this style more, I find it more readable and clear:

```java
            String fieldName = ElasticIndexUtils.fieldName(propertyName);
            if (pd != null && pd.isFlattened()) {
                return FieldNames.FLATTENED_FIELD_PREFIX +
                        ElasticIndexUtils.fieldName(pd.nodeName) + "." + fieldName;
            } else {
              return fieldName + ".keyword";
           }
```

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
```diff
@@ -324,13 +325,17 @@ public String getElasticKeyword(String propertyName) {
...
+            return fieldName + ".keyword";
         }
 
-        String field = propertyName;
+        String field = ElasticIndexUtils.fieldName(propertyName);
```
nfsantos: Again a repeated call to `ElasticIndexUtils.fieldName`

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
```diff
@@ -31,6 +31,31 @@ public class ElasticIndexUtils {
...
+     * @return the field name
+     */
+    public static String fieldName(String propertyName) {
+        String fieldName = propertyName;
+        // 99% property names don't contain a slash or dot,
+        // so for performance reason use indexOf
+        int slashIndex = fieldName.indexOf('|');
+        if (slashIndex >= 0) {
+            fieldName = fieldName.replaceAll("\\|", "\\|\\|");
+        }
```
nfsantos: The code seems to be inconsistent with the javadocs. The docs say that the method replaces "." and "/", but here the code is looking for "|". Which is pipe, not slash, so the variable name is also not consistent.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
```diff
@@ -31,6 +31,31 @@ public class ElasticIndexUtils {
...
+        int dotIndex = fieldName.indexOf('.');
+        if (dotIndex >= 0) {
+            fieldName = fieldName.replaceAll("\\.", "|dot|");
+        }
+        return fieldName;
```
nfsantos: This method is likely to be on the critical path for performance, so it would likely be better to avoid replaceAll, and instead use a StringBuilder and a simple loop over all the chars, after a first check to ensure that the String does not contain any of the characters to be replaced.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
```diff
@@ -78,10 +103,9 @@ public static List<Float> toFloats(byte[] array) {
      * @return byte array
      */
     public static byte[] toByteArray(List<Float> values) {
-        int blockSize = Float.SIZE / Byte.SIZE;
-        byte[] bytes = new byte[values.size() * blockSize];
+        byte[] bytes = new byte[values.size() * Float.BYTES];
         ByteBuffer wrap = ByteBuffer.wrap(bytes);
-        for (int i = 0, j = 0; i < values.size(); i++, j += blockSize) {
+        for (int i = 0; i < values.size(); i++) {
```
nfsantos, line 146: Nice!

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
```diff
@@ -649,11 +650,14 @@ private ObjectBuilder<BoolQuery> inference(BoolQuery.Builder b, String propertyN
...
+                .path(ElasticIndexUtils.fieldName(pd.nodeName))
                 .query(q -> q.functionScore(s -> s
                         .boost(DYNAMIC_BOOST_WEIGHT)
-                        .query(fq -> fq.match(m -> m.field(pd.nodeName + ".value").query(FieldValue.of(text))))
-                        .functions(f -> f.fieldValueFactor(fv -> fv.field(pd.nodeName + ".boost")))))
+                        .query(fq -> fq.match(m -> m.field(
+                                ElasticIndexUtils.fieldName(pd.nodeName) + ".value").
+                                query(FieldValue.of(text))))
+                        .functions(f -> f.fieldValueFactor(fv -> fv.field(
+                                ElasticIndexUtils.fieldName(pd.nodeName) + ".boost")))))
```
nfsantos: Repeated calls to `ElasticIndexUtils.fieldName(pd.nodeName)` in the same method. Call only once at the start of the `map()` handler.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/facets/ElasticStatisticalFacetAsyncProvider.java
```diff
@@ -74,11 +75,15 @@ public class ElasticStatisticalFacetAsyncProvider implements ElasticFacetProvide
 
         this.elasticResponseHandler = elasticResponseHandler;
         this.isAccessible = isAccessible;
-        this.facetFields = elasticRequestHandler.facetFields().collect(Collectors.toSet());
+        Set<String> elasticFieldNames = elasticRequestHandler.facetFields().
+                map(p -> ElasticIndexUtils.fieldName(p)).
+                collect(Collectors.toSet());
+        this.facetFields = elasticRequestHandler.facetFields().
+                collect(Collectors.toSet());
```
nfsantos: Can't this be replaced with `Set.copyOf(elasticRequestHandler.facetFields())`?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
```diff
@@ -31,6 +31,31 @@ public class ElasticIndexUtils {
...
+     * @param propertyName the property name
+     * @return the field name
+     */
+    public static String fieldName(String propertyName) {
+        String fieldName = propertyName;
+        // 99% property names don't contain a slash or dot,
+        // so for performance reason use indexOf
+        int slashIndex = fieldName.indexOf('|');
+        if (slashIndex >= 0) {
+            fieldName = fieldName.replaceAll("\\|", "\\|\\|");
```
chibulcu: shouldn't this be `\\/` (slash) instead of `\\|`?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
```diff
@@ -31,6 +31,31 @@ public class ElasticIndexUtils {
...
+     * "/" is converted to "||"
+     *
+     * @param propertyName the property name
+     * @return the field name
+     */
+    public static String fieldName(String propertyName) {
+        String fieldName = propertyName;
+        // 99% property names don't contain a slash or dot,
+        // so for performance reason use indexOf
+        int slashIndex = fieldName.indexOf('|');
```
chibulcu: same here, should it not be `/`?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/facets/ElasticStatisticalFacetAsyncProvider.java
thomasmueller: No, facetFields is a Stream, but yes it can be simplified.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
thomasmueller: The `\\` is for the regex... but now I changed the logic, because there are more characters to escape. 

We can't use `/` for escaping... we might be able to use `\\`, but I think it's a legal character in JCR, and `|` is illegal, and so safer.

There are actually quite many rules!

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
thomasmueller: Ah well, the field name `slashIndex` was wrong... but now changed.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
thomasmueller: You are right... I originally used slash, but then found out that this is not allowed because we need to support functions, and those can contain `/`...

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
thomasmueller: I changed the logic now... but 99.99% don't need escaping, so I think it's not on the critical path actually.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
```diff
@@ -31,6 +31,110 @@ public class ElasticIndexUtils {
...
+        }
+        String fieldName = propertyName;
+        boolean escape = false;
+        if (fieldName.isBlank()) {
+            // empty field name or field names that only consist of spaces
+            escape = true;
+        } else {
+            // 99.99% property names are OK,
+            // so we loop over the characters first
+            for (int i = 0; i < fieldName.length(); i++) {
```
nfsantos:
```suggestion
            for (int i = 0; i < fieldName.length() && !escape ; i++) {
```

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
```diff
@@ -31,6 +31,110 @@ public class ElasticIndexUtils {
...
+            StringBuilder buff = new StringBuilder(fieldName.length());
+            for (int i = 0; i < fieldName.length(); i++) {
+                char c = fieldName.charAt(i);
+                switch (c) {
+                case '|':
+                    buff.append("||");
+                    break;
+                case '.':
+                case '^':
+                    buff.append('|').append(Integer.toHexString(c)).append('|');
```
nfsantos: The characters here can only be `.` and `^`, so we could avoid the call to `toHexString(c)` and just append a constant string, like `|XXX|` and `|YYY|`. For instance:
```
                case '.':
                    buff.append("|XXX|"); break;
                case '^':
                     buff.append("|YYY|"); break;
```

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
```diff
@@ -31,6 +31,110 @@ public class ElasticIndexUtils {
...
+                default:
+                    buff.append(c);
+                }
+            }
+            fieldName = buff.toString();
+            // internal field start with a _
+            // we also support empty or just spaces
+            if (fieldName.startsWith("_") || fieldName.isBlank()) {
+                fieldName = "|" + fieldName;
+            }
```
nfsantos: Would be nice to avoid creating an additional String here if we have to prefix it with `|`. Can we do this check at the start, when the StringBuilder is empty, and add the pipe as the first character  of the StringBuilder if needed?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
thomasmueller: Yes, it's possible... but this is not on the critical path, and I wanted to allow extending the list if needed... it's really hard to get a good list... the `^` I found myself, and I first also had `#` and `-` and I'm not sure if those never need to be escaped...

nfsantos: I'm wondering if this should not be considered as being on the critical path. For most indexes, this logic will never be executed, sure. But for the very few indexes that contain one or more properties with characters that need to be escaped, won't this logic be called very frequently when processing that specific index? In that case, there could be a noticeable impact in performance when reindexing or processing that index.

thomasmueller: Sure, we can do that.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
```diff
@@ -31,6 +31,118 @@ public class ElasticIndexUtils {
...
+                case '|':
+                case '.':
+                case '^':
+                case '_':
+                    escape = true;
+                }
+            }
+        }
+        if (escape) {
+            StringBuilder buff = new StringBuilder(fieldName.length());
```
nfsantos, line 66: The final string will be always bigger than the original string, so presizing the StringBuilder with the size of the original string will always lead to at least one internal resizing. Maybe it would be better to allocate 2x the size of the original field, to avoid resizing in most of the cases.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
```diff
@@ -31,6 +31,118 @@ public class ElasticIndexUtils {
...
+     * @return the property name
+     */
+    public static String propertyNameFromFieldName(String fieldName) {
+        if (fieldName.indexOf('|') < 0) {
+            return fieldName;
+        }
+        if (fieldName.startsWith("|")) {
+            if (fieldName.equals("|")) {
+                return "";
+            } if (fieldName.startsWith("|_") || fieldName.substring(1).isBlank()) {
```
nfsantos: The second condition is creating a new String and it probably can be replaced with a check of the string length.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
```diff
@@ -31,6 +31,118 @@ public class ElasticIndexUtils {
...
+                fieldName = fieldName.substring(1);
+            }
+        }
+        StringBuilder buff = new StringBuilder(fieldName.length());
+        for (int i = 0; i < fieldName.length(); i++) {
+            char c = fieldName.charAt(i);
+            switch (c) {
+            case '|':
+                String next = fieldName.substring(i + 1);
```
nfsantos: Once again creating a new String instance. Would be better if this could be avoided, as this method may be on the critical path in some cases.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
```diff
@@ -31,6 +31,118 @@ public class ElasticIndexUtils {
...
+                    int end = next.indexOf('|');
+                    if (end < 0) {
+                        buff.append(next);
+                        break;
+                    }
+                    String code = next.substring(0, end);
+                    try {
+                        buff.append((char) Integer.parseInt(code, 16));
+                    } catch (NumberFormatException e) {
+                        buff.append(code);
```
nfsantos: Is it possible to avoid raising this exception? Especially in the case of converting to an hex char, this can be done more efficiently.

nfsantos (approved):
Seems correct, I just have a few suggestions for optimization.

chibulcuteanu (approved):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
thomasmueller: This method is not currently used in production, only for tests. It is optimized for performance. Hm, I'll move it to the unit test code.

isBlank() is checking for whitespace.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/facets/ElasticSecureFacetAsyncProvider.java
```diff
@@ -58,11 +60,14 @@ class ElasticSecureFacetAsyncProvider implements ElasticFacetProvider, ElasticRe
         this.elasticResponseHandler = elasticResponseHandler;
         this.isAccessible = isAccessible;
         this.facetFields = elasticRequestHandler.facetFields().collect(Collectors.toSet());
+        this.elasticFieldNames = elasticRequestHandler.facetFields().
+                map(p -> ElasticIndexUtils.fieldName(p)).
+                collect(Collectors.toSet());
```
fabriziofortino: no need to keep both facetFields and elasticFieldNames. We only need Set with the latter. By doing this, we can also remove the call `String elasticField = ElasticIndexUtils.fieldName(field);` at line 83.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/facets/ElasticStatisticalFacetAsyncProvider.java
```diff
@@ -74,11 +76,15 @@ public class ElasticStatisticalFacetAsyncProvider implements ElasticFacetProvide
 
         this.elasticResponseHandler = elasticResponseHandler;
         this.isAccessible = isAccessible;
-        this.facetFields = elasticRequestHandler.facetFields().collect(Collectors.toSet());
+        this.facetFields = elasticRequestHandler.facetFields().
+                collect(Collectors.toSet());
+        Set<String> elasticFieldNames = facetFields.stream().
```
fabriziofortino: Same comment made in `ElasticSecureFacetAsyncProvider`. We don't need to keep both Sets. Only the one with the fields converted using `ElasticIndexUtils ` is required.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
thomasmueller: Move to test

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
thomasmueller: Moved to test

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/util/ElasticIndexUtils.java
thomasmueller: Moved to test

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/facets/ElasticStatisticalFacetAsyncProvider.java
thomasmueller: I see! I'll try to simplify this.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/facets/ElasticSecureFacetAsyncProvider.java
thomasmueller: Done

fabriziofortino (approved):

PR #2147, opened 2025-03-07
OAK-11557 : removed usage of Guava's Iterables.getOnlyElements with S…

rishabhdaim:
…treams

### DIFF
--- a/oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndexTest.java
+++ b/oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndexTest.java
@@ -101,7 +101,6 @@ import org.apache.lucene.store.FSDirectory;
 import org.apache.lucene.store.FilterDirectory;
 import org.jetbrains.annotations.NotNull;
 import org.junit.After;
-import org.junit.Assert;
 import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
--- a/oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndexTest.java
+++ b/oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndexTest.java
@@ -60,6 +60,7 @@ import org.apache.jackrabbit.oak.api.Tree;
 import org.apache.jackrabbit.oak.api.Type;
 import org.apache.jackrabbit.oak.commons.collections.IterableUtils;
 import org.apache.jackrabbit.oak.commons.collections.ListUtils;
+import org.apache.jackrabbit.oak.commons.collections.StreamUtils;
 import org.apache.jackrabbit.oak.commons.concurrent.ExecutorCloser;
 import org.apache.jackrabbit.oak.plugins.index.AsyncIndexInfoService;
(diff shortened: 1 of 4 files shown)


oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndexTest.java
```diff
@@ -3123,7 +3125,8 @@ private String explain(String query){
     private String explainXpath(String query) throws ParseException {
         String explain = "explain " + query;
         Result result = executeQuery(explain, "xpath", NO_BINDINGS);
-        ResultRow row = Iterables.getOnlyElement(result.getRows());
+        Assert.assertEquals(1, StreamUtils.toStream(result.getRows()).count());
```
mbaedke: It's an explain query, so there should only be one row. If for some reason there are more, this might be expensive, no?

rishabhdaim: Yes, it should have only 1 result.

`Iterables.getOnlyElement` validates this as well internally. So that's I have added that check explicitly.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndexTest.java
```diff
@@ -3123,7 +3125,8 @@ private String explain(String query){
...
+        ResultRow row = StreamUtils.toStream(result.getRows()).findAny().orElseThrow();
```
mbaedke, line 3127: The type of the RuntimeException might change, which should be fine, I guess?

rishabhdaim: Since it is only a test case, it should be fine.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndexPlanner.java
```diff
@@ -591,7 +591,7 @@ private void visitTerm(String propertyName) {
                 log.debug("Following relative property paths are not index: {}", relPaths);
                 return false;
             }
-            result.setParentPath(Iterables.getOnlyElement(relPaths, ""));
+            result.setParentPath(relPaths.stream().findAny().orElse(""));
```
rishabhdaim, line 594: `relPaths` here is a `HashSet`.

oak-lucene/src/test/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndexTest.java
nfsantos: This seems convoluted. After the assertion, we already know that result.getRows() has 1 row. So we could just get that element like this: `result.getRows().iterator().next()`. Or even better, to convert `result.getRows()` to an array or list and use this for the checks on size and to get the single element. Here it's not about performance, but about readability.

rishabhdaim: @nfsantos I have changed this code, cause the underlying Iterable was one-time use only.

thomasmueller (approved):

mbaedke (approved):

PR #2154, opened 2025-03-07
OAK-11564 oak-run FullGC leaves background threads running

daniancu:
command will exit jvm when command finishes and the shutdown hook will close gc resources


rishabhdaim (approved):

Joscorbe (approved):

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -325,6 +339,10 @@ public void execute(String... args) throws Exception {
             } else {
                 System.err.println("unknown revisions command: " + subCmd);
             }
+            if (exitWhenDone) {
```
rishabhdaim, line 342: this change is causing the build to fail locally because it exits the running JDK instance.
cc @Joscorbe @daniancu

PR #2157, opened 2025-03-07
OAK-11570 oak-run fullGC fullGcMaxAge default value is wrongly calculated

daniancu:
fixed conversion where default value is converted to millis not seconds

### DIFF
--- a/oak-run/src/test/java/org/apache/jackrabbit/oak/plugins/document/RevisionsCommandTest.java
+++ b/oak-run/src/test/java/org/apache/jackrabbit/oak/plugins/document/RevisionsCommandTest.java
@@ -191,18 +191,18 @@ public class RevisionsCommandTest {
         ns.dispose();
 
         String output = captureSystemOut(new RevisionsCmd("fullGC", "--entireRepo"));
-        assertTrue(output.contains("DryRun is enabled : true"));
-        assertTrue(output.contains("ResetFullGC is enabled : false"));
-        assertTrue(output.contains("Compaction is enabled : false"));
-        assertTrue(output.contains("starting gc collect"));
-        assertTrue(output.contains("IncludePaths are : [/]"));
-        assertTrue(output.contains("ExcludePaths are : []"));
-        assertTrue(output.contains("FullGcMode is : 0"));
-        assertTrue(output.contains("FullGcDelayFactory is : 2.0"));
-        assertTrue(output.contains("FullGcBatchSize is : 1000"));
-        assertTrue(output.contains("FullGcProgressSize is : 10000"));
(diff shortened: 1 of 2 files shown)


Joscorbe (approved):

oak-run/src/test/java/org/apache/jackrabbit/oak/plugins/document/RevisionsCommandTest.java
```diff
@@ -201,7 +201,7 @@ public void fullGC() {
         assertTrue(output.contains("FullGcDelayFactory is : 2.0"));
         assertTrue(output.contains("FullGcBatchSize is : 1000"));
         assertTrue(output.contains("FullGcProgressSize is : 10000"));
-        assertTrue(output.contains("FullGcMaxAgeInSecs is : 86400"));
+        assertTrue(output.contains("FullGcMaxAgeInSecs is : 86400\n"));
```
Joscorbe, line 204: I assume the `\n` is to ensure the value is exactly **86400** and not something else containing that string, but I wonder if we should then apply the same to the rest of asserts to maintain consistency.

daniancu: yes, the wrong value was `86400000` and was still passing the test
i've updated the same for all numeric values checks

rishabhdaim (approved):

oak-run/src/test/java/org/apache/jackrabbit/oak/plugins/document/RevisionsCommandTest.java
reschke: This change breaks tests on Windows systems.

reschke: Nitpicking: why is there a space before the colons?

rishabhdaim: > This change breaks tests on Windows systems.

Do we need to change this before CF?

> Nitpicking: why is there a space before the colons?

I think this has been done to it consistent with other verifications done.

reschke: 1) It should be trivial; I'm on it.

2) Ack.

reschke: -> https://issues.apache.org/jira/browse/OAK-11630

PR #2172, opened 2025-03-11
OAK-11578 : removed usage of Guava's Iterables.getLast() with oak-com…

rishabhdaim:
…mons


rishabhdaim:
With this change, we are done with Guava's `Iterables`.

thomasmueller (approved):

PR #2173, opened 2025-03-11
OAK-11594 - Converting existing indexes from a type to another leaves the async lane the same

chibulcuteanu:
OAK-11594 - Converting existing indexes from a type to another leaves the async lane the same

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/IndexImporter.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/IndexImporter.java
@@ -36,6 +36,7 @@ import org.apache.jackrabbit.oak.plugins.index.MetricsFormatter;
 import org.apache.jackrabbit.oak.plugins.index.MetricsUtils;
 import org.apache.jackrabbit.oak.plugins.index.importer.AsyncIndexerLock.LockToken;
 import org.apache.jackrabbit.oak.plugins.index.upgrade.IndexDisabler;
+import org.apache.jackrabbit.oak.plugins.memory.PropertyStates;
 import org.apache.jackrabbit.oak.spi.commit.EditorDiff;
 import org.apache.jackrabbit.oak.spi.commit.VisibleEditor;
 import org.apache.jackrabbit.oak.spi.state.NodeBuilder;
@@ -58,6 +59,8 @@ import java.util.concurrent.TimeUnit;
 import static org.apache.jackrabbit.oak.commons.conditions.Validate.checkArgument;
 import static java.util.Objects.requireNonNull;
 import static org.apache.jackrabbit.oak.plugins.index.IndexConstants.REINDEX_COUNT;
+import static org.apache.jackrabbit.oak.plugins.index.IndexConstants.ASYNC_PROPERTY_NAME;
+import static org.apache.jackrabbit.oak.plugins.index.IndexConstants.TYPE_PROPERTY_NAME;
 import static org.apache.jackrabbit.oak.plugins.index.IndexUtils.INDEXING_PHASE_LOGGER;
 import static org.apache.jackrabbit.oak.plugins.index.importer.IndexDefinitionUpdater.INDEX_DEFINITIONS_JSON;
 import static org.apache.jackrabbit.oak.plugins.index.importer.NodeStoreUtils.mergeWithConcurrentCheck;
@@ -67,6 +70,10 @@ public class IndexImporter {
      * Symbolic name use to indicate sync indexes
      */
     static final String ASYNC_LANE_SYNC = "sync";
+    /**
+     * Symbolic name use to indicate elasticsearch index type
+     */
+    static final String TYPE_ELASTICSEARCH = "elasticsearch";
     /*
      * System property name for flag for preserve checkpoint. If this is set to true, then checkpoint cleanup will be skipped.
      * Default is set to false.
@@ -200,6 +207,20 @@ public class IndexImporter {
                 if (!indexInfo.newIndex) {
                     NodeBuilder idxBuilder = NodeStoreUtils.childBuilder(builder, indexInfo.indexPath);
                     indexPathsToUpdate.add(indexInfo.indexPath);
+                    String idxBuilderType = idxBuilder.getString(TYPE_PROPERTY_NAME);
+
+                    // check if provided index definitions is of different type than existing one
+                    // also check if one of them is an elasticsearch type
+                    if (idxBuilderType != null &&
+                            !idxBuilderType.equals(indexInfo.type) &&
+                            (idxBuilderType.equals(TYPE_ELASTICSEARCH) || indexInfo.type.equals(TYPE_ELASTICSEARCH))) {
+
+                        LOG.info("Provided index [{}] has a different type compared to the existing index." +
+                                " Using lane from the index definition provided", indexInfo.indexPath);
+
+                        PropertyState asyncProperty = PropertyStates.createProperty(ASYNC_PROPERTY_NAME, List.of(indexInfo.asyncLaneName), Type.STRINGS);
+                        idxBuilder.setProperty(asyncProperty);
+                    }
                     AsyncLaneSwitcher.switchLane(idxBuilder, AsyncLaneSwitcher.getTempLaneName(indexInfo.asyncLaneName));
                 }
             }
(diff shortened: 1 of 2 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/IndexImporter.java
```diff
@@ -200,6 +208,21 @@ void switchLanes() throws CommitFailedException {
                 if (!indexInfo.newIndex) {
                     NodeBuilder idxBuilder = NodeStoreUtils.childBuilder(builder, indexInfo.indexPath);
                     indexPathsToUpdate.add(indexInfo.indexPath);
+                    String idxBuilderType = idxBuilder.getString(TYPE_PROPERTY_NAME);
+
+                    // check if provided index definitions is of different type than existing one
+                    if (idxBuilderType != null &&
```
alagodasii, line 214: Can be replaced with `StringUtils.equals()`

alagodasii (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/IndexImporter.java
```diff
@@ -200,6 +208,21 @@ void switchLanes() throws CommitFailedException {
...
+                            !idxBuilderType.equals(indexInfo.type)) {
+
+                        LOG.info("Provided index [{}] has a different type compared to the existing index." +
+                                " Using lane from the index definition provided", indexInfo.indexPath);
```
fabriziofortino, line 219: I would improve this message:
```suggestion
                        LOG.info("Provided index [{}] has a different type [{}] compared to the existing index [{}]." +
                                " Using lane from the index definition provided", indexInfo.indexPath, indexInfo.type, idxBuilderType);
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/IndexImporter.java
```diff
@@ -200,6 +208,21 @@ void switchLanes() throws CommitFailedException {
...
+
+                        List<String> asyncLaneList = indexInfo.type.equals(TYPE_ELASTICSEARCH) ?
+                                List.of(indexInfo.asyncLaneName) : List.of(indexInfo.asyncLaneName, INDEXING_MODE_NRT);
```
fabriziofortino: I don't think this is always correct. When the provided index is not of type `elasticsearch`, we are always adding `nrt`. This is wrong when we would convert an elasticsearch to lucene that does not need to be `nrt`.

As an example, try to change the unit test at line 173 with

```
 builder.child("idx-a").setProperty("async", asList("async"), Type.STRINGS);
```

I would expect the async-previous at line 188 to be `async` and not `async, nrt`.

I think that when the index types differ, we should always use `List.of(indexInfo.asyncLaneName)`. If we want to make sure that this logic is executed only when there is a conversion from lucene to elastic or viceversa, we can check that at least one of the types (provided or existing) is elasticsearch and completely skip this logic.

fabriziofortino (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/IndexImporter.java
chibulcu: ack, done.

chibulcuteanu: ack, done

fabriziofortino (approved):

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/importer/IndexImporterTest.java
```diff
@@ -142,6 +144,50 @@ public void switchLanes() throws Exception{
...
+        builder.child("idx-a").setProperty("async", List.of("elastic-async"), Type.STRINGS);
+
+        store.merge(builder, EmptyHook.INSTANCE, CommitInfo.EMPTY);
+
+        IndexImporter importer = new IndexImporter(store, temporaryFolder.getRoot(), provider, NOOP_LOCK);
+        importer.switchLanes();
+
+        NodeState idxa = NodeStateUtils.getNode(store.getRoot(), "/idx-a");
+        assertEquals(AsyncLaneSwitcher.getTempLaneName("async"), idxa.getString(ASYNC_PROPERTY_NAME));
+        assertEquals(idxa.getStrings(ASYNC_PREVIOUS), List.of("async"));
```
tihom88, line 188: Please also add an assertion on "type" also.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/IndexImporter.java
```diff
@@ -200,6 +208,20 @@ void switchLanes() throws CommitFailedException {
                 if (!indexInfo.newIndex) {
                     NodeBuilder idxBuilder = NodeStoreUtils.childBuilder(builder, indexInfo.indexPath);
                     indexPathsToUpdate.add(indexInfo.indexPath);
+                    String idxBuilderType = idxBuilder.getString(TYPE_PROPERTY_NAME);
+
+                    // check if provided index definitions is of different type than existing one
+                    // also check if one of them is an elasticsearch type
+                    if (idxBuilderType != null &&
+                            !idxBuilderType.equals(indexInfo.type) &&
+                            (idxBuilderType.equals(TYPE_ELASTICSEARCH) || indexInfo.type.equals(TYPE_ELASTICSEARCH))) {
```
tihom88, line 216: Do we need this logic for elastic search? i think whatever be the async lane's value, we should update that in new index. Be it lucene or elastic.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/importer/IndexImporterTest.java
```diff
@@ -142,6 +144,50 @@ public void switchLanes() throws Exception{
...
+        builder.child("idx-a").setProperty("type", "lucene");
+        builder.child("idx-a").setProperty("async", asList("async", "nrt"), Type.STRINGS);
+
+        store.merge(builder, EmptyHook.INSTANCE, CommitInfo.EMPTY);
+
+        IndexImporter importer = new IndexImporter(store, temporaryFolder.getRoot(), provider, NOOP_LOCK);
+        importer.switchLanes();
+
+        NodeState idxa = NodeStateUtils.getNode(store.getRoot(), "/idx-a");
+        assertEquals(AsyncLaneSwitcher.getTempLaneName("elastic-async"), idxa.getString(ASYNC_PROPERTY_NAME));
```
tihom88, line 165: Please also add an assertion on "type" also.

tihom88 (changes requested):

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/importer/IndexImporterTest.java
chibulcuteanu: the type is not changed during this switch lane operation, so the type remains the same.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/importer/IndexImporterTest.java
chibulcuteanu: the type is not changed during this switch lane operation, so the type remains the same.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/IndexImporter.java
fabriziofortino: this logic is needed in case an index definition gets reindexed with a different `async` configuration from the one in the existing definition. This mainly happens when an index is converted to another type (eg: from lucene to elasticsearch or viceversa).

> i think whatever be the async lane's value, we should update that in new index.

I agree. We decided to change this behaviour only when there is a change of type and elastic is involved to be more conservative.

chibulcuteanu:
@tihom88 can you please re-review based on latest comments?

tihom88 (approved):

chibulcuteanu:
I've update the branch to remove unused imports. Can someone please approve the workflows?

PR #2181, opened 2025-03-13
 OAK-11571: commons: add Closer class (similar to Guava Closer)

reschke:
Note

  https://github.com/apache/jackrabbit-oak/pull/2181/commits/1b6043774c2d25e4df21298971dd415e6d8ab9be
  
which added tests for Guava Closer. The Oak implementation only was written after tests were there.


oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/pio/Closer.java
```diff
@@ -0,0 +1,130 @@
...
+ * Inspired by and replacing Guava's Closer.
+ */
+public class Closer implements Closeable {
+
+    private Closer() {
+        // no instances for you
+    }
+
+    // stack of closeables to close
+    private final Deque<Closeable> closeables = new ArrayDeque<>();
```
thomasmueller: I think the default of 16 is too high for our case. What about 3?
```suggestion
    private final Deque<Closeable> closeables = new ArrayDeque<>(3);

```

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/pio/Closer.java
```diff
@@ -0,0 +1,130 @@
...
+     * exception instead (wrapped into a {@link RuntimeException}
+     * when necessary).
+     */
+    public void close() throws IOException {
+        // keep track of the IOException to throw
+        IOException toThrow = null;
+
+        // close all in reverse order
+        while (!closeables.isEmpty()) {
+            Closeable closeable = closeables.removeLast();
```
thomasmueller, line 82: I think we are closing in the wrong order... We should have a unit test for that!

```suggestion
            Closeable closeable = closeables.removeFirst();

```

thomasmueller (changes requested):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/pio/Closer.java
rishabhdaim: To avoid this, shouldn't we use `Queue` instead of `Deque`.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/pio/Closer.java
```diff
@@ -0,0 +1,130 @@
...
+    public static Closer create() {
+        return new Closer();
+    }
+
+    /**
+     * Add a {@link Closeable} to the list.
+     * @param closeable {@link Closeable} object to be added
+     * @return the closeable param
+     */
+    public @Nullable <CLO extends Closeable> CLO register(@Nullable CLO closeable) {
```
rishabhdaim: I would make `CLO` as `NonNull` and avoid any risk of NPE.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/pio/Closer.java
reschke: Does this ever occur in code that is in the critical path wrt memory consumption???

https://m.xkcd.com/1691/

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/pio/Closer.java
reschke: Actually, we are closing in the same order as Guava. That's why I actually wrote unit tests against Guava in the first place.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/pio/Closer.java
reschke: That would be inconsistent with Guava's Closer.

reschke:
I believe all comments were based on misunderstandings about what Guava does; see again https://github.com/apache/jackrabbit-oak/pull/2181/commits/d254e8f64de003ec351153c44f68844ff9074985 which has the same tests running with Guava.

@thomasmueller - every time we instantiate a collection with a non-default size, readers of the code will ask themselves: why? and does it really matter here?

reschke:
Things IMHO left to do:

- [x] address some Sonar warnings
- [x] improve Javadoc
- [x] _rename the completely confusingly named method `rethrow()` to `rethrowOnClose()`_ - not done, because I misunterstood the semantics
- [ ] bikeshed about the package name - _any opinions here???_

Note that once this is in (and used throughout); we can get rid of the entire Guava common.io package.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/pio/Closer.java
reschke: ok, but I'll address it anyway.

thomasmueller (approved):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/pio/Closer.java
```diff
@@ -0,0 +1,120 @@
...
+            Closeable closeable = closeables.removeLast();
+            try {
+                closeable.close();
+            } catch (IOException exception) {
+                // remember the first one that occurred
+                if (toThrow == null) {
+                    toThrow = exception;
+                }
+            }
+        }
```
mbaedke, line 91: What about RuntimeExceptions (or any Throwable that is not an IOException)? If they are thrown from closeable.close(), they will not be catched, even if suppressExceptionsOnClose is true (see modified test in comment below).

oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/pio/CloserTest.java
```diff
@@ -0,0 +1,183 @@
...
+            fail("should not throw but got: " + e);
+        }
+
+        try {
+            guavaCloser.close();
+        } catch (Exception e) {
+            fail("should not throw but got: " + e);
+        }
+    }
+}
```
mbaedke, line 213:
```
    public void compareClosers() {
        // when rethrow was called, Exceptions that happened upon close will be swallowed

        com.google.common.io.Closer guavaCloser = com.google.common.io.Closer.create();
        Closer oakCloser = Closer.create();

        try {
            throw oakCloser.rethrow(new InterruptedException());
        } catch (Exception e) {}

        try {
            throw guavaCloser.rethrow(new InterruptedException());
        } catch (Exception e) {}

        try {
            oakCloser.register(() -> { throw new RuntimeException(); });
            oakCloser.close();
        } catch (Exception e) {
            fail("should not throw but got: " + e);
        }

        try {
            guavaCloser.register(() -> { throw new RuntimeException(); });
            guavaCloser.close();
        } catch (Exception e) {
            fail("should not throw but got: " + e);
        }
    }

```
would currently fail.

mbaedke (changes requested):
RuntimeException are not yet treated correctly, see inline comments.

mbaedke (approved):

PR #2182, opened 2025-03-14
OAK-11607 for ordered nodetypes read the child node names lazily

joerghoh:
Replace the existing logic of calculating the list of children for ordered children with an implementation which is lazy as well. This avoids to read all existing children upfront, which can be slow if a node has a lot of child nodes. The new implementation resembles the logic, but all wrapped in an iterator.

Testcases are provided to also show that the Iterator consumes the children list lazily.

I had to adjust the ``ChildOrderPropertyTest`` as it made some assumptions about implementation details, which are no longer true with the new implementation.


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/OrderedChildnameIterable.java
```diff
@@ -0,0 +1,143 @@
...
+ */
+package org.apache.jackrabbit.oak.plugins.tree.impl;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ *  Return the childrenNames in the order defined by the orderedChildren iterator, and merges it
```
reschke:
```suggestion
 *  Return the childrenNames in the order defined by the orderedChildren iterator, and merge it
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/OrderedChildnameIterable.java
```diff
@@ -0,0 +1,143 @@
...
+ *  with the existing children defined by allChildren.
+ *
```
reschke:
```suggestion
 *  <p>
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/OrderedChildnameIterable.java
```diff
@@ -0,0 +1,143 @@
...
+
+    final OrderedChildnameIterator iter;
+
+    /**
+     * Create an iterator, which returns all children in the correct order
+     * @param orderedChildren the children defined by the :childrenOrder property
+     * @param allChildren all children
+     */
+    public OrderedChildnameIterable (Iterable<String> orderedChildren, Iterable<String> allChildren) {
+        iter = new OrderedChildnameIterator(orderedChildren,allChildren);
```
reschke:
```suggestion
        iter = new OrderedChildnameIterator(orderedChildren, allChildren);
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
```diff
@@ -125,17 +120,7 @@ protected Iterable<String> getChildNames() {
...
-            for (String name : order.getValue(NAMES)) {
-                // only include names of child nodes that actually exist
-                if (names.remove(name)) {
-                    ordered.add(name);
-                }
-            }
-            // add names of child nodes that are not explicitly ordered
-            ordered.addAll(names);
-            return ordered;
+            return new OrderedChildnameIterable(order.getValue(NAMES), nodeBuilder.getChildNodeNames());
```
reschke: You can use `IteratorUtils.toIterable()` and avoid the wrapper class.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/OrderedChildnameIterable.java
```diff
@@ -0,0 +1,143 @@
...
+
+        final Iterator<String> orderedChildren;
+        final Iterator<String> allChildren;
+
+        private String nextResult;
+
+        private final List<String> nonOrderedChildren = new ArrayList<>();
+        private Iterator<String> nonOrderedChildrenIterator = null;
+
+        public OrderedChildnameIterator (Iterable<String> orderedChildren, Iterable<String> allChildren) {
```
reschke:
```suggestion
        public OrderedChildnameIterator(Iterable<String> orderedChildren, Iterable<String> allChildren) {
```

reschke (changes requested):

reschke (commented):
As this operates over Strings, maybe this should be renamed and documented as something generic.

reschke:
Note that I haven't looked at the tests (yet).

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/AbstractTree.java
joerghoh: Good hint, that makes things a bit easier.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/OrderedChildnameIterator.java
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * Consume the next element from the orderedChild list and validates that it's actually present
+     * @return the next ordered child or  {code null} if all ordered children have already been returned
+     */
+    @SuppressWarnings("java:S2589")
+    private String getNextOrderedChild() {
+        String current = null;
+        // check that this element is actually present in the allChildren iterable
+        while (current == null && orderedChildren.hasNext()) {
```
anchela: my IDE complains the current is always null and thus the check is redundant.

anchela (approved):
the changes look good to me.
my IDE spotted one problem as highlighted below in the review... the check for current = null is redundant... it think it could be simplified as follows:

```
// check that this element is actually present in the allChildren iterable
        while (orderedChildren.hasNext()) {
            String current = orderedChildren.next();
            if (isOrderedChildPresent(current)) {
                return current;
            } // else : skip this element, as it's not present in the allChildren iterator
        }
        return null;
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/tree/impl/OrderedChildnameIterator.java
joerghoh: that's a good spot, I even put in an explicit ignore, because I mis-understood it.

reschke:
https://sonarcloud.io/project/issues?id=org.apache.jackrabbit%3Ajackrabbit-oak&pullRequest=2182&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

Unused import it seems.

reschke (approved):

PR #2183, opened 2025-03-14
OAK-11545 - Use a single long-lived bulk processor for Elastic reindexing

nfsantos:
When writing to Elastic indexes, we use a `BulkIngester` (part of the Elastic Client Java API) to buffer operations before sending them to Elastic in a bulk request and to manage concurrent outgoing requests. This is essential to have good performance. In the current implementation, each instance of `ElasticIndexWriter` creates a new `ElasticBulkProcessorHandler`. Since index writers are created per-index, when indexing more than one Elastic index we are creating several `BulkIngester`s. This is wasteful because a BulkIngester can be used to write to several indexes, so a single instance can be shared by `ElasticIndexWriter`s. 

This PR moves the creation of the `ElasticBulkProcessor` into the `ElasticIndexerProvider` class so this single instance can be shared by all the index writers created by that indexer provider. This has the following advantages:

- Decreases the memory usage. Each bulk ingester contains buffers to keep the operations before sending them to Elastic. These buffers can take up significant space, especially if they are increased to optimize the size of the bulk requests. 
- Potentially improve performance because the single `BulkIngester` would fill up faster with operations from several indexes, leading to fewer but bigger bulk requests.
- In the incremental indexer, eliminates the cost of creating and destroying several BulkIngesters for every incremental cycle (every 5 seconds).

# Technical details

## Bulk ingester configuration is now global

The following properties used to configure the Bulk Ingester could be set per-index, in the index definition. Now they are global configuration properties:

|  Bulk Ingester property  |  New system property | Default |
| ---------- | ------------- | ------ |
| `maxOperations` | `oak.indexer.elastic.bulkProcessor.maxBulkOperations` | 8192 |
| `maxSize` | `oak.indexer.elastic.bulkProcessor.maxBulkSizeBytes` | 8MB |
| `flushInterval`  |  `oak.indexer.elastic.bulkProcessor.bulkFlushIntervalMs` |  2000 (2s) |
| `maxConcurrentRequests` | `oak.indexer.elastic.bulkProcessor.maxConcurrentRequests` | 1 |
| `failOnError` | `oak.indexer.elastic.bulkProcessor.failOnError` | true |

## Changes in default values
The current defaults for the maximum number of operations per bulk and maximum bulk size are too small for good performance. This PR increases them to the following values:

- `oak.indexer.elastic.bulkProcessor.maxBulkOperations`: 250 -> 8192
- `oak.indexer.elastic.bulkProcessor.maxBulkSizeBytes`: 1MB -> 8MB

## Closing writers without closing bulk ingester
The main complexity of this PR is to manage closing a index writer without closing the underlying bulk ingester. Until now, since each index writer contained a bulk ingester, closing the index writer and bulk ingester had the same lifecycle, so they could be closed together. But to use a single long lived bulk ingester for several writers, closing a index writer becomes more complex.

When closing a writer, we want to flush all the operations for that index. However, the bulk ingester will have in its buffers operations for several indexes and cannot easily distinguish between operations of each index. The solution used is to force a flush of the bulk ingester whenever a index writer is closed, and then wait until all bulk requests lower or equal to the one created by the flush request are processed. This will also flush operations for other indexes that are not being closed, but this will only slow down a little the time to close an index, and has no other problems.


oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
```diff
@@ -192,72 +258,222 @@ public void update(String id, ElasticDocument document) throws IOException {
...
     /**
-     * Closes the bulk ingester and waits for all the bulk requests to return.
+     * Closes an index. The underlying bulk ingestor will be flushed, to ensure that all pending operations for this
+     * index are sent to the server. If this index was registered with waitForESAcknowledgement set to true, then this
+     * method will wait until we receive an acknowledgement from the server for all the operations up to when this
+     * method was called.
+     * <p>
+     * Note: Closing an index will have the side-effect of flushing all pending operations for all indexes registered
+     * with the bulk processor. This should be transparent for the user, but it may mean that this method would take
+     * longer to return than if it was flusing only the operations for the index being closed.
```
fabriziofortino: typo
```suggestion
     * longer to return than if it was flushing only the operations for the index being closed.
```

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
```diff
@@ -38,148 +40,212 @@
...
+    /**
+     * Registers an ElasticIndex with the given index definition configuration.
+     * <p>
+     * The `sync-mode` property can be set to `rt` (real-time). In this case the returned handler will be real-time.
+     * This option is available for sync index definitions only.
+     */
+    public void registerIndex(String indexName, ElasticIndexDefinition indexDefinition, NodeBuilder definitionBuilder, CommitInfo commitInfo, boolean waitForESAcknowledgement) {
+        checkOpen();
+        if (registeredIndexes.containsKey(indexName)) {
+            LOG.warn("Index already registered: {}", indexName);
```
fabriziofortino, line 185: what happens when an existing index definition gets updated? Would the writer try to register it again? In that case we might miss changes in the index definition. In the IndexInfo we also store `NodeBuilder definitionBuilder`, would this be stale?

nfsantos: For each update cycle, writers are registered at the start and unregistered at the end. And if my understanding is correctly, when a writer is created, the index definition is read again. I try to explain the process below, could you please check if it makes sense?

The incremental indexing is managed by the `AsyncIndexUpdate`, and initiated by the method `runWhenPermittted()`
https://github.com/apache/jackrabbit-oak/blob/55bc3e17a6c1fcff3e570935ae028f1977a626a1/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java#L488

This calls `updateIndex()`, which creates a new `IndexUpdate` instance:
https://github.com/apache/jackrabbit-oak/blob/55bc3e17a6c1fcff3e570935ae028f1977a626a1/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java#L810

This is what actually does the index update. Each of these instances, creates new writers at the start, in the `enter()` method:
https://github.com/apache/jackrabbit-oak/blob/55bc3e17a6c1fcff3e570935ae028f1977a626a1/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java#L175

If I understand the code correctly, new writers will read the definition from the NodeStore and should pick up changes:
https://github.com/apache/jackrabbit-oak/blob/55bc3e17a6c1fcff3e570935ae028f1977a626a1/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexUpdate.java#L320
https://github.com/apache/jackrabbit-oak/blob/55bc3e17a6c1fcff3e570935ae028f1977a626a1/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexEditorProvider.java#L69

The writers are closed here in FulltextIndexEditor, by the `leave` method when leaving the root node:
https://github.com/apache/jackrabbit-oak/blob/55bc3e17a6c1fcff3e570935ae028f1977a626a1/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java#L153

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
```diff
@@ -192,72 +258,222 @@ public void update(String id, ElasticDocument document) throws IOException {
...
-        bulkIngester.close();
-        LOG.trace("Bulk Ingester {} closed", bulkIngester);
-
-        // de-register main controller
-        int phase = phaser.arriveAndDeregister();
-
-        if (totalOperations == 0) { // no need to invoke phaser await if we already know there were no operations
-            LOG.debug("No operations executed in this processor. Close immediately");
-            return false;
+    public boolean closeIndex(String indexName) throws IOException {
```
fabriziofortino: I think this method name is a bit misleading. No indexes are actually closed here. Should we call something like `flushIndex`?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
```diff
@@ -192,72 +258,222 @@ public void update(String id, ElasticDocument document) throws IOException {
...
+        LOG.info("Closing index: {}", indexName);
```
fabriziofortino: I guess this is called at every async lane run. Should we lower this to debug/trace?

fabriziofortino (approved):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
```diff
@@ -268,113 +484,75 @@ public void beforeBulk(long executionId, BulkRequest request, List<String> conte
         }
 
         @Override
-        public void afterBulk(long executionId, BulkRequest request, List<String> contexts, BulkResponse response) {
+        public void afterBulk(long executionId, BulkRequest request, List<OperationContext> contexts, BulkResponse response) {
             try {
-                LOG.debug("Bulk with id {} processed in {} ms", executionId, response.took());
+                LOG.debug("Bulk with id {} processed in {} ms", executionId, response.took() / 1_000_000);
```
fabriziofortino: `response.took()` returns milliseconds. Dividing it by 1M would result in `0` most of the times.

nfsantos: Fixed, removed the division by 1000000. But `response.took()` seems to be returning bogus values, in my tests, I was always seeing the same arbitrary value. This value is received from the server, so most likely the problem is not on the client side.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
nfsantos: Yes, it would have been too frequent. I reduced to level to debug.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
nfsantos: I renamed to `flushIndex`. Indeed it is a better description of what the method does.

thomasmueller (approved):

fabriziofortino (approved):

PR #2184, opened 2025-03-17
OAK-11605 : removed usage of Guava's ImmutableSet.copyOf with LinkedSet

rishabhdaim:


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/directory/CopyOnReadDirectory.java
```diff
@@ -332,7 +333,7 @@ private void removeDeletedFiles() throws IOException {
 
         Set<String> filesToBeDeleted =
                 // Files present locally
-                ImmutableSet.copyOf(local.listAll()).stream()
+                SetUtils.toLinkedSet(local.listAll()).stream()
```
rishabhdaim, line 336: we are simply iterating over `Set`, so there is no need for immutability here.

nfsantos: There is also no need to create a null-safe set as we do not do any look-up, so I think we can just remove the call to `ImmutableSet.copyOf()`. The set created by the stream is already immutable, so this call to Guava seems to be redundant.

mbaedke (approved):

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/directory/OakDirectory.java
```diff
@@ -341,7 +342,7 @@ private Set<String> getListing(){
         if (fileNames == null){
             fileNames = directoryBuilder.getChildNodeNames();
         }
-        Set<String> result = ImmutableSet.copyOf(fileNames);
+        Set<String> result = Collections.unmodifiableSet(SetUtils.toLinkedSet(fileNames));
```
nfsantos, line 345: Also no need for immutability of null-safe lookups. This method is called only from the constructor, which just iterates over the contents of the set to add it to another set. I think this could be simplified. Directory listings may be long, so it may be worth to avoid copies here if possible.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/directory/CopyOnReadDirectory.java
rishabhdaim: `ImmutableSet.copyOf()` is called to convert the array returned by `local.listAll()` to a collection, and then run stream operations on it.

It's either convert this to a collection OR use `Arrays.stream(local.listAll())`.
I don't have any strong opinion.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/directory/OakDirectory.java
rishabhdaim: I would rather keep the optimizations separate from this ticket.
In this PR, I would try to keep the behavior as close as possible to the original.

PR #2201, opened 2025-03-27
OAK-11625: Move o.a.j.o.plugins.bundlor.BundlingConfigInitializer to …

mbaedke:
…a suitable API package and export that package.

Done.


reschke (commented):
We'll need deprecation of the old class and a transition period, right?

mbaedke:
> We'll need deprecation of the old class and a transition period, right?

I'm not sure, after all this package was never exported.

mbaedke:
So the BundlingConfigInitializer is not part of any public API. Even internally it's only used in tests.
EDIT: Never mind, I just was confused for a minute. Yes, I guess we have to deprecate the old one.

reschke:
If we remove the old class, we'll break everybody who's been embedding it, as it seems was sort of recommended.

So we really need to deprecate (with forRemoval==true), and then wait for minimally one release cycle.

reschke (approved):

mbaedke:
@mreutegg , Julian gave his ok, but in this particular case I'd appreciate a nod from you.

mreutegg (approved):
Looks good to me.

PR #2202, opened 2025-03-31
OAK-11621: Reduce usage of Guava Ticker

reschke:
This is an interim step: ...next we'll implement a Stopwatch replacement that takes a Clock, not a Ticker.


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/TrackingCorruptIndexHandler.java
```diff
@@ -168,7 +168,13 @@ public class CorruptIndexInfo {
         private final String asyncName;
         private final String path;
         private final long lastIndexerCycleCount = indexerCycleCount;
-        private final Stopwatch watch = Stopwatch.createStarted(new ClockTicker(clock));
+        private final Stopwatch watch = Stopwatch.createStarted(new Ticker() {
+            @Override
+            public long read() {
+                return TimeUnit.MILLISECONDS.toNanos(clock.millis());
```
mbaedke, line 174: I didn't look deeper, but shouldn't we use clock#getTime()? (might be overridden while #millis() isn't)

reschke: `getTime()` does no exist in java.time.Clock, and I'd prefer new code to be generic.

reschke: Aha. Yes, that may be a problem, so we should update "our" clock impl. That actually has a ticket already. Will work on it in parallel.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexStatistics.java
```diff
@@ -26,13 +27,13 @@
 
 import co.elastic.clients.elasticsearch._types.query_dsl.Query;
 
+import org.apache.jackrabbit.guava.common.base.Ticker;
```
mbaedke, line 30: Maybe move back to the other guava imports?

mbaedke (approved):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexStatistics.java
reschke: Eventually, these will all go. The other ones are misplaced because they were automatically renamed.

PR #2204, opened 2025-04-01
OAK-11622: Clock improvements

reschke:


oak-core-spi/src/main/java/org/apache/jackrabbit/oak/stats/Clock.java
```diff
@@ -29,7 +31,12 @@
 /**
  * Mechanism for keeping track of time at millisecond accuracy.
```
mbaedke, line 33: s/accuracy/resolution   ?

mbaedke (approved):

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/stats/Clock.java
reschke: I didn't touch this :-)

mbaedke: Yes, but we could sneak it in :). Never mind.

PR #2205, opened 2025-04-02
OAK-11627: Fix NPE on isInheritedMember

Amoratinos:
Adding a check to prevent NullPointerException when a given Identity Provider name that has no auto membership mappings


anchela (approved):
lgtm thanks a lot

PR #2207, opened 2025-04-02
Oak 9762 improve principal provider alternative

Amoratinos:
Changes on how permissions are calculated on Principal permission provider. See Jira ticket for some benchmark results


Amoratinos:
As agreed this improvement won't be done

PR #2211, opened 2025-04-06
OAK-11642: oak-segment-tar uses unmaintained concurrentlinkedhashmap-lru

reschke:
1. Use commons-collections LRU map, and just wrap in into Immutable
2. Fix the only case where I write operation was attempted.

### DIFF
--- a/oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/CommitsTracker.java
+++ b/oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/CommitsTracker.java
@@ -22,7 +22,6 @@ import java.lang.ref.WeakReference;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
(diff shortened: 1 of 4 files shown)


oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/CommitsTracker.java
```diff
@@ -195,12 +196,11 @@ public Map<String, Long> getCommitsCountPerGroupLastMinute() {
                 }
             }
         }
-        return commitsPerGroup;
+        return Collections.unmodifiableMap(commitsPerGroup);
```
mbaedke, line 199: Unrelated change

mbaedke (commented):
I don't even understand how these changes relate to the issue. Why is no synchronization necessary? Why do these maps have to become immutable?

mbaedke (approved):

PR #2212, opened 2025-04-07
OAK-11640 : removed usage of ImmutableSet.builder

rishabhdaim:


thomasmueller (commented):
It looks OK as far as I see, but I'm not sure about the security part. 

Possibly memory usage could be much higher than with the current ImmutableSet.

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/AbstractLoginModule.java
```diff
@@ -212,9 +215,8 @@ public void initialize(Subject subject, CallbackHandler callbackHandler, Map<Str
     @Override
     public boolean logout() throws LoginException {
         boolean success = false;
-        Set<Object> creds = ImmutableSet.builder()
-                .addAll(subject.getPublicCredentials(Credentials.class))
-                .addAll(subject.getPublicCredentials(AuthInfo.class)).build();
+        Set<Object> creds = Stream.concat(subject.getPublicCredentials(Credentials.class).stream(), subject.getPublicCredentials(AuthInfo.class).stream())
```
reschke: Did you check that it's ok not to make this immutable?

reschke (approved):
...but see comments.

reschke:
> Possibly memory usage could be much higher than with the current ImmutableSet.

...because?

rishabhdaim:
> It looks OK as far as I see, but I'm not sure about the security part.
> 
> Possibly memory usage could be much higher than with the current ImmutableSet.

We have already done the micro benchmarks for `SetUtils.toLinkedSet` vs `ImmutableSet` and the results are here: https://github.com/apache/jackrabbit-oak/pull/2178#issuecomment-2723629655

it seems that using Java's `LinkedHashSet` uses lesser memory because it doesn't uses another data structure (i.e. ArrayList in case of ImmutableSet) to store insertion order.

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/AbstractLoginModule.java
rishabhdaim: Yes, it is only iterated in the same class to destroy these credentials, but I would say that we should play safe here as well. let me update the ticket to use `Collections.unmodifiableSet`

rishabhdaim: addressed in https://github.com/apache/jackrabbit-oak/pull/2212/commits/1c2d8a0e87263e3b7618119e901d34ec040cd478

oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/v8/AzureSegmentStoreV8Test.java
```diff
@@ -70,7 +72,7 @@ public class AzureSegmentStoreV8Test {
 
     private static final EnumSet<SharedAccessBlobPermissions> READ_ONLY = EnumSet.of(READ, LIST);
     private static final EnumSet<SharedAccessBlobPermissions> READ_WRITE = EnumSet.of(READ, LIST, CREATE, WRITE, ADD);
-    private static final ImmutableSet<String> BLOBS = ImmutableSet.of("blob1", "blob2");
+    private static final Set<String> BLOBS = ImmutableSet.of("blob1", "blob2");
```
mbaedke: What about the right side of the assignment?

mbaedke:
> It looks OK as far as I see, but I'm not sure about the security part.
> 
> Possibly memory usage could be much higher than with the current ImmutableSet.

AFAICT all the sets in question are pretty small (also I don't quite understand how the memory usage would increase). Could you elaborate on your security concerns?

oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/v8/AzureSegmentStoreV8Test.java
rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/2212/commits/89ac490e9d6ce011ce9ccecfaf7b8adb3b3f24b1

mbaedke: I still see ImmutableSet.of(...) there.

rishabhdaim: Sorry, I misread you comment in the first place. I have reverted the `left side` to previous value, cause I would be handling the `ImmutableSet.of` changes in next jira: OAK-11646

mbaedke: I see, my bad. Sry for the noise.

mbaedke (approved):

thomasmueller:
> > Possibly memory usage could be much higher than with the current ImmutableSet.
> 
> ...because?

Because the hash table implementation is very different. Guava seems to use an open-addressing hash table (power of 2 sized) with linear probing, where entries are very small. Java internally (inside of LinkedHashSet) uses a HashMap that uses separate chaining.... 

Did someone test memory usage?

thomasmueller (changes requested):
More investigation is needed:
* Verify memory usage
* Verify behavior of null entries
* Verify with someone that knows the security part

oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/basic/DefaultSyncConfig.java
```diff
@@ -203,9 +207,9 @@ public Authorizable setAutoMembershipConfig(@NotNull AutoMembershipConfig autoMe
          */
         @NotNull
         public Set<String> getAutoMembership(@NotNull org.apache.jackrabbit.api.security.user.Authorizable authorizable) {
-            return ImmutableSet.<String>builder().
-                    addAll(autoMembershipConfig.getAutoMembership(authorizable)).
-                    addAll(getAutoMembership()).build();
+            return Collections.unmodifiableSet((Set<String>)
+                    Stream.concat(autoMembershipConfig.getAutoMembership(authorizable).stream(), getAutoMembership().stream())
+                            .collect(Collectors.toCollection(LinkedHashSet::new)));
```
rishabhdaim, line 212: `getAutoMembership` has been annotated with `@NotNull`

oak-core/src/main/java/org/apache/jackrabbit/oak/security/authorization/composite/CompositeAccessControlManager.java
```diff
@@ -71,12 +74,11 @@ public CompositeAccessControlManager(@NotNull Root root,
     @NotNull
     @Override
     public Privilege[] getSupportedPrivileges(String absPath) throws RepositoryException {
-        ImmutableSet.Builder<Privilege> privs = ImmutableSet.builder();
+        Set<Privilege> privs = new LinkedHashSet<>();
         for (AccessControlManager acMgr : acMgrs) {
-            privs.add(acMgr.getSupportedPrivileges(absPath));
+            privs.addAll(Arrays.asList(acMgr.getSupportedPrivileges(absPath)));
```
rishabhdaim, line 79: Null not possible here as well since `getSupportedPrivileges` has been annotated as `@NotNull`

oak-exercise/src/main/java/org/apache/jackrabbit/oak/exercise/security/authorization/models/simplifiedroles/Role.java
```diff
@@ -44,7 +49,9 @@ private Role(long permissions, String... privilegeNames) {
 
     private Role(@NotNull Role base, long permissions, String... privilegeNames) {
         this.permissions = base.permissions|permissions;
-        this.privilegeNames = ImmutableSet.<String>builder().addAll(base.privilegeNames).add(privilegeNames).build();
+        this.privilegeNames = Collections.unmodifiableSet(
```
rishabhdaim, line 52: private constructor called from within the class, `privilegeNames` can't be null here.

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/AbstractLoginModule.java
```diff
@@ -212,9 +213,10 @@ public void initialize(Subject subject, CallbackHandler callbackHandler, Map<Str
     @Override
     public boolean logout() throws LoginException {
         boolean success = false;
-        Set<Object> creds = ImmutableSet.builder()
-                .addAll(subject.getPublicCredentials(Credentials.class))
-                .addAll(subject.getPublicCredentials(AuthInfo.class)).build();
+        Set<Object> builder = new LinkedHashSet<>(subject.getPublicCredentials(Credentials.class));
```
rishabhdaim, line 216: same case as above, `getPublicCredentials` can't return null.

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/privilege/ImmutablePrivilegeDefinition.java
```diff
@@ -37,11 +39,7 @@ public final class ImmutablePrivilegeDefinition implements PrivilegeDefinition {
     public ImmutablePrivilegeDefinition(@NotNull String name, boolean isAbstract, @Nullable Iterable<String> declaredAggregateNames) {
         this.name = name;
         this.isAbstract = isAbstract;
-        ImmutableSet.Builder<String> builder = ImmutableSet.builder();
-        if (declaredAggregateNames != null) {
-            builder.addAll(declaredAggregateNames);
-        }
-        this.declaredAggregateNames = builder.build();
+        this.declaredAggregateNames = declaredAggregateNames != null ? Collections.unmodifiableSet(SetUtils.toLinkedSet(declaredAggregateNames)) : Set.of();
```
rishabhdaim, line 42: Null case already handled.

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/privilege/PrivilegeBitsProvider.java
```diff
@@ -268,15 +269,15 @@ private Iterable<String> extractAggregatedPrivileges(@NotNull Iterable<String> p
 
     @NotNull
     private Set<String> resolveBuiltInAggregation(@NotNull String privilegeName) {
-        ImmutableSet.Builder<String> builder = ImmutableSet.builder();
+        Set<String> builder = new LinkedHashSet<>();
```
rishabhdaim, line 272: Only called when `AGGREGATE_PRIVILEGES` already has `privilegeName`, so null value not possible here as well.

oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/CompositionContext.java
```diff
@@ -67,10 +69,10 @@ class CompositionContext {
         this.prefetchNodeStore = globalStore instanceof PrefetchNodeStore ? (PrefetchNodeStore) globalStore : PrefetchNodeStore.NOOP;
         this.nonDefaultStores = nonDefaultStores;
 
-        ImmutableSet.Builder<MountedNodeStore> b = ImmutableSet.builder();
+        Set<MountedNodeStore> b = new LinkedHashSet<>();
         b.add(this.globalStore);
         b.addAll(this.nonDefaultStores);
```
rishabhdaim, line 74: Both values `globalStore` & `nonDefaultStores` can't be null else the code would have already failed in the higher tier.

oak-upgrade/src/main/java/org/apache/jackrabbit/oak/upgrade/RepositorySidegrade.java
```diff
@@ -477,7 +477,7 @@ private void copyWorkspace(NodeState sourceRoot, NodeBuilder targetRoot) {
 
         NodeStateCopier.builder()
             .include(includes)
-            .exclude(excludes.build())
+            .exclude(Collections.unmodifiableSet(excludes))
```
rishabhdaim, line 480: `excludes` can't be null.

rishabhdaim:
> More investigation is needed:
> 
>     * Verify memory usage
> 
>     * Verify behavior of null entries
> 
>     * Verify with someone that knows the security part

@thomasmueller please find reply as below:

1. Memory usage had already been tested along with performance and both Java's LinkedHashSet & ImmutableSet provides similar performance and memory usage.
2. ImmutableSet also throws NPE in case we put null values for `addAll` method, which is again same behaviour of `LinkedHashSet.addAll`. Also, I have added comments for each method explaining whether null value is possible or not.
3. @anchela @Amoratinos could you please review from security point of view.

oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/basic/DefaultSyncConfig.java
```diff
@@ -18,10 +18,14 @@
...
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.apache.jackrabbit.guava.common.collect.ImmutableSet;
+import org.apache.jackrabbit.oak.commons.collections.SetUtils;
```
Amoratinos, line 28: These two are unused in the class could you please remove them?

oak-core/src/main/java/org/apache/jackrabbit/oak/security/authorization/composite/CompositeAccessControlManager.java
```diff
@@ -18,7 +18,10 @@
 
 import java.security.Principal;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
```
Amoratinos, line 22: Unused import, could you please remove it from here?

oak-exercise/src/main/java/org/apache/jackrabbit/oak/exercise/security/authorization/models/simplifiedroles/Role.java
```diff
@@ -16,8 +16,13 @@
...
 package org.apache.jackrabbit.oak.exercise.security.authorization.models.simplifiedroles;
 
+import java.util.Arrays;
 import java.util.Collections;
+import java.util.LinkedHashSet;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
 import org.apache.jackrabbit.guava.common.collect.ImmutableSet;
```
Amoratinos, line 26: With the changes this import is no longer used

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/AbstractLoginModule.java
```diff
@@ -212,9 +213,10 @@ public void initialize(Subject subject, CallbackHandler callbackHandler, Map<Str
...
+        builder.addAll(subject.getPublicCredentials(AuthInfo.class));
+        Set<Object> creds = Collections.unmodifiableSet(builder);
```
Amoratinos, line 218: Could you review the imports as `import org.apache.jackrabbit.guava.common.collect.ImmutableSet;` is not longer used after the changes

oak-exercise/src/main/java/org/apache/jackrabbit/oak/exercise/security/authorization/models/simplifiedroles/Role.java
```diff
@@ -44,7 +49,9 @@ private Role(long permissions, String... privilegeNames) {
 
     private Role(@NotNull Role base, long permissions, String... privilegeNames) {
         this.permissions = base.permissions|permissions;
-        this.privilegeNames = ImmutableSet.<String>builder().addAll(base.privilegeNames).add(privilegeNames).build();
+        this.privilegeNames = Collections.unmodifiableSet(
+                (Set<String>) Stream.concat(base.privilegeNames.stream(), Arrays.stream(privilegeNames))
+                        .collect(Collectors.toCollection(LinkedHashSet::new)));
```
Amoratinos, line 54: There's a shortcut to create unmodifiable sets from Collectors class:

```suggestion
        this.privilegeNames = Stream.concat(base.privilegeNames.stream(), Arrays.stream(privilegeNames))
                .collect(Collectors.toUnmodifiableSet());
```

Amoratinos (approved):
to me looks good, there are a couple of trivial things like removing imports but nothing blocker

thomasmueller (commented):
It seems my concerns are addressed then! Thanks a lot!

oak-exercise/src/main/java/org/apache/jackrabbit/oak/exercise/security/authorization/models/simplifiedroles/Role.java
rishabhdaim: This would not retain the original order to elements and that would change the code behaviour since ImmutableSet retains the original order.

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/AbstractLoginModule.java
rishabhdaim: I will remove all unused `org.apache.jackrabbit.guava.common.collect.ImmutableSet` in a separate PR.

oak-exercise/src/main/java/org/apache/jackrabbit/oak/exercise/security/authorization/models/simplifiedroles/Role.java
rishabhdaim: Would remove them in next PR along with ImmutableSet imports

oak-core/src/main/java/org/apache/jackrabbit/oak/security/authorization/composite/CompositeAccessControlManager.java
rishabhdaim: Would remove them in next PR along with ImmutableSet imports

oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/basic/DefaultSyncConfig.java
rishabhdaim: Would remove them in next PR along with ImmutableSet imports

PR #2223, opened 2025-04-10
OAK-9586: SysViewImportHandler does not support expanded names in sv:name attributes

mbaedke:
Prefix generation is preliminary.


oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/xml/SysViewImportHandler.java
```diff
@@ -93,6 +95,42 @@ private void processNode(ImportState state, boolean start, boolean end)
...
+            String namespaceUri = svName.substring(svName.indexOf("{") + 1, svName.indexOf("}"));
+            String localName = svName.substring(svName.indexOf("}") + 1);
+            NamespaceRegistry namespaceRegistry = sessionContext.getWorkspace().getNamespaceRegistry();
+            String prefix;
+            try {
+                prefix = namespaceRegistry.getPrefix(namespaceUri);
+            } catch (NamespaceException expected) {
+                // this is an expanded svName using an unregistered namespace
+                // we need to make up a prefix for the namespace
+                prefix = "ns_" + UUID.randomUUID().toString().substring(0, 8);
```
mbaedke: That's temporary. We already have other places where we create ns prefixes, I'm going to clone from there.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/xml/SysViewImportHandler.java
```diff
@@ -93,6 +95,39 @@ private void processNode(ImportState state, boolean start, boolean end)
         }
     }
 
+    //TODO replace with library method
+    private static @NotNull String suggestPrefix(String namespace) {
+        return "ns_" + UUID.randomUUID().toString().substring(0, 8);
+    }
+
+    private NameInfo createNameInfo(String svName) throws RepositoryException {
```
kwin, line 102: I think this whole logic should be common across all JCR methods potentially creating new items (with potentially unregistered expanded names) such as 

- `Node.addNode(...)`
- `Node.setProperty(...)`
- `Session.move(...)`

Probably one needs to add a `SessionContext.getOrCreateOakPath(...)` similar to the `getOakPath(...)` which is called from all those methods. The implementation should delegate to a new method `PathMapper.getOrCreateOakPath(...)` which should automatically register all unknown prefixes.

kwin (changes requested):

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/xml/SysViewImportHandler.java
reschke: That should use jcr-commons NamespaceHelper

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/xml/SysViewImportHandler.java
```diff
@@ -93,6 +99,24 @@ private void processNode(ImportState state, boolean start, boolean end)
...
+        //name extraction algorithm taken from GlobalNameMapper#isExpandedName(String)
+        String namespaceUri = null;
+        if (svName.startsWith("{")) {
+            int brace = svName.indexOf('}', 1);
+            if (brace != -1) {
+                namespaceUri = svName.substring(1, brace);
+                // the empty namespace and "internal" are valid as well, otherwise it always contains a colon (as it is a URI)
+                // compare with RFC 3986, Section 3 (https://datatracker.ietf.org/doc/html/rfc3986#section-3)
+                if (namespaceUri.isEmpty() || namespaceUri.equals(NamespaceConstants.NAMESPACE_REP)|| namespaceUri.indexOf(':') != -1) {
```
reschke:
```suggestion
                if (namespaceUri.isEmpty() || namespaceUri.equals(NamespaceConstants.NAMESPACE_REP) || namespaceUri.indexOf(':') != -1) {
```

reschke (commented):
See one-line comment, otherwise approved.

reschke (approved):

PR #2224, opened 2025-04-11
OAK-11654: Remove usage of Guava Suppliers.ofInstance

reschke:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -117,7 +116,7 @@ public class DocumentNodeStoreBuilder<T extends DocumentNodeStoreBuilder<T>> {
      */
     static final int UPDATE_LIMIT = Integer.getInteger("update.limit", DEFAULT_UPDATE_LIMIT);
 
-    protected Supplier<DocumentStore> documentStoreSupplier = ofInstance(new MemoryDocumentStore());
+    protected Supplier<DocumentStore> documentStoreSupplier = () -> new MemoryDocumentStore();
```
reschke: Strange. This change is causing the test failures. What is going on here???

cc @mbaedke @rishabhdaim @thomasmueller

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
```diff
@@ -3803,11 +3801,11 @@ protected void execute(@NotNull DocumentNodeStore nodeStore) {
         }
 
         private static Supplier<Integer> getDelay(DocumentNodeStore ns) {
-            int delay = 0;
             if (ns.getAsyncDelay() != 0) {
-                delay = (int) SECONDS.toMillis(MODIFIED_IN_SECS_RESOLUTION);
+                return () -> (int) SECONDS.toMillis(MODIFIED_IN_SECS_RESOLUTION);
+            } {
```
rishabhdaim, line 3806: `{` around second `return` call unnecessary.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
rishabhdaim: This could be happening because `ofInstance(new MemoryDocumentStore())` returns the same value but `() -> new MemoryDocumentStore()` returns an new value of MemoryDocumentStore.

please check the below test where it fails for `documentStoreSupplier`.

> @Test
    public void test() {
        Supplier<DocumentStore> supplier = Suppliers.ofInstance(new MemoryDocumentStore());
        Supplier<DocumentStore> documentStoreSupplier = () -> new MemoryDocumentStore();
        Assert.assertTrue(supplier.get() == supplier.get());
        Assert.assertTrue(documentStoreSupplier.get() == documentStoreSupplier.get());
    }

rishabhdaim:
```suggestion
    private final DocumentStore ds = new MemoryDocumentStore();
    protected Supplier<DocumentStore> documentStoreSupplier = () -> ds;
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -534,13 +533,13 @@ public DiffCache getDiffCache(int clusterId) {
...
     public T setBlobStore(BlobStore blobStore) {
-        this.blobStoreSupplier = ofInstance(blobStore);
+        this.blobStoreSupplier = () -> blobStore;
         return thisBuilder();
     }
 
     public BlobStore getBlobStore() {
         if (blobStoreSupplier == null) {
-            blobStoreSupplier = ofInstance(new MemoryBlobStore());
+            blobStoreSupplier = () -> new MemoryBlobStore();
```
rishabhdaim, line 543: same issue could happen here, we need to create an instance of `new MemoryBlobStore` at instance level and pass that value in the `Supplier` lambda expression.

rishabhdaim, line 543:
```suggestion
            private final BlobStore mbs = new MemoryBlobStore(); // create at instance level
            blobStoreSupplier = () -> mbs;
```

rishabhdaim (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
reschke: Apparently because it gets overwritten in other classes. Backing out this change, but I still don't get what's wrong here.

@mreutegg maybe?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
reschke: I prefer return statements at the same identation level.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
reschke: I tried that and that did not solve the problem.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
reschke: yes, that code is similar but it does not appear to have test coverage...

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
reschke: Try:

~~~
diff --git a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
index 884ce2d881..4d760bedcc 100644
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
@@ -35,7 +35,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
 import java.util.function.Supplier;

-import org.apache.jackrabbit.guava.common.base.Suppliers;
 import org.apache.jackrabbit.guava.common.cache.Cache;
 import org.apache.jackrabbit.guava.common.cache.CacheBuilder;
 import org.apache.jackrabbit.guava.common.cache.RemovalCause;
@@ -117,8 +116,10 @@ public class DocumentNodeStoreBuilder<T extends DocumentNodeStoreBuilder<T>> {
      */
     static final int UPDATE_LIMIT = Integer.getInteger("update.limit", DEFAULT_UPDATE_LIMIT);

-    protected Supplier<DocumentStore> documentStoreSupplier = Suppliers.ofInstance(new MemoryDocumentStore());
-    protected Supplier<BlobStore> blobStoreSupplier;
+    private static DocumentStore defaultDocumentStore = new MemoryDocumentStore();
+    protected Supplier<DocumentStore> documentStoreSupplier = () -> defaultDocumentStore;
+    private static BlobStore defaultBlobStore = new MemoryBlobStore();
+    protected Supplier<BlobStore> blobStoreSupplier = () -> defaultBlobStore;
     private DiffCache diffCache;
     private int clusterId  = Integer.getInteger("oak.documentMK.clusterId", 0);
     private int asyncDelay = 1000;
@@ -539,9 +540,6 @@ public class DocumentNodeStoreBuilder<T extends DocumentNodeStoreBuilder<T>> {
     }

     public BlobStore getBlobStore() {
-        if (blobStoreSupplier == null) {
-            blobStoreSupplier = () -> new MemoryBlobStore();
-        }
         BlobStore blobStore = blobStoreSupplier.get();
         configureBlobStore(blobStore);
         return blobStore;
~~~

reschke: It might be caused that the same instance is expected, but it needs to be created later on; thus `memoize()` - trying that right now.

reschke: Maybe:

~~~
    private DocumentStore memoizedDocumentStore = null;
    protected Supplier<DocumentStore> documentStoreSupplier = () -> memoizedDocumentStore = Objects.requireNonNullElse(memoizedDocumentStore, new MemoryDocumentStore());
~~~

This might be the shortest way to do that just with JDK, at least for simple cases.

We have roughly 10 non-case uses of `memoize` - so we'll have to discuss how to address these.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -534,13 +534,13 @@ public DiffCache getDiffCache(int clusterId) {
...
```
rishabhdaim, line 543: It should be replaced with `memoize` as well.

reschke: Why? It seems that although this looks similar, the actual requirements are different.

rishabhdaim: okk, got it, thanks

reschke: ack :-)

PR #1761, opened 2024-10-04
OAK-9447: Upgrade Mongo java driver to 5.2

rgambelli:
https://issues.apache.org/jira/browse/OAK-9447

I've just finished with the replacement of java-mongo-driver with the latest mongodb-driver-sync 5.2.0 published on Sep 24, 2024.

Main changes are in oak-document-store project, mongodb-driver-sync has no CVE at all, that driver is also a dependency of the spring ecosystem, this means have no errors if you use oak inside a springboot application, spring actuator expects classes from mongodb-driver-sync which can't find in mongo-java-driver.

It should be more performant too, this is stated in web, I haven't done performance tests.

PR #2226, opened 2025-04-12
OAK-9447: Upgrade Mongo java driver to 5.2

Joscorbe:
Starting from base PR created by [rgambelli](https://github.com/rgambelli) #1761 

Some conflicts still need to be resolved before merging


Joscorbe:
This PR will likely contain many commits, so plan it to squash it into a single commit when ready.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/ClusterInfoTest.java
```diff
@@ -89,11 +89,11 @@ public void readWriteMode() throws InterruptedException {
         assertEquals(WriteConcern.MAJORITY, mem.getWriteConcern());
 
         op = new UpdateOp(list.get(0).getId(), false);
-        op.set("readWriteMode", "read:nearest, write:fsynced");
+        op.set("readWriteMode", "read:nearest, write:w2");
```
Joscorbe, line 92: FSYNCED doesn't exist anymore, the correct replacement is W2.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/BlobThroughPutTest.java
```diff
@@ -59,13 +60,11 @@ public class BlobThroughPutTest {
...
         Map<WriteConcern,String> bimap = new DualHashBidiMap<WriteConcern,String>();
-        bimap.put(WriteConcern.FSYNC_SAFE,"FSYNC_SAFE");
-        bimap.put(WriteConcern.JOURNAL_SAFE,"JOURNAL_SAFE");
+        bimap.put(WriteConcern.JOURNALED,"JOURNALED");
 //        bimap.put(WriteConcern.MAJORITY,"MAJORITY");
         bimap.put(WriteConcern.UNACKNOWLEDGED,"UNACKNOWLEDGED");
-        bimap.put(WriteConcern.NORMAL,"NORMAL");
 //        bimap.put(WriteConcern.REPLICAS_SAFE,"REPLICAS_SAFE");
-        bimap.put(WriteConcern.SAFE,"SAFE");
+        bimap.put(WriteConcern.ACKNOWLEDGED,"ACKNOWLEDGED");
```
Joscorbe, line 67: Those WriteConcerns are removed:

- FSYNC_SAFE and JOURNAL_SAFE replaced by JOURNALED
- NORMAL and SAFE replaced by ACKNOWLEDGED

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -2361,8 +2364,7 @@ private boolean withClientSession() {
     }
 
     private boolean secondariesWithinAcceptableLag() {
-        return getClient().getReplicaSetStatus() == null
-                || connection.getStatus().getReplicaSetLagEstimate() < acceptableLagMillis;
+        return connection.getStatus().getReplicaSetLagEstimate() < acceptableLagMillis;
```
Joscorbe, line 2367: getStatus() can't be null.

Joscorbe: getReplicaSetLagEstimate() returns 0 when Mongo is not a replica set.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoVersionGCSupport.java
```diff
@@ -18,21 +18,20 @@
...
+import static com.mongodb.client.model.Filters.and;
 import static com.mongodb.client.model.Filters.eq;
 import static com.mongodb.client.model.Filters.exists;
 import static com.mongodb.client.model.Filters.gt;
+import static com.mongodb.client.model.Filters.lt;
 import static com.mongodb.client.model.Filters.or;
 import static com.mongodb.client.model.Projections.include;
 import static com.mongodb.client.model.Sorts.ascending;
+
+import static java.util.Collections.emptyList;
```
Joscorbe, line 30: Not keeping the old imports ordering because they made no sense.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -99,19 +87,30 @@
 import org.slf4j.LoggerFactory;
 
 import com.mongodb.BasicDBObject;
+import com.mongodb.ConnectionString;
```
Joscorbe, line 90: Not keeping the old imports ordering because they made no sense.

Joscorbe:
A few references about the migration:
https://www.mongodb.com/docs/drivers/java/sync/v5.2/reference/upgrade/#breaking-changes
https://www.mongodb.com/docs/drivers/java/sync/v5.2/reference/legacy/#api-changes

nfsantos (approved):
The new Java driver seems to work correctly with indexing.

Joscorbe:
Thanks @nfsantos!

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
```diff
@@ -502,7 +502,7 @@ private void configureEstimators(IndexingProgressReporter progressReporter) {
     private long getEstimatedDocumentCount() {
         MongoConnection mongoConnection = indexHelper.getService(MongoConnection.class);
         if (mongoConnection != null) {
-            return mongoConnection.getDatabase().getCollection("nodes").count();
+            return mongoConnection.getDatabase().getCollection("nodes").countDocuments();
```
rishabhdaim: please use https://mongodb.github.io/mongo-java-driver/5.5/apidocs/driver-sync/com/mongodb/client/MongoCollection.html#estimatedDocumentCount(). `countDocuments` is pretty slow [1].

[1] https://mongodb.github.io/mongo-java-driver/5.5/apidocs/driver-sync/com/mongodb/client/MongoCollection.html#countDocuments()

oak-store-document/pom.xml
```diff
@@ -43,8 +43,8 @@
         <configuration>
           <instructions>
             <Import-Package>
-              com.mongodb*;version="[3.8, 4)";resolution:=optional,
-              org.bson*;version="[3.8, 4)";resolution:=optional,
+              com.mongodb*;version="[5.2, 5.3)";resolution:=optional,
+              org.bson*;version="[5.2, 5.3)";resolution:=optional,
```
rishabhdaim, line 47: we already have version `5.5.0` released for mongo-driver-sync [1]

[1] https://mvnrepository.com/artifact/org.mongodb/mongodb-driver-sync

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
```diff
@@ -340,7 +342,11 @@ private void registerNodeStore() throws IOException {
             String db = config.db();
             boolean soKeepAlive = config.socketKeepAlive();
 
-            MongoClientURI mongoURI = new MongoClientURI(uri);
+            ConnectionString mongoURI = new ConnectionString(uri);
+            MongoClientSettings settings = MongoClientSettings.builder()
+                    .applyConnectionString(mongoURI)
```
rishabhdaim, line 355: variable `mongoURI` can be inlined.
```suggestion
                    .applyConnectionString(new ConnectionString(uri))
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java
```diff
@@ -179,11 +198,8 @@ public static String toString(MongoClientOptions opts) {
      *      otherwise.
      */
     public static boolean hasWriteConcern(@NotNull String uri) {
-        MongoClientOptions.Builder builder = MongoClientOptions.builder();
-        builder.writeConcern(WC_UNKNOWN);
-        WriteConcern wc = new MongoClientURI(requireNonNull(uri), builder)
-                .getOptions().getWriteConcern();
-        return !WC_UNKNOWN.equals(wc);
+        ConnectionString connectionString = new ConnectionString(requireNonNull(uri));
```
rishabhdaim, line 201: why do we need need to change this behavior here, this might cause an regression.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceTest.java
```diff
@@ -155,18 +153,6 @@ public void setUpdateLimit() throws Exception {
         assertEquals(17, store.getUpdateLimit());
     }
 
-    @Test
-    public void keepAlive() throws Exception {
```
rishabhdaim, line 159: `keepAlive` config is only marked as deprecated not removed. I would suggest to keep these test till we remove that.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoConnectionTest.java
```diff
@@ -85,26 +95,6 @@ public void sufficientReadConcern() throws Exception {
         sufficientReadConcernSingleNode(ReadConcern.MAJORITY, true);
     }
 
-    @Test
-    public void socketKeepAlive() throws Exception {
```
rishabhdaim, line 89: Again, I would keep this alive till we remove the config.

rishabhdaim (changes requested):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/MongoConnection.java
Joscorbe: The behavior is the same: both implementations check whether the URI explicitly sets a `writeConcern`. The previous version used a workaround with `WC_UNKNOWN` to detect this, but the new API exposes this check directly via `getWriteConcern()`.

Since now there is a native method in the driver that does exactly what we need, I don't think we need the workaround anymore.

There are tests that checks this and I introduced a few more test cases in [MongoConnectionTest.java](https://github.com/apache/jackrabbit-oak/pull/2226/files/974372ee2b6d26cfb898864d58eaa45b0a352a7b#diff-b17e7e53cb9e04dbff6c53c1bb9b53ae708714314c21b2daf515acf1ee7b5e97) to ensure no regressions are introduced.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreServiceTest.java
Joscorbe: This test checks for `getMongoClientOptions().isSocketKeepAlive()`, which doesn't exist at all in modern driver, so this test would have no sense at all.

I marked `keepAlive` and all related properties as deprecated just to help to detect any uses in any application using Oak as dependency, to avoid breaking changes, but all those options should be completely removed.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoConnectionTest.java
Joscorbe: As before, Mongo doesn't have any related `isSocketKeepAlive` method anymore. The only way to keep this test is to make it a NOOP test, which doesn't make much sense to me.

Joscorbe: In [Mongo Java Driver 3.6](https://mongodb.github.io/mongo-java-driver/3.6/javadoc/?com/mongodb/MongoClientOptions.html) it was already deprecated, and in 4.0 removed.

oak-store-document/pom.xml
Joscorbe: Yes, but we should test it before allowing it. We can do it in a new PR after this is merged, to avoid further delays on this PR.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/DocumentStoreIndexerBase.java
Joscorbe: You are right, applied, thanks!

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreService.java
Joscorbe: Done

Joscorbe: I had to revert this because the `mongoURI` is [used a few lines below](https://github.com/apache/jackrabbit-oak/pull/2226/files/974372ee2b6d26cfb898864d58eaa45b0a352a7b#diff-0e315af2cc5cc1327210891e5ba43454d739fd8bcdcfefb016d985f071a35cb2R359) in a info log.

rishabhdaim (approved):

PR #2229, opened 2025-04-14
OAK-11657: JackrabbitSession.getExpandedName/Path need to return stable expanded names

reschke:
@kwin - this patch changes the precedence for the lookup and also introduces a RepositoryException if the expanded name would be invalid.

Could you please review?


oak-core/src/main/java/org/apache/jackrabbit/oak/namepath/impl/GlobalNameMapper.java
```diff
@@ -143,11 +143,11 @@ public String getExpandedJcrName(@NotNull String oakName) {
...
             String oakPrefix = oakName.substring(0, colon);
+            uri = getNamespacesProperty(oakPrefix);
             // local mapping must take precedence...
-            uri = getSessionLocalMappings().get(oakPrefix);
             if (uri == null) {
-                // ...over global mappings
-                uri = getNamespacesProperty(oakPrefix);
+                // ...over local mappings
+                uri = getSessionLocalMappings().get(oakPrefix);
             }
```
mbaedke, line 153: "local mapping must take precedence over local mapping"?

mbaedke (approved):
LGTM

reschke:
On hold; test failures...

mbaedke:
I'm confused: are the IteratorUtils changes really supposed to be here?

reschke:
please re-check

mbaedke (approved):
Reasonable change, I think it will not introduce more compatibility problems since it shouldn't matter much if the consumer gets an exception instead of a completely unusable name.

oak-core/src/main/java/org/apache/jackrabbit/oak/namepath/impl/GlobalNameMapper.java
```diff
@@ -143,17 +146,23 @@ public String getExpandedJcrName(@NotNull String oakName) {
...
             if (uri == null) {
                 throw new IllegalStateException(
                         "No namespace mapping found for " + oakName);
             }
             localName = oakName.substring(colon + 1);
+            // check namespace name for validity in Oak
+            if (!isValidNamespaceName(uri)) {
+                throw new IllegalStateException(
+                    new NamespaceException("Cannot determine expanded name for '" + oakName +
```
kwin, line 162: how can one register an invalid namespace in oak in the first place?

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/session/JackrabbitSessionTest.java
```diff
@@ -86,20 +88,39 @@ public void testGetExpandedName() throws RepositoryException {
...
         n = testRootNode.addNode("rep:bar");
         assertEquals("{internal}bar", s.getExpandedName(n));
     }
 
+    public void testGetExpandedNameBrokenNamespace() throws RepositoryException {
+        // empty namespace uri
+        assertEquals("{}testroot", s.getExpandedName(testRootNode));
+
+        // register broken namespace name
+        s.getWorkspace().getNamespaceRegistry().registerNamespace("foo", "bar");
```
kwin, line 105: shouldn't this fail (https://github.com/apache/jackrabbit-oak/blob/4b7c14a6dd4d36d546306503c397d0a1ee627f9e/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java#L74-L75)

kwin (approved):
LGTM, however I am not sure why Oak should support registering invalid namespaces (i.e. with invalid URIs in the first place). Shouldn't we rather throw an exception in that case?

kwin (changes requested):
The build fail currently with

```
Error:  testGetExpandedNameBrokenNamespace(org.apache.jackrabbit.oak.jcr.session.JackrabbitSessionTest)  Time elapsed: 0.015 s  <<< ERROR!
javax.jcr.NamespaceException: Failed to register namespace mapping foo -> bar
	at org.apache.jackrabbit.oak.api.CommitFailedException.asRepositoryException(CommitFailedException.java:228)
	at org.apache.jackrabbit.oak.plugins.name.ReadWriteNamespaceRegistry.registerNamespace(ReadWriteNamespaceRegistry.java:113)
	at org.apache.jackrabbit.oak.jcr.session.JackrabbitSessionTest.testGetExpandedNameBrokenNamespace(JackrabbitSessionTest.java:106)
....
Caused by: org.apache.jackrabbit.oak.api.CommitFailedException: OakNamespace0005: Namespace modification not allowed: foo
```

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/session/JackrabbitSessionTest.java
kwin: it actually does fail! Compare with https://github.com/apache/jackrabbit-oak/pull/2229#pullrequestreview-2797655309.

oak-core/src/main/java/org/apache/jackrabbit/oak/namepath/impl/GlobalNameMapper.java
reschke: Well, it is possible. That's bad, but changing that could be a breaking change. Note that this case already is error-logged.

reschke:
I'll have to check. Give me a few days.

oak-core/src/main/java/org/apache/jackrabbit/oak/namepath/impl/GlobalNameMapper.java
kwin: > Note that this case already is error-logged.

Have you found real use cases of that in life code? Otherwise I would tend to at least optionally toggle the behaviour to throw in case of invalid URI during namespace registration.

reschke:
FWIW, I do not see test failures locally (both Win and WSL).

@mbaedke @kwin - any ideas what could be going on here?

reschke:
Nor failed the Sonar build.

Maybe this just needed re-basing?

reschke:
@kwin - I do not see any failures. Please retest.

kwin (approved):
LGTM now

PR #2231, opened 2025-04-15
OAK-11658 - Nodes of type nt:resource in new index definitions must contain a jcr:uuid and it must be unique

nfsantos:
Index definitions may contain nodes of type `nt:resource`, which is referenceable, and thus must contain a `jcr:uuid` property. This property must be unique across the repository (it is a unique identifier). This invariant is not being enforced properly in all cases. In particular, the index importer logic does not ensure that this condition is respected and allows indexes to be imported even if they contain `nt:resource` nodes without `jcr:uuid`.

This PR changes the index importer to check for `nt:resource` nodes and either add a `jcr:uuid` property in case it is missing or replace it by a different value in case it is already present. This second step is to avoid duplicate uuids in the case where the new index is copied from an existing index without modifying existing `jcr:uuid` properties.

### DIFF
--- a/oak-core/src/test/resources/org/apache/jackrabbit/oak/plugins/index/importer/index-def-jcruuid-fix.json
+++ b/oak-core/src/test/resources/org/apache/jackrabbit/oak/plugins/index/importer/index-def-jcruuid-fix.json
@@ -1,0 +1,24 @@
+{
+  "/oak:index/barIndex": {
+    "jcr:primaryType": "nam:oak:QueryIndexDefinition",
+    "config_without_uuid.xml": {
+      "jcr:created": "dat:2024-07-11T11:37:51.727Z",
+      "jcr:primaryType": "nam:nt:file",
+      "jcr:content": {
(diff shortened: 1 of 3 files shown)


oak-core/src/test/resources/org/apache/jackrabbit/oak/plugins/index/importer/index-def-jcruuid-fix.json
```diff
@@ -0,0 +1,24 @@
+{
+  "/oak:index/barIndex": {
+    "jcr:primaryType": "nam:oak:QueryIndexDefinition",
+    "config_without_uuid.xml": {
+      "jcr:created": "dat:2024-07-11T11:37:51.727Z",
+      "jcr:primaryType": "nam:nt:file",
```
chibulcuteanu, line 6: shouldn't this be nt:resource?

chibulcuteanu (approved):
LGTM

PR #2234, opened 2025-04-17
OAK-11636: Add inference support.

tihom88:


oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexHelper.java
```diff
@@ -97,9 +104,47 @@ private static ObjectBuilder<TypeMapping> loadMappings(@NotNull TypeMapping.Buil
...
+        if (InferenceConfig.getInstance().isEnabled()) {
+            mapInferenceConfig(builder, indexDefinition, InferenceConfig.getInstance());
+        }
         return builder;
     }
 
+    private static void mapInferenceConfig(TypeMapping.Builder builder, @NotNull ElasticIndexDefinition indexDefinition, @NotNull InferenceConfig inferenceConfig) {
+        String indexName = PathUtils.getName(indexDefinition.getIndexName());
+        InferenceIndexConfig inferenceIndexConfig;
+//        try {
```
nit0906: remove commented line.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexHelper.java
```diff
@@ -97,9 +104,47 @@ private static ObjectBuilder<TypeMapping> loadMappings(@NotNull TypeMapping.Buil
...
+        // Store the inference configuration in the index metadata so that it can be used by the inference service
```
nit0906: inference service ? Do you mean enrichment service ? Also, without context, of what an inference or an enrichment service is - the comment can be confusing.

oak-query-spi/src/main/java/org/apache/jackrabbit/oak/spi/query/fulltext/InferenceQuery.java
```diff
@@ -0,0 +1,103 @@
...
+    private final String queryInferenceConfig;
+    private final String queryText;
+
+    public InferenceQuery(@NotNull String text) {
+        String[] components = parseText(text);
+        this.queryInferenceConfig = components[0];
+        this.queryText = components[1];
+    }
+
+    private String[] parseText(String inputText) {
```
nit0906, line 43: Can you make this static and write a unit test for this parser ?

nit0906: never mind - I see this is already being tested.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/FullTextSearchImpl.java
```diff
@@ -137,27 +138,29 @@ public FullTextExpression getFullTextConstraint(SelectorImpl s) {
             }
             String p2 = normalizePropertyName(p);
             String rawText = getRawText(v);
-            FullTextExpression e = FullTextParser.parse(p2, rawText);
+            InferenceQuery inferenceQuery = new InferenceQuery(rawText);
+            String queryText = inferenceQuery.getQueryText();
```
nit0906: Could you mention in a comment why exactly is the InferenceQuery object needed here ? I am assuming some parsing required for rawtext->queryText for some specific case ?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
```diff
@@ -569,11 +575,37 @@ private boolean visitTerm(String propertyName, String text, String boost, boolea
...
+
+                    if (InferenceConfig.getInstance().isEnabled()
+                            && !InferenceModelConfig.NOOP.equals(InferenceConfig.getInstance().getInferenceModelConfig(indexName, inferenceModelConfig))) {
+
+                        bqBuilder.must(m -> m.bool(b -> inferenceConfigQuery(b, propertyName, queryText, pr, includeDynamicBoostedValues, inferenceQuery, inferenceQueryConfig)));
+                    }
                     // Experimental support for inference queries
-                    if (elasticIndexDefinition.inferenceDefinition != null && elasticIndexDefinition.inferenceDefinition.queries != null) {
-                        bqBuilder.must(m -> m.bool(b -> inference(b, propertyName, text, pr, includeDynamicBoostedValues)));
+                    else if (elasticIndexDefinition.inferenceDefinition != null && elasticIndexDefinition.inferenceDefinition.queries != null) {
```
nit0906, line 653: Is this for backward compat with the PoC. ?

fabriziofortino:
@tihom88 I would the change the PR title with `Add Inference Support`

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexHelper.java
tihom88: done

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexHelper.java
tihom88: We have enricherConfig in inferenceConfig, hence we do have notion of enrichment service. Updated the comment.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/FullTextSearchImpl.java
tihom88: Sure... I am also adding a configuration in queryengine to enable inference. The same check I will add here.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/ElasticRequestHandler.java
tihom88: yes, it is backward compatible and we will remove older implementation after deprecating it.

oak-query-spi/src/main/java/org/apache/jackrabbit/oak/spi/query/fulltext/InferenceQuery.java
```diff
@@ -0,0 +1,103 @@
...
+                }
+                String candidateJson = text.substring(0, possibleEndIndex);
+                // Verify if this is valid JSON using Oak's JsopTokenizer
+                if (JsonUtils.isValidJson(candidateJson, false)) {
+                    jsonPart = candidateJson;
+                    jsonEndDelimiterIndex = possibleEndIndex;
+                    break;
+                } else {
+                    continue;
+                }
```
fabriziofortino: this block is not needed
```suggestion
                }
```

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/inference/InferenceConfig.java
```diff
@@ -0,0 +1,231 @@
...
+    private boolean isInferenceEnabled;
+
+    public boolean isInferenceEnabled() {
+        return isInferenceEnabled;
+    }
+
+    /**
+     * Loads configuration from the given NodeState
+     *
+     * @return InferenceConfiguration instance
```
fabriziofortino: `@return` is not allowed here
```suggestion
```

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/inference/InferenceConfig.java
```diff
@@ -0,0 +1,231 @@
...
+     */
+
+    private InferenceConfig() {
+        lock.writeLock().lock();
+        try {
+            if (INSTANCE == null) {
```
fabriziofortino: I don't think `INSTANCE` can ever be true. In general, I see a lot of complexity in the initialization/reInit. Can we simplify it?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexHelper.java
```diff
@@ -97,9 +107,48 @@ private static ObjectBuilder<TypeMapping> loadMappings(@NotNull TypeMapping.Buil
...
+                        vb.properties("id", p -> p.keyword(k -> k));
+                        vb.properties("vector", p -> p.denseVector(dv -> dv));
+                        vb.properties("metadata", p -> p.object(o -> o.enabled(false)));
+                        return vb;
+                    }));
+                }
+            }
+            return spaces;
+        }));
+    }
```
fabriziofortino, line 150: I don't think we should do this. We need a way to abstract the mapping and make it configurable.
One possibility is to have an elastic-specific index definition property that contains the actual mapping that will be sent to ES at index creation. The same approach can be used for the :enrich field. In that case, we can should also specify the default value (eg: PENDING) that will be sent out every time a doc gets updated. These changes are actually not related to the inference support but are generic. Can we create a separate PR for that?

fabriziofortino (changes requested):
Let's re-discuss how the mapping should be applied.

In general, I think we need to split this PR in multiple ones to make it easier to understand.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/inference/InferenceConfig.java
tihom88: The reason for complexity is because of  multithreaded use of this object. To make it simple we are keeping one element as volatile to know if we need to reinitialize this config in a multi threaded environemt.

paulalexwilson:
What happened to this work in the end?

PR #2236, opened 2025-04-17
OAK-11587 : added Iterators.asEnumeration replacement in oak-commons

rishabhdaim:


reschke (changes requested):
Why would we do it, when there's `Iterator` already?

rishabhdaim:
> Why would we do it, when there's `Iterator` already?

This method has already been used at multiple places inside OAK, thus I have added it inside OAK-COMMONS to replace Guava one.

In case, we want to remove this altogether, I would suggest opening a separate ticket for that and keeping this one's scope to just replacing/removing Guava usage.

Ticket to remove usage of enumeration: OAK-11662

rishabhdaim:
@reschke The `enumeration` is exposed by https://github.com/apache/jackrabbit-oak/blob/trunk/oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/security/principal/GroupPrincipal.java#L52 which is an public API, so I can't remove the usage of `enumeration` without causing the backward incompatibility.

Hence, we need to keep using `enumeration`.

reschke:
> @reschke The `enumeration` is exposed by https://github.com/apache/jackrabbit-oak/blob/trunk/oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/security/principal/GroupPrincipal.java#L52 which is an public exposed API, so I can't remove the usage of `enumeration` without causing the backward incompatibility.

ack

reschke (approved):

PR #2237, opened 2025-04-17
OAK-10952: improve autogenerated namespace prefixes - test with prototype

reschke:
PoC, hopefully illustrating the idea.

(needs more documentation and support for edge cases, final location in project to be decided)


stefan-egli:
* could the `makeUpPrefixTryEvenHarder` make a somewhat shorter (perhaps sha-hash) suggestion?
* generally I think this is useful indeed, to avoid later namespace remorse. How impossible would it be to look into support for changing namespaces (I know it's a challenge)

reschke:
> could the makeUpPrefixTryEvenHarder make a somewhat shorter (perhaps sha-hash) suggestion?

SHA-1 would be 160 bits, thus 20 bytes, right? That's not always less than the actual namespace name. I could however special case that case.

> generally I think this is useful indeed, to avoid later namespace remorse. How impossible would it be to look into support for changing namespaces (I know it's a challenge)

That's a separate issue. I believe we already found out that it can be done using the namespace registry, although risky. The point of *this* is to reduce the number of cases where it actually comes up.

stefan-egli:
I was thinking of something like the `cca2a4f` part of a `cca2a4f1dbde1e0f7e337615763ac20a64e39160` git commit

reschke:
> I was thinking of something like the `cca2a4f` part of a `cca2a4f1dbde1e0f7e337615763ac20a64e39160` git commit

-> https://github.com/apache/jackrabbit-oak/pull/2237/commits/7024038f2e79c3069164bf864f2d7652e8c860a1

reschke:
re valid prefixes: the JCR spec allows many characters here; but checking exactly might be hard and overkill.

As valid URIs, strictly speaking, are only ASCII, we should filter out all non-ASCII. That would leaves with a short "allow" list of characters.

stefan-egli:
+1, `"s-1578eb4"` looks nice!

oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/NamespaceHelperPrototypeTest.java
```diff
@@ -0,0 +1,190 @@
...
+public class NamespaceHelperPrototypeTest {
+
+    // map with 'optimal' prefix mappings; hard-wired should be ok
+    private static final Map<String, String> KNOWN_PREFIXES =
+            Map.of("http://purl.org/dc/terms/", "dc",
+                    "http://ns.adobe.com/exif/1.0/", "exif");
+
+    // public API: suggest an available prefix for the provided namespace
+    // (while considering pre-existing mappings)
+    private static @NotNull String suggestPrefix(@NotNull String namespace,
```
kwin, line 45: suggest is a bit fuzzy to me, why not `getUnusedPrefix(...)`

oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/NamespaceHelperPrototypeTest.java
```diff
@@ -0,0 +1,190 @@
...
+
+    // suggest an available prefix for the provided namespace, based on the characters
+    // in the namespace name
+    // (while considering pre-existing mappings)
+    private static @NotNull String makeUpPrefix(@NotNull String namespace,
+                                                @NotNull Function<String, String> lookupNamespace) {
+        String prefix = namespace.toLowerCase(Locale.ENGLISH);
+
+        // https://www.w3.org/guide/editor/namespaces.html
+        if (prefix.startsWith("http://www.w3c.org/ns/")) {
```
kwin, line 74: Let us use String constants and their length method to get rid of the magical size.

oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/NamespaceHelperPrototypeTest.java
```diff
@@ -0,0 +1,190 @@
...
+    // public API: suggest an available prefix for the provided namespace
+    public static @NotNull String suggestPrefix(String namespace) {
+        return suggestPrefix(namespace, n -> null);
+    }
+
+    // suggest an available prefix for the provided namespace, based on the characters
+    // in the namespace name
+    // (while considering pre-existing mappings)
+    private static @NotNull String makeUpPrefix(@NotNull String namespace,
+                                                @NotNull Function<String, String> lookupNamespace) {
```
kwin, line 70: `lookupNamespace` is hard to grasp why not `getRegisteredNamespaceForPrefix`

kwin (commented):
Overall looks good, some minor remarks only.

oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/NamespaceHelperPrototypeTest.java
mbaedke: Shouldn't it be rather named getRegisteredPrefix() here?

kwin: I think this returns a namespace URI for a given prefix (not the other way around). But this definitely calls for a non ambiguous name.

kwin:
I would place this in [`org.apache.jackrabbit.oak.namepath.impl.GlobalNameMapper`](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-core/src/main/java/org/apache/jackrabbit/oak/namepath/impl/GlobalNameMapper.java) and then call this from according `getOrCreateOakName(orNull)` which should be used from all JCR methods potentially creating items (with expanded names). This should be imho the only place dealing with name resolving in Oak.

PR #2240, opened 2025-04-18
OAK-11669 : added Iterators.concat replacement in oak-commons

rishabhdaim:


reschke (changes requested):
The title of the ticket and the PR ("concat") does not match the actual change (IMHO).

rishabhdaim:
> The title of the ticket and the PR ("concat") does not match the actual change (IMHO).

In apache commons-collections4, we have `chainedIterator` method for concatenation.

reschke (approved):
(I thought for a moment that it refers to the method added, not the Guava method being removed)

PR #2249, opened 2025-04-25
OAK-11664 : removed usage of Guava's Iterators.transform with oak-com…

rishabhdaim:
…mons


reschke (approved):
...assuming this is purely mechanical.

PR #2250, opened 2025-04-25
OAK-11680 : added Iterators.cycle replacement in oak-commons

rishabhdaim:


reschke (commented):
two different changes at once?

rishabhdaim:
> two different changes at once?

no, it is one change. Changes in `CollectionUtils` are required by `IteratorUtils.cycle` method

PR #2251, opened 2025-04-25
OAK-11681 Clean oak-segment-azure lib exporting

ierandra:
Remove the following:
- com.azure.identity,
- com.azure.blob,
- com.azure.storage.common,
- com.azure.storage.internal.avro.implementation

### DIFF
--- a/oak-segment-azure/pom.xml
+++ b/oak-segment-azure/pom.xml
@@ -72,10 +72,6 @@
                             com.microsoft.azure.storage,
                             com.microsoft.azure.storage.core,
                             com.microsoft.azure.storage.blob,
-                            com.azure.identity,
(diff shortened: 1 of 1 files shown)


jsedding:
I would also remove the packages
- com.microsoft.azure.storage
- com.microsoft.azure.storage.core
- com.microsoft.azure.storage.blob

I.e. it would be best to only export `org.apache.jackrabbit.oak.segment.azure.*`.

ierandra:
@jsedding 
The plan is to remove soon support for azure sdk 8 in oak-segment-azure ([OAK-11682](https://issues.apache.org/jira/browse/OAK-11682)). With this will remove all com.microsoft.azure.storage* references.

PR #2254, opened 2025-04-28
OAK-9584: Support expanded names in JCR method arguments for prefix

kwin:
"rep:"

Treat "{internal}" as namespace although it is no proper URI


anchela (approved):
lgtm

mbaedke (approved):

PR #2255, opened 2025-04-28
OAK-11684 : Send fullGC metrics to Prometheus pushgateway on each garbage collection iteration

shodaaan:
- moved code that exports fullGC metrics to Prometheus pushgateway so that it is called on each fullGC iteration in VersionGarbageCollector

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
@@ -994,9 +994,6 @@ public class VersionGarbageCollector {
                             // now remove the garbage in one go, if any
                             if (gc.hasGarbage() && phases.start(GCPhase.FULL_GC_CLEANUP)) {
                                 gc.removeGarbage(phases.stats);
-                                if (fullGCMetricsExporter != null) {
-                                    fullGCMetricsExporter.onIterationComplete();
-                                }
                                 phases.stop(GCPhase.FULL_GC_CLEANUP);
                             } else {
                                 if (log.isDebugEnabled()) {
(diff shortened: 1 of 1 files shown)


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
```diff
@@ -1007,6 +1004,9 @@ protected boolean removeEldestEntry(Entry<Path, Boolean> eldest) {
                                 fromModifiedMs = lastDoc.getModified() == null ? oldModifiedMs : SECONDS.toMillis(lastDoc.getModified());
                                 fromId = lastDoc.getId();
                             }
+                            if (fullGCMetricsExporter != null) {
+                                fullGCMetricsExporter.onIterationComplete();
+                            }
```
daniancu, line 1009: test coverage?

Joscorbe (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
rishabhdaim: from what I understand is that, you are sending the metrics everytime instead of sending it only in case of having a garbage as done earlier, correct ?

rishabhdaim (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollector.java
shodaaan: @rishabhdaim yes, that is exactly what this change does, it sends metrics after each batch traversal is complete.

PR #2257, opened 2025-04-28
OAK-11686: Provide method to lookup value from OSGi framework property

kwin:
without passing BundleContext

This falls back to using System properties when running outside OSGi


kwin:
The same logic should be used also in the SystemPropertySupplier.

kwin:
Closed for the reasons outlined in https://issues.apache.org/jira/browse/OAK-11686?focusedCommentId=17948145&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17948145

PR #2258, opened 2025-04-28
OAK-11683: Optionally disallow registration of invalid namespace URIs

kwin:
This feature is off by default and can be enabled by setting framework/system property "oak.allowInvalidNamespaceUris" to "false".

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java
@@ -42,14 +42,13 @@ public abstract class ReadWriteNamespaceRegistry
     /**
      * Feature flag to allow registering invalid namespace URIs (without a colon).
      * Set the system property {@code oak.allowInvalidNamespaceUris} to {@code false} to disable this feature.
+     * Cannot be static in order to allow testing with different values.
      */
-    private final boolean allowInvalidNamespaceUris;
+    private final boolean allowInvalidNamespaceUris = SystemPropertySupplier.create("oak.allowInvalidNamespaceUris", true)
+            .loggingTo(LOG).get();;
 
     protected ReadWriteNamespaceRegistry(Root root) {
         super(root);
-        // cannot be static in order to allow testing with different values
-        allowInvalidNamespaceUris = SystemPropertySupplier.create("oak.allowInvalidNamespaceUris", true)
-                .loggingTo(LOG).get();
     }
 
     /**
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java
@@ -41,14 +41,14 @@ public abstract class ReadWriteNamespaceRegistry
 
     /**
      * Feature flag to allow registering invalid namespace URIs (without a colon).
-     * Set the system property {@code oak.allowInvalidNamespaceUris} to {@code true} to enable this feature.
+     * Set the system property {@code oak.allowInvalidNamespaceUris} to {@code false} to disable this feature.
      */
     private final boolean allowInvalidNamespaceUris;
 
     protected ReadWriteNamespaceRegistry(Root root) {
         super(root);
         // cannot be static in order to allow testing with different values
-        allowInvalidNamespaceUris = SystemPropertySupplier.create("oak.allowInvalidNamespaceUris", false)
+        allowInvalidNamespaceUris = SystemPropertySupplier.create("oak.allowInvalidNamespaceUris", true)
                 .loggingTo(LOG).get();
     }
 
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java
@@ -25,6 +25,7 @@ import org.apache.jackrabbit.oak.api.CommitFailedException;
 import org.apache.jackrabbit.oak.api.PropertyState;
 import org.apache.jackrabbit.oak.api.Root;
 import org.apache.jackrabbit.oak.api.Tree;
+import org.apache.jackrabbit.oak.commons.properties.SystemPropertySupplier;
 import org.apache.jackrabbit.oak.spi.namespace.NamespaceConstants;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java
```diff
@@ -71,11 +71,13 @@ public void registerNamespace(String prefix, String uri)
 
         // sanity check for legal namespace names (excluding the "internal"
         // namespace, see OAK-74)
-        if (!NamespaceConstants.NAMESPACE_REP.equals(uri)) {
-            if (!uri.contains(":")) {
+        if (!NamespaceConstants.NAMESPACE_REP.equals(uri) && !uri.contains(":")) {
+            if (Boolean.parseBoolean(System.getProperty("oak.allowInvalidNamespaceUris", "true"))) {
```
mbaedke: I'd suggest using a SystemPropertySupplier.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java
```diff
@@ -71,11 +72,15 @@ public void registerNamespace(String prefix, String uri)
 
         // sanity check for legal namespace names (excluding the "internal"
         // namespace, see OAK-74)
-        if (!NamespaceConstants.NAMESPACE_REP.equals(uri)) {
-            if (!uri.contains(":")) {
+        if (!NamespaceConstants.NAMESPACE_REP.equals(uri) && !uri.contains(":")) {
+            final boolean allowInvalidNamespaceUris = SystemPropertySupplier.create("oak.allowInvalidNamespaceUris", false)
```
mbaedke: Isn't a static member preferrable here?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java
reschke: @kwin - can we move this into a static variable and a short comment about what the system property does?

reschke (changes requested):
@kwin - can we move this into a static variable and a short comment about what the system property does?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java
kwin: Done in https://github.com/apache/jackrabbit-oak/pull/2258/commits/e90fc3bf65d965728c8f7a74183ab8098ffe8acb

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java
kwin: Done in https://github.com/apache/jackrabbit-oak/pull/2258/commits/e90fc3bf65d965728c8f7a74183ab8098ffe8acb.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java
kwin: Making it static is conflicting with testing this feature flag...
Compare with https://github.com/apache/jackrabbit-oak/blob/bb5ca5a6e011f07385ff965d3166ad8d7e636bf1/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/nodetype/write/NodeTypeRegistry.java#L104.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java
```diff
@@ -38,8 +39,17 @@ public abstract class ReadWriteNamespaceRegistry
 
     private static final Logger LOG = LoggerFactory.getLogger(ReadWriteNamespaceRegistry.class);
 
-    public ReadWriteNamespaceRegistry(Root root) {
+    /**
+     * Feature flag to allow registering invalid namespace URIs (without a colon).
+     * Set the system property {@code oak.allowInvalidNamespaceUris} to {@code true} to enable this feature.
+     */
+    private final boolean allowInvalidNamespaceUris;
```
mbaedke, line 46: I'd make it static.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java
```diff
@@ -38,8 +39,17 @@ public abstract class ReadWriteNamespaceRegistry
...
+
+    protected ReadWriteNamespaceRegistry(Root root) {
         super(root);
+        // cannot be static in order to allow testing with different values
```
mbaedke: I don't understand how this is supposed to work. The system property value will not change for the given ClassLoader.

kwin: The unit tests is reusing the same classloader to test with the feature being on and off. Therefore it cannot be static.

mbaedke: Never mind, I was totally confused here. All good.

mbaedke (approved):

kwin:
@reschke Are you also fine with toggling the default (i.e. throw for invalid namespaces except "internal") and just opt-in to get the old behaviour (just log but accept invalid namespaces)?

reschke (commented):
> @reschke Are you also fine with toggling the default (i.e. throw for invalid namespaces except "internal") and just opt-in to get the old behaviour (just log but accept invalid namespaces)?

Hm, no. That would break existing deployments, and it might remain unnoticed for a long time (until it breaks).

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java
```diff
@@ -38,8 +39,17 @@ public abstract class ReadWriteNamespaceRegistry
...
+        allowInvalidNamespaceUris = SystemPropertySupplier.create("oak.allowInvalidNamespaceUris", false)
```
reschke: can we move that up to the declaration of the variable?

reschke (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/name/ReadWriteNamespaceRegistry.java
kwin: Done in https://github.com/apache/jackrabbit-oak/pull/2258/commits/b83396b5a8ba371745a7a55cdf30e0cd279e1fda.

kwin:
> Hm, no. That would break existing deployments, and it might remain unnoticed for a long time (until it breaks).

Ok, toggled default behaviour to what it was before again in https://github.com/apache/jackrabbit-oak/pull/2258/commits/1d02f81331a4f9b0b37af02e9eeb758c259d7dfd

reschke (approved):

PR #2261, opened 2025-04-29
OAK-11688 : removed usage of Guava's Iterators.getOnlyElements with o…

rishabhdaim:
…ak-commons

### DIFF
--- a/oak-blob-plugins/src/test/java/org/apache/jackrabbit/oak/plugins/blob/UploadStagingCacheTest.java
+++ b/oak-blob-plugins/src/test/java/org/apache/jackrabbit/oak/plugins/blob/UploadStagingCacheTest.java
@@ -51,6 +51,7 @@ import org.apache.commons.io.FileUtils;
 import org.apache.jackrabbit.core.data.DataStoreException;
 import org.apache.jackrabbit.oak.commons.FileIOUtils;
 import org.apache.jackrabbit.oak.commons.collections.IteratorUtils;
+import org.apache.jackrabbit.oak.commons.collections.ListUtils;
 import org.apache.jackrabbit.oak.commons.concurrent.ExecutorCloser;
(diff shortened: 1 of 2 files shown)


oak-blob-plugins/src/test/java/org/apache/jackrabbit/oak/plugins/blob/UploadStagingCacheTest.java
```diff
@@ -358,7 +359,9 @@ public void testGetAllIdentifiers() throws Exception {
 
         // Check getAllIdentifiers
         Iterator<String> idsIter = stagingCache.getAllIdentifiers();
-        assertEquals(ID_PREFIX + 0, Iterators.getOnlyElement(idsIter));
+        final List<String> idsList = ListUtils.toList(idsIter);
```
rishabhdaim, line 362: converted the Iterator to list because it can only be iterated once.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/DocumentSplitTest.java
```diff
@@ -104,7 +105,8 @@ public void splitRevisions() throws Exception {
         // check if document is still there
         assertNotNull(ns.getNode(Path.ROOT, RevisionVector.fromString(head)));
 
-        NodeDocument prevDoc = Iterators.getOnlyElement(doc.getAllPreviousDocs());
+        NodeDocument prevDoc = IteratorUtils.get(doc.getAllPreviousDocs(), 0);
+        assertEquals(1, IteratorUtils.size(doc.getAllPreviousDocs()));
```
rishabhdaim, line 109: Added size check since `Iterators.getOnlyElement` does that as well.

rishabhdaim:
Since it is only used inside the test cases, so I didn't added the new API for this.

PR #2293, opened 2025-05-13
OAK-11720 : added OSGI config to avoid taking exclusive merge lock in…

rishabhdaim:
… case of conflicts

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
@@ -34,6 +34,7 @@ import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreBuilde
 import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreBuilder.DEFAULT_NODE_CACHE_PERCENTAGE;
 import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreBuilder.DEFAULT_PREV_DOC_CACHE_PERCENTAGE;
 import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreBuilder.DEFAULT_UPDATE_LIMIT;
+import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreService.DEFAULT_AVOID_MERGE_LOCK;
 import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreService.DEFAULT_FULL_GC_ENABLED;
 import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreService.DEFAULT_EMBEDDED_VERIFICATION_ENABLED;
 import static org.apache.jackrabbit.oak.plugins.document.DocumentNodeStoreService.DEFAULT_FULL_GC_GENERATION;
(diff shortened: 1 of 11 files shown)


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
```diff
@@ -417,4 +418,11 @@
         name = "Enable Full GC Persistent Audit Logging",
         description = "This parameter will enable/disable the saving of deleted document IDs and properties during FullGC into a persistent storage, e.g Mongo collection")
     boolean fullGCAuditLoggingEnabled() default false;
+
+    @AttributeDefinition(
+            name = "Avoid Exclusive Merge lock",
+            description = "Boolean value indicating whether we need to avoid the exclusive merge lock while " +
+                    "merging the changes in case of a conflict. The Default value is " + DEFAULT_AVOID_MERGE_LOCK +
+                    " Note that this value can be overridden via framework property 'oak.documentstore.avoidMergeLock'")
+    boolean avoidMergeLock() default DEFAULT_AVOID_MERGE_LOCK;
```
stefan-egli: doesn't have to be everywhere else, but just wondered if the config param should contain the word `exclusive` - just to be more in line with the code, so eg

```suggestion
    boolean avoidExclusiveMergeLock() default DEFAULT_AVOID_MERGE_LOCK;
```

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBranch.java
```diff
@@ -122,9 +125,9 @@ public NodeState merge(@NotNull CommitHook hook, @NotNull CommitInfo info)
                 throw e;
             }
         }
-        // retry with exclusive lock, blocking other
+        // retry with exclusive lock (if avoidMergeLock is not enabled), blocking other
         // concurrent writes
-        return merge0(hook, info, true);
+        return merge0(hook, info, !avoidMergeLock);
```
stefan-egli: in case of `!avoidMergeLock` do you even want to do another retry? `merge0` already does a number of retries (depending on value of `maximumBackoff`) - I'd argue that if that fails and `avoidMergeLock` is true it should just rethrow the `e` above

Joscorbe: Yeah, I also think if it failed the number of retries, we could just throw the exception, no need for an additional retry.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
Joscorbe: I think this change makes sense to be more informative. Also in the property if possible:
`oak.documentstore.avoidExclusiveMergeLock`.

reschke (commented):
(outside my experience area)

stefan-egli (approved):
+1

Joscorbe (approved):

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/Configuration.java
Joscorbe: Done

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBranch.java
Joscorbe: Done

rishabhdaim:
thanks @stefan-egli for the review and @Joscorbe for addressing the review comments

PR #2294, opened 2025-05-13
OAK-11721: experimental and inferenceConfig implementation should be …

tihom88:
…backward compatible + resolve default config implementation

PR #2295, opened 2025-05-13
OAK-11719: vector query should not break normal queries

tihom88:


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
```diff
@@ -1477,6 +1478,14 @@ public boolean visit(FullTextTerm term) {
 
             private boolean visitTerm(String propertyName, String text, String boost, boolean not) {
                 String p = getLuceneFieldName(propertyName, pr);
+                // below condition is for contains(*, 'xyz')
+                if (propertyName == null) {
+                    // Lucene indexes don't support inference, so we should remove queryInferenceConfig
```
nit0906, line 1483: Should we rather write - don't support vector queries - it's easier to understand.

Also an example of queryInferenceConfig that is getting removed in comment would be helpful.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/FullTextSearchImpl.java
```diff
@@ -144,10 +144,9 @@ public FullTextExpression getFullTextConstraint(SelectorImpl s) {
             // e.g. ?{"inferenceModelConfig": "ada-test-model"}?little red fox
             // So here we split the query into text part of query and inferenceConfig part of query.
             // Afterwards we only parse text part of query as this part of query is what we want to search.
-            if (query.getSettings().isInferenceEnabled()) {
-                VectorQuery vectorQuery = new VectorQuery(rawText);
-                queryText = vectorQuery.getQueryText();
-            }
+            // We are explicitly removing vectorQueryConfig from raw query
```
nit0906, line 147: Comment messages are not consistent. In the LucenePropertyIndex, similar code has comment 

`Lucene indexes don't support inference, so we should remove queryInferenceConfig`

tihom88: FulltextSearchImpl is executed prior to specific implementation of lucene or elastic. There we are removing the queryVectorConfig completely. Here are only providing query part to be parsed by FullTextParser but later send full query as  raw text.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
tihom88: Updated

tihom88:
Moved to https://github.com/apache/jackrabbit-oak/pull/2294

PR #2300, opened 2025-05-21
OAK-11733: Adding information to access denied

Amoratinos:
When getting an access denied exception it's hard to identify the path and what was the options evaluated.

### DIFF
--- a/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/security/AccessManager.java
+++ b/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/security/AccessManager.java
@@ -24,8 +24,12 @@ import org.apache.jackrabbit.oak.api.Tree;
 import org.apache.jackrabbit.oak.jcr.delegate.SessionDelegate;
 import org.apache.jackrabbit.oak.jcr.session.operation.SessionOperation;
 import org.apache.jackrabbit.oak.spi.security.authorization.permission.PermissionProvider;
+import org.apache.jackrabbit.oak.spi.security.authorization.permission.Permissions;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+
+import java.util.Locale;
+import java.util.Set;
 
 /**
  * AccessManager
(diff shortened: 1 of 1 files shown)


anchela (approved):
lgtm

cschneider:
lgtm

reschke:
@anchela - did you consider whether this could disclose information that the caller shouldn't get?

Amoratinos:
To my understanding it shouldn't be any risk as it's only revealing the path. Initially more information was exposed like the actions or the permissions requested but that was removed. 

In my opinion, the method we should pay the most attention to is the one that receives the tree, the property state, and the actions, since this is the one used in multiple places. This checkPermissions is used on `NodeImpl.addNode removeMixin addMixin` methods and in `ImporterImpl.createTree`.   The only scenario where I could foresee a risk is when the path of a node is used like an ID, e.g. node path represents a token or sth like, and I think that's an edge case.

anchela:
@reschke , i looked at all instances where the methods are being called today and didn't find how they would reveal the existence of a path it it would happen with itemExists or in the permissionvalidator.
if you happen to find a case, i am happy to be proven wrong.

having said that:
please check as well.... will wait for your explicit approval before merging it

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/security/AccessManager.java
```diff
@@ -62,13 +64,13 @@ public Boolean perform() {
 
     public void checkPermissions(@NotNull String oakPath, @NotNull String actions) throws RepositoryException {
         if (!hasPermissions(oakPath, actions)) {
-            throw new AccessDeniedException("Access denied.");
+            throw new AccessDeniedException(String.format(Locale.ROOT, "Access denied at path '%s'", oakPath));
```
kwin, line 67: Why don't we include actions/permissions (and potentially also used principal names)? I don't think this exposes anything security relevant.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/security/AccessManager.java
```diff
@@ -62,13 +64,13 @@ public Boolean perform() {
...
         }
     }
 
     public void checkPermissions(@NotNull Tree tree, @Nullable PropertyState property, long permissions) throws RepositoryException {
         if (!hasPermissions(tree, property, permissions)) {
-            throw new AccessDeniedException("Access denied.");
+            throw new AccessDeniedException(String.format(Locale.ROOT, "Access denied at path '%s'", tree.getPath()));
```
kwin, line 73:
```suggestion
            throw new AccessDeniedException(String.format(Locale.ROOT, "Access denied to path '%s'", tree.getPath()));
```

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/security/AccessManager.java
anchela: @kwin , imho this would leak security relevant information and potentially violate PII

kwin: For the principals I tend to agree. For the actions/permissions requested I tend to disagree. Why do you consider this sensitive?

Amoratinos: maybe we could add a trace logging with the privileges and permissions so if the information its useful then it could be gathered but we do not expose too much information to the callers. wdyt?

PR #2302, opened 2025-05-22
OAK-11735 Index merge: merge aggregation definitions

thomasmueller:


oak-run/src/main/java/org/apache/jackrabbit/oak/index/merge/IndexDefMergerUtils.java
```diff
@@ -287,6 +292,81 @@ private static JsonObject mergeChild(String path, String child, int level, JsonO
...
+    private static long getNextIncludeId(JsonObject json) {
+        long max = 0;
+        for(String n : json.getChildren().keySet()) {
+            if (n.startsWith("include")) {
+                n = n.substring("include".length());
+                try {
+                    long id = Long.parseLong(n);
+                    max = Math.max(max, id);
+                } catch (NumberFormatException e) {
+                    // ignore
```
tihom88: Log this info.

oak-run/src/main/java/org/apache/jackrabbit/oak/index/merge/IndexDefMergerUtils.java
```diff
@@ -287,6 +292,81 @@ private static JsonObject mergeChild(String path, String child, int level, JsonO
...
+            nextIncludeId = Math.max(nextIncludeId, getNextIncludeId(custom.getChildren().get(child)));
+            nextIncludeId = Math.max(nextIncludeId, getNextIncludeId(product.getChildren().get(child)));
+
+            // loop over conflicts, and find + remove these
+            // the aggregateConflicts will contain entries that look like this:
+            // "Could not merge value; path=/oak:index/assets-11/aggregates/asset/include11
+            // property=path; ancestor=null; custom=...; product=..."
+            // and we need to extract the path
+            for (String n : aggregateConflicts) {
+                String regex = "path=([^\\s]+)\\sproperty=";
```
tihom88, line 326: I think instead of conflict string, we should conflict json or a serialised object and get rid of regex conditions.

tihom88 (commented):
We should simplify it by using conflict json or an a java object. We can pick this as an enhancement if we don't update it in this PR.

oak-run/src/main/java/org/apache/jackrabbit/oak/index/merge/IndexDefMergerUtils.java
thomasmueller: Yes it's a bit a hack. But this string is coming from the same class; it is not a dependency to another library. So if we change the format, the test would break; and a version bump of another library will not cause issues.

tihom88 (approved):

PR #2306, opened 2025-05-22
OAK-11657: JackrabbitSession.getExpandedName/Path need to return stab…

reschke:
…le expanded names, getExpandedName needs to fail for broken namespace names

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/namepath/impl/GlobalNameMapper.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/namepath/impl/GlobalNameMapper.java
@@ -35,6 +35,7 @@ import static org.apache.jackrabbit.oak.spi.namespace.NamespaceConstants.REP_URI
 import java.util.Map;
 import java.util.Map.Entry;
 
+import javax.jcr.NamespaceException;
 import javax.jcr.RepositoryException;
 
 import org.apache.jackrabbit.oak.api.Root;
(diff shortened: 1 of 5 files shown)


oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/session/JackrabbitSessionTest.java
```diff
@@ -86,20 +89,46 @@ public void testGetExpandedName() throws RepositoryException {
...
+            // there is no expanded name, thus we expect an exception here
+            String result = s.getExpandedName(n);
+            fail("there is no expanded name in this case, so we expect the call to fail, however we get: " + result);
+        } catch (NamespaceException ex) {
+            // expected
+        } finally {
+            s.getWorkspace().getNamespaceRegistry().unregisterNamespace(randomNamespacePrefix);
+        }
+    }
+
```
mbaedke, line 125: Maybe #assertThrows() is more readable.

oak-core/src/main/java/org/apache/jackrabbit/oak/namepath/impl/GlobalNameMapper.java
```diff
@@ -143,17 +145,23 @@ public String getExpandedJcrName(@NotNull String oakName) {
...
             if (uri == null) {
-                // ...over global mappings
-                uri = getNamespacesProperty(oakPrefix);
+                // ...over local mappings
+                uri = getSessionLocalMappings().get(oakPrefix);
             }
             if (uri == null) {
                 throw new IllegalStateException(
                         "No namespace mapping found for " + oakName);
             }
```
mbaedke, line 157: Maybe also use a NamespaceException as root cause?

mbaedke (approved):
LGTM, some comments inline.

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/session/JackrabbitSessionTest.java
reschke: Junit4 does not have `assertThrows`, right? In any case, in case of no exception I wanted the return value to be dumped.

oak-core/src/main/java/org/apache/jackrabbit/oak/namepath/impl/GlobalNameMapper.java
reschke: yes

PR #2307, opened 2025-05-23
OAK-11737: Properly set embeddedVerification value on FullGC revisions command

Joscorbe:
So there were 2 issues:

1. `EmbeddedVerification` is enabled by default, so it was impossible to disable it because it only set the value to the builder if it is true, in case of false it was never executed, so it remained enabled.
2. `collectDocument` method was not setting `EmbeddedVerification` value at all.


rishabhdaim (approved):

rishabhdaim:
Could you please add a test case for this?

Joscorbe:
I have added some improvements to the FullGC command output, and now it outputs actually applied settings to the VersionGarbageCollector object, instead of just the options passed to the command.

In fact, this helped me to detect another issue: included/excluded paths were not set, so any value sent to the RevisionsCommand was ignored. This is also fixed now in the same PR.

rishabhdaim (approved):

PR #2339, opened 2025-06-18
OAK-11766 Write Throttling Mechanism - Session.save() delay

thomasmueller:
A minimal configuration for this is:
Staring the process using `-Doak.sessionSaveDelayer=true`
RepositoryManagement - SessionSaveDelayerConfig: `{"entries":[{"delayMillis":1,"threadNameRegex":".*"}]}`

### DIFF
--- a/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionSaveDelayer.java
+++ b/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionSaveDelayer.java
@@ -27,7 +27,6 @@ import org.apache.jackrabbit.oak.spi.toggle.Feature;
 import org.apache.jackrabbit.oak.spi.whiteboard.Whiteboard;
 import org.apache.jackrabbit.oak.spi.whiteboard.WhiteboardUtils;
 import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -55,7 +54,7 @@ public class SessionSaveDelayer implements Closeable {
     private final Whiteboard whiteboard;
     private final AtomicBoolean closed = new AtomicBoolean();
 
-    private RepositoryManagementMBean mbean;
+    private RepositoryManagementMBean cachedMbean;
     private String lastConfigJson;
     private SessionSaveDelayerConfig lastConfig;
     private volatile boolean logNextDelay;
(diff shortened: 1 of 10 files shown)


oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionSaveDelayer.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    private RepositoryManagementMBean getRepositoryMBean() {
+        if (cachedMbean == null) {
+            cachedMbean = WhiteboardUtils.getService(whiteboard, RepositoryManagementMBean.class);
+        }
+        return cachedMbean;
+    }
+
+    public long delayIfNeeded() {
```
mkapalka: Can this method be called by multiple threads?

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionSaveDelayer.java
```diff
@@ -0,0 +1,146 @@
...
+        this.whiteboard = whiteboard;
+    }
+
+    /**
+     * Gets the name of the current thread.
+     *
+     * @return the current thread name
+     */
+    @NotNull
+    public static String getCurrentThreadName() {
```
mkapalka: Detail: not sure if this method fits here, or is useful if you don't use it, e.g., in line 120 below

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionSaveDelayer.java
```diff
@@ -0,0 +1,146 @@
...
+            long millis = delayNanos / 1_000_000;
+            int nanos = (int) (delayNanos % 1_000_000);
+            if (logNextDelay) {
+                LOG.info("Sleep {} ms {} ns", millis, nanos);
+                logNextDelay = false;
+            }
+            try {
+                Thread.sleep(millis, nanos);
+            } catch (InterruptedException e) {
+                // ignore
```
mkapalka, line 122: Detail: we don't ignore it (anymore), but delay the exception to a next blocking call

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionSaveDelayer.java
thomasmueller: Yes. By default it is disabled. We need to be extremely careful that there is no error, or performance issue, if _disabled_.

If enabled, I think we don't need synchronization. It is fine if the configuration is read multiple times in some edge cases. But even in this case, we should try to prevent errors (NPE etc, setting the interrupt flag). But performance issues are not the highest priority in this case.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionSaveDelayer.java
thomasmueller: Ah I thought I removed the method (having one line is not useful).

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionSaveDelayer.java
thomasmueller: I forgot to set the interrupt flag sorry.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionSaveDelayer.java
mkapalka: Good point. I think that the disabled path is fine. The enabled path is a bit tricky though: since we access non-volatile fields from multiple threads without any synchronization, we have race conditions (according to the Java memory model). This might work fine but is hard to analyze. For example, if one thread sets `lastConfig` and another thread reads the new value, I'm not sure if we don't risk NPE accessing individual fields of the referenced object (since there was no safe publish of the object). IMHO, protecting the "enabled" path with a single lock, even a simple `synchronized { ... }` block, has less chance of introducing errors than unsafe access.

thomasmueller: It is sufficient if the fields of SessionSaveDelayerConfig are final; see https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.5.1
that means the JVM memory model guarantees that other threads see the fully constructed object if the constructor returns, via memory barrier.

We actually had an issue where we missed this, and this caused issues for ARM CPUs: https://issues.apache.org/jira/browse/OAK-9634

PR #2340, opened 2025-06-18
OAK-11768: Repository on read-only NodeStore

mreutegg:
Call merge only when there are non-default initializers.
Add test and update one test to now force an initializer merge.


mreutegg:
Looks like builds from forks do not work (?)

Closing this PR and opened https://github.com/apache/jackrabbit-oak/pull/2341 instead.

PR #2343, opened 2025-06-19
OAK-11765 - BulkProcessor unable to insert after a failure

nfsantos:
# Recovering from Elasticsearch connection errors 

## Handling of errors on Elasticsearch bulk processor
The Bulk processor sends requests to elastic asynchronously using the `BulkIngester` class of the Elasticsearch Java driver. When the driver receives a response from a bulk request, the `BulkIngester` invokes a callback in our `ElasticBulkProcessorHandler`, indicating the result of the bulk. In case of errors, this sets internal state in `ElasticBulkProcessorHandler` with the indication of the error. The next time the client of the bulk processor executes an operation on the processor (adding a document or flushing/closing an index), this error is propagated.

Previously, the error status was not being cleared after being propagated to the callers. So once an error occurred, any further call to the bulk processor handler would fail with the same error. This is incorrect, because the bulk processor in the Elastic Java driver is capable of recovering from failed connections and other errors, and errors on the operations of a particular index should not affect the other non-faulty indexes being updated in the same bulk. This PR fixes this issue, by clearing the error status after it is propagated to the caller.

A bulk can fail with two types of errors:
- **A connection or server error** - The server cannot be reached or cannot process the bulk. This means that no operation in the bulk was processed. In this case, we log the error and propagate the exception. It is up to the caller to decide to retry or to fail.
- **Individual documents in the bulk failed to process** - This is a partial failure, as other operations in the bulk might have been successful. Additionally, these errors affect only the operations on a particular index (e.g., trying to add documents to an index that does not exist). We log this error and optionally (controlled by `oak.indexer.elastic.bulkProcessor.failOnError`) propagate the error to the callers that are performing operations on the index of the operations that failed. If the operations on other indexes in the same bulk succeeded, the callers for these indexes do not observe any error.

The incremental indexing ([AsyncIndexUpdate](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java) already retries failed operations. If a cycle fails, the checkpoint is not advanced and the next cycle will retry the operations from the original checkpoint. Therefore, if there are transient failures, once the underlying connection problem is resolved, the incremental indexer will resume. So this PR does not make any changes to this logic, it is enough to fix the issue that prevented the bulk processor from clearing up the internal faulty state.

## Retry policy
Added a retry policy to [ElasticIndexWriter](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticIndexWriter.java). This policy will retry the operations either for a maximum number of times or a maximum time, using exponential backoff with a cap on the maximum interval between retries.

The policy can be configured in two ways, depending on whether the instance of `ElasticIndexWriter` is created for off-line reindexing or by the `ElasticIndexProviderService` for incremental indexing.

### Incremental indexing
The following property is exposed in the [ElasticIndexProviderService](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexProviderService.java):

- `maxRetryTime` - for how long to retry after a failure. Default: 0 seconds.

Disabled by default because the incremental indexer already retries failed indexing cycles. In the future, we can enable retry at the level of the ElasticIndexWriter as an optimization.

### Off-line reindexing

Currently, the Elastic off-line reindexers will abort reindexing in the case of a connection error. This PR adds the following system property that is read by [ElasticDocumentStoreIndexer](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-run-elastic/src/main/java/org/apache/jackrabbit/oak/index/ElasticDocumentStoreIndexer.java) and [ElasticOutOfBandIndexer](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-run-elastic/src/main/java/org/apache/jackrabbit/oak/index/ElasticOutOfBandIndexer.java):

- `oak.indexer.elastic.connectionRetrySeconds` - for how long to retry after a failure. Default: 30 seconds

Additional changes:
- Do not save the paths of failed documents in index definition in the node store. This feature is not being used and adds significant complexity.
- Upgrade ES Java driver from 8.18.1 to 8.18.2.
- Throttle logs of failed operations on bulks.

### DIFF
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
@@ -297,7 +297,6 @@ public class ElasticBulkProcessorHandler {
 
         // Some of the operations for this index pending may be buffered for sending in the bulk ingester.
         // Force sending them now.
-        LOG.trace("Flushing bulk ingester {}", bulkIngester);
         bulkIngester.flush();
 
         if (indexInfo.waitForESAcknowledgement) {
@@ -348,6 +347,7 @@ public class ElasticBulkProcessorHandler {
             }
         }
 
+        checkConnectionFailures();
         checkFailuresForIndex(indexInfo);
         LOG.trace("Bulk identifier -> update status = {}", registeredIndexes);
         return indexInfo.indexModified;
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
@@ -374,7 +374,7 @@ public class ElasticBulkProcessorHandler {
 
     private void checkConnectionFailures() throws IOException {
         if (lastConnectionError != null) {
-            IOException ioe = new IOException("Exception while indexing.", lastConnectionError);
+            IOException ioe = new IOException("Service error while indexing.", lastConnectionError);
             lastConnectionError = null; // Clear the last connection error after throwing it
             throw ioe;
         }
@@ -387,7 +387,7 @@ public class ElasticBulkProcessorHandler {
             String overflowMessage = suppressedErrors.size() >= MAX_SUPPRESSED_ERROR_CAUSES ?
                     ". (Too many failed operations in last bulk request, including only " + suppressedErrors.size() + " errors)"
                     : "";
-            IOException ioe = new IOException("Exception while indexing " + indexInfo.indexName + ". See suppressed for details" + overflowMessage);
+            IOException ioe = new IOException("Error indexing documents for index: " + indexInfo.indexName + ". See suppressed errors for details" + overflowMessage);
             suppressedErrors.forEach(ec -> ioe.addSuppressed(new IllegalStateException(ec.reason())));
             throw ioe;
         }
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
@@ -26,11 +26,11 @@ import co.elastic.clients.elasticsearch.core.bulk.BulkResponseItem;
 import co.elastic.clients.json.JsonData;
 import org.apache.jackrabbit.oak.api.PropertyState;
 import org.apache.jackrabbit.oak.api.Type;
+import org.apache.jackrabbit.oak.commons.log.LogSilencer;
 import org.apache.jackrabbit.oak.plugins.index.ConfigHelper;
 import org.apache.jackrabbit.oak.plugins.index.FormattingUtils;
 import org.apache.jackrabbit.oak.plugins.index.elastic.ElasticConnection;
 import org.apache.jackrabbit.oak.plugins.index.elastic.ElasticIndexDefinition;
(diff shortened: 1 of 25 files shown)


fabriziofortino (approved):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
```diff
@@ -532,14 +494,11 @@ public void afterBulk(long executionId, BulkRequest request, List<OperationConte
...
+            // Called in case of a connection failure or other error that prevented the full bulk request from being executed
             try {
-                LOG.error("ElasticIndex Update Bulk Failure : Bulk with id {} threw an error", executionId, failure);
-                globalSuppressedErrorCauses.add(ErrorCause.of(ec -> {
-                    StringWriter sw = new StringWriter();
-                    PrintWriter pw = new PrintWriter(sw);
-                    failure.printStackTrace(pw);
-                    return ec.reason(failure.getMessage()).stackTrace(sw.toString());
-                }));
+                LOG.warn("ElasticIndex Update Bulk Failure : Bulk with id {} threw an error", executionId, failure);
```
chibulcuteanu, line 499: shouldn't we still log as error here?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
```diff
@@ -481,41 +449,35 @@ public void beforeBulk(long executionId, BulkRequest request, List<OperationCont
...
-                                item.id(), item.index(), item.error());
+                        String type = item.error().type() != null ? item.error().type() : "type-unknown";
+                        String reason = item.error().reason() != null ? item.error().reason() : "reason-unknown";
+                        if (reason.length() > 20) {
+                            reason = reason.substring(0, 20) + "...";
+                        }
+                        String logSilenceKey = indexInfo.indexName + ":" + type + ":" + reason;
+                        if (!LOG_SILENCER.silence(logSilenceKey)) {
+                            // Log entry to be used to parse logs to get the failed doc id/path if needed
+                            LOG.warn("Failure Details: BulkItem ID: {}, Index: {}, Failure Cause: {} - {}",
```
chibulcuteanu, line 476: error level here as well?

chibulcuteanu (approved):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
nfsantos: I'm not sure. This message is printed when a document fails to index, so it is an error caused by the user providing a wrong/invalid document or index rule. This can happen quite often and pollute the logs with ERROR level messages which are not that critical. As it is not a failure on our side, I think WARN level is more appropriate. But I don't have a strong opinion here, so I'm willing to change it to ERROR if you think it's better.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/index/ElasticBulkProcessorHandler.java
nfsantos: In this case, it is a connection failure to Elastic. This can happen under "normal" conditions, when Elastic is scaling up/down, there is a failure on Elastic or there is some connection problem. All these are normal conditions that we handle. The exception is saved internally in the ElasticBulkProcessorHandler and then propagated to client code, which may retry or decide to give up. I think since at this point we are not sure there is a fatal error, we should just log as warn and let the higher levels decide what to do: retry or give up and log an ERROR. Once again, I don't have strong feelings about this, so I can change it to ERROR.

PR #2349, opened 2025-06-23
OAK-11764 - Query planner: when multiple indexes have same cost, planner should choose index deterministically

nfsantos:
When choosing an index for a query, when several indexes have the same optimal cost, pick an index in a deterministic way. As Oak makes no guarantee as to what index to choose in this cases, the selection rule can be arbitrary. For simplicity, this PR selects the first in alphabetic order.


tihom88 (approved):

mkapalka (approved):
LGTM

PR #2356, opened 2025-06-25
OAK-11498: Added getPrincipals default implementation

Amoratinos:


oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/JackrabbitSession.java
```diff
@@ -284,9 +287,33 @@ default Node getParentOrNull(@NotNull Item item) throws RepositoryException {
...
+        }
+        
+        if (!authorizable.isGroup() && ((User) authorizable).isSystemUser()) {
+            throw new IllegalStateException("Unable to calculate effective set of principals for system user " + userId);
+        }
+        
+        Principal userPrincipal = authorizable.getPrincipal();
+        Set<Principal> principals = new java.util.HashSet<>();
+        principals.add(userPrincipal);
+        PrincipalIterator iterator =  getPrincipalManager().getGroupMembership(userPrincipal);
```
kwin, line 311: maybe use `.forEachRemaining(principals::add)`

kwin (approved):

oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/JackrabbitSession.java
Amoratinos: I tried but PrincipalIterator is a raw iterator so it cannot use `principals::add` without doing an explicit cast to principal

anchela (approved):
@Amoratinos the patch looks good to me.
one thing that came to my mind: ideally the default implementation is not executed. so, we might want to add some log out put that would allow us to spot if this happens. e.g. there is a bug in oak leads to us falling back to the default.
wdyt?

Amoratinos:
If I'm not mistaken, adding log sentences need to introduce a new dependency that, at the moment, it's not there. I'm a bit against it but I'm open to do if needed.

kwin:
Thanks for the PR, I took the liberty to rebase and merge and will continue to work on the enhanced PR #2093

PR #2358, opened 2025-06-30
OAK-11780 calculate documentPrefixMap only when required

joerghoh:


reschke (changes requested):
Thinking about it, I believe this would break imports where different sub trees need different mappings...

This might be fixable by implementing these as some kind of stack.

(and we should add test coverage unless it's there already)

mbaedke (changes requested):
Maybe I miss something, but I think we do not need documentPrefixMap at all. We can just pass createCurrentPrefixMap() to the LocalNameMapper constructor without any previous assignments.
EDIT: Also the ITs in oak-jcr don't complain, so either I'm right or we need a test.

joerghoh:
@reschke I don't understand your concerns. From what I see, the field ```documentPrefixMap``` is consumed within the TargetImportHandler just once: in the method ```currentNamePathMapper()```. Also the method ```createCurrentPrefixMap()```  does not have any side effects and consistently rebuilds the prefixMap from scratch.

Based on these 2 observations I think it's fine to call ```createCurrentPrefixMap``` only when this map is required, and not at (arbitrary) intermediate steps.

mbaedke (approved):

PR #2361, opened 2025-07-01
OAK-11783: DocViewImport fails for content not using namespace prefixes

reschke:
This adds two similar tests, one using prefixes (passing), one not using prefixes (failing).


kwin (approved):
LGTM, where does the second test fail exactly?

mbaedke (approved):

PR #2370, opened 2025-07-06
OAK-11792: mark commons.cache package as internal

reschke:


oak-store-spi/src/main/java/org/apache/jackrabbit/oak/plugins/memory/AbstractBlob.java
```diff
@@ -189,4 +182,39 @@ public String toString() {
         return getSha256().toString();
     }
 
+    public static class HashCode {
```
rishabhdaim: Could you please revert this change. I don't think this is part of this PR.

reschke: Indeed; thanks.

rishabhdaim (approved):

rishabhdaim (approved):

PR #2371, opened 2025-07-06
OAK-11784 avoid repeated instantiation of the NamespaceRegistry

joerghoh:
avoid using the ``NamespaceHelper.registerNamespace()`` in a tight loop, as it consistently needs to create the NamespaceRegistry object, which requires the read of the respective Oak Tree objects.

Instead adapt this method, remove the not required functionality and inline it into the ImportHandler, so it will create the NamespaceRegistry just once.

### DIFF
--- a/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/xml/ImportHandler.java
+++ b/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/xml/ImportHandler.java
@@ -19,15 +19,17 @@ package org.apache.jackrabbit.oak.jcr.xml;
 import java.util.HashMap;
 import java.util.Map;
 
+import javax.jcr.NamespaceException;
+import javax.jcr.NamespaceRegistry;
 import javax.jcr.RepositoryException;
 
-import org.apache.jackrabbit.commons.NamespaceHelper;
 import org.apache.jackrabbit.oak.api.CommitFailedException;
 import org.apache.jackrabbit.oak.api.Root;
 import org.apache.jackrabbit.oak.jcr.delegate.SessionDelegate;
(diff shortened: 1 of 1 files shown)


reschke (commented):
We need to improve that, but optimallty in NamespaceHelper (which is in jcr-commons and it used outside Oak as well).

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/xml/ImportHandler.java
```diff
@@ -136,8 +140,7 @@ public void endDocument() throws SAXException {
     public void startPrefixMapping(String prefix, String uri)
             throws SAXException {
         try {
-            new NamespaceHelper(sessionContext.getSession()).registerNamespace(
-                    prefix, uri);
+            registerNamespace(prefix, uri);
```
reschke, line 143: Once `NamespaceHelper` avoids fetching the registry every time, ImportHandler should should get the NamespaceRegistry reference only once. -- see https://github.com/apache/jackrabbit/pull/259

reschke (changes requested):

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/xml/ImportHandler.java
mbaedke: As discussed elsewhere: NamespaceHelper has to fetch the registry every time because ReadonlyNamepaceRegistry internally uses Tree objects retrieved from the Root passed to the constructor. So in order to come up with a registry based on the current root, NamespaceHelper should create a new one.

joerghoh:
I close this PR in favor of #2401 which addresses the problem much more elegant.

PR #2372, opened 2025-07-06
OAK-11793: remove usage of Guava common.hash

reschke:
this could probably could be simplified to get rid of the internal HashCode class as well.


oak-store-spi/src/main/java/org/apache/jackrabbit/oak/plugins/memory/AbstractBlob.java
```diff
@@ -189,4 +182,39 @@ public String toString() {
...
+            try {
+                return new HashCode(MessageDigest.getInstance("SHA-256").digest(bytes));
+            } catch (NoSuchAlgorithmException ex) {
+                throw new IllegalStateException("no SHA256 digest", ex);
+            }
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (o == null || getClass() != o.getClass()) return false;
```
mbaedke, line 177: That is no good. The actual param could be of a derived class. I'd use "if (!(o instanceof AbstractBlob))"

reschke: It was generated by IntelliJ :-)

PR #2378, opened 2025-07-08
 OAK-11793: remove usage of Guava common.hash

reschke:


oak-store-spi/src/main/java/org/apache/jackrabbit/oak/plugins/memory/AbstractBlob.java
```diff
@@ -66,7 +66,7 @@ public static boolean equal(Blob a, Blob b) {
 
         //Check for identity first. If they are same then its
         //definitely same blob. If not we need to check further.
-        if (ai != null && bi != null && ai.equals(bi)){
+        if (ai != null && ai.equals(bi)){
             return true;
         }
```
mbaedke, line 72: On that occasion, we could also simplify the length check to "if (a.length() != b.length())".

mbaedke (approved):

rishabhdaim (approved):
If possible, could you please run the build locally with only test changes to make sure they passed with older changes?

reschke:
> If possible, could you please run the build locally with only test changes to make sure they passed with older changes?

Yes, I can do that.

EDIT: yes, done.

rishabhdaim (approved):

oak-store-spi/src/main/java/org/apache/jackrabbit/oak/plugins/memory/AbstractBlob.java
```diff
@@ -152,7 +122,7 @@ public String getContentIdentity() {
 
     /**
      * To {@code Blob} instances are considered equal iff they have the
-     * same SHA-256 hash code  are equal.
+     * same SHA-256 hash code or are equal.
```
mbaedke, line 125: s/or are equal/or are identical/ would make clearer that we refer to == here and not to #equals() .

mbaedke: Still totally redundant though.

PR #2390, opened 2025-07-14
OAK-11672 log warnings when fetching large result sets

joerghoh:


reschke (approved):

thomasmueller (commented):
This is more a question:

We already log warnings if a query reads many nodes. What is different is:

* (A) in the current code (without the PR), warnings are logged when reading many entries from the index / from the node store
* (B) in this PR, warnings are logged while iterating over the results

It is possible that we see warnings from (A) but not from (B) for queries that don't have a proper index. Additionally, if many of the results are not accessible (missing access rights). Also, if a query has unindexed "order by" and only the first few entries are read. 

So, I wonder, were you aware that we already log warnings in (A)? If yes, I do not see this information in the Jira issue. It would be good to add. Or maybe it is not working?

Then, having two kinds of warnings might possibly be a bit confusing? For me it's OK.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/query/QueryResultDebugIterator.java
```diff
@@ -0,0 +1,83 @@
...
+    public K next() {
+        mark();
+        if (count >= FIRST_LOG_THRESHOLD) {
+            potentiallyLog();
+        }
+        return iter.next();
+    }
+
+    protected void mark() {
+        count++;
```
thomasmueller: Why is this protected? (Why even have a method?)

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/query/QueryResultDebugIterator.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+public class QueryResultDebugIterator<K> implements Iterator<K> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(QueryResultDebugIterator.class);
+
+    private Iterator<K> iter;
+    private int count;
+    private String query;
+    private String queryLanguage;
```
thomasmueller:
```suggestion
    private final Iterator<K> iter;
    private final String query;
    private final String queryLanguage;
    private int count;
```

joerghoh:
@thomasmueller I am aware of the warnings, when many nodes are read during a query (for whatever reason); these warnings are useful, as they indicate a slowness, but just from looking at them it's normally unclear why that many nodes are read.

This PR adds logs for one of those cases (reading many results).

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/query/QueryResultDebugIterator.java
joerghoh: good point, inlined it.

PR #2392, opened 2025-07-14
OAK-11807: Duplicate header after failed SegmentBufferWriter.flush()

mreutegg:
Do not modify the instance field length when the total length including the header is calculated.

### DIFF
--- a/oak-segment-tar/src/test/java/org/apache/jackrabbit/oak/segment/FailedFlushTest.java
+++ b/oak-segment-tar/src/test/java/org/apache/jackrabbit/oak/segment/FailedFlushTest.java
@@ -71,6 +71,10 @@ public class FailedFlushTest {
             writer.writeString(randomStrings.nextAlphanumeric(16));
         }
 
+        // Repeatedly fail flush before segment write to provoke OAK-11807.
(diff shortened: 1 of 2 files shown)


oak-segment-tar/src/test/java/org/apache/jackrabbit/oak/segment/FailedFlushTest.java
```diff
@@ -0,0 +1,173 @@
...
+                writer.flush();
+                fail("This flush must fail");
+            } catch (IOException e) {
+                // expected
+            }
+        }
+        failBeforeSegmentWrite = false;
+
+        // must succeed now
+        writer.flush();
```
jelmini, line 91: It's not clear why the flush succeeds now. It could seem that it now succeeds because trivially you set `failBeforeSegmentWrite = false`. I suppose the intention of the test is to show that this last flush doesn't fail with 'Too much data for a segment`. Can we make it more explicit?

mreutegg: Added comments to the test.

smiroslav (approved):

PR #2396, opened 2025-07-21
OAK-11812 : added missing osgi dependencies to fix the classnotfounde…

rishabhdaim:
…xception


amit-jain (approved):

reschke:
Rishab: when did this start to fail, and is the failure reproducible?

PR #2398, opened 2025-07-22
OAK-11813 : updated h2database to 2.3.232 to fix a security issue

rishabhdaim:


reschke:
please see https://issues.apache.org/jira/browse/OAK-10701

PR #2402, opened 2025-07-25
OAK-11816: oak-run should ignore unknown properties in --fds config b…

yuansc:
…y default

This change improves the usability of the oak-run `datastore` command by allowing it to tolerate unknown configuration properties in the FileDataStore config file (e.g., Sling-specific keys like `org.apache.sling.installer.configuration.persist`).

A new optional flag `--strict-config-check` was added to explicitly enforce strict validation. When enabled, oak-run will fail on unknown properties; otherwise, it logs a warning and proceeds (default behavior).

This aligns oak-run with real-world usage in Oak-based systems like AEM, where OSGi config files may include unrelated metadata keys.

Includes:
* `strictConfigCheck` flag support added to `BlobStoreFixtureProvider.populate(...)`
* CLI support for `--strict-config-check`
* Help text for the new flag
* Warning log on ignored properties


scottyuancoc:
Hello, I hope  this minor change would be helpful.  Also, please use scottyuancoc as the committer.  Thanks.

PR #2410, opened 2025-07-29
OAK-11814 - Improve error handling in ElasticResultRowAsyncIterator to deal with timeouts of both slow Elastic queries and slow readers.

nfsantos:
The async iterator that is reading results from Elastic was not handling correctly a few errors cases:
- The Elastic query takes too long to execute and the reader time outs. The iterator might not be closed in this case, leaving the request to Elasticsearch running even though there is no thread waiting for this result 
- The reader thread is too slow or crashes without closing the iterator. In this case, the threads managed by the Elasticsearch Java driver that are performing the request may hang forever trying to produce results when there is no thread consuming them

This PR fixes these issues, adding error handling to close resources and propagate errors in all these cases. 

It introduces a new configuration property:
- `oak.index.elastic.query.asyncIteratorEnqueueTimeoutMs` - Controls how long the Elasticsearch Java driver threads will wait trying to enqueue results.


fabriziofortino (approved):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
```diff
@@ -241,6 +288,7 @@ class ElasticQueryScanner {
 
         // Semaphore to guarantee only one in-flight request to Elastic
         private final Semaphore semaphore = new Semaphore(1);
+        private CompletableFuture<SearchResponse<ObjectNode>> ongoingRequest;
```
mkapalka: It's not obvious from the code if this is always accessed by the same thread, maybe it's safer to make it `volatile`?

nfsantos: It's also not clear to me. It may be ok as non-volatile as the class uses other synchronization mechanisms (semaphore and AtomicBoolean) before accessing this field which may trigger a memory barrier. But as it would take me too long to figure out, I'll just go with your suggestion of making it volatile. It is not performance critical, so no harm in being cautious here.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
```diff
@@ -176,7 +214,10 @@ public boolean on(Hit<ObjectNode> searchHit) {
     @Override
     public void endData() {
         try {
-            queue.put(POISON_PILL);
+            boolean success = queue.offer(POISON_PILL, enqueueTimeoutMs, TimeUnit.MILLISECONDS);
+            if (!success) {
+                LOG.warn("Timeout waiting to insert poison pill into the iterator queue. The iterator might not be closed properly.");
```
mkapalka, line 219: When this happens, the following call to `hasNext` will time out and throw an exception. I'm wondering if it's not better to throw an exception here directly.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
```diff
@@ -415,31 +466,72 @@ private void scan() {
...
+                }
+                try {
+                    if (!queue.offer(POISON_PILL, enqueueTimeoutMs, TimeUnit.MILLISECONDS)) {
+                        LOG.warn("Timeout waiting to enqueue poison pill after error processing search response. The iterator might not be closed properly.");
+                    }
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();  // restore interrupt status
+                    LOG.warn("Interrupted while trying to enqueue poison pill after error processing search response", e);
+                }
+                throw t;
```
mkapalka: Method `handleResponse` is used only as argument to `whenComplete` which javadoc is saying: "this method is not designed to translate completion outcomes, so the supplied action should not throw an exception", and also the resulting stage is not used later to get results (it's only used for cancelling the task). Would it be safer to close the scanner here instead of throwing `t` to propagate the error quickly to the iterator?

mkapalka (approved):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
nfsantos: `endData()` is called from two locations:
- `ElasticInsecureFacetAsyncProvider.on` - This is called at the end of the results, so there will be no more calls to the iterator. And if we throw an exception here, it will be swallowed by the driver as it executes on the event handling thread of the driver.
- `ElasticQueryScanner.close` - The scanner is closing, so any further calls to `hasNext()` will see `isClosed` set to true and will no try to read from the queue.

I prefer to leave it just as a warning, as throwing an exception could lead to unexpected results.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/ElasticResultRowAsyncIterator.java
nfsantos: I changed the method to not throw an exception, to respect the contract of `whenComplete`. 
I want to preserve the semantics that the consumer will see all the results that were received until the error occurred. So any results on the queue should be given to the consumer, before we close the iterator due to errors. That's why in this error handler I enqueue the poison pill to schedule throwing an error to the consumer when it finishes reading the results.

PR #2399, opened 2025-07-23
OAK-11809 : replaced guava's bloom filter with apache commons-collect…

rishabhdaim:
…ions4 implementation

PR #1345, opened 2024-03-05
OAK-10674: DocumentStore: verify that we could use Oak's Bloom filter

reschke:
(temporary playground for testing)

PR #2412, opened 2025-07-29
OAK-10674 Use Oak's Bloom filter

thomasmueller:
The PR aims to be an alternative to:
* https://github.com/apache/jackrabbit-oak/pull/2399 (using apache commons-collections4)
* https://github.com/apache/jackrabbit-oak/pull/1345 (draft PR)

This PR is larger because (unlike the other two PRs) it also changes oak-run-commons


oak-blob/src/main/java/org/apache/jackrabbit/oak/spi/blob/split/BlobIdSet.java
```diff
@@ -25,36 +25,34 @@
...
 
-    private final BloomFilter<CharSequence> bloomFilter;
+    private final BloomFilter bloomFilter;
 
     private final Cache<String, Boolean> cache;
 
     BlobIdSet(String repositoryDir, String filename) {
         store = new File(new File(repositoryDir), filename);
-        bloomFilter = BloomFilter.create(Funnels.stringFunnel(StandardCharsets.UTF_8), 9000000); // about 8MB
+        bloomFilter = BloomFilter.construct(9000000, 0.01); // 9M entries, 1% false positive rate
```
rishabhdaim: The default for Guava is 3%.

oak-commons/pom.xml
```diff
@@ -102,6 +102,10 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-collections4</artifactId>
     </dependency>
+    <dependency>
```
rishabhdaim, line 105: Don't think this is required anymore.

rishabhdaim (changes requested):

oak-commons/pom.xml
thomasmueller: It is needed for MurmurHash3

oak-blob/src/main/java/org/apache/jackrabbit/oak/spi/blob/split/BlobIdSet.java
thomasmueller: Oh! That is surprisingly high. OK, I'll also use 3%.

reschke (commented):
I'll leave review of the actual implementation to others.

But I'll have a look at the API aspects in commons.

reschke (approved):

rishabhdaim (approved):

PR #2415, opened 2025-07-30
OAK-11827 check the permission of the mixin tree only once

joerghoh:
Introducing a simple cache structure on a session level, which is invalidated on every session.refresh().

### DIFF
--- a/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/NodeImpl.java
+++ b/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/NodeImpl.java
@@ -128,6 +128,8 @@ public class NodeImpl<T extends NodeDelegate> extends ItemImpl<T> implements Jac
      */
     private static final Logger LOG = LoggerFactory.getLogger(NodeImpl.class);
 
+    private static final String CACHE_KEY_MIXIN_TPYES_READABLE = NodeImpl.class.getName() + ".MIXIN_TYPES_READABLE";
+
     private final int logWarnStringSizeThreshold;
 
     @Nullable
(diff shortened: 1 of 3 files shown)


oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/NodeImpl.java
```diff
@@ -1342,8 +1344,11 @@ protected Tree createValue() {
     }
 
     private boolean canReadMixinTypes(@NotNull Tree tree) {
-        return sessionContext.getAccessManager().hasPermissions(
-                tree, EMPTY_MIXIN_TYPES, Permissions.READ_PROPERTY);
+        // cache this information per SessionContext, as it is valid independent of the node
```
anchela, line 1347: that is actually not correct.
whether or not the mixin type property can be read for a particular tree may vary from tree to tree.
so caching it for the first tree you encounter is not correct.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/SessionContext.java
```diff
@@ -152,6 +155,7 @@ public SessionContext(
         this.valueFactory = new ValueFactoryImpl(
                 delegate.getRoot(), namePathMapper, this.blobAccessProvider);
         this.sessionQuerySettings = sessionQuerySettings;
+        this.cacheMap = new HashMap<>();
```
anchela, line 158: hm.... not sure how that works now... you would need to keep it for every node you take a look at which will result for the cache to get really big. that doesn't sound good to me.
on the other hand: storing a boolean for each node might also come with some overhead as one would have to invalidate it upon save/refresh.

anchela (changes requested):
hi @joerghoh , i don't think the current approach works reliably as the readability of the mixin type property may vary from node to node.
i think we need a second round of thoughts on this one....

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/session/NodeImpl.java
joerghoh: hmpf, for some reason I oversaw this ``tree`` parameter. 

But on a different view, does it even make sense to disallow reading the mixin types in the context of ``Node.isNodeType()``? In that case this API call could return a different result, depending if I have read access to that property or not.

joerghoh:
closing because this approach is incorrect and won't work.

PR #2436, opened 2025-08-07
OAK-11847 - AzureJournalFile fails to rotate journal

jsedding:


oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureJournalFile.java
```diff
@@ -33,16 +33,19 @@
...
+import java.util.List;
+import java.util.Map;
 import java.util.stream.Collectors;
 
 public class AzureJournalFile implements JournalFile {
 
     private static final Logger log = LoggerFactory.getLogger(AzureJournalFile.class);
 
-    private static final int JOURNAL_LINE_LIMIT = Integer.getInteger("org.apache.jackrabbit.oak.segment.azure.journal.lines", 40_000);
+    static final int JOURNAL_LINE_LIMIT = Integer.getInteger("org.apache.jackrabbit.oak.segment.azure.journal.lines", 40_000);
```
reschke, line 48: this should use SystemPropertyProvider...

reschke (approved):
I *believe* this is good.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzurePersistence.java
```diff
@@ -89,7 +96,7 @@ public boolean segmentFilesExist() {
 
     @Override
     public JournalFile getJournalFile() {
-        return new AzureJournalFile(readBlobContainerClient, writeBlobContainerClient, rootPrefix + "/journal.log", writeAccessController);
+        return new AzureJournalFile(readBlobContainerClient, writeBlobContainerClient, rootPrefix + "/journal.log", writeAccessController, journalLineLimit);
```
ierandra, line 99: `journalLineLimit` is already available in  `AzureJournalFile`

ierandra: this is set here https://github.com/apache/jackrabbit-oak/blob/trunk/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureJournalFile.java#L66

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureJournalFile.java
jsedding: agreed, but I would say that's for a different JIRA ticket.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzurePersistence.java
jsedding: I wanted to use the way `AzurePersistence` constructs `AzureJournalFile`, because the bug was introduced by changing the arguments of the `AzureJournalFile` constructor.

However, by doing that, I needed a way to feed the `journalLineLimit` through `AzurePersistence`. I'm more than happy if you see a better way to achieve this.

smiroslav (approved):

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzurePersistence.java
ierandra: the logic for setting the `journalLineLimit` has not changed.

The problem indeed is because of the `rootPrefix + "/journal.log"` in the constructor

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureJournalFile.java
```diff
@@ -245,23 +248,22 @@ private void createNextFile(int suffix) throws IOException {
         }
 
         private int parseCurrentSuffix() {
-            String name = AzureUtilities.getName(currentBlob);
```
ierandra, line 248: the issue should be fixed to change this line with:
```
String name = currentBlob.getBlobName();
```

ierandra: this is because the `journalNamePrefix` = `rootPrefix + "/journal.log"` 

and the result of `AzureUtilities.getName(currentBlob);` = `journal.log` or `journal.log.001`

however is we change it to `currentBlob.getBlobName();` = `rootPrefix + "/journal.log"` or `rootPrefix + "/journal.log.001"`

ierandra:
Created this https://github.com/apache/jackrabbit-oak/pull/2438 with the fix

jsedding:
Closing as #2438 was merged.

PR #2438, opened 2025-08-07
OAK-11847 - AzureJournalFile fails to rotate journal

ierandra:
The issue was introduced because when [AzureJournalFile](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzurePersistence.java#L92) is created the `journalNamePrefix` is set to `rootPrefix + "/journal.log"` and `AzureUtilities.getName(currentBlob)` is returning `journal.log`. The result is taht `matcher.find()` will always return false since `journalNamePrefix` = `rootPrefix + "/journal.log"` and `name` = `journal.log`

### DIFF
--- a/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureJournalFile.java
+++ b/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureJournalFile.java
@@ -33,7 +33,12 @@ import org.slf4j.LoggerFactory;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
-import java.util.*;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.Iterator;
+import java.util.HashMap;
+import java.util.ArrayList;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -245,7 +250,7 @@ public class AzureJournalFile implements JournalFile {
         }
 
         private int parseCurrentSuffix() {
-            String name = AzureUtilities.getName(currentBlob);
+            String name = currentBlob.getBlobName();
             Pattern pattern = Pattern.compile(Pattern.quote(journalNamePrefix) + "\\.(\\d+)");
             Matcher matcher = pattern.matcher(name);
             int parsedSuffix;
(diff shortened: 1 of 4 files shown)


oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/AzureJournalFileTest.java
```diff
@@ -53,13 +53,15 @@ public class AzureJournalFileTest {
...
+    private final String rootPrefix = "oak";
+
     @Before
     public void setup() throws BlobStorageException {
         readBlobContainerClient = azurite.getReadBlobContainerClient("oak-test");
         writeBlobContainerClient = azurite.getWriteBlobContainerClient("oak-test");
         WriteAccessController writeAccessController = new WriteAccessController();
         writeAccessController.enableWriting();
-        journal = new AzureJournalFile(readBlobContainerClient, writeBlobContainerClient, "journal.log", writeAccessController, 50);
+        journal = new AzureJournalFile(readBlobContainerClient, writeBlobContainerClient, rootPrefix + "journal.log", writeAccessController, 50);
```
jsedding: With this change your test does not cover the fact that the issue was introduced by a change in `AzurePersistence`.

Of course it can be argued that this is a unit test and it should only test the class `AzureJournalFile`. However, as we have seen, this test did not catch the regression, because this aspect was untested.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureJournalFile.java
```diff
@@ -245,7 +250,7 @@ private void createNextFile(int suffix) throws IOException {
         }
 
         private int parseCurrentSuffix() {
-            String name = AzureUtilities.getName(currentBlob);
+            String name = currentBlob.getBlobName();
```
jsedding, line 254: Granted, this is simpler. My feeling was just that using a regexp is overkill here, and doesn't help the clarity of the code.

jsedding:
This looks pretty much like my first iteration did 🙂 I hope I managed to explain why I opted for a slightly more complicated change.

oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/AzureJournalFileTest.java
ierandra: you are correct, thanks! added the changes

smiroslav (approved):

PR #2442, opened 2025-08-12
OAK-11851 - Log performance metrics of statistical facets and simplify logic

nfsantos:
- Print timings on success at DEBUG level and on failure at ERROR level. Example:

```
16:14:21.608 DEBUG [ForkJoinPool.commonPool-worker-1] ElasticStatisticalFacetAsyncProvider.java:132 Facet computation times: {query: 46 ms, processAggregations: 0 ms, filterByAcl: 5 ms, processHits: 0 ms, computeStatisticalFacets: 0 ms}. Total hits: 4001, samples: 1000
```

- Cancel ES search request if consumer times out waiting for results.
- Simplify logic by removing CountdownLatch and using instead the search future to wait for completion of ES request.

The coupling between facet evaluation and query results retrieval complicates further improvements:
-  In case of failure, we log an exception and return null. This is likely not the ideal behavior, as the caller has no way to distinguish between a failure and empty results. I'm leaving this behavior as is to not introduce further changes into this PR. With the current architecture, the layers above facet evaluation that also manage iteration over the results are swallowing exceptions from facet evaluation, so to propagate exceptions to callers we would have to also change these layers.
- There are not tests for timeouts or failures. Once again, the layers above facet evaluation have their own blocking points, where they wait for the first results, which interfere with timeouts set for facet evaluation.

We should revise these behaviors once facet evaluation and query results retrieval are decoupled.

### DIFF
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/facets/ElasticStatisticalFacetAsyncProvider.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/facets/ElasticStatisticalFacetAsyncProvider.java
@@ -75,7 +75,7 @@ public class ElasticStatisticalFacetAsyncProvider implements ElasticFacetProvide
     // same thread or by the client thread that waits for the latch to complete. Since the latch causes a memory barrier,
     // the updated values will be visible to the client thread.
     private long queryTimeNanos;
-    private long processAggregationsTimeNannos;
+    private long processAggregationsTimeNanos;
     // It is written by multiple threads, so we use LongAdder for better performance than AtomicLong
     private final LongAdder aclTestTimeNanos = new LongAdder();
     private long processHitsTimeNanos;
@@ -107,7 +107,7 @@ public class ElasticStatisticalFacetAsyncProvider implements ElasticFacetProvide
         );
 
         this.queryStartTimeNanos = System.nanoTime();
-        LOG.trace("Kicking search query with random sampling {}", searchRequest, new Throwable());
+        LOG.trace("Kicking search query with random sampling {}", searchRequest);
         this.searchFuture = connection.getAsyncClient()
                 .search(searchRequest, ObjectNode.class)
                 .thenApplyAsync(this::computeFacets);
@@ -141,7 +141,7 @@ public class ElasticStatisticalFacetAsyncProvider implements ElasticFacetProvide
     }
 
     private Map<String, List<FulltextIndex.Facet>> computeFacets(SearchResponse<ObjectNode> searchResponse) {
-        LOG.info("SearchResponse: {}", searchResponse);
+        LOG.trace("SearchResponse: {}", searchResponse);
         this.queryTimeNanos = System.nanoTime() - queryStartTimeNanos;
         List<Hit<ObjectNode>> searchHits = searchResponse.hits().hits();
         this.sampled = searchHits != null ? searchHits.size() : 0;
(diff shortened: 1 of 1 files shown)


oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/facets/ElasticStatisticalFacetAsyncProvider.java
```diff
@@ -63,14 +64,23 @@ public class ElasticStatisticalFacetAsyncProvider implements ElasticFacetProvide
...
+    private final CompletableFuture<Map<String, List<FulltextIndex.Facet>>> searchFuture;
     private int sampled;
     private long totalHits;
 
+    private final long queryStartTimeNanos;
+    // All these variables are updated only by the event handler thread of Elastic. They are read either by that
+    // same thread or by the client thread that waits for the latch to complete. Since the latch causes a memory barrier,
+    // the updated values will be visible to the client thread.
+    private long queryTimeNanos;
+    private long processAggregationsTimeNannos;
```
fabriziofortino: typo
```suggestion
    private long processAggregationsTimeNanos;
```

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/facets/ElasticStatisticalFacetAsyncProvider.java
```diff
@@ -96,87 +106,120 @@ public class ElasticStatisticalFacetAsyncProvider implements ElasticFacetProvide
...
-                        searchResponse.hits().hits().forEach(this::processHit);
-                        computeStatisticalFacets();
-                    }
-                }
-            } finally {
-                latch.countDown();
-            }
-        });
+        this.queryStartTimeNanos = System.nanoTime();
+        LOG.trace("Kicking search query with random sampling {}", searchRequest, new Throwable());
```
fabriziofortino: any specific reason to print the stacktrace here?

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/facets/ElasticStatisticalFacetAsyncProvider.java
```diff
@@ -96,87 +106,120 @@ public class ElasticStatisticalFacetAsyncProvider implements ElasticFacetProvide
...
-                                    v.increment();
-                                    return v;
-                                }
-                            });
-                            return facetValues;
-                        }
-                    });
-                }
+    private Map<String, List<FulltextIndex.Facet>> computeFacets(SearchResponse<ObjectNode> searchResponse) {
+        LOG.info("SearchResponse: {}", searchResponse);
```
fabriziofortino: debug/trace?

nfsantos: Set level to TRACE.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/query/async/facets/ElasticStatisticalFacetAsyncProvider.java
nfsantos: Leftover from local debugging during development. I removed logging of the stack trace.

fabriziofortino (approved):

PR #2447, opened 2025-08-13
OAK-11833 : replaced Guava's Traverser with OAK commons

rishabhdaim:


oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/Traverser.java
```diff
@@ -169,5 +167,107 @@ public T next() {
...
+     * @param root the root node of the tree, may be null
+     * @param childExtractor function to extract children from a node, must not be null
+     * @return an iterator that traverses the tree in post-order
+     * @throws NullPointerException if childExtractor or any child is null
+     */
+    public static <T> FluentIterable<T> postOrderTraversal(final T root, final @NotNull Function<T, Iterable<? extends T>> childExtractor) {
+        Objects.requireNonNull(childExtractor, "Children extractor function must not be null");
+        if (root == null) {
+            return FluentIterable.empty();
+        }
```
mbaedke: Shouldn't it throw if root==null?

thomasmueller: (I don't know if it should or should not throw in this specific case...)

I think it would be good to have a randomized test case where we can compare the exact behavior of the old vs the new implementation. Specially edge cases like this.

rishabhdaim: @thomasmueller randomized test case has been added for this traversal order as well.

rishabhdaim: @mbaedke Yes, you are right, Guava code does throw NPE in case the `root` is null. I will make the change to throw NPE in case the `root` is null.

rishabhdaim: @mbaedke addressed in https://github.com/apache/jackrabbit-oak/pull/2447/commits/7e017e82d0d7cf433c6fe21e322e439d3eedb552

mbaedke: Randomized tests are great, as long as we also have tests that specifically address obvious edge cases.

mbaedke (approved):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/Traverser.java
thomasmueller: The randomized tests are _specially_ useful to test edge cases. Specially randomized tests that compare the old with the new implementation. (The problem was that the randomized test didn't include edge cases I guess).

oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/TraverserTest.java
```diff
@@ -348,10 +346,179 @@ public void testBreadthFirstOrderTraversalWithRandomTree() {
...
+
+    // TODO remove this test when we remove guava dependency
+    @Test
+    public void testPostOrderTraversalWithRandomTree() {
+
+        final Node root = getRoot(10000);
+
+        List<Integer> result = Traverser.postOrderTraversal(root, Node::getChildren)
+                .transform(Node::getValue)
+                .toList();
```
thomasmueller, line 547: What the random data does not seem to include is edge cases, like null entries, null iterators, iterators that throw exceptions.

rishabhdaim: @thomasmueller in that, how can we predict the test behaviour. Like what to `assert` for ?

Would it be ok, that I add `Guava` validations in other edge test cases like `null` check ?

rishabhdaim: @thomasmueller added these test with https://github.com/apache/jackrabbit-oak/pull/2447/commits/244592135e11d082d8f5f2bd80cdf303c2465370

PR #2451, opened 2025-08-14
OAK-11859: update Tomcat dependency to 9.0.108

reschke:


kwin:
What about raising to 10.1 now that we require Java 11?

reschke:
> What about raising to 10.1 now that we require Java 11?

Good question. Note we use this for the webapp (demo only) and for JDBC/RDB, so that might need a bit more testing.

PR #2452, opened 2025-08-14
OAK-11846 oak-segment-azure - recovery will recover from .bak

ierandra:
How was the issue introduced:
- in v8 we had the concept of `CloudBlobDirectory` so using the path (rootPrefix+archiveName) worked.
- in v12 we have `BlobContainerClient`, we have the container and we must rely on prefix and there can be archives with the same prefix (eg. `aem/data0000a.tar`, `aem/data0000a.tar.bak`, `aem/data0000a.tar.2.bak`) so using path (rootPrefix+archiveName) didn't worked. To fix this the prefix is must end in `/` (eg. `aem/data0000a.tar/`)

I added a new class with unit tests to test all cases where this can happen.

### DIFF
--- a/oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManagerOverflowTest.java
+++ b/oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManagerOverflowTest.java
@@ -1,0 +1,192 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.jackrabbit.oak.segment.azure;
+
+import com.azure.core.util.BinaryData;
+import com.azure.storage.blob.BlobContainerClient;
+import com.azure.storage.blob.models.BlobStorageException;
+import com.azure.storage.blob.models.ListBlobsOptions;
+import org.apache.jackrabbit.oak.segment.remote.WriteAccessController;
+import org.apache.jackrabbit.oak.segment.spi.monitor.FileStoreMonitorAdapter;
+import org.apache.jackrabbit.oak.segment.spi.monitor.IOMonitorAdapter;
+import org.apache.jackrabbit.oak.segment.spi.monitor.RemoteStoreMonitorAdapter;
+import org.apache.jackrabbit.oak.segment.spi.persistence.SegmentArchiveManager;
+import org.apache.jackrabbit.oak.segment.spi.persistence.SegmentArchiveWriter;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.ClassRule;
+import org.junit.contrib.java.lang.system.ProvideSystemProperty;
+import org.junit.rules.TemporaryFolder;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.security.InvalidKeyException;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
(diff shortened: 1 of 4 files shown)


oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManagerOverflowTest.java
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        writer.flush();
+        writer.close();
+
+        readBlobContainerClient.getBlobClient("oak/" + archiveName + "/0005." + uuids.get(5).toString()).delete();
+
+        LinkedHashMap<UUID, byte[]> recovered = new LinkedHashMap<>();
+        manager.recoverEntries(archiveName, recovered);
+        assertEquals(new ArrayList<>(uuids.subList(0, 5)), new ArrayList<>(recovered.keySet()));
```
smiroslav:
```suggestion
        assertEquals(uuids.subList(0, 5), new ArrayList<>(recovered.keySet()));
```

oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManagerOverflowTest.java
```diff
@@ -0,0 +1,192 @@
...
+
+    @Test
+    public void testExistsArchiveOverflow() {
+        SegmentArchiveManager manager = azurePersistence.createArchiveManager(false, false, new IOMonitorAdapter(), new FileStoreMonitorAdapter(), new RemoteStoreMonitorAdapter());
+
+        assertFalse(manager.exists(archiveName));
+    }
+
+    @Test
+    public void testRenameToOverflow() {
```
smiroslav: From its name, it is not clear what this test does.  Maye something like `testRenameToIgnoresBackupArchiveBlobs`.

I would also prefer that the setup in terms of data present in the archive is done in the test, so that I do not need to constantly scroll up to remind myself what is in the archive.

oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManagerOverflowTest.java
```diff
@@ -0,0 +1,192 @@
...
+        manager.copyFile(archiveName, targetArchiveName);
+
+        boolean blobExistsInTargetArchive = readBlobContainerClient.listBlobs(new ListBlobsOptions().setPrefix(rootPrefix + "/" + targetArchiveName), null)
+                .iterator().hasNext();
+
+        assertFalse("blob from backup tar archive should not be copied", blobExistsInTargetArchive);
+    }
+
+    @Test
+    public void testDeleteOverflow() {
```
smiroslav: Similar comments like for the test `testExistsArchiveOverflow `

smiroslav (commented):
Changes look good. I would only make changes in the test so that it is more obvious what each does.

oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManagerOverflowTest.java
ierandra: I would like to have something more generic. How about `testRenameToIgnoreBlobsSamePrefix`.
WDYT?

ierandra: or `testRenameToIgnoreArchivesSamePrefix`

oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManagerOverflowTest.java
ierandra: done

oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManagerOverflowTest.java
ierandra: >I would also prefer that the setup in terms of data present in the archive is done in the test, so that I do not need to constantly scroll up to remind myself what is in the archive.

Done 

Also renamed the test to `testRenameToIgnoreBlobsSamePrefix`

oak-segment-azure/src/test/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManagerOverflowTest.java
ierandra: Renamed to `testExistsArchiveIgnoreArchiveSamePrefix`

smiroslav (approved):

PR #2453, opened 2025-08-14
OAK-11860 - oak-segment-azure should perform I/O in parallel

jsedding:


jelmini:
I wonder if parallelizing `AzureArchiveManager#delete` is safe, in case Oak crashes while deleting the archive. Currently `AzureArchiveManager#listArchives` considers an archive to be deleted if no blob starting with `0000.` is present. By parallelizing, the segment starting with `0000.` may be deleted after other blobs (other segments or the `closed` blob). In case of crash, the `0000.` may still be there: `TarReader`will then attempt to open the archive and fail, preventing the repository to initialize.

However, I'm not even sure the current sequential code is safe either, as the order of the blobs returned by `AzureUtilities#getBlobs` is not specified. 

In any case, I suggest to validate the behavior in case of crash before introducing this change in `AzureArchiveManager#delete`.

jsedding:
Interesting consideration, @jelmini ! Even though not related to my change, as you acknowledge.

How about we write a marker file `deleted` into the archive? And then we kick off a daemon thread to perform the actual delete. If we check for this file at startup, we could also start a daemon thread for the deletion(s). That way deletions would be somewhat atomic. WDYT? But that warrants its own ticket IMHO.

jsedding:
I have created OAK-11867 to track atomic deletes.

jsedding:
I'll close this PR for now. This is becoming very complex, possibly for little gain.

PR #2457, opened 2025-08-18
 OAK-11656: Remove usage of Guava Suppliers.memoize 

reschke:


reschke:
Can somebody suggest a way to make this thread-safe (which I believe it is already), *and* which Sonar does not complain about, *and* which does not have to go through through one indirection when the result value already is computed?

EDIT: ok, I see now.

mbaedke:
I don't think it is thread-safe if there is an unsynchronized null check. Can't you just synchronize the getter itself?

mbaedke:
Alternatively, use a volatile boolean flag that you can use instead of the null check.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/function/SuppliersUtils.java
```diff
@@ -0,0 +1,55 @@
...
+ * under the License.
+ */
+package org.apache.jackrabbit.oak.commons.function;
+
+import java.util.function.Supplier;
+
+/**
+ * Utility methods for {@link java.util.function.Supplier} handling.
+ */
+public class SuppliersUtils {
```
rishabhdaim: IMO we should use with name `Supplier` here.

reschke: I was going to say, that might cause confusion with the JDK class, but that's name "Supplier". So ack.

mbaedke:
Added new test SuppliersTest#concurrentSupplierAccess() which fails with previous versions of Suppliers#memoize().

nfsantos:
I found this alternative solution with an AtomicReference here:
https://www.baeldung.com/java-lambda-lazy-field-initialization#thread-safe-solution
```
    private final AtomicReference<T> data;

    public LazyLambdaThreadSafeSupplier(Supplier<T> expensiveData) {
        super(expensiveData);
        data = new AtomicReference<>();
    }

    public T getData() {
        if (data.get() == null) {
            synchronized (data) {
                if (data.get() == null) {
                    data.set(expensiveData.get());
                }
            }
        }
        return data.get();
    }
```
Seems to be a bit cleaner than the current solution: only a single AtomicReference variable instead of two variables, no need for a volatile boolean guard variable as that synchronization happens inside the AtomicReference. I find it more elegant and easier to understand. 
But it's just an idea for consideration, as the current solution also seems to be correct.

reschke:
> I found this alternative solution with an AtomicReference here: https://www.baeldung.com/java-lambda-lazy-field-initialization#thread-safe-solution
> 
> ```
>     private final AtomicReference<T> data;
> 
>     public LazyLambdaThreadSafeSupplier(Supplier<T> expensiveData) {
>         super(expensiveData);
>         data = new AtomicReference<>();
>     }
> 
>     public T getData() {
>         if (data.get() == null) {
>             synchronized (data) {
>                 if (data.get() == null) {
>                     data.set(expensiveData.get());
>                 }
>             }
>         }
>         return data.get();
>     }
> ```
> 
> Seems to be a bit cleaner than the current solution: only a single AtomicReference variable instead of two variables, no need for a volatile boolean guard variable as that synchronization happens inside the AtomicReference. I find it more elegant and easier to understand. But it's just an idea for consideration, as the current solution also seems to be correct.

I *believe* this would have the same issue wrt not returning from the synchronized block, at which time it gets more complex, too.

nfsantos:
> I _believe_ this would have the same issue wrt not returning from the synchronized block, at which time it gets more complex, too.

What is the concurrency issue that would happen if it does not return from within the synchronized block? The comment in the source code does not provide details.

reschke:
> What is the concurrency issue that would happen if it does not return from within the synchronized block? The comment in the source code does not provide details.

Without that, it failed concurrent tests (oak-segment-tar). Let me retry.

And yes, your proposal looks cleaner; will test.

nfsantos:
Hmm... I think I understand why it might be incorrect, but I don't see why the return would fix this. Scenario:
- Thread 1 enters the synchronized block and executes up to `initialized = true`. It is suspended at this point
- Thread 2 enters the method and sees `initialized = true` because this variable is volatile. However, `result` is not volatile so we have no guarantees that it will be updated until thread 1 leaves the synchronized block.
- Thread 3 skips the initialization block and returns result, which is still null.

I believe this can happen, but returning from within the block will not prevent this from happening. Unless there's something about putting the return there that makes it more likely that thread 1 will leave the synchronized block faster. Or maybe non-determinism in the tests. The solution with AtomicReference should not have the problem described above.

reschke:
The solution with `AtomicReference` will fail with a `null` return value from the supplier, right? So that would need another check.

nfsantos:
> The solution with `AtomicReference` will fail with a `null` return value from the supplier, right? So that would need another check.

In what scenario would it fail with a null return? I don't see how it can return null.

reschke:
> > The solution with `AtomicReference` will fail with a `null` return value from the supplier, right? So that would need another check.
> 
> In what scenario would it fail with a null return? I don't see how it can return null.

If the supplier returns `null`, the null check would always be true, and it would keep calling the supplier. Test case coming.

-> https://github.com/apache/jackrabbit-oak/pull/2457/commits/4f88425e2175f4586a58d2efb186259811212887

So we'd have to disallow it (check and throw), or need another boolean flag.

reschke:
> I believe this can happen, but returning from within the block will not prevent this from happening. Unless there's something about putting the return there that makes it more likely that thread 1 will leave the synchronized block faster. Or maybe non-determinism in the tests. The solution with AtomicReference should not have the problem described above.

It's possible that I borrowed this from Guava's implementation without checking whether it is needed.

nfsantos:
> > I believe this can happen, but returning from within the block will not prevent this from happening. Unless there's something about putting the return there that makes it more likely that thread 1 will leave the synchronized block faster. Or maybe non-determinism in the tests. The solution with AtomicReference should not have the problem described above.
> 
> It's possible that I borrowed this from Guava's implementation without checking whether it is needed.

Oh, I see. I would still prefer to understand why the extra return statement is needed, but ok, I guess we can trust the Guava approach.

nfsantos (approved):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/function/Suppliers.java
```diff
@@ -0,0 +1,62 @@
...
+    private Suppliers() {
+    }
+
+    /**
+     * Transforms a {@link Supplier} based on a wrapper that evaluates
+     * the given {@code Supplier} at most one time.
+     * @param <T> return type
+     * @return Supplier based on the given Supplier
+     */
+    public static <@Nullable T> Supplier<T> memoize(final Supplier<@Nullable T> computeOnce) {
```
jsedding: Nullability annotations don't go on type parameters.

```suggestion
    public static <T> Supplier<@Nullable T> memoize(final Supplier<@Nullable T> computeOnce) {
```

oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/function/SuppliersTest.java
```diff
@@ -0,0 +1,111 @@
...
+                } catch (InterruptedException ignored) {}
+            }
+        });
+        synchronized (concurrencyTestMonitor) {
+            for (int k = 0; k < threadCount; k++) {
+                threads.get(k).start();
+            }
+        }
+        waitForAll.start();
+        waitForAll.join();
```
jsedding: This test seems overly complicated. I think it could be simplified a lot by using a `CountDownLatch` to let all threads wait before calling the memoizer.

TBH, I have never seen this pattern with an empty synchronize block, and I don't understand what it does.

The "waitForAll" functionality can happen in the "main" thread, no need to create a new thread and join it.

Finally, it would be more elegant to move all the class fields (e.g. `concurrencyTestFailed`,  `sourceSupplier`, etc) into the test method (this may require changing `concurrencyTestFailed` to `AtomicBoolean`). This makes the test method more self-contained and thus easier to read. It is also more maintainable, because it prevents test methods written in the future from re-using these fields, which might introduce flaky tests that depend on the ordering of test execution.

jsedding (commented):
No blockers, but I would like to see my two comments addressed.

jsedding:
Just for fun, this implementation would be `null`-safe. The trick is to use the `ref` object as a `null`-marker.

However, it might lead to `computeOnce` getting called more than once when called concurrently. It may be possible to add a guard against this, but I don't think it's worthwhile.

```

    public static <T> Supplier<@Nullable T> memoize(final Supplier<@Nullable T> computeOnce) {
        return new Supplier<>() {
            final AtomicReference<Object> ref = new AtomicReference<>(computeOnce);

            @Override
            @SuppressWarnings( "unchecked")
            public @Nullable T get() {
                Object o = ref.updateAndGet(prev -> Objects.requireNonNullElseGet(
                        prev,
                        () -> Optional.<Object>ofNullable(computeOnce.get()).orElse(ref)
                ));
                return o == ref ? null : (T) o;
            }
        };
    }
```

oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/function/SuppliersTest.java
reschke: -> @mbaedke - can you check?

reschke:
> However, it might lead to `computeOnce` getting called more than once when called concurrently. It may be possible to add a guard against this, but I don't think it's worthwhile.

If we did, we'd be back to the original PR, just with `AtomicReference`.

oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/function/SuppliersTest.java
mbaedke: @jsedding: An empty synchronized block is commonly used to let multiple threads start from the same point in time. The test does not work without that. There are certainly other solutions. I am extremely old school when it comes to elementary concurrent programming (in particular in testing). I find it extremely readable, but apparently not everyone shares this opinion. It's a test. It works. I don't want to spend too much time on it. But I'll look into it.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/function/Suppliers.java
```diff
@@ -0,0 +1,62 @@
...
+            public @Nullable T get() {
+                if (!initialized) {
+                    synchronized (this) {
+                        if (!initialized) {
+                            result = computeOnce.get();
+                            // only set initialized once value is computed and assigned
+                            initialized = true;
+                            // return from inside the synchronized block in order to
+                            // avoid concurrency problems
+                            return result;
```
rishabhdaim: @nfsantos @reschke Just for my understanding, why do we need to return from inside synchronized block.

rishabhdaim:
> * Thread 1 enters the synchronized block and executes up to `initialized = true`. It is suspended at this point
> 
> * Thread 2 enters the method and sees `initialized = true` because this variable is volatile. However, `result` is not volatile so we have no guarantees that it will be updated until thread 1 leaves the synchronized block.
> 
> * Thread 3 skips the initialization block and returns result, which is still null.

@nfsantos This can't happen. Any thread which reads the updated value of `initialized = true` which only happens after updating `result`, it is bound to see all the changes of the thread till updating `initialized` (which is thread 1 in this case) including the `result` variable.

rishabhdaim (approved):

oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/function/SuppliersTest.java
mbaedke: Pushed a new version of the test case addressing @jsedding's last two issues.

mbaedke: Pushed another version, now using CountDownLatch. Did not know that, very practical, thx.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/function/Suppliers.java
mbaedke: With this implementation I think it's actually not necessary.

mbaedke: I think it's from a previous version and now redundant. Test also work fine without it.

mbaedke (approved):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/function/Suppliers.java
reschke: I'm currently double-checking.

reschke:
```
[INFO] Running org.apache.jackrabbit.oak.segment.ReadOnlyStoreBlobReferencesTest
[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.471 s <<< FAILURE! - in org.apache.jackrabbit.oak.segment.ReadOnlyStoreBlobReferencesTest
[ERROR] collectReferencesAfterGC(org.apache.jackrabbit.oak.segment.ReadOnlyStoreBlobReferencesTest)  Time elapsed: 0.316 s  <<< FAILURE!
java.lang.AssertionError: Read only store visible references different expected:<0> but was:<1>
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:647)
        at org.apache.jackrabbit.oak.segment.ReadOnlyStoreBlobReferencesTest.assertReferences(ReadOnlyStoreBlobReferencesTest.java:140)
        at org.apache.jackrabbit.oak.segment.ReadOnlyStoreBlobReferencesTest.collectReferencesAfterGC(ReadOnlyStoreBlobReference
```

(sometimes)

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/function/Suppliers.java
```diff
@@ -0,0 +1,64 @@
...
+     * the given {@code Supplier} at most one time.
+     * <p>
+     * Suppliers returning {@code null} values are allowed; the memoized
+     * Supplier will be return {@code null} when the wrapped one does on
+     * first invocation.
+     * @param <T> return type
+     * @return Supplier based on the given Supplier
+     */
+    public static <T> Supplier<T> memoize(final Supplier<T> computeOnce) {
+        return new Supplier<>() {
```
kwin, line 42: why anonymous inner class? I would prefer a static final nested class.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/function/Suppliers.java
reschke: ok: https://issues.apache.org/jira/browse/OAK-10422

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/function/Suppliers.java
mbaedke: Why?

kwin: easier to read, easier to reuse.

reschke:
ok, I removed the early `return` - and it still appears to work.

I believe we have reached the point of deminishing returns; please quickly review so I can merge.

(after running CI)

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/function/Suppliers.java
mbaedke: I don't find it easier to read, but don't really care, so that's also fine.

reschke: I don't find it easier to read. And if we need to re-use it (why?), we can still change it.

mbaedke:
The two issues Sonar detected will be addressed once CI is done (one is actually relevant and could in theory lead to a hanging test).

oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/function/SuppliersTest.java
mbaedke: FYI https://github.com/apache/jackrabbit-oak/pull/2457#issuecomment-3205782142

nfsantos:
In the near future, this issue will be solved by Stable Values, which will first appear in Java 25 as preview:
https://openjdk.org/jeps/502

```
    private final StableValue<Logger> logger = StableValue.of();

    Logger getLogger() {
        return logger.orElseSet(() -> Logger.create(OrderController.class));
    }
```
This guarantees lazy and thread-safe initialization. For Oak, it will take a long time until we can target a Java version with a finalized version of Stable Values, but it's nice to know that we will no longer need to write double-checked locking logic.

PR #2461, opened 2025-08-20
OAK-11869 : removed usage of Guava's hashing with apache commons-codec

rishabhdaim:


rishabhdaim:
Checked the exact outputs from Guava & Apache by following code.

```
        byte mask = (byte) 0xFF;
        long length = 12345678987654332L;
        byte[] data = "data".getBytes();
        
// Guava implementation
        Hasher hasher = Hashing.murmur3_32().newHasher();
        long guavaResult = hasher.putByte(mask).putLong(length).putBytes(data).hash().padToLong();

// Commons implementation - use the right byteBuffer size and byte order
        ByteBuffer byteBuffer = ByteBuffer.allocate(1 + 8 + data.length)
                .order(ByteOrder.LITTLE_ENDIAN)  // Important for long values
                .put(mask)
                .putLong(length)
                .put(data);
        byteBuffer.flip();  // Reset position to start
        byte[] bytes = new byte[byteBuffer.limit()];
        byteBuffer.get(bytes);
        long commonsResult = Integer.toUnsignedLong(MurmurHash3.hash32x86(bytes));

        Assert.assertEquals(guavaResult, commonsResult);
```


&&

```
Assert.assertEquals(Hashing.sha256().hashString("4:/indexPath/local-1234", StandardCharsets.UTF_8).toString(), DigestUtils.sha256Hex("4:/indexPath/local-1234"));

```

&& 

```
        byte[] data = "data".getBytes();

        Hasher hasher = Hashing.murmur3_32().newHasher();
        
        Assert.assertEquals(hasher.putBytes(data).hash().padToLong(), Integer.toUnsignedLong(MurmurHash3.hash32x86(data)));
```

reschke:
In the comment above, the following is missing, right?
~~~
bytes = byteBuffer.get(bytes);
~~~

reschke (approved):
LGTM.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/standby/codec/ChunkedBlobStream.java
```diff
@@ -143,8 +144,17 @@ private ByteBuf decorateRawBuffer(ByteBufAllocator allocator, ByteBuf buffer) {
...
+        final ByteBuffer byteBuffer = ByteBuffer.allocate(1 + 8 + data.length)
+                .order(ByteOrder.LITTLE_ENDIAN)  // To align with Guava that uses Little Endianess
+                .put(mask)
+                .putLong(length)
+                .put(data);
+        byteBuffer.flip();  // Reset position to start to read data from beginning
+        final byte[] bytes = new byte[byteBuffer.limit()];
+        byteBuffer.get(bytes);
+
+        long hash = Integer.toUnsignedLong(MurmurHash3.hash32x86(bytes));
```
thomasmueller: I think we should avoid copy & paste of code...
here and ResponseDecoder.java and StandbyTestUtils.java

thomasmueller (changes requested):

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/standby/codec/ResponseDecoder.java
```diff
@@ -203,11 +205,21 @@ private static void decodeGetReferencesResponse(int length, ByteBuf in, List<Obj
...
-        return Hashing.murmur3_32().newHasher().putByte(mask).putLong(blobLength).putBytes(data).hash().padToLong();
+
+        final ByteBuffer byteBuffer = ByteBuffer.allocate(1 + 8 + data.length)
+                .order(ByteOrder.LITTLE_ENDIAN)  // To align with Guava that uses Little Endianess
+                .put(mask)
+                .putLong(blobLength)
+                .put(data);
+        byteBuffer.flip();  // Reset position to start to read data from beginning
+        final byte[] bytes = new byte[byteBuffer.limit()];
+        byteBuffer.get(bytes);
```
mbaedke: Can't we just use ByteBuffer#array() here?

mbaedke (approved):
lgtm, assuming that both are correct MumurHash3 implementations.
Please also see my inline question.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/standby/codec/ResponseDecoder.java
rishabhdaim: Yes, it can be used, the above test passes with `ByteBuffer#array()` as well.

Reason for not using it is to avoid any uninitialized values to interfere with hash calculation i.e. the default values in `ByteBuffer`.

It is working because of we are initializing the ByteBuffer with exact size and the `ByteBuffer#flip()` sets the `limit` value to the end of bytebuffer to avoid reading un-initialized values. (where it has been filled).

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/standby/codec/ChunkedBlobStream.java
rishabhdaim: Let me create a common util function to handle this logic.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/standby/codec/HashUtils.java
```diff
@@ -0,0 +1,56 @@
...
+     * This method combines a byte mask, a long length value, and a byte array into a single
+     * byte sequence using little-endian byte ordering, then computes a 32-bit MurmurHash3
+     * hash of this sequence, returned as an unsigned long.
+     *
+     * @param mask    A byte value to include in the hash computation
+     * @param length  A long value to include in the hash computation (stored in little-endian order)
+     * @param data    The byte array data to include in the hash computation
+     * @return        The computed MurmurHash3 value as an unsigned 32-bit integer converted to long
+     */
+    public static long hash(byte mask, long length, byte[] data) {
```
thomasmueller: For the caller it's quite important to know which hash algorithm is used: the old code explicitly used murmur hash... for in-memory hashing (Bloom filter, hash tables, etc) that's not important, but here it seems it is, because the result is persisted (right?)

```suggestion
    public static long hashMurmur32(byte mask, long length, byte[] data) {
```

thomasmueller (changes requested):

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/standby/codec/HashUtils.java
rishabhdaim: addressed in https://github.com/apache/jackrabbit-oak/pull/2461/commits/9ff5339d8539eeeebc23d01078637f69195dd6a7

reschke (approved):
LGTM; commons codec indeed default to UTF-8 although the Javadoc for sha256hex does not mention it.

PR #2464, opened 2025-08-21
OAK-11873 also log the caller for Fulltext searches

joerghoh:
(see also OAK-9624, where I added this for other queries in Oak)


thomasmueller (changes requested):
Please _also_ add the call stack.

PR #2465, opened 2025-08-21
OAK-11870: - allow passing fullGCGeneration parameter to RevisionsCommand

shodaaan:
- added fullGCGeneration parameter for RevisionsCommand revisions (MongoDB garbage collection) command


Joscorbe (approved):
It looks good, but please add a test case that covers the new parameter [with the default value here](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-run/src/test/java/org/apache/jackrabbit/oak/plugins/document/RevisionsCommandTest.java#L190).

And a new test case that covers the non-default value, [like this](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-run/src/test/java/org/apache/jackrabbit/oak/plugins/document/RevisionsCommandTest.java#L376-L391).

Otherwise the coverage checks on new code will fail the pipeline.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
```diff
@@ -429,6 +437,7 @@ private VersionGarbageCollector bootstrapVGC(RevisionsOptions options, Closer cl
         System.out.println("IncludePaths are : " + sortedSet(gc.getFullGCIncludePaths()));
         System.out.println("ExcludePaths are : " + sortedSet(gc.getFullGCExcludePaths()));
         System.out.println("FullGcMode is : " + VersionGarbageCollector.getFullGcMode());
+        System.out.println("FullGCGeneration is : " + options.getFullGcGeneration());
```
rishabhdaim:
```suggestion
        System.out.println("FullGCGeneration is : " + gc.getFullGcGeneration());
```

rishabhdaim: To read values from `VersionGarbageCollector`

rishabhdaim (changes requested):
Please add test cases for fullGcGeneration as well.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/RevisionsCommand.java
shodaaan: Done, thank you for the suggestion.

shodaaan:
Added unit tests.

Joscorbe:
Ok, build passed but the sonarqube checks failed because it doesn't have permissions on the fork.

I will merge it then.

PR #2466, opened 2025-08-22
OAK-11300: shaded-guava: remove packages that are not used (anymore) …

reschke:
…(ongoing) - primitives


reschke:
- still used by concurrent (Monitor) -

PR #2467, opened 2025-08-22
OAK-9212 AzureArchiveManage.listArchives() should not delete segments

smiroslav:
https://issues.apache.org/jira/browse/OAK-9212

### DIFF
--- a/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
+++ b/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
@@ -120,6 +120,12 @@ public class AzureArchiveManager implements SegmentArchiveManager {
         return readBlobContainerClient.getBlobClient(getDirectory(archiveName) + DELETED_ARCHIVE_MARKER).exists();
     }
 
+    private void checkWriteOperation(String operation) {
+        if (readOnly) {
+            throw new UnsupportedOperationException("Operation " + operation + " is not allowed in read-only mode");
+        }
+    }
+
     @Override
     public SegmentArchiveReader open(String archiveName) throws IOException {
         try {
@@ -140,11 +146,13 @@ public class AzureArchiveManager implements SegmentArchiveManager {
 
     @Override
     public SegmentArchiveWriter create(String archiveName) throws IOException {
+        checkWriteOperation("create");
         return new AzureSegmentArchiveWriter(writeBlobContainerClient, rootPrefix, archiveName, ioMonitor, monitor, writeAccessController);
     }
 
     @Override
     public boolean delete(String archiveName) {
+        checkWriteOperation("delete");
         try {
             uploadDeletedMarker(archiveName);
             getBlobs(archiveName)
@@ -186,6 +194,7 @@ public class AzureArchiveManager implements SegmentArchiveManager {
 
     @Override
     public boolean renameTo(String from, String to) {
+        checkWriteOperation("renameTo");
         try {
             String targetDirectory = getDirectory(to);
             getBlobs(from)
@@ -206,6 +215,7 @@ public class AzureArchiveManager implements SegmentArchiveManager {
 
     @Override
     public void copyFile(String from, String to) throws IOException {
+        checkWriteOperation("copyFile");
         String targetDirectory = getDirectory(to);
         getBlobs(from)
                 .forEach(blobItem -> {
@@ -231,6 +241,7 @@ public class AzureArchiveManager implements SegmentArchiveManager {
 
     @Override
     public void recoverEntries(String archiveName, LinkedHashMap<UUID, byte[]> entries) throws IOException {
+        checkWriteOperation("recoverEntries");
         Pattern pattern = Pattern.compile(RemoteUtilities.SEGMENT_FILE_NAME_PATTERN);
         List<RecoveredEntry> entryList = new ArrayList<>();
 
@@ -288,6 +299,7 @@ public class AzureArchiveManager implements SegmentArchiveManager {
      */
     @Override
     public void backup(@NotNull String archiveName, @NotNull String backupArchiveName, @NotNull Set<UUID> recoveredEntries) throws IOException {
+        checkWriteOperation("backup");
         copyFile(archiveName, backupArchiveName);
         delete(archiveName, recoveredEntries);
     }
--- a/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
+++ b/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
@@ -80,36 +80,14 @@ public class AzureArchiveManager implements SegmentArchiveManager {
     @Override
     public List<String> listArchives() throws IOException {
         try {
-            List<String> archiveNames = readBlobContainerClient.listBlobsByHierarchy(rootPrefix).stream()
+            return readBlobContainerClient.listBlobsByHierarchy(rootPrefix).stream()
                     .filter(BlobItem::isPrefix)
                     .map(AzureUtilities::getName)
                     .filter(blobName -> blobName.endsWith(".tar"))
                     .collect(Collectors.toList());
-
-            Iterator<String> it = archiveNames.iterator();
-            while (it.hasNext()) {
-                String archiveName = it.next();
-                if (isArchiveEmpty(archiveName)) {
-                    delete(archiveName);
-                    it.remove();
-                }
-            }
-            return archiveNames;
         } catch (BlobStorageException e) {
             throw new IOException(e);
         }
-    }
-
-    /**
-     * Check if there's a valid 0000. segment in the archive
-     * @param archiveName
-     * @return true if the archive is empty (no 0000.* segment)
-     */
-    private boolean isArchiveEmpty(String archiveName) throws BlobStorageException {
-        String fullBlobPrefix = getDirectory(archiveName) + "0000.";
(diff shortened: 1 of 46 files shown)


oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureSegmentArchiveWriter.java
```diff
@@ -28,6 +28,8 @@
 import org.apache.jackrabbit.oak.segment.remote.WriteAccessController;
 import org.apache.jackrabbit.oak.segment.spi.monitor.FileStoreMonitor;
 import org.apache.jackrabbit.oak.segment.spi.monitor.IOMonitor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
mreutegg: These are not used.
```suggestion
```

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
```diff
@@ -80,38 +80,16 @@ public AzureArchiveManager(BlobContainerClient readBlobContainerClient, BlobCont
     @Override
     public List<String> listArchives() throws IOException {
         try {
-            List<String> archiveNames = readBlobContainerClient.listBlobsByHierarchy(rootPrefix).stream()
+            return readBlobContainerClient.listBlobsByHierarchy(rootPrefix).stream()
```
mreutegg: This method now returns empty archives. This is probably OK, but the code in trunk also filters out and attempts to delete archives that were in the process of being deleted and got interrupted. IIUC, `AzureArchiveManager.delete(String)` lists blobs in the archive and deletes them in that sequence. When interrupted, the first segment starting with `0000` has likely been deleted, hence `isArchiveEmpty()` returns `true`. This means `listArchives()` now also returns half-deleted archives. Isn't this a problem?

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/v8/AzureSegmentArchiveWriterV8.java
```diff
@@ -52,11 +55,23 @@ public class AzureSegmentArchiveWriterV8 extends AbstractRemoteSegmentArchiveWri
...
         this.writeOptimisedBlobRequestOptions = AzureRequestOptionsV8.optimiseForWriteOperations(archiveDirectory.getServiceClient().getDefaultRequestOptions());
+        this.created = hasBlobs();
+    }
+
+    private boolean hasBlobs() {
+        try {
+            return this.archiveDirectory.listBlobs().iterator().hasNext();
+        } catch (StorageException | URISyntaxException | NoSuchElementException e) {
+            LOG.error("Error listing blobs", e);
+            return false;
```
mreutegg: Wouldn't it be better to fail with an exception? Returning `false` in case of an exception seems wrong to me when there are actually blobs, but listing them failed.

smiroslav: Yes, makes sense.

 Test `AzureSegmentArchiveWriterV8Test` fails, but will mock the request that lists blobs in a container during archive writer initialisation.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
smiroslav: The code in `trunk` right now assumes the archive is empty if the segment "0000.*" is absent.
If we do filtering based on that, the question is how to distinguish intentional archive deletion from data corruption, when a segment is deleted or not successfully uploaded for any reason. 

 Let's handle archive deletion in a separate task, where we could signal intent with a special blob ("deleted" marker), which we can use for filtering. 
 WDYT?

mreutegg: I agree on the need to improve the archive delete operation and we can certainly do it in a separate task.

> The code in `trunk` right now assumes the archive is empty if the segment "0000.*" is absent. If we do filtering based on that, the question is how to distinguish intentional archive deletion from data corruption, when a segment is deleted or not successfully uploaded for any reason.

The first segment with position 0000 should only be deleted when the archive is deleted. I'm not aware of another code path where this happens. Or you mean when someone manually deletes the blob in Azure? If the segment is not successfully uploaded, then the segment write should simply fail or retried. Can you clarify why this is relevant in this context? 

My concern is that not filtering out those archives anymore in AzureArchiveManager.listArchives() is a change in behavior. There must be a reason why the code in trunk filters out those archives. Is this not relevant anymore? Why?

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureSegmentArchiveWriter.java
```diff
@@ -57,6 +57,7 @@ public AzureSegmentArchiveWriter(BlobContainerClient blobContainerClient, String
         this.archiveName = AzureUtilities.ensureNoTrailingSlash(archiveName);
         this.archivePathPrefix = AzureUtilities.asAzurePrefix(rootPrefix, archiveName);
         this.writeAccessController = writeAccessController;
+        this.created = AzureUtilities.archiveExists(blobContainerClient, archivePathPrefix);
```
mreutegg, line 60: Can you please explain why this is necessary? I don't understand how this is related to the reported issue.

smiroslav: Without that, the "empty" archive (no segments inside, only metadata) is not closed properly during recovery - the "closed" marker is not uploaded.
https://github.com/apache/jackrabbit-oak/blob/19f29c6e4e3001c2d7237c184588d5513f7b80e4/oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/tar/TarWriter.java#L212-L212

You can see that in the test 
org.apache.jackrabbit.oak.segment.azure.AzureArchiveManagerTest#testUncleanStopWithEmptyArchive

```java
java.io.IOException: Failed to open recovered tar file data00001a.tar
	at org.apache.jackrabbit.oak.segment.file.tar.TarReader.open(TarReader.java:113)
	at org.apache.jackrabbit.oak.segment.file.tar.TarFiles.lambda$init$0(TarFiles.java:425)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceTask.doLeaf(ReduceOps.java:952)
	at java.base/java.util.stream.ReduceOps$ReduceTask.doLeaf(ReduceOps.java:926)
	at java.base/java.util.stream.AbstractTask.compute(AbstractTask.java:327)
	at java.base/java.util.concurrent.CountedCompleter.exec(CountedCompleter.java:746)
	at java.base/java.util.concurrent.ForkJoinTask.doExec$$$capture(ForkJoinTask.java:290)

```

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
smiroslav: > Or you mean when someone manually deletes the blob in Azure? 

Another Oak process could do it. Alternatively, the segment may not have been uploaded due to the abrupt termination of the current process. Upon restarting the application, the archive might contain 0001 segment but not  0000 segment.  By filtering out based on the absence of segment 0000, the new process is unaware that the problem has happened.

> My concern is that not filtering out those archives anymore in AzureArchiveManager.listArchives() is a change in behavior.

Yes, like getting rid of the part that deletes the segments :). 
I can add filter logic back, and when we introduce a marker for the delete operation, we can do filtering based on that.

mreutegg (approved):

jsedding:
Wow, that's a huge change. And it addresses the issue only for "read-only" `SegmentArchiveManagers`.

Have you considered triggering an asynchronous delete instead? There _may_ be use-cases where we need to block until a delete is complete. But this is a cleanup from a previous, incomplete delete, or possibly another malfunction. Processing this asynchronously would move it off the critical path for all use-cases, and I dare say the API change that causes such a big changeset, would not be necessary.

smiroslav:
hi @jsedding , I noticed your comment only after merging the PR.

smiroslav:
> Have you considered triggering an asynchronous delete instead? 

Yes, I have, but decided to deal with that in a separate issue. 

> Processing this asynchronously would move it off the critical path for all use-cases


The primary concern was to prevent a read-only process from deleting the archive and not performance optimisation. Making any modification while retrieving the list of archives seems wrong in any case, especially when, for example, instantiating a read-only file store.

smiroslav:
After discussing with @jsedding , I have decided to revert this PR. Here I have propagated flag `readOnly` and had to touch many files. 
Also, the flag propagated to the segment tar implementation was not utilised. 

Instead of that, I could have used `WriteAccessController` in the archive manager, which is already present, to the same effect. 

Below is the new PR. @mreutegg @jsedding I would appreciate if you could have a look.
https://github.com/apache/jackrabbit-oak/pull/2480

PR #2469, opened 2025-08-25
OAK-11878 ClusterNodeStoreLock LOCK_TIMEOUT is too large

thomasmueller:


fabriziofortino (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/ClusterNodeStoreLock.java
```diff
@@ -38,10 +38,14 @@
...
-     * in between the import process which can take some time
+     * Use a long lease time to ensure that async indexer does not start
+     * in between the import process which can take some time.
+     * For a very large repository (about 2 billion nodes) the longest
+     * index import can take 2.5 hours. 6 hours has then a safety margin of 100%
+     * above that. A diff of 6 hours is typically OK with the document node store,
+     * but beyond that it gets harder.
      */
-    private static final long LOCK_TIMEOUT = TimeUnit.DAYS.toMillis(100);
+    private static final long LOCK_TIMEOUT = TimeUnit.HOURS.toMillis(6);
```
mkapalka, line 48: How about making this configurable (but still change the default) so that one can quickly override this if it causes issues?

mkapalka (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/ClusterNodeStoreLock.java
thomasmueller: There's also cost of making things configurable. And how could it be configured, and where would the configuration be?

6 hours timeout would mean there is an indexing job where the import alone takes more than 6 hours... I actually can't imagine such a case, because that would mean indexing takes multiple days. (The risk of the indexing process to die on it's own is very large in this case... but let's ignore that.) Let's assume there is such a case, eg. manually run, on a separate machine... In this case, the async indexing can be stopped.

So, for this extreme case, there are already two workarounds: 
* stop the async indexing.
* create a custom Oak version with increased timeout.

PR #2472, opened 2025-08-25
OAK-11875 : added unit cases for EmptyProps and split document issue

rishabhdaim:


rishabhdaim:
We are skipping the deletion of props in case the document has any empty split properties.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/VersionGarbageCollectorIT.java
```diff
@@ -129,6 +129,7 @@
 import org.jetbrains.annotations.NotNull;
 import org.junit.After;
 import org.junit.AfterClass;
+import org.junit.Assert;
```
Joscorbe, line 132: I think this import is not used

Joscorbe (approved):

PR #2487, opened 2025-09-02
OAK-11889 : replaced Guavas's Futures.successfulAsList with JDK comple…

rishabhdaim:
…tableFuture

### DIFF
--- a/oak-blob-plugins/src/test/java/org/apache/jackrabbit/oak/plugins/blob/FileCacheTest.java
+++ b/oak-blob-plugins/src/test/java/org/apache/jackrabbit/oak/plugins/blob/FileCacheTest.java
@@ -30,7 +30,7 @@ import org.apache.jackrabbit.guava.common.util.concurrent.MoreExecutors;
 import org.apache.jackrabbit.guava.common.util.concurrent.SettableFuture;
 import org.apache.jackrabbit.oak.commons.StringUtils;
 import org.apache.jackrabbit.oak.commons.concurrent.ExecutorCloser;
-import org.apache.jackrabbit.oak.commons.concurrent.FutureUtils;
+import org.apache.jackrabbit.oak.commons.internal.concurrent.FutureUtils;
 import org.apache.jackrabbit.oak.commons.internal.concurrent.FutureConverter;
 import org.apache.jackrabbit.oak.commons.pio.Closer;
 import org.junit.After;
--- a/oak-blob-plugins/src/test/java/org/apache/jackrabbit/oak/plugins/blob/FileCacheTest.java
+++ b/oak-blob-plugins/src/test/java/org/apache/jackrabbit/oak/plugins/blob/FileCacheTest.java
@@ -25,13 +25,13 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.io.FileUtils;
-import org.apache.jackrabbit.guava.common.util.concurrent.Futures;
-import org.apache.jackrabbit.guava.common.util.concurrent.ListenableFuture;
 import org.apache.jackrabbit.guava.common.util.concurrent.ListeningExecutorService;
 import org.apache.jackrabbit.guava.common.util.concurrent.MoreExecutors;
 import org.apache.jackrabbit.guava.common.util.concurrent.SettableFuture;
 import org.apache.jackrabbit.oak.commons.StringUtils;
 import org.apache.jackrabbit.oak.commons.concurrent.ExecutorCloser;
+import org.apache.jackrabbit.oak.commons.concurrent.FutureUtils;
+import org.apache.jackrabbit.oak.commons.internal.concurrent.FutureConverter;
 import org.apache.jackrabbit.oak.commons.pio.Closer;
 import org.junit.After;
 import org.junit.Before;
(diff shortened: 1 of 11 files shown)


oak-blob-plugins/src/test/java/org/apache/jackrabbit/oak/plugins/blob/FileCacheTest.java
```diff
@@ -25,13 +25,13 @@
...
 
 import org.apache.commons.io.FileUtils;
-import org.apache.jackrabbit.guava.common.util.concurrent.Futures;
-import org.apache.jackrabbit.guava.common.util.concurrent.ListenableFuture;
 import org.apache.jackrabbit.guava.common.util.concurrent.ListeningExecutorService;
 import org.apache.jackrabbit.guava.common.util.concurrent.MoreExecutors;
 import org.apache.jackrabbit.guava.common.util.concurrent.SettableFuture;
 import org.apache.jackrabbit.oak.commons.StringUtils;
 import org.apache.jackrabbit.oak.commons.concurrent.ExecutorCloser;
+import org.apache.jackrabbit.oak.commons.concurrent.FutureUtils;
```
reschke: shouldn't this go into `internal.concurrent` as well?

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/concurrent/FutureUtils.java
```diff
@@ -0,0 +1,60 @@
+/*
```
reschke: -> internal.concurrent?

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/concurrent/FutureUtils.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private FutureUtils() {
+        // no instance for you
+    }
+
+    /**
+     * Returns a {@link CompletableFuture} that completes when all of the given futures complete.
+     * <p>
+     * The resulting list contains the results of each input future, or {@code null} for any future that completed exceptionally.
+     * This mimics Guava's {@code Futures.successfulAsList} behavior.
```
reschke: that comment will fail when Guava is out of the classpath

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/concurrent/FutureUtils.java
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * @param futures the list of CompletableFutures to aggregate
+     * @param <T> the result type
+     * @return a CompletableFuture containing a list of results (or null for failed futures)
+     */
+    public static <T> CompletableFuture<List<T>> successfulAsList(final List<CompletableFuture<T>> futures) {
+
+        // need to handle each future case indivudually since if failed inside CompletableFuture.allOf(),
+        // it would throws an exception rather than returning null
```
reschke:
```suggestion
        // it would throw an exception rather than returning null
```

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/concurrent/FutureUtils.java
```diff
@@ -0,0 +1,60 @@
...
+        final List<CompletableFuture<T>> handled = futures.stream()
```
reschke: magic :-)

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/internal/concurrent/package-info.java
```diff
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-@Version("1.0.0")
+@Version("1.0.1")
```
reschke: do we need this change?

reschke (changes requested):

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/concurrent/FutureUtils.java
thomasmueller: I have a question: why do we need "...commons.internal.concurrent" and can't use only "org.apache.jackrabbit.oak.commons.concurrent."?

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/concurrent/FutureUtils.java
rishabhdaim: It is not linking to the Guava code, only adding it as a reference. So, I don't think it would fail.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/internal/concurrent/package-info.java
rishabhdaim: Let me revert it.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/concurrent/FutureUtils.java
reschke: Because we don't want to add more functionality to packages that are already public; API changes in these packages become very hard; in particular if we want to remove something (maybe because a newer JDK provides that functionality).

See https://issues.apache.org/jira/browse/OAK-11872

reschke (approved):

PR #2492, opened 2025-09-04
Revert "OAK-11224: shaded guava is only a test dependency in commons, adjust …"

rishabhdaim:
Reverts apache/jackrabbit-oak#2488


rishabhdaim:
Guava is still required due to https://github.com/apache/jackrabbit-oak/pull/2487

PR #2494, opened 2025-09-05
OAK-11894 resolve the IdentifierManager lazily

joerghoh:


oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/xml/ImporterImpl.java
reschke, line 1: Looks good to me.

There may be other places where we'd want to do that...

reschke (approved):

PR #2509, opened 2025-09-12
OAK-11899 - use default value in case config value can not be parsed

waldoro:
If the value in the OSGi console is cleared out, the value in the configuration file is stored with empty value :
propName=""
This makes issue that the empty string is not a null (then the default value would be used) and throws an parse exception which causes that the token is not correctly generated. 
The default value is used only when the propName is not present.


Amoratinos (approved):
Looks good to me

reschke (commented):
2 cents:

- is this really the only place with this problem?
- maybe it would be better to extend the ConfigurationParameters API instead? (that would also avoid having to catch exceptions)

Amoratinos (commented):
What's the point to have 2 different methods where both receives a default value but only one of them has the `OrDefault` suffix. Looks kind of a duplication. wdyt @reschke

reschke (approved):

PR #2510, opened 2025-09-15
OAK-11929: NullPointerException in SecureSortedSetDocValuesFacetCounts

ChlineSaurus:
This PR resolves a NullPointerException in the SecureSortedSetDocValuesFacetCounts class which occurs when a document with facets is inaccessible, but not in the topN results. 

The class counts the facets of the topN results and afterwards reduces the facet counts of the inaccessible documents. However, if a document wasn't in the topN in the first place, it tries removing facets from a document that was not considered when counting, which is therefore not in the hash map, which therefore throws a NullPointerException.

Fix: A simple Null-Check resolves the problem, as the facet counts doesn't have to be reduced, as it was not considered for those documents in the first place.


nfsantos (approved):

fabriziofortino (approved):
Fix looks good to me.
I would also suggest to add a test in FacetCommonTest. These tests are executed against both Lucene and Elastic. In this way we guarantee the issue is fixed everywhere.

ChlineSaurus:
Moved both tests to FacetCommonTest. This also reduces duplication in setUp code for the tests

thomasmueller (approved):

PR #2511, opened 2025-09-15
OAK-11930 prevent binaries from being read and returned as string

joerghoh:
if ``event.toString()`` is invoked (e.g. during logging), do not serialize the binary values as part of the returned string, but use something less memory-intensive.

The question remains if ``event.getInfo().get("beforeValue")`` or ``event.getInfo().get("afterValue")`` is used and useful as it is implemented today ... but at least that won't happen accidentally.

### DIFF
--- a/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/observation/EventFactory.java
+++ b/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/observation/EventFactory.java
@@ -107,7 +107,7 @@ public class EventFactory {
                 return Collections.unmodifiableMap(builder);
             }
             @Override
-            public Map<?,?> getInfoToString() {
+            public Map<?, ?> getInfoToString() {
                 if (!after.getType().equals(Type.BINARY)) {
                     return getInfo();
                 }
@@ -138,7 +138,7 @@ public class EventFactory {
                 return Collections.unmodifiableMap(builder);
             }
             @Override
-            public Map<?,?> getInfoToString() {
+            public Map<?, ?> getInfoToString() {
                 if (!before.getType().equals(Type.BINARY)) {
                     return getInfo();
                 }
@@ -168,7 +168,7 @@ public class EventFactory {
                 return Collections.unmodifiableMap(builder);
             }
             @Override
-            public Map<?,?> getInfoToString() {
+            public Map<?, ?> getInfoToString() {
                 if (!before.getType().equals(Type.BINARY)) {
                     return getInfo();
                 }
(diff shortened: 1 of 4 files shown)


anchela (approved):
@joerghoh , i am not the original author of the event machinery and thus cannot assess if that change is prone to any regression as people might have relied on the original behavior in some strange way.
having said that: the proposed change looks reasonable to well done to me.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/observation/EventFactory.java
```diff
@@ -298,6 +334,17 @@ public boolean isExternal() {
 
         //--------------------------------------------------------< Object >--
 
+        /** A custom version of getInfo(), which is only used
+         * for the toString() implementation; the goal is to limit the amount
+         * of data, which is returned, no binary data should be logged
```
reschke:
```suggestion
         * of data that is returned; no binary data should be logged.
```

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/observation/EventFactory.java
```diff
@@ -101,6 +103,17 @@ public int getType() {
                 builder.put("afterValue", createValue(after));
                 return Collections.unmodifiableMap(builder);
             }
+            @Override
+            public Map<?,?> getInfo_ToString() {
```
reschke: I was wondering whether we can make all these maps String-keyd; but maybe that would be a different ticket. (see Sonar warnings)

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/observation/EventFactory.java
```diff
@@ -298,6 +334,17 @@ public boolean isExternal() {
 
         //--------------------------------------------------------< Object >--
 
+        /** A custom version of getInfo(), which is only used
+         * for the toString() implementation; the goal is to limit the amount
+         * of data, which is returned, no binary data should be logged
+         * This is not included in hash calculation and equality checks.
+         * 
+         * @return
```
reschke: ...returns what?

reschke (commented):
a few nits.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/observation/EventFactory.java
joerghoh: indeed, especially that more or less directly wires through to Event.getInfo(), which is like this (thank you JCR 1.0 where generics were not available yet):

```
Map getInfo();
```

PR #2513, opened 2025-09-16
 OAK-11932: Segment prefetching for CachingSegmentReader

smiroslav:
https://issues.apache.org/jira/browse/OAK-11932

### DIFF
--- a/oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/CachingSegmentArchiveReader.java
+++ b/oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/CachingSegmentArchiveReader.java
@@ -19,16 +19,29 @@
 package org.apache.jackrabbit.oak.segment.spi.persistence.persistentcache;
 
 import org.apache.jackrabbit.oak.commons.Buffer;
+import org.apache.jackrabbit.oak.commons.concurrent.ExecutorCloser;
 import org.apache.jackrabbit.oak.segment.file.tar.SegmentGraph;
 import org.apache.jackrabbit.oak.segment.spi.persistence.SegmentArchiveEntry;
 import org.apache.jackrabbit.oak.segment.spi.persistence.SegmentArchiveReader;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.util.Collections;
 import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 public class CachingSegmentArchiveReader implements SegmentArchiveReader {
+
+    private static final Logger LOG = LoggerFactory.getLogger(CachingSegmentArchiveReader.class);
+
 
     @NotNull
     private final PersistentCache persistentCache;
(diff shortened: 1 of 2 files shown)


oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/CachingSegmentArchiveReader.java
```diff
@@ -19,34 +19,123 @@
...
+        var data = SegmentData.newSegmentData(buffer);
+        int refs = data.getSegmentReferencesCount();
+        ArrayList<UUID> out = new ArrayList<>(refs);
+        for (int i = 0; i < refs; i++) {
+            out.add(new UUID(data.getSegmentReferenceMsb(i), data.getSegmentReferenceLsb(i)));
+        }
+        return out;
+    }
+
+    private void schedulePrefetch(long msb, long lsb, Buffer buffer) {
```
smiroslav, line 93: @nfsantos made a good remark that this method, from the start, should be executed in a separate thread context, so it does not impact execution time for the thread invoking CachingSegmentArchiveReader#readSegment

jsedding: Indeed, my tests with #2519, which implements a similar mechanism, but on a different level of abstraction, showed that dispatching from the caller thread hurts performance.

I even opted to trigger preloading only from within the load-callback, so preloading is completely off the critical path for cache hits.

Of course that's a trade-off, as it requires the caller thread to load one segment before any preloading happens.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/CachingSegmentArchiveReader.java
```diff
@@ -19,34 +19,123 @@
...
+                final long rMsb = ref.getMostSignificantBits();
+                final long rLsb = ref.getLeastSignificantBits();
+
+                // Skip if already present in cache
+                if (persistentCache.containsSegment(rMsb, rLsb)) {
+                    continue;
+                }
+
+                // Drop prefetch if already in progress for this segment
+                boolean registered = inFlightPrefetch.add(ref);
```
jsedding, line 108: Nice idea! I missed this trick so far in #2519 🙂

jsedding (commented):
Looks good! I think some benchmarking is needed to ensure the added overhead is worthwhile.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/CachingSegmentArchiveReader.java
```diff
@@ -19,34 +19,123 @@
...
+            out.add(new UUID(data.getSegmentReferenceMsb(i), data.getSegmentReferenceLsb(i)));
+        }
+        return out;
+    }
+
+    private void schedulePrefetch(long msb, long lsb, Buffer buffer) {
+        try {
+            List<UUID> refs = extractReferences(buffer);
+            int limit = Math.min(refs.size(), prefetchMaxRefs);
+            for (int i = 0; i < limit; i++) {
```
nfsantos, line 97: You are getting a list with all the references but then potentially iterate over only a subset of them. You could save some work by extracting only the references that will be prefetched, this could also avoid allocating an array with all the refs. Using streams with `limit()`may be an easy way of implementing this optimization, as streams are computed lazily.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/CachingSegmentArchiveReader.java
```diff
@@ -19,34 +19,123 @@
...
+                final UUID ref = refs.get(i);
+                final long rMsb = ref.getMostSignificantBits();
+                final long rLsb = ref.getLeastSignificantBits();
+
+                // Skip if already present in cache
+                if (persistentCache.containsSegment(rMsb, rLsb)) {
+                    continue;
+                }
```
nfsantos, line 105: Maybe it's better to do this check in the worker thread, just before the segment is downloaded? The task that is scheduled to download the segment might not execute for a while, if the worker pool is busy, so from this point until the actual download, the segment might have been added to the cache. Or we can leave the check here and do another one before trying to download. 
It would also be better to have a mechanism similar to the Guava loading cache, where thread requesting a segment that is not in the cache but is being downloaded will block waiting for the first download to complete, instead of starting a new download. This would avoid duplicate downloads.

jsedding:
As indicated in #2569, this approach is limited in that a `CachingSegmentArchiveReader` has only got access to a single archive, and thus cannot read segments from other archives.

PR #2534, opened 2025-09-23
OAK-11944 : upgraded aws-java-sdk-s3 version to 1.12.791

rishabhdaim:


reschke (approved):

reschke (commented):
Does this *add* exports, or just sets version numbers?

(A related question would be *why* we are exporting those, which will be relevant when we update to the new JDK).

rishabhdaim:
> Does this _add_ exports, or just sets version numbers?
> 
> (A related question would be _why_ we are exporting those, which will be relevant when we update to the new JDK).

This only sets the version numbers as suggested by the baseline check. Without those, the build wouldn't succeed.
This export was added by Alvaro Dias in OAK-8827, when he added support for AWS in this bundle.

reschke (approved):

PR #2540, opened 2025-09-25
OAK-11949: Sort union queries without "order-by" by score

ChlineSaurus:
Union queries without a "order by" clause so far simply concatenate the left and the right query to one another. With this PR the behavior changes to merge the two queries based on their jcr:scores. Since this changes the query behavior, and therefore could cause issues with backwards compatibility for users/customers, we placed this feature behind a feature toggle which is disabled by default. 

Issue: https://issues.apache.org/jira/projects/OAK/issues/OAK-11949?filter=allopenissues

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/UnionQueryImpl.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/UnionQueryImpl.java
@@ -313,7 +313,7 @@ public class UnionQueryImpl implements Query {
         FacetMerger facetMerger = new FacetMerger(left, right);
 
         Iterator<ResultRowImpl> it;
-        final Iterator<ResultRowImpl> leftRows = facetMerger.getLeftIterator();;
+        final Iterator<ResultRowImpl> leftRows = facetMerger.getLeftIterator();
         final Iterator<ResultRowImpl> rightRows = facetMerger.getRightIterator();
         Iterator<ResultRowImpl> leftIter = leftRows;
         Iterator<ResultRowImpl> rightIter = rightRows;
@@ -324,7 +324,21 @@ public class UnionQueryImpl implements Query {
             rightIter = ((MeasuringIterator) rightRows).getDelegate();
         }
         if (orderBy == null) {
-            it = IteratorUtils.chainedIterator(leftIter, rightIter);
+            boolean sortByScoreEnabled = settings.isSortUnionQueryByScoreEnabled();
+            boolean leftHasScore = isScorePresent(left);
+            boolean rightHasScore = isScorePresent(right);
+
+            if (sortByScoreEnabled && leftHasScore && rightHasScore) {
+                // Both sides have scores => merge
+                Comparator<ResultRowImpl> scoreComparator = createScoreBasedComparator();
+                it = IteratorUtils.mergeSorted(List.of(leftIter, rightIter), scoreComparator);
+            } else if (sortByScoreEnabled && rightHasScore) {
+                // Only right has score => start with right
+                it = IteratorUtils.chainedIterator(rightIter, leftIter);
+            } else {
+                // Default old behavior, only left has score, or neither has scores => start with left
+                it = IteratorUtils.chainedIterator(leftIter, rightIter);
+            }
         } else {
             // This would suggest either the sub queries are sorted by index or explicitly by QueryImpl (in case of traversing index)
             // So use mergeSorted here.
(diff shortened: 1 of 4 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/query/UnionQueryImpl.java
```diff
@@ -324,7 +324,21 @@ public Iterator<ResultRowImpl> getRows() {
...
+                // Both sides have scores => merge
+                Comparator<ResultRowImpl> scoreComparator = createScoreBasedComparator();
+                it = IteratorUtils.mergeSorted(List.of(leftIter, rightIter), scoreComparator);
+            } else if (sortByScoreEnabled && rightHasScore) {
+                // Only right has score => start with right
+                it = IteratorUtils.chainedIterator(rightIter, leftIter);
+            } else {
+                // Default old behavior, only left has score, or neither has scores => start with left
+                it = IteratorUtils.chainedIterator(leftIter, rightIter);
+            }
```
thomasmueller: I usually start with the case where the feature is disabled, so that no new code (that might have some bugs / forgotten edge cases) can cause issues; something like this:

```suggestion
            boolean sortByScoreEnabled = settings.isSortUnionQueryByScoreEnabled();
            if (!sortByScoreEnabled) {
                it = IteratorUtils.chainedIterator(leftIter, rightIter);
            } else {
                boolean leftHasScore = isScorePresent(left);
                boolean rightHasScore = isScorePresent(right);
                if (leftHasScore && rightHasScore) {
                    // Both sides have scores => merge
                    Comparator<ResultRowImpl> scoreComparator = createScoreBasedComparator();
                    it = IteratorUtils.mergeSorted(List.of(leftIter, rightIter), scoreComparator);
                } else if (rightHasScore) {
                    // Only right has score => start with right
                    it = IteratorUtils.chainedIterator(rightIter, leftIter);
                } else {
                    // only left has score, or neither has scores => start with left
                    it = IteratorUtils.chainedIterator(leftIter, rightIter);
                }
            }
```

thomasmueller (commented):
Looks very good to me, just one recommendation.

oak-core/src/test/java/org/apache/jackrabbit/oak/query/UnionQueryTest.java
```diff
@@ -54,6 +58,7 @@
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mockito;
+import org.apache.jackrabbit.oak.spi.toggle.Feature;
```
reschke: This import seems to be in the wrong place.

reschke:
Do we have test coverage for both sides of the feature toggle (did not check)?

Note that we don't have Sonar coverage from PRs coming out of Apache Jackrabbit space. For anything non-trivial, it's good to first merge into an Oak branch.

ChlineSaurus:
> Do we have test coverage for both sides of the feature toggle (did not check)?

The PR has coverage for the feature flag enabled and disabled. Furthermore, I now also added one more test for the case that the feature flag is null.

ChlineSaurus:
> Note that we don't have Sonar coverage from PRs coming out of Apache Jackrabbit space. For anything non-trivial, it's good to first merge into an Oak branch.

How do you mean this? Should I create a new branch inside the original Oak-Repo and merge it into that branch?
Something similar to those branches?
https://github.com/apache/jackrabbit-oak/tree/OAK-11787
https://github.com/apache/jackrabbit-oak/tree/OAK-11785

fabriziofortino (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/query/UnionQueryImpl.java
```diff
@@ -527,4 +544,41 @@ public Optional<Long> getLimit() {
...
+        };
+    }
+
+    /**
+     * @param row the result row
+     * @return the jcr:score as a Double
+     * Precondition: {@link #isScorePresent(Query)} must be true. If the row lacks a jcr:score, 0.0 is returned and
+     * the issue is logged.
+     */
+    private Double getScoreFromRow(ResultRowImpl row) {
```
thomasmueller: Nitpick: Double can be "null" which is not what we want here.

```suggestion
    private double getScoreFromRow(ResultRowImpl row) {
```

thomasmueller (changes requested):

thomasmueller (approved):

PR #2560, opened 2025-10-06
OAK-11969 do not check the existence of the tree twice

joerghoh:
``tree.getChildren()`` returns only accessible tree objects, for that these tree objects always exist, and there is no need to check for their existence here as well.

on the implementation side ``AbstractTree.getChildren()`` already performs these checks for existence of the child nodes, so this check is really redundant.


reschke (approved):
Change LGTM, but I haven't thought about the semantics. If it passes tests it should be good.

anchela (approved):
hi @joerghoh , i verified that the check for tree.exists() is already executed in Tree.getChildren and thus is superfluous here.
good catch!

PR #2519, opened 2025-09-17
OAK-11934 - segment prefetching for segmentstore cache

jsedding:

PR #2569, opened 2025-10-09
OAK-11934 - segment preloading for PersistentCache

jsedding:


jsedding:
@smiroslav I have experimented with various approaches to the cache preloading mechanism.

First, I dismissed my initial approach in #2519 to preload segments into the in-memory `SegmentCache`, because it busts the cache and there is no good way to know if a segment was already cached in a `PersistentCache` (which is the main objective).

I also dismissed the approach you took in #2513, where you implement preloading in the `CachingSegmentArchiveReader`, because the reader is bound to a single archive. This means that the preload mechanism cannot load referenced segments from other archives, which makes it a lot less useful.

The approach implemented in this PR allows adding a `PersistentCache` instance to a `*FileStore` via the `FileStoreBuilder`. When this is done, prefetching can additionally be configured (number of async threads and "depth", i.e. how many levels of referenced segments to follow). With this approach, the `SegmentPreloader` has access to the `PersistentCache`, which allows it to check if a segment is already present. It also has access to the `TarFiles` object, which allows it to read segments from _any_ archive, and it can retrieve the segment graph from the `TarFiles` as well, which avoids having to read the references from the segment's data.

PR #2572, opened 2025-10-10
OAK-11979: Clean up *StringValue in JCR XML import

reschke:


reschke:
Seems we have mo coverage for Base64 encoded binaries.

rishabhdaim (commented):
Can we create unit cases for these files, just to compare the output before & after.

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/xml/StringValue.java
```diff
@@ -63,12 +61,11 @@ public String getString() {
     @Override @SuppressWarnings("deprecation")
     public Value getValue(int type) throws RepositoryException {
         if (type == PropertyType.BINARY) {
-            ByteArrayOutputStream baos = new ByteArrayOutputStream();
             try {
-                Base64.decode(value, baos);
+                byte[] bytes = Base64.getDecoder().decode(value);
```
mbaedke, line 65: Maybe inline that?

mbaedke (approved):

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/xml/StringValue.java
reschke: inline?

reschke:
Before we proceed here, we need test coverage. Optimally separate ticket / PR.

PR #2573, opened 2025-10-13
OAK-11914 : removed usage of Guava's HashBasedTable

rishabhdaim:


oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/binary/fixtures/nodestore/DocumentMemoryNodeStoreFixture.java
```diff
@@ -87,11 +86,11 @@ public NodeStore createNodeStore() {
 
             // track all main components
             if (dataStore != null) {
-                components.put(nodeStore, DataStore.class.getName(), dataStore);
-                components.put(nodeStore, DataStore.class.getName() + ":folder", dataStoreFolder);
+                components.computeIfAbsent(nodeStore, k -> new LinkedHashMap<>()).put(DataStore.class.getName(), dataStore);
+                components.computeIfAbsent(nodeStore, k -> new LinkedHashMap<>()).put(DataStore.class.getName() + ":folder", dataStoreFolder);
             }
```
mbaedke, line 92: I'm confused. How are these equivalent? The old code sets two different cells in the same row, the new code sets a row and then the second call is a noop because the row is already there, right?

rishabhdaim: No, the new code works as follows:

`components.computeIfAbsent(nodeStore, k -> new LinkedHashMap<>())`

In case the value associated with `nodeStore` key is null, then create an empty `LinkedHashMap`

and then `.put(DataStore.class.getName(), dataStore);`

would put above values into that newly created `LinkedHashMap`.

So essentially it is creating an store which is equivalent to above like: `nodeStore` is row name and values added into that are cells.

mbaedke: Yes. My comment is about the second call, which then does nothing at all, because the value has just been set.

rishabhdaim: Internal implementation of `HashBasedTable.create`

`public static <R, C, V> HashBasedTable<R, C, V> create() {
    return new HashBasedTable<>(new LinkedHashMap<R, Map<C, V>>(), new Factory<C, V>(0));
  }`

rishabhdaim: > Yes. My comment is about the second call, which then does nothing at all, because the value has just been set.

Ahh, got it. Yes, that can we removed.

thomasmueller: I think this should work as well:

```
                components.computeIfAbsent(nodeStore, k -> new LinkedHashMap<>()).put(DataStore.class.getName(), dataStore);
                components.get(nodeStore).put(DataStore.class.getName() + ":folder", dataStoreFolder);
```

because the first call will add a `LinkedHashMap` - but I understand why you didn't want to do that.

mbaedke: As the value for the key nodestore, you want to create one LinkedHashMap with two values. That does not happen.

thomasmueller (approved):
It looks correct to me

oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/binary/fixtures/nodestore/DocumentMemoryNodeStoreFixture.java
thomasmueller: I think it does work: the first call
`components.computeIfAbsent(nodeStore, k -> new LinkedHashMap<>())` 
will add a LinkedHashMap (because it is absent) and the second call will re-use it...

mbaedke: I also think that computeIfAbsent isn't the right choice, because in case the value has been already set, you can't override it.

mbaedke: Never mind, misread the code. All good. Sorry for the noise.

rishabhdaim: thanks for review, but I have made it simpler this time with https://github.com/apache/jackrabbit-oak/pull/2573/commits/6b4d5c920c78c1a0d87b95a48999246c04deda37

rishabhdaim: > I think it does work: the first call `components.computeIfAbsent(nodeStore, k -> new LinkedHashMap<>())` will add a LinkedHashMap (because it is absent) and the second call will re-use it...

Yes, that was the intention, but since it was confusing so I have updated the code.

reschke (approved):
I'll trust @mbaedke and @thomasmueller

mbaedke (approved):

PR #2574, opened 2025-10-13
OAK-11980 - improve parallelization of I/O during segment-azure initialization

jsedding:


oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
```diff
@@ -91,16 +94,21 @@ public List<String> listArchives() throws IOException {
...
+                            .parallel()
+                            .filter(this::deleteInProgress)
+                            .peek(archiveName -> {
+                                if (writeAccessController.isWritingAllowed()) {
+                                    delete(archiveName);
+                                }
+                            })
+                            .collect(Collectors.toUnmodifiableSet()));
+
+            archiveNames.removeAll(archivesToDelete);
```
nfsantos: What about something like the snippet below? It avoids calling peek, which I feel is not intended for this type of situations. And it also checks only once if writing is allowed. The code is likely not syntactically valid, I did not check it on a IDE.

```suggestion
                    () -> archiveNames.stream()
                            .parallel() // Maybe not needed if deleteInProgress is fast?
                            .filter(this::deleteInProgress)
                            .collect(Collectors.toUnmodifiableSet()));

             if (writeAccessController.isWritingAllowed()) {
                  archiveNames.stream().parallel().forEach(archiveName -> delete(archiveName));
             }
            archiveNames.removeAll(archivesToDelete);
```

jsedding: Thanks, @nfsantos.  I addressed this in 424b93d. Does that work for you?

nfsantos: Yes, looks good, cleaner.

jsedding: Agreed, thanks!

PR #2578, opened 2025-10-13
Upgrade azure sdk 12

seropian:


reschke:
Note to committers: please do not merge into trunk without getting Sonar results first - and that requires merging into a branch over here.

PR #2581, opened 2025-10-14
OAK-11984 Support UserId Change for External Users

nscendoni:


oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/ExternalLoginModule.java
```diff
@@ -228,7 +230,22 @@ public boolean login() throws LoginException {
             // before into the repository.
             UserManager userManager = getUserManager();
             SyncedIdentity sId = getSyncedIdentity(userId, userManager);
-
+            if (sId ==null && userManager != null && creds != null) {
```
Amoratinos: Trivial typo

```suggestion
            if (sId == null && userManager != null && creds != null) {
```

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/token/TokenConstants.java
```diff
@@ -34,11 +35,15 @@ public interface TokenConstants {
...
     String TOKEN_NT_NAME = "rep:Token";
+    
+    String EXTERNAL_ID_ATTRIBUTE = ":externalId";
 
     Set<String> RESERVED_ATTRIBUTES = Set.of(
             TOKEN_ATTRIBUTE,
             TOKEN_ATTRIBUTE_EXPIRY,
-            TOKEN_ATTRIBUTE_KEY);
+            TOKEN_ATTRIBUTE_KEY,
+            AbstractLoginModule.SHARED_KEY_LOGIN_NAME,
```
Amoratinos: Looks a bit odd to me that we're saving an attribute from other class. Wouldn't be better to specify here the attribute name?

If we do not want to remove the constant from `AbstractLoginModule` it can be initialized with the TokenConstant value:

```
public static final String SHARED_KEY_LOGIN_NAME = TokenConstants.SHARED_KEY_LOGIN_NAME;
```

oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/ExternalLoginModule.java
```diff
@@ -228,7 +230,22 @@ public boolean login() throws LoginException {
...
+                Object externalAttribute = credentialsSupport.getAttributes(creds).get(ExternalIdentityConstants.EXTERNAL_ID_ATTRIBUTE);
+                if (externalAttribute != null ) {
+                    @NotNull Iterator<Authorizable> authIterator = userManager.findAuthorizables(ExternalIdentityConstants.REP_EXTERNAL_ID, externalAttribute + ";" + idp.getName(), UserManager.SEARCH_TYPE_USER);
+                    if (authIterator.hasNext()) {
+                        //modify credentials to reflect the login name stored in oak
+                        Authorizable authorizable = authIterator.next();
+                        sId = getSyncedIdentity(authorizable.getID(), userManager);
+                        Map<String, ?> attributes = credentialsSupport.getAttributes(creds);
+                        HashMap<String, Object> newAttributes = new HashMap<>(attributes);
+                        newAttributes.put(AbstractLoginModule.SHARED_KEY_LOGIN_NAME, authorizable.getID());
```
Amoratinos: This class already inherits from `AbstractLoginModule` so it's not needed to add it while using the constant.

```suggestion
                        newAttributes.put(SHARED_KEY_LOGIN_NAME, authorizable.getID());
```

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/token/TokenConstants.java
```diff
@@ -34,11 +35,15 @@ public interface TokenConstants {
     String TOKENS_NT_NAME = NodeTypeConstants.NT_REP_UNSTRUCTURED;
 
     String TOKEN_NT_NAME = "rep:Token";
+    
+    String EXTERNAL_ID_ATTRIBUTE = ":externalId";
```
anchela: this constant is now duplicated here and in the externalidentity constants. i wonder if it would make sense to place it in a single common place, either in security-spi or in oak-jackrabbit-api. 
we could also add the 'attribute-login-name' there and then reuse it in AbstractLoginModule and here in TokenConstants.Reserved-Attributes.
wdyt?

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/token/TokenConstants.java
anchela: see above for a suggestion that would make both new constants generic and not have abstractloginmodule, external-identity-constants and tokenconstants depend on each other or duplicate constants.

oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/ExternalLoginModule.java
```diff
@@ -228,7 +230,22 @@ public boolean login() throws LoginException {
             // before into the repository.
             UserManager userManager = getUserManager();
             SyncedIdentity sId = getSyncedIdentity(userId, userManager);
-
+            if (sId ==null && userManager != null && creds != null) {
+                // Check if the external user was registered with a different userId, and the same externalId
```
anchela, line 237: imho it would make sense to add a tmp toggle for this change and clean it up if we see that it doesn't cause any trouble.
if i remember correctly we had to do this with system-properties in the past. let's check with the rest of the oak-folks if there a better alternative available now

anchela (commented):
just minor remarks

oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/ExternalLoginModule.java
nscendoni: I agree to have a toggle here. We can have a configuration parameter to enable/disable this feature.

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/token/TokenConstants.java
anchela: @Amoratinos , @nscendoni  i reshuffled the constants a bit. please take a look if you are good with that.

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/token/TokenConstants.java
anchela: see above.... i reshuffled the constants a bit.

oak-security-spi/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/token/TokenConstants.java
anchela: btw: i attemted to replace the constants in AbstractLoginModule but this resulted in a major version increase if making AbstractLoginModule implement the new constants

oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/ExternalLoginModule.java
nscendoni: I created a FeatureFlag and added test for it.

anchela (approved):
lgtm

anchela:
@Amoratinos , can you review again with the latest changes made by @nscendoni and myself?

Amoratinos (approved):
Looks better now with the introduction of the AuthenticationConstants 👍 LTGM

PR #2585, opened 2025-10-15
OAK-11982 move the calculation of the WARN limit out of the constructor

joerghoh:


anchela (approved):
hi @joerghoh , i had a look and didn't spot anything to nitpick :)

PR #2604, opened 2025-11-03
OAK-11991 Optimize the oak-segment recovery process

ierandra:


oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
```diff
@@ -336,6 +331,73 @@ private void copyBlob(BlobItem blob, String newParent) throws IOException {
...
+            try {
+                BlockBlobClient blobClient = readBlobContainerClient.getBlobClient(blob.getName()).getBlockBlobClient();
+
+                BlockBlobClient destinationBlobClient = writeBlobContainerClient.getBlobClient(destinationBlob).getBlockBlobClient();
+
+                SyncPoller<BlobCopyInfo, Void> copy = destinationBlobClient.beginCopy(blobClient.getBlobUrl(), null);
+
+                copyBlobs.add(new CopyBlob(copy, destinationBlob));
+            } catch (Exception e) {
+                log.error("Failed to start copying of blob {} to {}", blob.getName(), destinationBlob, e);
```
smiroslav, line 373: The issue here is that if the copy fails, it will not in any way disturb the caller. 
The original code was also not better.
How about throwing IOException?  The method ```backup``` has it in its signature.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
```diff
@@ -336,6 +331,73 @@ private void copyBlob(BlobItem blob, String newParent) throws IOException {
 
     }
 
+    private void batchCopyBlobs(List<BlobItem> from, String to) {
+        String newParent = getDirectory(to);
+
+        log.info("Start tp copy {} blobs to {}", from.size(), newParent);
```
smiroslav:
```suggestion
        log.info("Start to copy {} blobs to {}", from.size(), newParent);
```

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
```diff
@@ -336,6 +331,73 @@ private void copyBlob(BlobItem blob, String newParent) throws IOException {
...
+
+    private void processBeginCopy(List<CopyBlob> copyBlobs) {
+        for (CopyBlob copy : copyBlobs) {
+            try {
+                CopyStatusType statusType = readBlobContainerClient.getBlobClient(copy.blobName).getBlockBlobClient().getProperties().getCopyStatus();
+                if (statusType == CopyStatusType.PENDING) {
+                    statusType = copy.poller.waitForCompletion().getValue().getCopyStatus();
+                }
+                if (statusType != CopyStatusType.SUCCESS) {
+                    log.warn("Failed to copy blob {}, status {}", copy.blobName, statusType.toString());
```
smiroslav, line 388: Like the previous comment, in case of an issue, the caller won't be notified.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
```diff
@@ -336,6 +331,73 @@ private void copyBlob(BlobItem blob, String newParent) throws IOException {
...
+        for (int i = 0; i < batches; i++) {
+            log.info("Start batch {}/{}: {} to {}", i + 1, batches, start, end);
+            List<BlobItem> blobItemsBatch = from.subList(start, end);
+            copyBlobs(blobItemsBatch, newParent);
+            start = end;
+            end += COPY_BATCH;
+        }
+
+        log.info("Copy {} to {}", start, from.size());
+        copyBlobs(from.subList(start, from.size()), newParent);
```
smiroslav: This call is not necessary when blob count is an exact multiple of COPY_BATCH (e.g., 2000).  Code makes an unnecessary call with an empty list. 

When this method is called, if we have a remainder, it will not be displayed in the logs. 
For example, if the list is 2500 blobs, we will see in the log "Start batch 2/2 ..." (from the for loop), but the method will be invoked once more,

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
ierandra: I think we should open a new task and add details why do we want to change this behavior. I the existing code the IOException is not thrown.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
ierandra: I think we should open a new task and add details why do we want to change this behavior. I the existing code the IOException is not thrown.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java
ierandra: >This call is not necessary when blob count is an exact multiple of COPY_BATCH (e.g., 2000). Code makes an unnecessary call with an empty list.

for this I can add and if like 
```
if (start < from.size()) {
  log.info("Copy {} to {}", start, from.size());
  copyBlobs(from.subList(start, from.size()), newParent);
}
```
this way will not print the log and will not do the unnecessary call with an empty list.

> For example, if the list is 2500 blobs, we will see in the log "Start batch 2/2 ..." (from the for loop), but the method will be invoked once more,

yes, this is how I wanted to be. we have 2 batches of 1000 and then copy the remaining (in this case the 500).
I was thinking to set the log like 
```log.info("Copy remaining  {} to {}", start, from.size());```
but this is not ok because we can have the situation where we have less then 1000 blobs and the batch `for` loop will not be executed.

File: oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java (hunk 1 above)
Comment (smiroslav) [reply oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java:373]: Yes, makes sense.

File: oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java (hunk 4 above)
Comment (smiroslav) [reply oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java]: Yes, that change would work if we add the size of the remaining batch to the log.

Comment (ierandra) [reply oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java]: changed the logic

File: oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java (hunk 2 above)
Comment (ierandra) [reply oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java]: fixed

File: oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureArchiveManager.java [hunk 5]
```diff
@@ -61,6 +62,12 @@ public class AzureArchiveManager implements SegmentArchiveManager {
 
     private static final String CLOSED_ARCHIVE_MARKER = "closed";
 
+    public static final String COPY_BATCH_SIZE_PROP = "segment.azure.batch.copy.size";
+
+    private static final int DEFAULT_COPY_BATCH = 1000;
+
+    private static final int COPY_BATCH = Integer.getInteger(COPY_BATCH_SIZE_PROP, DEFAULT_COPY_BATCH);
```
smiroslav: I would suggest removing `final` here. It might be hard to manipulate its value in tests, since, as it is now, it will be initialized when the class is loaded and cannot be changed.

ierandra: you mean `static` not `final` right? then I agree, I will remove

smiroslav: Yes, sorry, I indeed meant ```static```.

ierandra: done

smiroslav (approved):

nfsantos:
I'm just starting to explore the segment store, so I may be missing the full context, but while looking at this PR and at the recovery process, I think the logic can be further improved. IIUC, the current logic will process a batch of blobs, wait until it is fully processed, then advance to the next one. This will cause a burst of quick requests, followed by a slow down until the last elements of the current batch are processed. Any stragglers in a batch will slow down the total time taken by the batch.

One possible alternative is to instead of batching, we keep a sliding window of requests that are being processed. We could keep a count of how many ongoing requests are being processed and launch requests until reaching this value. When we reach this limit, wait until one request completes and then launch the next one. And repeat until all blobs are processed. This way, as soon as the earliest request is processed, we can launch a new one. 

Ideally, we should wait for all sync pollers at the same time, to send a new request as soon as the fastest one completes, but I'm not sure it is possible to do this without launching a thread for each sync poller, which would add a lot of complexity, for potentially small gains. Or does Azure support callbacks when a request is completed?

Anyway, the code looks good as is, but I just wanted to share this idea. Maybe it is worth considering in the future.

ierandra:
@nfsantos thanks for sharing the idea. I added in https://issues.apache.org/jira/browse/OAK-11991 (comments section) why I chose this option. 
In addition to what I mentioned there: in practice we never wait for blobs to be copied because by the time we finish the start copy batch and checking the first blob the copy operation is already done.
Also, based on my investigation the solution you proposed is not more eficient because adds more read-write interleave
and "when you interleave PUTs with HEADs (reads), Azure’s load balancers and caching layers can’t optimize as effectively — you’re forcing read-after-write consistency checks repeatedly.
Azure is more efficient when doing “pure” PUT operations (writes) or pure HEAD operations (reads) in bulk, rather than interleaving them."

PR #2607, opened 2025-11-05
OAK-12004 Datastore: speedup datastore copy

thomasmueller:
* Increasing the number of workers in the executor service by a factor of 4-8 (using two executor services, and each a factor of 4).
* When large files are downloaded you use multiple threads and range queries.

### DIFF
--- a/oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCommand.java
+++ b/oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCommand.java
@@ -698,9 +698,11 @@ public class DataStoreCommand implements Command {
         if (hashIndex < 0) {
             return 0;
         }
+        String lengthString = blobId.substring(hashIndex + 1);
         try {
-            return Long.parseLong(blobId.substring(hashIndex + 1));
+            return Long.parseLong(lengthString);
         } catch (NumberFormatException e) {
+            log.warn("Can not parse length for blob id {}", blobId);
             return 0;
         }
     }
--- a/oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCommand.java
+++ b/oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCommand.java
@@ -644,12 +644,8 @@ public class DataStoreCommand implements Command {
 
             String id = list.get(0);
             // Split b47b58169f121822cd4a0a0a153ba5910e581ad2bc450b6af7e51e6214c2b173#123311 on # to get the id
-            List<String> idLengthSepList = Arrays.stream(id.split(HASH))
-                    .map(String::trim)
-                    .filter(s -> !s.isEmpty())
-                    .collect(Collectors.toList());
-            String blobId = idLengthSepList.get(0);
-
+            String blobId = id.split(HASH)[0];
+            long length = getBlobLengthOrZero(id);
             if (dsType == FAKE || dsType == FDS) {
                 // 0102030405... => 01/02/03/0102030405...
                 blobId = String.join(System.getProperty("file.separator"), blobId.substring(0, 2), blobId.substring(2, 4),
@@ -665,7 +661,7 @@ public class DataStoreCommand implements Command {
             // In case of blob ids dump, the list size would be 1 (Consisting of just the id)
             if (list.size() > 1) {
                 // Join back the encoded blob ref and the path on which the ref is present
-                return String.join(DELIM, blobId, EscapeUtils.unescapeLineBreaks(list.get(1)));
+                return String.join(DELIM, blobId, EscapeUtils.unescapeLineBreaks(list.get(1)), "" + length);
             } else {
                 // return the encoded blob id
                 return blobId;
@@ -684,6 +680,28 @@ public class DataStoreCommand implements Command {
                     writeStrings(iterator, outFile, true, log, "Transformed to verbose ids - ");
                 }
             }
+        }
+    }
+
+    /**
+     * Try to read the blob length from the blobId. It is typically after the '#'.
+     * It never throws an exception.
+     *
+     * @param blobId the blob id, which may contain a '#'
+     * @return the length, or 0 if unknown.
+     */
+    public static long getBlobLengthOrZero(String blobId) {
+        if (blobId == null) {
+            return 0;
+        }
+        int hashIndex = blobId.indexOf('#');
+        if (hashIndex < 0) {
+            return 0;
+        }
+        try {
+            return Long.parseLong(blobId.substring(hashIndex + 1));
+        } catch (NumberFormatException e) {
+            return 0;
         }
     }
 
--- a/oak-run/src/main/java/org/apache/jackrabbit/oak/run/Downloader.java
+++ b/oak-run/src/main/java/org/apache/jackrabbit/oak/run/Downloader.java
@@ -56,10 +56,32 @@ import java.util.stream.Collectors;
 public class Downloader implements Closeable {
 
     private static final Logger LOG = LoggerFactory.getLogger(Downloader.class);
+
+    /**
+     * The maximum size of what is considered a "small file".
+     * At the same time, this is the block size for large files.
+     */
     private static final long MAX_LENGTH_SINGLE_THREADED = 16 * 1024 * 1024;
 
+    /**
+     * The executor service used for small files,
+     * and to coordinate download of large files.
+     * Large files are split into parts, which are downloaded
+     * concurrently using range headers.
+     *
+     * The parts of large files may not use this service,
+     * otherwise download might deadlock: all threads
+     * might wait for parts, but the parts themselves
+     * can't be downloaded because the pool is full.
+     * The easiest solution is to use two pools.
+     */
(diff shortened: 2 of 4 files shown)


oak-run/src/main/java/org/apache/jackrabbit/oak/run/Downloader.java
```diff
@@ -170,29 +183,86 @@ public ItemResponse call() throws Exception {
...
-            try (InputStream inputStream = sourceUrl.getInputStream();
-                 FileOutputStream outputStream = new FileOutputStream(destinationPath.toFile())) {
-                byte[] buffer = new byte[bufferSize];
-                int bytesRead;
-                while ((bytesRead = inputStream.read(buffer)) != -1) {
-                    if (md != null) {
-                        md.update(buffer, 0, bytesRead);
+            if (item.length >= segmentSize) {
+                size = item.length;
+                LOG.info("Downloading large file {}: {} bytes", destinationPath.toString(), item.length);
```
joerghoh: how frequently does that happen? If too frequently I would suggest to move this to DEBUG otherwise it will pollute the log.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/Downloader.java
```diff
@@ -170,29 +183,86 @@ public ItemResponse call() throws Exception {
...
+                if (allSuccess) {
+                    try (OutputStream fileOut = Files.newOutputStream(destinationPath)) {
+                        OutputStream out = md == null ? fileOut : new DigestOutputStream(fileOut, md);
+                        for (Path segmentFile : segmentFiles) {
+                            if (Files.exists(segmentFile)) {
+                                Files.copy(segmentFile, out);
+                                Files.delete(segmentFile);
+                            }
+                        }
+                        LOG.info("Download {} size {}, {} parts", destinationPath.toString(), size, downloadTasks.size());
```
joerghoh: given the async nature of these steps, we end up with the 2 log statements for a large file being written quite apart from each other (this line + line 190 above). Wouldn't it make sense to just log once per file (still the question if INFO or DEBUG).

oak-run/src/main/java/org/apache/jackrabbit/oak/run/Downloader.java
```diff
@@ -170,29 +183,86 @@ public ItemResponse call() throws Exception {
...
+                boolean allSuccess = true;
+                for (int i = 0; i < downloadTasks.size(); i++) {
+                    try {
+                        boolean success = downloadTasks.get(i).get();
+                        if (!success) {
+                            allSuccess = false;
+                            break;
+                        }
+                    } catch (Exception e) {
+                        allSuccess = false;
```
joerghoh, line 262: With this we loose the actual exception. I would suggest to convert this ``boolean`` into an ``Exception`` type, and populate it here, so you can later log it.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/Downloader.java
```diff
@@ -170,29 +183,86 @@ public ItemResponse call() throws Exception {
...
             long size = 0;
-            try (InputStream inputStream = sourceUrl.getInputStream();
-                 FileOutputStream outputStream = new FileOutputStream(destinationPath.toFile())) {
-                byte[] buffer = new byte[bufferSize];
-                int bytesRead;
-                while ((bytesRead = inputStream.read(buffer)) != -1) {
-                    if (md != null) {
-                        md.update(buffer, 0, bytesRead);
+            if (item.length >= segmentSize) {
+                size = item.length;
```
joerghoh, line 221: the code in this branch is quite lengthy... would it make sense to extract it into a dedicated method?

joerghoh (commented):
the change looks good; from what I see the 2 major changes are:
* increasing the number of workers in the executor service by a factor of 4 
* when large files are downloaded you use multiple threads and range queries

I just a few minor nitpicks, mostly to regarding to logging.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/Downloader.java
thomasmueller: It is not very frequent: we have typically around 100 large files that need to be split. For me it is great to know if the large files are typically downloaded at the end of the process. But I think I have enough data for now, and I can change this to debug. I had two log statements: when did the download start, and when did it end.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/Downloader.java
thomasmueller: We already log exceptions in the task itself. If we log it here as well, then that would duplicate the logic. We don't currently have the case "return false": either we return true, or we throw an exception; but the API requires we return something. I just tried to have the most simple solution that is possible, given the API.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/Downloader.java
thomasmueller: I thought about that, but it would add additional complexity, because we need to have fields for the various variables then... I think in this case it's not worth it, because we only use the code in one place. This is the discussion "spagetti code vs ravioli code", which is described in detail here: https://wiki.c2.com/?RavioliCode

In general, I try to use separate function if an algorithm can be / is called from multiple places. Also, static functions are nice, because they can be tested more easily. But here, this doesn't apply.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/Downloader.java
thomasmueller: You are right, it's no longer needed. I changed it to debug now.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/DataStoreCommand.java
```diff
@@ -687,6 +683,28 @@ public void log() throws IOException {
...
+            return 0;
+        }
+        int hashIndex = blobId.indexOf('#');
+        if (hashIndex < 0) {
+            return 0;
+        }
+        try {
+            return Long.parseLong(blobId.substring(hashIndex + 1));
+        } catch (NumberFormatException e) {
+            return 0;
```
fabriziofortino, line 706: perhaps this never happens but should we log the exception just in case?

oak-run/src/main/java/org/apache/jackrabbit/oak/run/Downloader.java
```diff
@@ -101,16 +107,21 @@ public Downloader(int concurrency, int connectTimeoutMs, int readTimeoutMs, int
...
-                (int) Math.ceil(concurrency * .1), concurrency, 60L, TimeUnit.SECONDS,
+                corePoolSize, concurrency, 60L, TimeUnit.SECONDS,
                 new LinkedBlockingQueue<>(),
                 BasicThreadFactory.builder().namingPattern("downloader-%d").daemon().build()
         );
+        this.executorServiceForParts = new ThreadPoolExecutor(
+                corePoolSize, concurrency, 60L, TimeUnit.SECONDS,
+                new LinkedBlockingQueue<>(),
+                BasicThreadFactory.builder().namingPattern("partDownloader-%d").daemon().build()
+        );
```
fabriziofortino, line 150: by adding this thread pool executor, we are actually doubling the number of concurrent requests the downloader does. This might be confusing since `concurrency` is an option defined as you can see here.

```
OptionSpec<Integer> concurrencyOpt = parser.accepts("concurrency",
                        "Max number of concurrent requests that can occur (the default value is equal to 16 multiplied by the number of cores)")
                .withRequiredArg().ofType(Integer.class).defaultsTo(16 * Runtime.getRuntime().availableProcessors());
```

thomasmueller: Well, because we use the default LinkedBlockingQueue(), in reality each of the ThreadPoolExecutor will only create corePoolSize threads at most, and not more. All other tasks are waiting in the queue.

(We could replace LinkedBlockingQueue with eg. SynchronousQueue, but then we would get a RejectedExecutionException when we schedule more than max concurrency tasks... So that wouldn't simplify things).

I wanted to keep the changes to the minimum, so I didn't try to fix other things... So with 0.4 times concurrency corePoolSize, we have twice that number of threads, so still below the limit.

oak-run/src/main/java/org/apache/jackrabbit/oak/run/Downloader.java
```diff
@@ -54,7 +57,31 @@ public class Downloader implements Closeable {
 
     private static final Logger LOG = LoggerFactory.getLogger(Downloader.class);
 
+    /**
+     * The maximum size of what is considered a "small file".
+     * At the same time, this is the block size for large files.
+     */
+    private static final long MAX_LENGTH_SINGLE_THREADED = 16 * 1024 * 1024;
```
fabriziofortino, line 64: it would be nice to make this configurable through a command option

fabriziofortino (approved):
just a potential small improvement. LGTM 🍔 🍟

PR #2613, opened 2025-11-13
OAK-12007 XPath with or conditions should not always be converted to union

thomasmueller:


oak-core/src/test/java/org/apache/jackrabbit/oak/query/xpath/XPathToSQL2Test.java
```diff
@@ -0,0 +1,171 @@
...
+import org.mockito.Mockito;
+
+/**
+ * Tests for XPathToSQL2Converter focusing on the OR-to-UNION conversion behavior
+ * controlled by the optimizeXPathUnion feature toggle.
+ */
+public class XPathToSQL2Test {
+
+    @Test
+    public void testOrToUnionConversionFeatureDisabled() throws ParseException {
```
alagodasii, line 35: How about parameterizing the tests 
```suggestion
    class ConversionTripleParams implements ArgumentsProvider {
    @Override
    public Stream<Arguments> provideArguments(ExtensionContext context) {
        return Stream.of(
            Arguments.of("//*[@a = 1 or @b = 2]", false, "select ... where [a] = 1 union select ... where [b] = 2 "),
            Arguments.of("//*[@a = 1 or @b = 2]", true,  "select ... where [a] = 1 or [b] = 2 "),
            Arguments.of("//*[@a = 1 or @b = 2 or @c = 3]", false, "select ... where [a] = 1 union select ... where [b] = 2 union select ... where [c] = 3 "),
            Arguments.of("//*[@a = 1 or @b = 2 or @c = 3]", true,  "select ... where [a] = 1 or [b] = 2 or [c] = 3 "),
            Arguments.of("//*[@x = 1 and (@a = 1 or @b = 2)]", false, "select ... where [x] = 1 and [a] = 1 union select ... where [x] = 1 and [b] = 2 "),
            Arguments.of("//*[@x = 1 and (@a = 1 or @b = 2)]", true,  "select ... where [x] = 1 and ([a] = 1 or [b] = 2) "),
            Arguments.of("//*[@name = 'foo' or @name = 'bar']", false, "select ... where [name] in('foo', 'bar') "),
            Arguments.of("//*[@name = 'foo' or @name = 'bar']", true,  "select ... where [name] in('foo', 'bar') "),
            Arguments.of("//*[(@a = 1 and (@b = 2 or @c = 3)) or @d = 4]", false, "select ... where [a] = 1 and [b] = 2 union select ... where [a] = 1 and [c] = 3 union select ... where [d] = 4 "),
            Arguments.of("//*[(@a = 1 and (@b = 2 or @c = 3)) or @d = 4]", true,  "select ... where [a] = 1 and ([b] = 2 or [c] = 3) or [d] = 4 "),
            Arguments.of("//*[@title = 'Test' or @desc = 'Test']", false, "select ... where [title] = 'Test' union select ... where [desc] = 'Test' "),
            Arguments.of("//*[@title = 'Test' or @desc = 'Test']", true,  "select ... where [title] = 'Test' or [desc] = 'Test' "),
            Arguments.of("//*[jcr:contains(., 'test') or @type = 'page']", false, "select ... where contains(*, 'test') union select ... where [type] = 'page' "),
            Arguments.of("//*[jcr:contains(., 'test') or @type = 'page']", true,  "select ... where contains(*, 'test') or [type] = 'page' ")
        );
    }
}
```

thomasmueller: That's an option, but I don't see how it would be better.

thomasmueller:
I'll check which tests fail (in all of oak) if the default is changed.

oak-core/src/test/java/org/apache/jackrabbit/oak/query/xpath/XPathToSQL2Test.java
alagodasii: It's a matter of preference, but for maintainability mainly.
It's easier to compare inputs vs outputs and detect possible duplicate/missing scenario. It also makes purpose of this method more understandable to the reader, that it only serves single purpose and doesn't require additional setup.
But as I said at the beginning - matter of preference.

alagodasii (approved):

fabriziofortino (approved):

PR #2616, opened 2025-11-17
OAK-12008 : Embed aws sdk 2.x jars into oak-blob-cloud bundle

rishabhdaim:


rishabhdaim:
FYI, the oak-blob-cloud bundle is getting started with these embedded jars.

amit-jain (approved):

PR #2618, opened 2025-11-17
OAK-10844: Skip FullGC BETWEEN_CHECKPOINTS tests by default

Joscorbe:
Disabled by default, added a flag in case we ever need to run them again: 

`mvn test -Doak.test.enableBetweenCheckpointsModes=true`

These tests are currently flaky and increases significatively the time spent by FullGC ITs.


rishabhdaim:
@Joscorbe I would ignore the whole BranchCommitGCTest class since we aren't using this mode anymore and it has been superseded by KEEP_ONE.

PR #2621, opened 2025-11-21
OAK-12009 : provided support for GCP for new AWS sdk 2.x

rishabhdaim:


reschke (commented):
I really can't review; I'm not familiar with AWS bindings.

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
```diff
@@ -528,15 +530,6 @@ public void addMetadataRecord(final InputStream input, final String name) throws
             if (contextClassLoader != null) {
                 Thread.currentThread().setContextClassLoader(contextClassLoader);
             }
-            executor.shutdown();
```
amit-jain, line 531: Why has the closing of the executor been removed? This looks an unintended change

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/Utils.java
```diff
@@ -410,12 +473,39 @@ private static SdkHttpClient getSdkHttpClient(Properties prop) {
...
+
+        builder.connectionTimeout(Duration.ofMillis(config.connectionTimeout)) // Connection timeouts
                 .readTimeout(Duration.ofMillis(config.socketTimeout))
                 .writeTimeout(Duration.ofMillis(config.socketTimeout))
-                .maxConcurrency(config.maxConnections);
+                .maxConcurrency(concurrency)  // Connection pool - increased for better concurrency
+                .connectionMaxIdleTime(Duration.ofSeconds(idleTimeSeconds))
+                .connectionTimeToLive(Duration.ofSeconds(ttlSeconds))
+                .useIdleConnectionReaper(true)
+                .connectionAcquisitionTimeout(Duration.ofSeconds(10)) // Don't wait too long for a connection from pool
```
amit-jain: No config, changing the defaults os Ok?

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/Utils.java
```diff
@@ -482,6 +574,17 @@ private static void configureBuilder(final S3BaseClientBuilder builder, final Pr
 
         builder.endpointOverride(getEndPointUri(prop, accReq, region));
         builder.crossRegionAccessEnabled(Boolean.parseBoolean(prop.getProperty(S3Constants.S3_CROSS_REGION_ACCESS)));
+
+        // Disable checksums (replaces deprecated checksumValidationEnabled)
+        builder.requestChecksumCalculation(RequestChecksumCalculation.WHEN_REQUIRED);
+        builder.responseChecksumValidation(ResponseChecksumValidation.WHEN_REQUIRED);
+
```
amit-jain, line 582: Again defaults changed?

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
```diff
@@ -514,11 +514,13 @@ public void addMetadataRecord(final InputStream input, final String name) throws
         try {
             Thread.currentThread().setContextClassLoader(getClass().getClassLoader());
             // Specify `null` for the content length when you don't know the content length.
-            final AsyncRequestBody body = AsyncRequestBody.fromInputStream(input, null, executor);
+            byte[] bytes = input.readAllBytes();
```
amit-jain: Reading all input to memory will cause poetntial OOMs, these files can be quite big

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/Utils.java
```diff
@@ -372,16 +408,24 @@ static String calculateMD5(final String customerKey) {
...
     private static ClientOverrideConfiguration getClientConfiguration(Properties prop) {
+        final boolean isS3 = Objects.equals(RemoteStorageMode.S3, prop.get(S3Constants.MODE));
+
         int maxErrorRetry = Integer.parseInt(prop.getProperty(S3Constants.S3_MAX_ERR_RETRY));
         int connectionTimeOut = Integer.parseInt(prop.getProperty(S3Constants.S3_CONN_TIMEOUT));
         String encryptionType = prop.getProperty(S3Constants.S3_ENCRYPTION);
 
+        // API timeout should be much longer than connection timeout for large file uploads
+        // Use at least 5 minutes, or 10x connection timeout, whichever is larger
+        int apiTimeout = Math.max(connectionTimeOut * 10, 300000); // At least 5 minutes
```
amit-jain, line 419: Some heuristic applied, is there any best practice/recommendation on this?

amit-jain (changes requested):
I don't really know the nitty gritties of the aws & gcp and how their play along but I have highlighted some issues which should be fixed.

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
rishabhdaim: I have tried with files up to 200+ MB, and it worked fine with them.

The real reason for doing it this way is to allow GCP to work, which doesn't work when the length is not known.

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
rishabhdaim: Actually, the creation of Executor hasn't been removed. Missed that, would remove it.

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/Utils.java
rishabhdaim: These values have been suggested by the cursor and confirmed the same with perplexity.

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/Utils.java
rishabhdaim: This is an leftover change, would remove it.

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/Utils.java
rishabhdaim: Good catch, these needs to be disabled only for GCP.

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
rishabhdaim: I will make this code change for GCP only and keep the original one for AWS.

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
```diff
@@ -513,12 +513,26 @@ public void addMetadataRecord(final InputStream input, final String name) throws
...
+                    .key(addMetaKeyPrefix(name));
+
             // Specify `null` for the content length when you don't know the content length.
-            final AsyncRequestBody body = AsyncRequestBody.fromInputStream(input, null, executor);
+            final AsyncRequestBody body;
+            if (Objects.equals(RemoteStorageMode.S3, properties.get(S3Constants.MODE))) {
+                body = AsyncRequestBody.fromInputStream(input, null, executor);
+            } else {
+                // for GCP we need to know the length in advance, else it won't work.
+                byte[] bytes = input.readAllBytes();
```
joerghoh: when you are handling large binaries, this is likely to cause an OOM

rishabhdaim: @joerghoh Yes, it might cause OOM, but there is no other option. 
GCP doesn't work with unknown-length streams, so we had to find the length.

FYI: I have tested with 200+MB files and they are working fine.

joerghoh: You could stream it into a temp file and calculate the length from there.

rishabhdaim: thanks @joerghoh for the review and suggestion. Please find the fix here: https://github.com/apache/jackrabbit-oak/pull/2621/commits/34786fbc17d680d01a56d1855eae5289a92b5564

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
```diff
@@ -513,12 +516,55 @@ public void addMetadataRecord(final InputStream input, final String name) throws
...
+                    .bucket(bucket)
+                    .contentType("application/octet-stream")
+                    .key(addMetaKeyPrefix(name));
+
             // Specify `null` for the content length when you don't know the content length.
-            final AsyncRequestBody body = AsyncRequestBody.fromInputStream(input, null, executor);
+            final AsyncRequestBody body;
+            if (Objects.equals(RemoteStorageMode.S3, properties.get(S3Constants.MODE))) {
+                body = AsyncRequestBody.fromInputStream(input, null, executor);
+            } else {
```
joerghoh: Can you refactor that code into a dedicated method? Having all of these different cases inline makes the method very long and also harder to understand and test.

rishabhdaim: addressed in https://github.com/apache/jackrabbit-oak/pull/2621/commits/663652d3a5046da1b00fd488ca98741055733552

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
```diff
@@ -1323,6 +1331,48 @@ private static String getIdentifierName(String key) {
...
+            } else if (input instanceof ByteArrayInputStream) {
+                length = input.available();
+                body = AsyncRequestBody.fromInputStream(input, length, executor);
+            } else if (input.markSupported()) {
+                // in case the inputStream supports mark & reset
+                input.mark(Integer.MAX_VALUE);
+                length = IOUtils.consume(input);
+                input.reset();
+                body = AsyncRequestBody.fromInputStream(input, length, executor);
+            } else {
```
joerghoh, line 1357: thinking again about that: Do we really need to store the entire stream on disk if we just the length? Can't we consume it in chunks?

```
long length = 0;
byte[] buffer = new byte[2048];
int sizeRead = buffer.read(buffer);
while (sizeRead != -1) {
  length+=sizeRead;
  sizeRead = buffer.read(buffer);
}
```

rishabhdaim: In case we consume the stream, then we can't use that again to upload to AWS/GCP.

joerghoh: indeed, that's correct. 

I am not 100% happy with that temp file approach either especially when these files are cleaned up on shutdown; meaning that they accumulate over time and can consume all disk space.

rishabhdaim: @joerghoh tempfiles would only be created as a last resort.

Apart from above, if you are fine with PR, could you please approve it.

joerghoh (approved):
+1 on the stream handling, cannot say much about the general AWS specific things.

Joscorbe (approved):
Not very familiar with this part of the code, but looks good now.

PR #2622, opened 2025-11-24
OAK-11916 : added exiting executor service in oak-commons

rishabhdaim:


oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/internal/concurrent/ExecutorUtils.java
```diff
@@ -36,4 +43,27 @@ public static Executor directExecutor() {
     public static ExecutorService newDirectExecutorService() {
         return new DirectExecutorService();
     }
+
+    public static ExecutorService getExitingExecutorService(ThreadPoolExecutor executor) {
+        setDeamonThreadFactory(executor);
```
mbaedke, line 48: Why is that important? Does it even make a difference?

mbaedke:
What is the use case? Why the fixed timeout?

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/internal/concurrent/ExecutorUtils.java
rishabhdaim: This threadpool shouldn't stop the JVM shutdown, so marking all the threads as `deamon`.

rishabhdaim:
> What is the use case? Why the fixed timeout?

@mbaedke fixed timeouts are taken from Guava's implementation to keep them as similar as possible.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/internal/concurrent/ExecutorUtils.java
mbaedke: It's a shutdown hook, so the threads are running only during shutdown. Since I have no idea how the deamon flag works in this situation, I was just wondering.

mbaedke (approved):

PR #2626, opened 2025-11-26
OAK-11697: Indexing / query limit / traversal error: improve diagnostics 

mbaedke:


oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -1267,6 +1269,9 @@ public void verifyNotPotentiallySlow() {
             }
             String caller = IndexUtils.getCaller(settings.getIgnoredClassNamesInCallTrace());
             String message = "Traversal query (query without index): " + statement + "; called by " + caller + "; consider creating an index";
+            if (traversal == Traversal.FAIL || traversal == Traversal.WARN && !potentiallySlowTraversalQueryLogged) {
```
reschke: I would swap plan and index information, as the latter might be more relevant

reschke (commented):
and the extended exception messsage will propagate up? maybe add a simple test case?

reschke (approved):

reschke (commented):
LGTM, but it would be nice to hear from @thomasmueller

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -1267,6 +1273,26 @@ public void verifyNotPotentiallySlow() {
             }
             String caller = IndexUtils.getCaller(settings.getIgnoredClassNamesInCallTrace());
             String message = "Traversal query (query without index): " + statement + "; called by " + caller + "; consider creating an index";
```
reschke: would indeed be nice to refactor this slightly

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -1285,6 +1285,37 @@ public void verifyNotPotentiallySlow() {
...
+        String caller = IndexUtils.getCaller(settings.getIgnoredClassNamesInCallTrace());
+        String message = "Traversal query (query without index): " + statement + "; called by " + caller + "; consider creating an index";
+        if (traversal == Traversal.FAIL || traversal == Traversal.WARN && !potentiallySlowTraversalQueryLogged) {
+            Set<String> reindex = getNamesOfReindexingIndexes();
+            message += "\n\nExecution plan:\n" + getPlan();
+            if (!reindex.isEmpty()) {
+                String reindexNames = reindex.stream().map(name -> name + ",").collect(Collectors.joining());
+                message += "\n\nNote that the following indexes were unavailable because of re-indexing:\n"
+                        + reindexNames.substring(0, reindexNames.length() - 1);
+            }
```
thomasmueller, line 1306: Even if the indexes are reindexing currently, I think the old version can still be used for queries (no?)

thomasmueller: So I probably would not add this here... instead, I think we should probably add a metric for "indexes that are currently reindexing", which should in theory be zero. And if there are indexes that are indexing, specially if that's for longer than eg. 1 hour, then we should investigate that.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -1285,6 +1285,37 @@ public void verifyNotPotentiallySlow() {
...
+        Set<String> reindex = new HashSet<>();
+        Iterable<Tree> indexes = context.getRoot().getTree("/" + INDEX_DEFINITIONS_NAME).getChildren();
+        for (Tree index : indexes) {
+            String name = index.getName();
+            PropertyState primaryType = index.getProperty(JCR_PRIMARYTYPE);
+            if (primaryType != null && INDEX_DEFINITIONS_NODE_TYPE.equals(primaryType.getValue(Type.STRING))) {
+                PropertyState reindexProp = index.getProperty(REINDEX_PROPERTY_NAME);
+                if (reindexProp != null && reindexProp.getValue(Type.BOOLEAN)) {
+                    reindex.add(name);
+                }
```
thomasmueller: I saw some cases where an index had reindex = true, but type = disabled... So that would need to be ignored.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -1257,16 +1263,10 @@ public boolean isPotentiallySlow() {
...
-                traversal = settings.getFailTraversal() ? Traversal.FAIL : Traversal.WARN;
-            } else {
-                // explicitly set in the query
-                traversal = queryOptions.traversal;
-            }
-            String caller = IndexUtils.getCaller(settings.getIgnoredClassNamesInCallTrace());
-            String message = "Traversal query (query without index): " + statement + "; called by " + caller + "; consider creating an index";
+            Traversal traversal = queryOptions.traversal == Traversal.DEFAULT ?
+                    settings.getFailTraversal() ? Traversal.FAIL : Traversal.WARN :
+                    queryOptions.traversal;
```
thomasmueller: I actually find the earlier version a bit more readable....

thomasmueller:
I think it's a very good improvement, thanks a lot! Just a few minor points / points for discussion.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
reschke: > So I probably would not add this here... instead, I think we should probably add a metric for "indexes that are currently reindexing", which should in theory be zero.
Yes, absolutely (separate ticket?)

>And if there are indexes that are indexing, specially if that's for longer than eg. 1 hour, then we should investigate that.

But that's "just" an operational issue, right?

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
reschke: If that's a weird case, should we add a DEBUG log for this somewhere?

reschke:
Ok. Can we close the loop on this soon? Would be good to have it in 1.92.0....

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
mbaedke: I just made the message more generic.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -1285,6 +1292,38 @@ public void verifyNotPotentiallySlow() {
...
+    private Set<String> getNamesOfReindexingIndexes() {
+        Set<String> reindex = new HashSet<>();
+        Iterable<Tree> indexes = context.getRoot().getTree("/" + INDEX_DEFINITIONS_NAME).getChildren();
+        for (Tree index : indexes) {
+            String name = index.getName();
+            PropertyState primaryType = index.getProperty(JCR_PRIMARYTYPE);
+            if (primaryType != null && INDEX_DEFINITIONS_NODE_TYPE.equals(primaryType.getValue(Type.STRING))) {
+                PropertyState reindexProp = index.getProperty(REINDEX_PROPERTY_NAME);
+                PropertyState typeProp = index.getProperty(TYPE_PROPERTY_NAME);
+                if (reindexProp != null && reindexProp.getValue(Type.BOOLEAN) && !(typeProp != null && TYPE_DISABLED.equals(typeProp.getValue(Type.STRING)))) {
```
thomasmueller: sorry I have a hard time understanding the "not not" part here:

!(typeProp != null && TYPE_DISABLED.equals(typeProp.getValue(Type.STRING))

what about:


```suggestion
                if (reindexProp != null && reindexProp.getValue(Type.BOOLEAN) && typeProp != null && !TYPE_DISABLED.equals(typeProp.getValue(Type.STRING))) {
```

it changes the logic of typeProp == null... normally, it is never null (the node type disallows that) but it's now only there to protect against NPE.

mbaedke: Since this only makes a difference in the case of typeProp == null, I don't really care.

PR #2635, opened 2025-11-30
OAK-12017: Add path information to CommitFailedException in EditorDiff

cschneider:
- Track current path during tree traversal in EditorDiff
- Automatically enhance CommitFailedException with path information
- Works generically for all editors, not just VersionEditor
- Path is appended to exception message: 'message (path: /path/to/node)'
- Avoids duplication if path already present in exception message


rishabhdaim (commented):
Could you please create a unit case for this change.

cschneider:
@rishabhdaim This is not yet at the level to be merged. I would like to first get feedback from Oak committers if this is a good idea and will work.

joerghoh:
implemented with #3050

PR #2641, opened 2025-12-02
OAK-10844: Disable BranchCommitGCTest due to flakiness

Joscorbe:


rishabhdaim (approved):
Since we are not using BRANCH_COMMIT mode, I think its best to disable these tests.

reschke (approved):

PR #2646, opened 2025-12-04
OAK-11921 : replaced Guava's AbstractListeningExecutorService with na…

rishabhdaim:
…tive java implementation


reschke (approved):
This hints that we can remove things like `FutureConverter` at some point?

PR #2651, opened 2025-12-05
OAK-11731: expose NodeCounter (getEstimatedChildNodeCounts) in Jackra…

reschke:
…bbitNode - register bean just once with the compatible interface


reschke:
Problem: other than expected, the mbean can only be found under the "new" class name.

Maybe this is not a problem at the end of the day? Tests are passing.

PR #2653, opened 2025-12-08
OAK-11975: Use Mockito as an agent in Oak

mbaedke:
Replaced the dynamic inline mocking with a static agent dependency.

### DIFF
--- a/oak-api/pom.xml
+++ b/oak-api/pom.xml
@@ -84,5 +84,10 @@
       <artifactId>junit</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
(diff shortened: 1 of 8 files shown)


mbaedke:
The surefire und failsafe configs are part of the parent pom so that they are reused in all subproject. I left it that way, which has the slightly ugly side effect that the agent is used even in projects that do not use mockito, which now need a mockito-core dependency. Of course, we can also decide to leave the parent pom unchanged and override the config in the subprojects that actually use mockito. I prefer the parent pom solution, which produces much less boilerplate and automatically makes all subprojects mockito ready. 
Or we use mockito-junit-jupiter as rishab mentioned in the Jira. I didn't test that yet.

reschke (approved):

PR #2657, opened 2025-12-11
OAK-12036 - Expose stats in the PersistentCache

nfsantos:
And add method to FileStore to retrieve persistent cache stats in case it is configured.

### DIFF
--- a/oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/AbstractFileStore.java
+++ b/oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/AbstractFileStore.java
@@ -192,6 +192,11 @@ public abstract class AbstractFileStore implements SegmentStore, Closeable {
         return segmentCache.getCacheStats();
     }
 
+    @Nullable
+    public CacheStatsMBean getPersistentCacheStats() {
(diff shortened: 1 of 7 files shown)


jsedding (approved):
LGTM.

The only thing that could be improved, and it isn't introduced in this PR, is the field `AbstractPersistentCache#segmentCacheStats`. The getter is annotated as `@NotNull` and no null checks are present in the implementation. It would be more robust to make the field private and require a `@NonNull SegmentCacheStats` argument in the constructor. I know, that would require adjusting implementations as well. So maybe not the most practical idea.

nfsantos:
> The only thing that could be improved, and it isn't introduced in this PR, is the field `AbstractPersistentCache#segmentCacheStats`. The getter is annotated as `@NotNull` and no null checks are present in the implementation. It would be more robust to make the field private and require a `@NonNull SegmentCacheStats` argument in the constructor. I know, that would require adjusting implementations as well. So maybe not the most practical idea.

I see the problem, there is no mechanism to enforce that derived classes set the `segmentCacheStats`. Would be cleaner if there was some check, but the base class will fail almost immediately with a NPE if this field is null, so this issue would be caught in testing. I prefer to not touch this on this PR. Maybe another PR.

smiroslav (approved):

PR #2658, opened 2025-12-12
Find matching index for query

markdaugherty:
@thomasmueller I know this needs refinement (and many more tests etc), but I didn't want to take it too far before reviewing the overall strategy.


thomasmueller:
Looks very good to me, I'll merge it.

PR #2660, opened 2025-12-14
OAK-12037 : remove usage of Guava's Monitor from oak-lucene

rishabhdaim:


thomasmueller (approved):

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/IndexCopier.java
```diff
@@ -363,28 +366,31 @@ public boolean isCopyInProgress(LocalIndexFile file) {
...
-                notCopying = copyCompletionMonitor.enterWhen(notCopyingGuard, timeoutMillis, TimeUnit.MILLISECONDS);
-                if (notCopying) {
-                    copyCompletionMonitor.leave();
+                while (isCopyInProgress(file)) {
+                    long remaining = deadline - System.nanoTime();
+                    if (remaining <= 0) {
+                        // timeout
+                        break;
+                    }
+                    notCopyingCondition.awaitNanos(remaining);
```
mbaedke, line 386: This whole nano calculation seems clumsy. Why not just `notCopyingCondition.await(timeoutMillis, TimeUnit.MILLISECONDS)` ?

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/IndexCopier.java
```diff
@@ -363,28 +366,31 @@ public boolean isCopyInProgress(LocalIndexFile file) {
...
-                return !isCopyInProgress(file);
-            }
-        };
         long localLength = file.actualSize();
         long lastLocalLength = localLength;
 
         boolean notCopying = !isCopyInProgress(file);
         while (!notCopying) {
+            final long deadline = System.nanoTime() + TimeUnit.MILLISECONDS.toNanos(timeoutMillis);
+            copyCompletionLock.lock();
```
mbaedke, line 375: How is that working with the timeout?

mbaedke (changes requested):
See comments inline.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/IndexCopier.java
thomasmueller: AFAIK the thread could be woken up in some cases, and so a loop is needed, and so a calculation is needed.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/IndexCopier.java
rishabhdaim: by following this logic, we ensure that we block for only an absolute time in future.

On each inner-loop iteration you recompute `remaining = deadline - System.nanoTime()`

This is to ensure that regardless of how many times the threads are woken, we only wait for `timeoutMillis`.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/IndexCopier.java
mbaedke: Yes, a loop is needed . But just use notCopyingCondition.await(timeoutMillis, TimeUnit.MILLISECONDS). There is no reason to convert to nanos, is there?

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/IndexCopier.java
mbaedke: I'm not asking what the timeout is there for, I'm asking how copyCompletionLock.lock() is supposed to work here. It might break the whole timeout idea. Shouldn't we use copyCompletionLock.tryLock(long, TimeUnit) there?

mbaedke (changes requested):
Comments inline

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/IndexCopier.java
rishabhdaim: `copyCompletionLock.lock()` is required to wait for the condition i.e. `isCopyInProgress(file)` in this case. I am not sure why do we need `copyCompletionLock.tryLock(long, TimeUnit)` here.

Also, `nanos` is used to for better time management and it was suggested by `cursor` and other `AI` tools over timeout in millis.

mbaedke: Ok. I though that copyCompletionLock.lock() might lock indefinitely, but with the logic implemented here that's actually not possible, afaics.

mbaedke (approved):

PR #2663, opened 2025-12-16
OAK-12039 Treat transient client-side exceptions as recoverable in AzureRepositoryLock

smiroslav:
https://issues.apache.org/jira/browse/OAK-12039

### DIFF
--- a/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureRepositoryLock.java
+++ b/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureRepositoryLock.java
@@ -220,6 +220,8 @@ public class AzureRepositoryLock implements RepositoryLock {
      * This includes timeouts and IO/network errors that can occur when communicating with Azure.
      * <p>
      * Per Azure SDK documentation, the timeout parameter causes a RuntimeException to be raised.
+     * Reactor-core throws IllegalStateException with message "Timeout on blocking read" when
+     * the timeout expires (see BlockingSingleSubscriber.blockingGet in reactor-core).
      * @param e the exception to check
      * @return true if this is a transient exception that should be retried
      */
@@ -227,7 +229,8 @@ public class AzureRepositoryLock implements RepositoryLock {
         Throwable current = e;
         while (current != null) {
             if (current instanceof java.util.concurrent.TimeoutException ||
-                current instanceof java.io.IOException) {
+                current instanceof java.io.IOException ||
+                current instanceof IllegalStateException) {
                 return true;
             }
             current = current.getCause();
--- a/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureRepositoryLock.java
+++ b/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureRepositoryLock.java
@@ -36,7 +36,9 @@ public class AzureRepositoryLock implements RepositoryLock {
     private static final Logger log = LoggerFactory.getLogger(AzureRepositoryLock.class);
 
     private static final int TIMEOUT_SEC = Integer.getInteger("oak.segment.azure.lock.timeout", 0);
-    private static final Integer LEASE_RENEWAL_TIMEOUT_MS = 5000;
+
+    public static final String LEASE_RENEWAL_TIMEOUT_PROP = "oak.segment.azure.lock.leaseRenewalTimeoutInMs";
+    private static final int LEASE_RENEWAL_TIMEOUT_MS = Integer.getInteger(LEASE_RENEWAL_TIMEOUT_PROP, 5000);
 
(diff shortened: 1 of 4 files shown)


oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureRepositoryLock.java
```diff
@@ -211,6 +213,26 @@ private boolean isInError() {
...
+        Throwable current = e;
+        while (current != null) {
+            if (current instanceof java.util.concurrent.TimeoutException ||
+                current instanceof java.io.IOException) {
+                return true;
+            }
+            current = current.getCause();
+        }
+        return false;
+    }
```
jsedding, line 239: @smiroslav looking at the [code referenced by the stacktrace](https://github.com/reactor/reactor-core/blob/60e3e4d8e5f10992d8d7acee75a5caa58eda1e68/reactor-core/src/main/java/reactor/core/publisher/BlockingSingleSubscriber.java#L128) you shared in OAK-12039, I don't think this check will work.

As far as I can see, an `IllegalStateException` without any causes is constructed if a timeout occurs.

The case where an `IOException` is expected as the cause _may_ work. I cannot tell immediately from the code. A few more stack traces/thread dumps may be helpful to get a better understanding of the failing code-paths.

smiroslav: @jsedding thanks for chiming in. 

I have updated the issue description to include the code where the timeout exception is thrown and a discussion thread.

ierandra:
I think the fix should also be added to V8 implementation.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureRepositoryLock.java
jsedding: @smiroslav I completely missed that they create a new `TimeoutException` and set it as the cause. I only saw the new `IllegalStateException`. Sorry, my bad!

jsedding (approved):

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureRepositoryLock.java
jsedding: I wonder why the cause is not shown in the stack trace you shared in the JIRA ticket? Could that stacktrace be from an older version of the reactor-core library?

jsedding: Indeed, the `TimeoutException` cause was added [reactor-core 8331346](https://github.com/reactor/reactor-core/commit/8331346774c6fd397aa5646949d5d7800312bb60), which was first released in version v3.6.4.

I think you need to update `reactor-core` before your fix can work. WDYT?

smiroslav: Yes, definitely, thank you!

jsedding: Looks good now!

smiroslav:
I have changed the test to cause ```TimeoutException``` instead of mocking it.

smiroslav:
> I think the fix should also be added to V8 implementation.

@ierandra will address it in a separate PR

smiroslav:
I decided not to update reactor-core, as the Azure documentation advises against using a version with a different major or minor version than the one specified in the Azure SDK dependencies.
Azure SDK v12 version should be updated, but I would not do that in this PR. 

https://learn.microsoft.com/en-us/azure/developer/java/sdk/troubleshooting-dependency-version-conflict#understand-compatible-dependency-versions
https://github.com/reactor/.github/blob/main/SUPPORT.adoc#our-policy-on-deprecations

Instead, IllegalStateException is treated as recoverable.

smiroslav:
> > I think the fix should also be added to V8 implementation.
> 
> @ierandra will address it in a separate PR

The AzureRepositoryLockV8 is already correctly handling client-side timeout exceptions in the correct way:
https://github.com/apache/jackrabbit-oak/pull/2667

ierandra:
@smiroslav 
> The AzureRepositoryLockV8 is already correctly handling client-side timeout exceptions in the correct way:

that is very strange because when we did the analysis we observe the same issue for customers running on v8

smiroslav:
@ierandra, looking at the code, all exceptions that happen in AzureRepositoryLockV8#refreshLease should be logged. 

Do you perhaps have the stacktrace?

I suggest we move the discussion for v8 to the JIRA or #2667.

jsedding:
> I decided not to update reactor-core ...

I hope you're still merging the rest of the PR, because eventually we will update azure and reactor 🙂

PR #2665, opened 2025-12-17
OAK-12040 - segment-azure: reduce HTTP requests for writes

jsedding:


oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzureSegmentArchiveWriter.java
```diff
@@ -77,8 +79,10 @@ protected void doWriteArchiveEntry(RemoteSegmentArchiveEntry indexEntry, byte[]
         ioMonitor.beforeSegmentWrite(new File(blob.getBlobName()), msb, lsb, size);
         Stopwatch stopwatch = Stopwatch.createStarted();
         try {
-            blob.upload(BinaryData.fromBytes(Arrays.copyOfRange(data, offset, offset + size)), true);
-            blob.setMetadata(AzureBlobMetadata.toSegmentMetadata(indexEntry));
+            BinaryData binaryData = BinaryData.fromStream(new ByteArrayInputStream(data, offset, size), (long) size);
```
joerghoh, line 85: there's quite a bit of complexity behind this, can you just add a comment, e.g.:

```suggestion
            // upload the binary and set the metadata in a single call to the blobstore
            BinaryData binaryData = BinaryData.fromStream(new ByteArrayInputStream(data, offset, size), (long) size);
```

joerghoh: Nice finding. Can you outline the measured improvement from that in the ticket?

jsedding: > can you just add a comment
makes sense, done

jsedding: > Nice finding. Can you outline the measured improvement from that in the ticket?

Will do once I get around to measuring something 🙂 However, reducing from 2 to 1 HTTP call cannot make things worse.

From analysis of a log file, I saw that the upload calls had a median of 10ms (and slightly higher avg), while the metadata calls had a median of 8ms (also slightly higher avg). Based on this, I expect a speedup of writes in the range of 30-40%. How that manifests on the JCR level remains to be seen.

Clearly, write-heavy use-cases should benefit most. E.g. I expect compaction to run faster.

jsedding: @joerghoh I added some [performance numbers in the ticket](https://issues.apache.org/jira/browse/OAK-12040?focusedCommentId=18046172&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-18046172). For a write-heavy use-case I measured an improvement of ~45%.

PR #2672, opened 2025-12-23
OAK-12047 - CachingSegmentArchiveReader#readSegment should check if archive contains segment before looking up in the PersistentCache

nfsantos:
See rationale here [OAK-12047](https://issues.apache.org/jira/browse/OAK-12047)


mreutegg (commented):
The change looks good to me, but I'd like to see a test that validates the change.

nfsantos:
> The change looks good to me, but I'd like to see a test that validates the change.

Added a check in an existing test: https://github.com/apache/jackrabbit-oak/pull/2672/commits/9f555b80f5702e03e3feb77b164416f27ed7efec

mreutegg (approved):

PR #2687, opened 2026-01-19
OAK-12051: Fix NPE when ordering union query by jcr:score

ChlineSaurus:
Fixes the NullPointerException during UnionQuery sorting by the JCR score. The problem occurs if there is a score column, but accessing a score of that column returns Null. This can happen if, for example, there is a union between three queries, of which at least one has scores, and at least one doesn't have, and those two results are merged first. 

Issue: https://issues.apache.org/jira/browse/OAK-12051

New behavior: If null values occur in such a column, we replace it with a zero. The warning from the illegalArgumentException was removed, as returning 0 occurs frequently and is nothing to worry about.

Question:
- Should this PR first be merged to a separate branch?

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/UnionQueryImpl.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/UnionQueryImpl.java
@@ -324,8 +324,8 @@ public class UnionQueryImpl implements Query {
             rightIter = ((MeasuringIterator) rightRows).getDelegate();
         }
         if (orderBy == null) {
-            if(!settings.isSortUnionQueryByScoreEnabled()) {
-                // Default old behavior
+            if(settings.isSortUnionQueryLegacyModeEnabled()) {
+                // Legacy mode: concatenate results without score-based merging
                 it = IteratorUtils.chainedIterator(leftIter, rightIter);
             } else {
                 boolean leftHasScore = isScorePresent(left);
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/UnionQueryImpl.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/UnionQueryImpl.java
@@ -568,16 +568,16 @@ public class UnionQueryImpl implements Query {
 
     /**
      * @param row the result row
-     * @return the jcr:score as a double
-     * Precondition: {@link #isScorePresent(Query)} must be true. If the row lacks a jcr:score, 0.0 is returned and
-     * the issue is logged.
+     * @return the jcr:score as a double, or 0.0 if the row lacks a score value
      */
     private double getScoreFromRow(ResultRowImpl row) {
         try {
             PropertyValue scoreValue = row.getValue(QueryConstants.JCR_SCORE);
+            if (scoreValue == null) {
+                return 0.0;
+            }
             return scoreValue.getValue(Type.DOUBLE);
         } catch (IllegalArgumentException e) {
-            LOG.warn("Failed to get jcr:score for path={}", row.getPath(), e);
             return 0.0;
         }
     }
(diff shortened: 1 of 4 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/query/UnionQueryImpl.java
```diff
@@ -568,16 +568,16 @@ public int compare(ResultRowImpl left, ResultRowImpl right) {
 
     /**
      * @param row the result row
-     * @return the jcr:score as a double
-     * Precondition: {@link #isScorePresent(Query)} must be true. If the row lacks a jcr:score, 0.0 is returned and
-     * the issue is logged.
+     * @return the jcr:score as a double, or 0.0 if the row lacks a score value
```
thomasmueller, line 571: The problem is now that the feature toggle needs to be changed. Otherwise we can never enable it, as there is no way to distinguish between the old and the new version.

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/query/UnionQueryImpl.java
ChlineSaurus: Oh, fair enough. How would you do this? Add a second feature toggle? Or is renaming the existing one enough?

thomasmueller: Renaming the existing one is enough.

thomasmueller: There is a large delay between us changing Oak and then enabling the feature toggle. Also, the changed behavior is not tested currently before the customer. I think we should try a different approach: we should _enable_ the new behavior by default, and keep the feature toggle to be able to switch _back_ to the old behavior where needed.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/UnionQueryImpl.java
```diff
@@ -568,16 +568,16 @@ public int compare(ResultRowImpl left, ResultRowImpl right) {
...
      */
     private double getScoreFromRow(ResultRowImpl row) {
         try {
             PropertyValue scoreValue = row.getValue(QueryConstants.JCR_SCORE);
+            if (scoreValue == null) {
+                return 0.0;
+            }
             return scoreValue.getValue(Type.DOUBLE);
         } catch (IllegalArgumentException e) {
-            LOG.warn("Failed to get jcr:score for path={}", row.getPath(), e);
```
thomasmueller: Hm, I wouldn't remove this line, it is can happen.

oak-core/src/main/java/org/apache/jackrabbit/oak/query/UnionQueryImpl.java
ChlineSaurus: I responded to the issues. However, I currently have problems when running the tests locally (OOM errors in Oak Segment AWS and Oak Segment Azure, it only works when running with -Dbaseline.skip=true, and even then it is sometimes flakey)

thomasmueller:
@ChlineSaurus  could you merge the changes from trunk, and then open a PR against the OAK-12051 branch I just created? That way we will be able to run the tests.

thomasmueller:
(This branch has conflicts that must be resolved)

thomasmueller:
I can only run the maven build, but would like to run the Sonar workflow as well, and that can only be done if we merge it into a branch in Oak.

ChlineSaurus:
> @ChlineSaurus could you merge the changes from trunk, and then open a PR against the [OAK-12051](https://issues.apache.org/jira/browse/OAK-12051) branch I just created? That way we will be able to run the tests.

I opened a new PR from my forks trunk to the new branch you created. https://github.com/apache/jackrabbit-oak/pull/2746 I closed this PR.

ChlineSaurus:
> (This branch has conflicts that must be resolved)

Ah yes, the removed feature toggles caused merge conflicts. I now rebased everything on top of trunk and I'm currently rerunning the tests locally.

PR #2688, opened 2026-01-19
OAK-12062 : fixed NPE while uploading metadata for AWS S3

rishabhdaim:


reschke:
I'm a bit concerned that we have an File leak; `deleteOnExit()` will not be sufficient unless the VM really terminates.

Deleting the File after use should be sufficient.

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
```diff
@@ -1334,42 +1334,38 @@ private static String getIdentifierName(String key) {
...
-                input.reset();
-                body = AsyncRequestBody.fromInputStream(input, length, executor);
-            } else {
-                // we have to read all the stream to get the actual length
-                // last else block: store to temp file and re-read
-                final File tempFile = File.createTempFile("inputstream-", ".tmp");
-                tempFile.deleteOnExit(); // Clean up after JVM exits
+            // we have to read all the stream to get the actual length
+            // last else block: store to temp file and re-read
+            final File tempFile = File.createTempFile("inputstream-", ".tmp");
```
reschke: Prefix should show which class created the temp file.

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
```diff
@@ -1334,42 +1334,38 @@ private static String getIdentifierName(String key) {
...
+            tempFile.deleteOnExit(); // Clean up after JVM exits
```
thomasmueller: This will delete on exit, but not delete after reading. I would delete after reading as well, so that in the normal case (where the process is not killed) the file is deleted. Otherwise we might have a large number of temp files, if the process is not terminated for a long time.

(This could cause out-of-disk space, and if the number of files is huge, also out-of-memory in kubernetes)

thomasmueller (changes requested):

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
rishabhdaim: addressed in https://github.com/apache/jackrabbit-oak/pull/2688/commits/8f64b8a8e32ceb92024fde9998bdfba63220e5f3

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
rishabhdaim: addressed in https://github.com/apache/jackrabbit-oak/pull/2688/commits/8f64b8a8e32ceb92024fde9998bdfba63220e5f3

thomasmueller: How is it addressed? I do not see where "delete()" is called. It is still only calling "deleteOnExit()", which will not delete immediately.

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
```diff
@@ -536,6 +562,13 @@ public void addMetadataRecord(final InputStream input, final String name) throws
             if (contextClassLoader != null) {
                 Thread.currentThread().setContextClassLoader(contextClassLoader);
             }
+            if (tempFile != null) {
+                try {
+                    Files.deleteIfExists(tempFile.toPath());
+                } catch (IOException e) {
+                    LOG.warn("Failed to delete temp file {}", tempFile, e);
+                }
+            }
```
rishabhdaim, line 571: @thomasmueller here I am calling `delete` inside `finally` block immediately after uploading.

thomasmueller (approved):

PR #2689, opened 2026-01-19
OAK-12010 Simplified index management (without optimizer)

thomasmueller:
This is a subset of the (much) larger draft PR https://github.com/apache/jackrabbit-oak/pull/2652

This is just support for the "diff.index"; without the optimizer.

### DIFF
--- a/oak-core/DiffIndex.java
+++ b/oak-core/DiffIndex.java
@@ -30,7 +30,6 @@ import org.apache.jackrabbit.oak.commons.json.JsonObject;
 import org.apache.jackrabbit.oak.plugins.index.IndexConstants;
 import org.apache.jackrabbit.oak.plugins.index.IndexName;
 import org.apache.jackrabbit.oak.plugins.tree.TreeConstants;
-import org.apache.jackrabbit.oak.spi.nodetype.NodeTypeConstants;
 import org.apache.jackrabbit.oak.spi.state.NodeBuilder;
 import org.apache.jackrabbit.oak.spi.state.NodeStore;
 import org.slf4j.Logger;
@@ -47,8 +46,6 @@ public class DiffIndex {
 
     private static final Logger LOG = LoggerFactory.getLogger(DiffIndex.class);
 
-    private final static DiffIndexMerger MERGER = new DiffIndexMerger();
-
     /**
      * Apply changes to the index definitions. That means merge the index diff with
      * the existing indexes, creating new index versions. It might also mean to
@@ -58,25 +55,8 @@ public class DiffIndex {
      * @param indexDefinitions the /oak:index node
      */
     public static void applyDiffIndexChanges(NodeStore store, NodeBuilder indexDefinitions) {
-        JsonObject diffs = collectDiffs(indexDefinitions);
-        if (diffs == null) {
-            // nothing todo
-            return;
-        }
-        processDiffs(store, indexDefinitions, diffs);
-    }
-
-    /**
-     * Collect the diffs from the diff.index and diff.index.optimizer.
-     *
-     * @param indexDefinitions the node builder for /oak:index
-     * @return the diffs, or null if none
-     */
-    public static JsonObject collectDiffs(NodeBuilder indexDefinitions) {
-        JsonObject diffs = null;
-        for (String diffIndex : new String[] {
-                DiffIndexMerger.DIFF_INDEX,
-                DiffIndexMerger.DIFF_INDEX_OPTIMIZER }) {
+        JsonObject newImageLuceneDefinitions = null;
+        for (String diffIndex : new String[] { DiffIndexMerger.DIFF_INDEX, DiffIndexMerger.DIFF_INDEX_OPTIMIZER }) {
             if (!indexDefinitions.hasChildNode(diffIndex)) {
                 continue;
             }
@@ -85,12 +65,12 @@ public class DiffIndex {
             if (!diffContent.exists()) {
                 continue;
             }
-            PropertyState lastMod = diffContent.getProperty(NodeTypeConstants.JCR_LASTMODIFIED);
+            PropertyState lastMod = diffContent.getProperty("jcr:lastModified");
             if (lastMod == null) {
                 continue;
             }
             String modified = lastMod.getValue(Type.DATE);
-            PropertyState lastProcessed = diffContent.getProperty(DiffIndexMerger.LAST_PROCESSED);
+            PropertyState lastProcessed = diffContent.getProperty(":lastProcessed");
             if (lastProcessed != null) {
                 if (modified.equals(lastProcessed.getValue(Type.STRING))) {
                     // already processed
@@ -98,7 +78,7 @@ public class DiffIndex {
                 }
             }
             // store now, so a change is only processed once
-            diffContent.setProperty(DiffIndexMerger.LAST_PROCESSED, modified);
+            diffContent.setProperty(":lastProcessed", modified);
             PropertyState jcrData = diffContent.getProperty("jcr:data");
             String diff = tryReadString(jcrData);
             if (diff == null) {
@@ -107,47 +87,39 @@ public class DiffIndex {
             try {
                 JsonObject diffObj = JsonObject.fromJson("{\"diff\": " + diff + "}", true);
                 diffIndexDefinition.removeProperty("error");
-                if (diffs == null) {
-                    diffs = new JsonObject();
+                if (newImageLuceneDefinitions == null) {
+                    newImageLuceneDefinitions = new JsonObject();
                 }
-                diffs.getChildren().put("/oak:index/" + diffIndex, diffObj);
+                newImageLuceneDefinitions.getChildren().put("/oak:index/" + diffIndex, diffObj);
             } catch (Exception e) {
-                String message = "Error parsing " + diffIndex;
-                LOG.warn("{}: {}", message, e.getMessage(), e);
+                String message = "Error parsing diff.index";
+                LOG.warn(message + ": {}", e.getMessage(), e);
                 diffIndexDefinition.setProperty("error", message + ": " + e.getMessage());
             }
         }
-        return diffs;
-    }
-
-    /**
-     * Process the diffs.
-     *
-     * @param store the node store
-     * @param indexDefinitions the node builder for /oak:index
-     * @param diffs the json object with the combined diffs
-     */
-    private static void processDiffs(NodeStore store, NodeBuilder indexDefinitions, JsonObject diffs) {
-        LOG.info("Processing a diffs");
+        if (newImageLuceneDefinitions == null) {
+            // not a valid diff index, or already processed
+            return;
+        }
+        LOG.info("Processing a new diff.index with node store {}", store);
         JsonObject repositoryDefinitions = RootIndexesListService.getRootIndexDefinitions(indexDefinitions);
-        LOG.debug("Index list {}", repositoryDefinitions);
+        LOG.debug("Index list {}", repositoryDefinitions.toString());
         try {
-            MERGER.merge(diffs, repositoryDefinitions, store);
-            for (String indexPath : diffs.getChildren().keySet()) {
+            DiffIndexMerger.instance().merge(newImageLuceneDefinitions, repositoryDefinitions, store);
+            for (String indexPath : newImageLuceneDefinitions.getChildren().keySet()) {
                 if (indexPath.startsWith("/oak:index/" + DiffIndexMerger.DIFF_INDEX)) {
                     continue;
                 }
-                JsonObject newDef = diffs.getChildren().get(indexPath);
+                JsonObject newDef = newImageLuceneDefinitions.getChildren().get(indexPath);
                 String indexName = PathUtils.getName(indexPath);
-                JsonNodeUpdater.addOrReplace(indexDefinitions, store, indexName,
-                        IndexConstants.INDEX_DEFINITIONS_NODE_TYPE, newDef.toString());
+                JsonNodeBuilder.addOrReplace(indexDefinitions, store, indexName, IndexConstants.INDEX_DEFINITIONS_NODE_TYPE, newDef.toString());
                 updateNodetypeIndexForPath(indexDefinitions, indexName, true);
                 disableOrRemoveOldVersions(indexDefinitions, indexPath, indexName);
             }
             removeDisabledMergedIndexes(indexDefinitions);
             sortIndexes(indexDefinitions);
         } catch (Exception e) {
-            LOG.warn("Error merging diffs: {}", e.getMessage(), e);
+            LOG.warn("Error merging diff.index: {}", e.getMessage(), e);
             NodeBuilder diffIndexDefinition = indexDefinitions.child(DiffIndexMerger.DIFF_INDEX);
             diffIndexDefinition.setProperty("error", e.getMessage());
         }
@@ -165,7 +137,8 @@ public class DiffIndex {
         if (jcrData == null) {
             return null;
         }
-    try (InputStream in = jcrData.getValue(Type.BINARY).getNewStream()) {
+        InputStream in = jcrData.getValue(Type.BINARY).getNewStream();
+        try {
             return new String(in.readAllBytes(), StandardCharsets.UTF_8);
         } catch (IOException e) {
             LOG.warn("Can not read jcr:data", e);
@@ -185,12 +158,10 @@ public class DiffIndex {
     private static void removeDisabledMergedIndexes(NodeBuilder definitions) {
         ArrayList<String> toRemove = new ArrayList<>();
         for (String child : definitions.getChildNodeNames()) {
-            if (!definitions.getChildNode(child).hasProperty(DiffIndexMerger.MERGE_CHECKSUM)) {
+            if (!definitions.getChildNode(child).hasProperty("mergeChecksum")) {
                 continue;
             }
-            if (IndexConstants.TYPE_DISABLED.equals(definitions.
-                    getChildNode(child).
-                    getString(IndexConstants.TYPE_PROPERTY_NAME))) {
+            if ("disabled".equals(definitions.getChildNode(child).getString("type"))) {
                 toRemove.add(child);
             }
         }
@@ -223,9 +194,7 @@ public class DiffIndex {
             String childBaseName = IndexName.parse(child).getBaseName();
             if (baseName.equals(childBaseName)) {
                 if (indexName.equals(child)) {
-                    if (!IndexConstants.TYPE_DISABLED.equals(definitions.
-                            getChildNode(indexName).
-                            getString(IndexConstants.TYPE_PROPERTY_NAME))) {
+                    if (!"disabled".equals(definitions.getChildNode(indexName).getString("type"))) {
                         continue;
                     }
                 }
@@ -233,7 +202,7 @@ public class DiffIndex {
             }
         }
         for (String r : toRemove) {
-            LOG.info("Removing old index  {}", r);
+            LOG.info("Removing old index " + r);
             definitions.child(r).remove();
             updateNodetypeIndexForPath(definitions, r, false);
         }
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexName.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexName.java
@@ -234,7 +234,7 @@ public class IndexName implements Comparable<IndexName> {
         return result;
     }
 
-    public static Collection<String> filterNewestIndexes(Collection<String> indexPaths, NodeState rootState) {
+    public static Collection<String> filterNewestIndexes(Collection<String> indexPaths) {
         HashMap<String, IndexName> latestVersions = new HashMap<>();
         for (String p : indexPaths) {
             IndexName indexName = IndexName.parse(p);
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexName.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexName.java
@@ -234,6 +234,23 @@ public class IndexName implements Comparable<IndexName> {
         return result;
     }
 
+    public static Collection<String> filterNewestIndexes(Collection<String> indexPaths, NodeState rootState) {
+        HashMap<String, IndexName> latestVersions = new HashMap<>();
+        for (String p : indexPaths) {
+            IndexName indexName = IndexName.parse(p);
+            IndexName stored = latestVersions.get(indexName.baseName);
+            if (stored == null || stored.compareTo(indexName) < 0) {
+                // no old version, or old version is smaller: use
+                latestVersions.put(indexName.baseName, indexName);
+            }
+        }
+        ArrayList<String> result = new ArrayList<>(latestVersions.size());
+        for (IndexName n : latestVersions.values()) {
+            result.add(n.nodeName);
+        }
+        return result;
+    }
+
     public String nextCustomizedName() {
         return baseName + "-" + productVersion + "-custom-" + (customerVersion + 1);
     }
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
@@ -137,8 +137,7 @@ public class DiffIndex {
         if (jcrData == null) {
             return null;
         }
-        InputStream in = jcrData.getValue(Type.BINARY).getNewStream();
-        try {
+    try (InputStream in = jcrData.getValue(Type.BINARY).getNewStream()) {
             return new String(in.readAllBytes(), StandardCharsets.UTF_8);
         } catch (IOException e) {
             LOG.warn("Can not read jcr:data", e);
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
@@ -201,7 +201,7 @@ public class DiffIndex {
             }
         }
         for (String r : toRemove) {
-            LOG.info("Removing old index " + r);
+            LOG.info("Removing old index  {}", r);
             definitions.child(r).remove();
             updateNodetypeIndexForPath(definitions, r, false);
         }
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
@@ -116,6 +116,9 @@ public class DiffIndex {
                 LOG.warn("{}: {}", message, e.getMessage(), e);
                 diffIndexDefinition.setProperty("error", message + ": " + e.getMessage());
             }
+            if (!diffIndexDefinition.hasProperty("info")) {
+                diffIndexDefinition.setProperty("info", "This diff is are automatically merged with other indexes. See https://oak-indexing.github.io/oakTools/simplified.html");
+            }
         }
         return diffs;
     }
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
@@ -92,7 +92,7 @@ public class DiffIndex {
                 }
                 newImageLuceneDefinitions.getChildren().put("/oak:index/" + diffIndex, diffObj);
             } catch (Exception e) {
-                String message = "Error parsing diff.index";
+                String message = "Error parsing " + diffIndex;
                 LOG.warn(message + ": {}", e.getMessage(), e);
                 diffIndexDefinition.setProperty("error", message + ": " + e.getMessage());
             }
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
@@ -103,7 +103,7 @@ public class DiffIndex {
         }
         LOG.info("Processing a new diff.index with node store {}", store);
         JsonObject repositoryDefinitions = RootIndexesListService.getRootIndexDefinitions(indexDefinitions);
-        LOG.debug("Index list {}", repositoryDefinitions.toString());
+        LOG.debug("Index list {}", repositoryDefinitions);
         try {
             DiffIndexMerger.instance().merge(newImageLuceneDefinitions, repositoryDefinitions, store);
             for (String indexPath : newImageLuceneDefinitions.getChildren().keySet()) {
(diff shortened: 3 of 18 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
```diff
@@ -0,0 +1,242 @@
...
+        for (String diffIndex : new String[] { DiffIndexMerger.DIFF_INDEX, DiffIndexMerger.DIFF_INDEX_OPTIMIZER }) {
+            if (!indexDefinitions.hasChildNode(diffIndex)) {
+                continue;
+            }
+            NodeBuilder diffIndexDefinition = indexDefinitions.child(diffIndex);
+            NodeBuilder diffContent = diffIndexDefinition.getChildNode("diff.json").getChildNode("jcr:content");
+            if (!diffContent.exists()) {
+                continue;
+            }
+            PropertyState lastMod = diffContent.getProperty("jcr:lastModified");
```
bhabegger: Is there not a constant definition for this type of "well-known" entries ? (same for :lastProcessed, jcr:data, /oak:index/, etc.)

In my opinion this helps a lot to understand, identify in one go what are the 'system' properties, paths, etc., and know where they are used. But also have them documented in the code base as javadoc.

(Not a blocker, just a curiosity 😉)

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
```diff
@@ -0,0 +1,242 @@
...
+            }
+            try {
+                JsonObject diffObj = JsonObject.fromJson("{\"diff\": " + diff + "}", true);
+                diffIndexDefinition.removeProperty("error");
+                if (newImageLuceneDefinitions == null) {
+                    newImageLuceneDefinitions = new JsonObject();
+                }
+                newImageLuceneDefinitions.getChildren().put("/oak:index/" + diffIndex, diffObj);
+            } catch (Exception e) {
+                String message = "Error parsing diff.index";
```
bhabegger:
```suggestion
                String message = "Error parsing " + diffIndex;
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
```diff
@@ -0,0 +1,242 @@
...
+                LOG.warn(message + ": {}", e.getMessage(), e);
```
bhabegger:
```suggestion
                LOG.warn("{}: {}", message, e.getMessage(), e);
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
```diff
@@ -0,0 +1,242 @@
...
+
+    /**
+     * Apply changes to the index definitions. That means merge the index diff with
+     * the existing indexes, creating new index versions. It might also mean to
+     * remove old (merged) indexes if the diff no longer contains them.
+     *
+     * @param store            the node store
+     * @param indexDefinitions the /oak:index node
+     */
+    public static void applyDiffIndexChanges(NodeStore store, NodeBuilder indexDefinitions) {
```
bhabegger, line 60: To ease readability maybe we could split this method in :

-   parsing entries
-   diffing entries

and delegate the orchestration to a caller method ?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
```diff
@@ -0,0 +1,242 @@
...
+        }
+        if (newImageLuceneDefinitions == null) {
+            // not a valid diff index, or already processed
+            return;
+        }
+        LOG.info("Processing a new diff.index with node store {}", store);
+        JsonObject repositoryDefinitions = RootIndexesListService.getRootIndexDefinitions(indexDefinitions);
+        LOG.debug("Index list {}", repositoryDefinitions.toString());
+        try {
+            DiffIndexMerger.instance().merge(newImageLuceneDefinitions, repositoryDefinitions, store);
```
bhabegger: `merge` also seams to do something with diff.index and diff.index.optimizer couldn't we call the `mergeDiff` directly instead (without a dependency to the NodeStore)?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
```diff
@@ -0,0 +1,833 @@
...
+        // index definitions in the repository
+        combined.getChildren().putAll(repositoryDefinitions.getChildren());
+
+        // read the diff.index.optimizer explicitly,
+        // because it's a not a regular index definition,
+        // and so in the repositoryDefinitions
+        if (repositoryNodeStore != null) {
+            Map<String, JsonObject> diffInRepo = readDiffIndex(repositoryNodeStore, DIFF_INDEX_OPTIMIZER);
+            combined.getChildren().putAll(diffInRepo);
+        }
```
bhabegger, line 113: To favor separation of responsibilities, couldn't we allow this class to be agnostic of the NodeStore by pulling out of this class these few lines of code and the `readDiff` method ? This should be possible by passing `diffInRepo` as an argument instead.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
```diff
@@ -0,0 +1,833 @@
...
+            return null;
+        }
+        // extract either the file, or the nested json
+        JsonObject file = diffIndex.getChildren().get("diff.json");
+        JsonObject diff;
+        if (file != null) {
+            // file
+            JsonObject jcrContent = file.getChildren().get("jcr:content");
+            if (jcrContent == null) {
+                String message = "jcr:content child node is missing in diff.json";
```
bhabegger, line 200:
```suggestion
                String message = "jcr:content child node is missing in " + name;
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
```diff
@@ -0,0 +1,833 @@
...
+        obj.getProperties().remove(":version");
+        obj.getProperties().remove(":nameSeed");
+        obj.getProperties().remove(":mappingVersion");
+        obj.getProperties().remove("refresh");
+        obj.getProperties().remove("reindexCount");
+        obj.getProperties().remove("reindex");
+        obj.getProperties().remove("seed");
+        obj.getProperties().remove("merges");
+        obj.getProperties().remove("mergeInfo");
+        obj.getProperties().remove("mergeChecksum");
```
bhabegger, line 558: To me, this looks like a nice candidate for an enum, no ?


```suggestion
        for (ImportantProperty p : ImportantProperty.values()) {
            obj.getProperties().remove(p.getKey());
        }
```

And somewhere (with likely a better name):
```
    enum ImportantProperty {
        VERSION(":version"),
        MAPPING_VERSION(":mappingVersion"),
        NAME_SEED(":nameSeed"),
        REFRESH("refresh"),
        REINDEX_COUNT("reindexCount"),
        REINDEX("reindex"),
        SEED("seed"),
        MERGES("merges"),
        MERGE_INFO("mergeInfo"),
        MERGE_CHECKSUM("mergeChecksum");

        private final String key;
        ImportantProperty(String key) {
            this.key = key;
        }
        public String getKey() {
              return this.key;
        }
    }
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
```diff
@@ -0,0 +1,833 @@
...
+     */
+    private void mergeInto(String path, JsonObject diff, JsonObject target) {
+        for (String p : diff.getProperties().keySet()) {
+            if (path.isEmpty()) {
+                if ("jcr:primaryType".equals(p)) {
+                    continue;
+                }
+            }
+            if (target.getProperties().containsKey(p)) {
+                // we do not currently allow to overwrite most existing properties
```
bhabegger, line 684:
```suggestion
                // we only support overwriting a limited number of properties
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
```diff
@@ -0,0 +1,833 @@
...
+                    // only create the child (properties are added below)
+                    target.getChildren().put(c, new JsonObject());
+                }
+            }
+            mergeInto(path + "/" + targetChildName, diff.getChildren().get(c), target.getChildren().get(targetChildName));
+        }
+        if (target.getProperties().isEmpty() && target.getChildren().isEmpty()) {
+            if (deleteCreatesDummyIndex) {
+                // dummy index
+                target.getProperties().put("async", "\"async\"");
```
bhabegger, line 729: Am I correctly understanding that this is to provide a basic sample index definition in case the merging led to completely drop all properties ? Shouldn't the merge fail in case it leads to such a situation ?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
```diff
@@ -0,0 +1,833 @@
...
+     *
+     * @param indexName                 the name, eg. "damAssetLucene"
+     * @param indexDiff                 the diff with the new properties
+     * @param newImageLuceneDefinitions the new Lucene definitions (input + output)
+     * @param combined                  the definitions in the repository, including
+     *                                  the one in the customer repo and new ones
+     *                                  (input)
+     * @return whether a new version of an index was added
+     */
+    public boolean processMerge(String indexName, JsonObject indexDiff, JsonObject newImageLuceneDefinitions, JsonObject combined) {
```
bhabegger, line 295: This method is a bit hard to digest when reading it. Could we maybe outsource some of the logic ?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
```diff
@@ -0,0 +1,833 @@
...
+            if (includesUnsupportedPaths(includedPaths)) {
+                LOG.warn("Customizing index {} is not supported because it contains an unsupported path ({})",
+                        latestProductKey, Arrays.toString(unsupportedIncludedPaths));
+                return false;
+            }
+        }
+
+        // merge
+        JsonObject merged = null;
+        if (indexDiff == null) {
```
bhabegger, line 363: As an example, maybe this premerge null checking logic could be moved down into the processMerge itself ? (The "nothing to do" case could be handled by a null or Optional merged result)

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/JsonNodeBuilder.java
```diff
@@ -0,0 +1,279 @@
...
+ *
+ * A "jcr:uuid" is automatically added for nodes of type "nt:resource".
+ *
+ * String, string arrays, boolean, blob, long, and double values are supported.
+ * Values that start with ":blobId:...base64..." are stored as binaries. "str:",
+ * "nam:" and "dat:" prefixes are removed.
+ *
+ * "null" entries are not supported.
+ */
+public class JsonNodeBuilder {
```
bhabegger: The `Builder` in the name could be misleading as, unless I missed something. this class doesn't seem to follow the "Builder" pattern, and seams more like a Helper class (but then there are different style tastes out their).

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/RootIndexesListService.java
```diff
@@ -0,0 +1,112 @@
...
+ * limitations under the License.
+ */
+package org.apache.jackrabbit.oak.plugins.index.diff;
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.ArrayList;
+
+import org.apache.felix.inventory.Format;
+import org.apache.jackrabbit.oak.commons.json.JsonObject;
```
bhabegger, line 24: I'm guessing that OAK has it's own JSON processing library for historical reasons ?

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexTest.java
```diff
@@ -0,0 +1,307 @@
...
+                + "                        \"name\": \"str:jcr:modified\",\n"
+                + "                        \"propertyIndex\": true\n"
+                + "                    }\n"
+                + "                }\n"
+                + "            }\n"
+                + "        }\n"
+                + "    } }");
+
+        repositoryDefinitions = RootIndexesListService.getRootIndexDefinitions(store, "lucene");
+        assertSameJson("{\n"
```
bhabegger, line 217: It's a shame Java 11 is a bit too early for text blocks :)

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/diff/MergeTest.java
```diff
@@ -0,0 +1,437 @@
...
+                + "    }\n"
+                + "  }\n"
+                + "}", merged);
+    }
+
+    @Test
+    public void createDummy() {
+        // when enabling "deleteCreatesDummyIndex", then a dummy index is created
+        // (that indexes /dummy, which doesn't exist)
+        String merged = new DiffIndexMerger(new String[0], true, true, false).processMerge(JsonObject.fromJson("{}"
```
bhabegger, line 221: Which true is it ? 

```suggestion
        boolean deleteCreatesDummyIndex = true;
        String merged = new DiffIndexMerger(new String[0], deleteCreatesDummyIndex, true, false).processMerge(JsonObject.fromJson("{}"
```

Would make it explicit

bhabegger (commented):
Nice clean work 👍 

I like the test coverage and the comments among the rest !

I don't see anything blocking and just have a few comments/suggestions.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/diff/MergeTest.java
thomasmueller: Hm, "list of boolean flags" is not user friendly... thinking about an alternative.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexTest.java
thomasmueller: I agree (btw. text blocks are not "raw" and still need escaping). 

But I think it's readable as is.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/RootIndexesListService.java
thomasmueller: Yes.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/JsonNodeBuilder.java
thomasmueller: Right... It's a converter from "Json" to "NodeBuilder". It's confusing. 

What about JsonNodeWriter, or JsonToNodeConverter?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
thomasmueller: There probably is. I agree we are not consistently using these constants, so there is a risk of typos.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
```diff
@@ -0,0 +1,242 @@
...
+     *
+     * @param jcrData the "jcr:data" property
+     * @return the string, or null if reading fails
+     */
+    public static String tryReadString(PropertyState jcrData) {
+        if (jcrData == null) {
+            return null;
+        }
+        InputStream in = jcrData.getValue(Type.BINARY).getNewStream();
+        try {
```
amit-jain:
```suggestion 
    try (InputStream in = jcrData.getValue(Type.BINARY).getNewStream()) {
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/IndexName.java
```diff
@@ -234,6 +234,23 @@ public static Collection<String> filterReplacedIndexes(Collection<String> indexP
         return result;
     }
 
+    public static Collection<String> filterNewestIndexes(Collection<String> indexPaths, NodeState rootState) {
```
amit-jain: rootState ununsed parameter, consider removing

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
```diff
@@ -0,0 +1,833 @@
...
+     * @return the SHA-256 checksum
+     */
+    private static String computeMergeChecksum(JsonObject json) {
+        byte[] bytes = json.toString().getBytes(StandardCharsets.UTF_8);
+        try {
+            MessageDigest md = MessageDigest.getInstance("SHA-256");
+            return StringUtils.convertBytesToHex(md.digest(bytes));
+        } catch (NoSuchAlgorithmException e) {
+            // SHA-256 is guaranteed to be available in standard Java platforms
+            throw new RuntimeException("SHA-256 algorithm not available", e);
```
amit-jain: Maybe IllegalStateException is better?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
```diff
@@ -0,0 +1,833 @@
...
+            }
+        }
+        merged.getProperties().put("mergeInfo", JsopBuilder.encode(MERGE_INFO));
+        merged.getProperties().put("mergeChecksum", JsopBuilder.encode(mergeChecksum));
+        merged.getProperties().put("merges", "[" + JsopBuilder.encode("/oak:index/" + indexName) + "]");
+        merged.getProperties().remove("reindexCount");
+        merged.getProperties().remove("reindex");
+        if (!deleteCopiesOutOfTheBoxIndex && indexDiff.toString().equals("{}")) {
+            merged.getProperties().put("type", "\"disabled\"");
+            merged.getProperties().put("mergeComment", "\"This index is superseeded and can be removed\"");
```
amit-jain:
```suggestion
            merged.getProperties().put("mergeComment", "\"This index is superseded and can be removed\"");
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
```diff
@@ -0,0 +1,242 @@
...
+                if (indexName.equals(child)) {
+                    if (!"disabled".equals(definitions.getChildNode(indexName).getString("type"))) {
+                        continue;
+                    }
+                }
+                toRemove.add(child);
+            }
+        }
+        for (String r : toRemove) {
+            LOG.info("Removing old index " + r);
```
amit-jain:
```suggestion
            LOG.info("Removing old index  {}", r);
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
amit-jain: yes recommend breaking this method, example (from comments)
- findLatestVersions
- performMerge
- isChanged
- addMetadata

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
```diff
@@ -0,0 +1,242 @@
...
+                diffIndexDefinition.setProperty("error", message + ": " + e.getMessage());
+            }
+        }
+        if (newImageLuceneDefinitions == null) {
+            // not a valid diff index, or already processed
+            return;
+        }
+        LOG.info("Processing a new diff.index with node store {}", store);
+        JsonObject repositoryDefinitions = RootIndexesListService.getRootIndexDefinitions(indexDefinitions);
+        LOG.debug("Index list {}", repositoryDefinitions.toString());
```
amit-jain:
```suggestion
        LOG.debug("Index list {}", repositoryDefinitions);
```

amit-jain (approved):
Overall looks good to me!!
A few minor suggestions added.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/JsonNodeBuilder.java
bhabegger: I would prefer Converter among the two as Writer is quite connotated (PrintWriter, StringWriter, etc.). However, both Writer and Converter still suggests some sort of instantiation and don't highlight that this class is a utility class.

bhabegger: JsonNodeConversionUtil ? 
JsonNodeConversionHelper ?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/JsonNodeBuilder.java
```diff
@@ -0,0 +1,279 @@
...
+     * @param nodeStore the target node store
+     * @param targetPath the target path where the node(s) is/are replaced
+     * @param nodeType the node type of the new node (eg. "nt:unstructured")
+     * @param jsonString the json string with the node data
+     * @throws CommitFailedException if storing the nodes failed
+     * @throws IOException if storing a blob failed
+     */
+    public static void addOrReplace(NodeBuilder builder, NodeStore nodeStore, String targetPath, String nodeType, String jsonString) throws CommitFailedException, IOException {
+        LOG.info("Storing {}: {}", targetPath, jsonString);
+        if (nodeType.indexOf("/") >= 0) {
```
bhabegger:
```suggestion
        if (nodeType.contains("/")) {
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
thomasmueller: We want to re-use the "merge" feature in another project (when processing diffs that are sent via script, so come from external sources). So it's better to keep `DiffIndexMerger#merge`

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
thomasmueller: There is some code executed in this case (latestCustomized == null...) so that would require changing `processMerge`, which I would avoid.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/JsonNodeBuilder.java
thomasmueller: Hm, it's not really a converter... the nodes are persisted in Oak, and "converter" (for me) would imply that the nodes are just converted and the result is returned.

It's mostly about creating, and possibly updating and deleting, nodes... So what about:
* JsonNodeSmith (recommended by a tool; sounds right to me)
* JsonNodeCreator (because it's mostly about creating new nodes... but not purely)
* JsonNodeWriter, for the case it's updating / deleting (currently, I think it's not, but possibly in the future)

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
thomasmueller: I now split it into two parts that I think make sense.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/diff/MergeTest.java
thomasmueller: I added setters (there no real harm to not have the fields final in this case). Sure we could also add a Builder. Or a Factory, or AdapterFactoryBuilder... :-)

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/JsonNodeBuilder.java
bhabegger: Ok, then. Writer could pass, because it does I/O even though it's not in the usual sense. JsonNodeSaver ? JsonNodeStorer ? JsonNodeManager ? JsonNodeUpdater ? 

I'm not really convinced by any... But maybe it triggers need ideas ? Otherwise just go with your preference.

bhabegger: Ok, then. Writer could pass, because it does I/O even though it's not in the usual sense. JsonNodeSaver ? JsonNodeStorer ? JsonNodeManager ? JsonNodeUpdater ? 

I'm not really convinced by any... But maybe it triggers need ideas ? Otherwise just go with your preference.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
thomasmueller: I used constants now where available, and did add new ones where I think it is useful (which -- granted -- is subjective).

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/JsonNodeBuilder.java
thomasmueller: OK so I use JsonNodeUpdater.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
bhabegger:
```
        if (indexDiff == null) {
            // no diff definition: use to the OOTB index
            if (latestCustomized == null) {
                log("Only a product index found, nothing to do");
                return false;
            }
            merged = latestProductIndex;
        } else {
            merged = processMerge(latestProductIndex, indexDiff);
        }
```

Could be rewritten:

```
           if(indexDiff == null && latestCustomized == null) {
                log("Only a product index found, nothing to do");
                return false;
           }
           merged = processMerge(latestProductIndex, indexDiff);
```
with processMerge returning latestProductIndex if indexDiff is null. 

we could even let it return null if both are null and exit fast if merged is null with the log.

thomasmueller: Well it would move part of the logic to processMerge, which I would like to avoid. Also, the comment is lost.

bhabegger: Ok, that's fine then.

bhabegger: Ok, that's fine then.

bhabegger: Ok, that's fine then.

oak-core/DiffIndex.java
ChlineSaurus, line 1: What's the purpose of the four files on the oak-core root level? For me it looks like duplicate code to code in /oak/plugins/diff and to the test code 🤔

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
```diff
@@ -0,0 +1,853 @@
...
+                    latestProduct = name;
+                    latestProductKey = key;
+                }
+            }
+        }
+        log("Latest product: {}", latestProductKey);
+        log("Latest customized: {}", latestCustomizedKey);
+        if (latestProduct == null) {
+            if (indexName.indexOf('.') >= 0) {
+                // a fully custom index needs to contains a dot
```
ChlineSaurus, line 331: Detail: Typo, should be "contain"

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
```diff
@@ -0,0 +1,853 @@
...
+            if (latestCustomized == null) {
+                log("Only a product index found, nothing to do");
+                return false;
+            }
+            merged = latestProductIndex;
+        } else {
+            merged = processMerge(latestProductIndex, indexDiff);
+        }
+
+        // compare to the latest version of the this index
```
ChlineSaurus, line 374: Detail: Typo "the this index" to "the index"

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
```diff
@@ -0,0 +1,276 @@
...
+            String modified = lastMod.getValue(Type.DATE);
+            PropertyState lastProcessed = diffContent.getProperty(DiffIndexMerger.LAST_PROCESSED);
+            if (lastProcessed != null) {
+                if (modified.equals(lastProcessed.getValue(Type.STRING))) {
+                    // already processed
+                    continue;
+                }
+            }
+            // store now, so a change is only processed once
+            diffContent.setProperty(DiffIndexMerger.LAST_PROCESSED, modified);
```
ChlineSaurus, line 101: In my understanding, we always (even in case of failure) set the LAST_PROCESSED property, as it is set before the processing, and in the case of failure will be committed when the error is committed. Thus we don't have a retry in case of failure. Do I understand this right? And is this intentional?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
ChlineSaurus: I see a different problem.
In my understanding "if (indexDiff.getProperties().isEmpty() && ..." on line 341 could trigger a NPE before this null check happens. I'm not sure if it could occur in a full flow, but it's a public method and in a unit test I was able to get an NPE.

ChlineSaurus (commented):
Overall looks good to me, 😌 good clean code
I added a few comments

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndex.java
thomasmueller: Retry wouldn't make sense: it would fail again in the same way.

oak-core/DiffIndex.java
thomasmueller: Thanks a lot! I committed them by mistake, now removed.

PR #2690, opened 2026-01-19
OAK-12063: Remove obsolete feature toggles FT_NAME_IMPROVED_IS_NULL_COST, FT_OPTIMIZE_IN_RESTRICTIONS_FOR_FUNCTIONS, and FT_NAME_PREFETCH_FOR_QUERIES

bhabegger:
See [OAK-12063 - Feature toggle cleanup](https://issues.apache.org/jira/browse/OAK-12063)

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
@@ -61,12 +61,6 @@ public class QueryEngineSettings implements QueryEngineSettingsMBean, QueryLimit
 
     public static final String OAK_QUERY_PREFETCH_COUNT = "oak.prefetchCount";
 
-    public static final String FT_NAME_PREFETCH_FOR_QUERIES = "FT_OAK-10490";
-
-    public static final String FT_NAME_IMPROVED_IS_NULL_COST = "FT_OAK-10532";
-
-    public static final String FT_OPTIMIZE_IN_RESTRICTIONS_FOR_FUNCTIONS = "FT_OAK-11214";
-
     public static final String FT_SORT_UNION_QUERY_BY_SCORE = "FT_OAK-11949";
 
     public static final String FT_OPTIMIZE_XPATH_UNION = "FT_OAK-12007";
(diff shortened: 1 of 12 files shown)


oak-query-spi/src/main/java/org/apache/jackrabbit/oak/spi/query/package-info.java
```diff
@@ -18,7 +18,7 @@
 /**
  * This package contains oak query index related classes.
  */
-@Version("3.2.0")
+@Version("4.0.0")
```
bhabegger: Not sure about this...

joerghoh: removing methods from the public API is an incompatible change, which requires a bump of the major package version. I would like to avoid as much as possible.

bhabegger: Ok, I understand. How about I mark them as deprecated and remove their effect ? The goal here is to remove obsolete code. 

Or is there some deprecation strategy that's in place ?

joerghoh: the official deprecation strategy is just to mark functionality as such, I am not aware of a process to actually remove deprecate functionality (which is part of the exposed API) from the OAK codebase.

bhabegger: Rolled this back

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java
```diff
@@ -63,10 +63,6 @@ public class QueryEngineSettings implements QueryEngineSettingsMBean, QueryLimit
 
     public static final String FT_NAME_PREFETCH_FOR_QUERIES = "FT_OAK-10490";
```
thomasmueller: Could you also remove FT_OAK-10490 please? Thanks a lot!

thomasmueller (approved):

PR #2692, opened 2026-01-20
OAK-12065 : added class to split locks similar to guava's Striped

rishabhdaim:


oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/internal/concurrent/StripedLocks.java
```diff
@@ -0,0 +1,52 @@
...
+        for (int i = 0; i < stripesCount; i++) {
+            this.stripes[i] = new ReentrantLock();
+        }
+    }
+
+    public Lock get(Object key) {
+        int h = (key == null) ? 0 : key.hashCode();
+        // Spread bits to reduce collisions, similar to ConcurrentHashMap
+        h ^= (h >>> 16);
+        int index = (h & 0x7fffffff) % stripes.length;
```
thomasmueller:
```suggestion
        // Spread bits to reduce collisions, using
        // https://stackoverflow.com/questions/664014/what-integer-hash-function-are-good-that-accepts-an-integer-hash-key/12996028#12996028
        h *= 0x45d9f3b;
        // http://lemire.me/blog/2016/06/27/a-fast-alternative-to-the-modulo-reduction/
        int index = (int) (((h & 0xffffffffL) * len) >>> 32);
```

thomasmueller (commented):
I have two small improvements:

* the reminder operation (%) is slow on most CPUs (all except for Apple silicon)
* the "xor shifted by 16 bit" doesn't fully resolve collisions for short strings / doubles (see example below); multiplication is better

Test case:

```
    public static void main(String... args) {
        for(int repeat = 0; repeat < 5; repeat++) {
            int test = 1_000_000;
            for(int len=1; len<100; len+=3) {
                int[] counts = new int[len];
                for(int i=0; i<test / 2; i++) {
                    counts[reduce("" + i, len)]++;
                    counts[reduce(Double.valueOf(i), len)]++;
                }
                for(int i=0; i<len; i++) {
                    if (Math.abs(counts[i] - test / len) > 0.03 * (test / len)) {
                        System.out.println("len " + len + " slot " + i + " count " + counts[i] + " expected " + (test / len));
                    }
                }
            }
            long time = System.nanoTime() - start;
            System.out.println("time used: " + time / test + " ns/key");
        }
    }

    public static int reduce1(Object key, int len) {
        int h = (key == null) ? 0 : key.hashCode();
        // Spread bits to reduce collisions, similar to ConcurrentHashMap
        h ^= (h >>> 16);
        int index = (h & 0x7fffffff) % len;
        return index;
    }

    public static int reduce(Object key, int len) {
        int h = (key == null) ? 0 : key.hashCode();
        // Spread bits to reduce collisions, using
        // https://stackoverflow.com/questions/664014/what-integer-hash-function-are-good-that-accepts-an-integer-hash-key/12996028#12996028
        h *= 0x45d9f3b;
        // http://lemire.me/blog/2016/06/27/a-fast-alternative-to-the-modulo-reduction/
        int index = (int) (((h & 0xffffffffL) * len) >>> 32);
        return index;
    }
```

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/internal/concurrent/StripedLocks.java
rishabhdaim: added in https://github.com/apache/jackrabbit-oak/pull/2692/commits/a8546e7120c75e5a2b5be5a7f017c53a09730008

thomasmueller (approved):

PR #2721, opened 2026-02-02
OAK-12072 : removed Guava's Monitor and Guard

rishabhdaim:


rishabhdaim:
@kwin, could you please have a look at this PR ?

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/observation/ChangeProcessor.java
```diff
@@ -460,9 +464,15 @@ public synchronized boolean stopAndWait(int timeOut, TimeUnit unit) {
      */
     public synchronized void stop() {
         Validate.checkState(registration != null, "Change processor not started");
-        if (running.stop()) {
-            registration.unregister();
-            runningMonitor.leave();
+
+        if (stopFlagSet()) {
+            // Wait until no contentChanged is in the critical section
+            runningLock.lock();
```
mbaedke, line 470: At first glance, this is exactly what to original method avoids (JavaDoc: "...this method returns immediately without waiting for pending listeners to complete...").

mbaedke (changes requested):

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/observation/ChangeProcessor.java
rishabhdaim: I am sorry, I didn't understand the comment.

Original code was :

```
if (running.stop()) {
            registration.unregister();
            runningMonitor.leave();
        }
```

And I have changed it to below

```
if (stopFlagSet()) {
            // Wait until no contentChanged is in the critical section
            runningLock.lock(); 
            try {
                registration.unregister();
            } finally {
                runningLock.unlock();
            }
        }
```

and it is only `unregisters` if the current thread is setting the flag to stop.

mbaedke: My point is that the original code explicitly does _not_ lock before calling `registration.unregister()`.

mbaedke (changes requested):

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/observation/ChangeProcessor.java
rishabhdaim: Yes, it doesn't enter the `Monitor`, but IMO, it should.
this `stop()` is called from [ObservationManagerImpl#stop()](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/observation/ObservationManagerImpl.java#L462) when we are not able to acquire the lock within `timeout` and then proceed to `stop` it without any timeout.

rishabhdaim: Created a test case to confirm this:
```

@Test(expected=IllegalStateException.class)
    public void testStopWithoutEnterThrows() {
        // Create a real ChangeProcessor instance with minimal setup using mocks
        ContentSession contentSession = Mockito.mock(ContentSession.class);
        NamePathMapper namePathMapper = Mockito.mock(NamePathMapper.class);
        ListenerTracker tracker = Mockito.mock(ListenerTracker.class);
        FilterProvider filter = Mockito.mock(FilterProvider.class);
        StatisticManager statisticManager = Mockito.mock(StatisticManager.class);
        int queueLength = 1;
        CommitRateLimiter commitRateLimiter = Mockito.mock(CommitRateLimiter.class);
        BlobAccessProvider blobAccessProvider = Mockito.mock(BlobAccessProvider.class);
        ChangeProcessor changeProcessor = new ChangeProcessor(
            contentSession,
            namePathMapper,
            tracker,
            filter,
            statisticManager,
            queueLength,
            commitRateLimiter,
            blobAccessProvider
        );
        // Use reflection or a test hook to set registration to a dummy CompositeRegistration
        CompositeRegistration registration = Mockito.mock(CompositeRegistration.class);
        java.lang.reflect.Field regField;
        try {
            regField = ChangeProcessor.class.getDeclaredField("registration");
            regField.setAccessible(true);
            regField.set(changeProcessor, registration);
        } catch (Exception e) {
            throw new RuntimeException(e);
        }
        // Now call stop(), which should throw due to leave() without enter()
        changeProcessor.stop();
    }
```

We must own the `lock` beforehand while calling `unlock`.

Now, I am really confused here, whether this code has been written incorrectly since the beginning OR I am missing something.

rishabhdaim: Change was added in https://svn.apache.org/viewvc/jackrabbit/oak/trunk/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/observation/ChangeProcessor.java?pathrev=1551502&r1=1551502&r2=1551501#content

rishabhdaim: @mbaedke added test cases for the change.

reschke (approved):
In this case, I would just rely on CI passing.

PR #2722, opened 2026-02-03
OAK-12057 - Select same plan with or without LIMIT option

bhabegger:


oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
```diff
@@ -1117,12 +1115,6 @@ private SelectorExecutionPlan getBestSelectorExecutionPlan(
                     if (p.getSupportsPathRestriction()) {
                         entryCount = scaleEntryCount(rootState, filter, entryCount);
                     }
-                    if (sortOrder == null || p.getSortOrder() != null) {
-                        // if the query is unordered, or
-                        // if the query contains "order by" and the index can sort on that,
-                        // then we don't need to read all entries from the index
-                        entryCount = Math.min(maxEntryCount, entryCount);
-                    }
```
bhabegger, line 1125: @thomasmueller This is the part that leads to the change in costs of the broken test

oak-jcr/src/test/java/org/apache/jackrabbit/oak/IndexCostEvaluationTest.java
```diff
@@ -36,66 +37,59 @@
...
     }
 
     // In cases where two indexes have same min cost i.e. both indexes are on par, we don't skip cost evaluation
     // even of cost from previous index is less than min cost of new index.
     @Test
     public void costEvaluationTest() throws Exception {
+        String query = "SELECT * FROM [rep:Authorizable] WHERE [rep:principalName] = 'anonymous'";
+        session.getLatestRoot().getQueryEngine().executeQuery(query, JCR_SQL2, 1, 0, null, null);
+
         boolean evaluationContinueLogPresent = false;
```
bhabegger, line 92: I simplified the test a bit to reduce it to one explicit query rather than 3 implicit with one of them generating the expected log entries.

(It's in a separate commit so that we can check that it passed before, and that now it no longer passes)

oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryImpl.java
bhabegger, line 1093: Concerning the broken test.

I've try to look into this a bit deeper. In the original case, as there was a limit of 1 so the best cost of test-index1 up around the minimal cost of test-index3  with the latter still bigger than the former but then passing through this if.

Currently, dropping the limit all indices are evaluated as their minCost is definitely below the bestCost.

So, I wondering, is the scenario that lead to this situation still valid ? If so, what case would that be and how to simulate it ?

thomasmueller: I'll check. Maybe the test needs to be changed.

thomasmueller:
Merging it, but only into the branch

PR #2744, opened 2026-02-16
OAK-12098: blob-plugins - remove use of TransientFileFactory (part 2)

reschke:


reschke:
OK, this fails on Linux because the file *entries* are removed earlier than expected by the test.

However, `TransientFileFactory` does not guarantee when the `File` objects are closed. It could happen late, it could happen at VM shutdown, or it could happen *earlier* than expected. This was verified with a version of the Factory that adds a method to run the cleanup immediately. I'll conclude that the tests checking the contents of the temp directory are broken.

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/AbstractSharedCachingDataStore.java
```diff
@@ -323,11 +322,16 @@ public InputStream getStream() throws DataStoreException {
             try {
                 // If cache configured to 0 will return null
                 if (cached == null || !cached.exists()) {
-                    TransientFileFactory fileFactory = TransientFileFactory.getInstance();
-                    File tmpFile = fileFactory.createTransientFile("temp0cache", null, temp);
+                    final File tmpFile = Files.createTempFile(temp.toPath(), "blob-cache-", null).toFile();
                     try (InputStream in = backend.getRecord(getIdentifier()).getStream()) {
                         copyInputStreamToFile(in, tmpFile);
-                        return new LazyFileInputStream(tmpFile);
+                        return new FileInputStream(tmpFile);
```
jsedding: You might  have better luck returning an `InputStream` that deletes the `tmpFile` on `InputStream#close()`. However, that requires all users of the API to actually close the stream.

Of course this could be combined with the `TransientFileFactory` as a safeguard. I.e. the `TransientFileFactory` would only clean up after non-conforming users.

Out of curiosity, why do you want to get rid of the `TransientFileFactory`? Are there problems with it?

rishabhdaim: Won't this cause issues, since we are deleting the file before returning the `FileInputStream`?

cc @amit-jain

rishabhdaim: Also, this won't work well in Windows:

- On Windows, you cannot delete a file that is open by any process.
- On Unix-like systems, you can delete an open file, and the file's data remains accessible to processes that have it open until all handles are closed.

So basically, on Windows, we would be having garbage files under tmp folders since we are opening a Stream before calling delete.

See : https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-deletefilea

rishabhdaim: Created a new to check this on MacOS:

```
private Object[] createTempFileAndReturnStream() throws Exception {
        File tempFile = File.createTempFile("test", ".txt");
        System.out.println(tempFile.getAbsolutePath());
        try (FileOutputStream fos = new FileOutputStream(tempFile)) {
            fos.write("hello world".getBytes());
        }
        FileInputStream fis = new FileInputStream(tempFile);
        // Delete the file immediately after opening the stream
        tempFile.delete();
        return new Object[]{fis, tempFile};
    }

    @Test
    public void testWaitForBucketNotFound() throws Exception {
        FileInputStream fis = null;
        String name = null;
        try {
            Object[] tempFileAndReturnStream = createTempFileAndReturnStream();
            fis = (FileInputStream) tempFileAndReturnStream[0];
            name = tempFileAndReturnStream[1].toString();
            byte[] buf = new byte[11];
            int read = fis.read(buf);
            Assert.assertEquals(11, read);
            Assert.assertEquals("hello world", new String(buf));
            System.out.println(Path.of(name));
            Assert.assertFalse(Files.exists(Path.of(name)));
        } finally {
            if (fis != null) {
                fis.close();
            }
            Assert.assertFalse(Files.exists(Path.of(name)));
        }
    }
```

On Mac, it is working fine, but I couldn't test on Windows since I don't have one.

@Joscorbe could you please try the above test on Windows if possible?

Joscorbe: @rishabhdaim 

```
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.357 s <<< FAILURE! -- in org.apache.jackrabbit.oak.plugins.blob.TransientFileFactoryTest
[ERROR] org.apache.jackrabbit.oak.plugins.blob.TransientFileFactoryTest.testWaitForBucketNotFound -- Time elapsed: 0.320 s <<< FAILURE!
java.lang.AssertionError
        at org.junit.Assert.fail(Assert.java:87)
        at org.junit.Assert.assertTrue(Assert.java:42)
        at org.junit.Assert.assertFalse(Assert.java:65)
        at org.junit.Assert.assertFalse(Assert.java:75)
        at org.apache.jackrabbit.oak.plugins.blob.TransientFileFactoryTest.testWaitForBucketNotFound(TransientFileFactoryTest.java:67)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
        at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
        at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
        at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
        at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
        at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
        at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
        at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
        at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:316)
        at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:240)
        at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:214)
        at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:155)
        at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:385)
        at org.apache.maven.surefire.booter.ForkedBooter.execute(ForkedBooter.java:162)
        at org.apache.maven.surefire.booter.ForkedBooter.run(ForkedBooter.java:507)
        at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:495)

[INFO]
[INFO] Results:
[INFO]
[ERROR] Failures:
[ERROR]   TransientFileFactoryTest.testWaitForBucketNotFound:67
[INFO]
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0
```

reschke:
1) @jsedding - we (I) want to get rid of uses of `TransientFileFactory` because it gives a false feeling of safety where there is none (GC, VM shutdown..)
2) @rishabhdaim  - It works on Windows. Despite urban legends, Windows can delete open files. The only difference is that directory entries remain visible until the last open handle is gone (EDIT: and the file was opened with lock, which is not the case here)
3) @Joscorbe - Tests (currently) fail because they make incorrent assumptions about when file entries go away (in this case, caused by the way `TransientFileFactory` works; they rely on file entries stay there despite `TransientFileFactory` might have removed them early).
4) @rishabhdaim - timing of file removal - let me check - where do you see that?

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/AbstractSharedCachingDataStore.java
reschke: @Joscorbe - can you please share the test?

reschke:
ok, on Windows, the file is indeed removed only one the process terminates (if it's open locked, which is the case here, with shared access it would work).

(and because the test VM did terminate....)

Otherwise, on non-Windows systems it will only disappear once removed by TransientFileWhatever. If that happens to soon, tests will fail because they expect the file still to be present (despite it was closed).

reschke:
This:

```
    private Object[] createTempFileAndReturnStream() throws Exception {
        File tempFile = File.createTempFile("test", ".txt");
        System.out.println(tempFile.getAbsolutePath());
        try (FileOutputStream fos = new FileOutputStream(tempFile)) {
            fos.write("hello world".getBytes());
        }
        InputStream fis = Files.newInputStream(tempFile.toPath(), StandardOpenOption.DELETE_ON_CLOSE);
        return new Object[]{fis, tempFile};
    }
```
seems to work.

oak-blob-plugins/src/main/java/org/apache/jackrabbit/oak/plugins/blob/AbstractSharedCachingDataStore.java
Joscorbe: @reschke I have attached it to the ticket: https://issues.apache.org/jira/browse/OAK-12098

PR #2745, opened 2026-02-17
OAK-12099 add AGENTS.md

smiroslav:
OAK-12099 initial version of AGENTS.md


thomasmueller (approved):

AGENTS.md
```diff
@@ -0,0 +1,236 @@
...
+
+Apache Jackrabbit Oak is a scalable, high-performance hierarchical content repository
+implementing the JCR (Java Content Repository) specification. It is a multi-module Maven
+project with ~47 modules written in Java 11.
+
+## General Guidelines
+
+- When working on a specific area, first read the area-specific documentation before
+  making changes (see doc links in the module tables below)
+- New code must have high test coverage — all critical paths must be covered by tests
```
anchela: i would add a number here (>90%) and also mention that AI should never modify existing tests in order to make the build pass after changing something.
specially in the security area we have up to 100% test coverage to maximize scrutiny on modifications as this is crucial code

AGENTS.md
```diff
@@ -0,0 +1,236 @@
...
+- Write self-descriptive, easy-to-read code. Avoid trivial comments. Only add comments
```
anchela: i would also add here that changes should be done with minimal amount of code. AI has a tendency to maximize the number of lines not to be as minimalistic as possible.

on top: i believe this file should describe what is considered trivial. certain areas in oak are crucial to overall behavior of the repository and what looks like a trivial change to AI may introduce massive regressions.

AGENTS.md
```diff
@@ -0,0 +1,236 @@
...
+  where the logic is complex and not obvious from the code itself
+- Use feature toggles for non-trivial changes. If the change is a bug fix, the toggle
+  should be enabled by default. If the change introduces a new feature, the toggle should
+  be disabled by default
```
anchela, line 24: on top we may want to specify that public API and SPI contracts cannot be automatically modified.

finally: in oak we want every change to be backed by a JIRA ticket. i would bake that into the instructions here.

AGENTS.md
```diff
@@ -0,0 +1,236 @@
...
+
+```bash
+# Full build (skip tests for speed)
+mvn clean install -DskipTests
+
+# Fast build (no tests, no coverage)
+mvn clean install -Pfast
+
+# Build a single module
+mvn clean install -pl oak-core -DskipTests
```
anchela, line 36: why skipping tests?

AGENTS.md
```diff
@@ -0,0 +1,236 @@
...
+
+# Build a module and its dependencies
+mvn clean install -pl oak-core -am -DskipTests
```
anchela, line 39: same here: i would encourage AI to always run the tests.

AGENTS.md
```diff
@@ -0,0 +1,236 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+```
+
+### Annotations
+- Use `@NotNull` / `@Nullable` from `org.jetbrains.annotations` for API contracts
+- Use OSGi annotations (`@Component`, `@Activate`, etc.) for service components
```
anchela, line 218: this should come with an indication which version of OSGI to use

AGENTS.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Git Workflow
+
+- **Main branch:** `trunk` (not master/main). Never commit directly to trunk
+- **Issue tracker:** Apache Jira, project key `OAK` (e.g., OAK-12345)
+- **Branch naming:** `issue/OAK-<issue_number>` (e.g., `issue/OAK-12345`). If no Jira
+  issue is specified, ask the user what the branch name should be
+- **Commit message format:** Start with Jira issue key: `OAK-XXXXX: Description of change`
```
anchela, line 226: the JIRA issue needs to point to an existing JIRA ticket from issues.apache.org/jira/Oak and not invent some number that don't exist

AGENTS.md
```diff
@@ -0,0 +1,236 @@
...
+- **PR target:** PRs should target `trunk`
+- All changes must be committed to the issue branch, never directly to trunk
```
anchela, line 228: additional hint: sent review request to committers that have contributed to the code base before

anchela (changes requested):
hi @smiroslav , i believe this PR could benefit from additional extensions like context.md, active_context.md, product_evolution.md, and tech_debt.md and project level skills for cursor and claude-code already configured.

i would also love for the project to come with written specifications (proposal.md and other specs/artifcacts)... while we didn't build OAK with AI we can still provide the information as if we had built it with spec-driven development.

pom.xml
```diff
@@ -92,6 +92,7 @@
           <excludes>
             <exclude>release.properties</exclude>
             <exclude>CONTRIBUTING.md</exclude>
+            <exclude>AGENTS.md</exclude>
```
thomasmueller, line 95:
```suggestion
            <exclude>AGENTS.md</exclude>
            <exclude>CLAUDE.md</exclude>
```

AGENTS.md
```diff
@@ -0,0 +1,236 @@
...
+- Document store toggles (throttling, full GC, embedded verification) in
+  `oak-store-document/.../DocumentNodeStoreBuilder.java`
+- Some features also support a system property fallback (e.g., `oak.classicMove`)
+
+## Code Conventions
+
+### Style
+- No wildcard imports (import each class individually)
+- Follow existing code style in the module you are modifying
+- OSGi bundle compliance is required - modules produce OSGi bundles
```
thomasmueller, line 190:
```suggestion
- Follow existing code style in the module you are modifying
- OSGi bundle compliance is required - modules produce OSGi bundles
- Never use regular expression parsing and find / replace for JSON or XML data
```

AGENTS.md
thomasmueller: I added a comment "use -DskipTests only for modules you did not change"

AGENTS.md
thomasmueller: same

AGENTS.md
thomasmueller: What do you suggest we add?

AGENTS.md
thomasmueller: See new comment "If no Jira issue is specified, ask the user what the branch name should be"

AGENTS.md
thomasmueller: See comment below

AGENTS.md
thomasmueller: > public API and SPI contracts cannot be automatically modified.

Sometimes this is hard to avoid, because we have many modules in Oak.

AGENTS.md
thomasmueller: This should be fixed now

AGENTS.md
thomasmueller: We use 80% in Sonar I think, and I think that's OK, but if you want to increase it to 90% then we can surely discuss this... with others.

smiroslav:
Hi @anchela  @thomasmueller  thanks for the comments. 
This change was only an initial effort during some experimentation. I think we should find a way to collaborate on this PR. 
Maybe the best would be to check out the `issue/OAK-12099` branch and make direct edits. Or PR into this one... not sure what is the most optimal.

It does not need to be perfect the first time, but if we agree that we want md files that aim to enable agents to better cope with the source code, we can add an initial version and make updates with improvements.

thomasmueller:
I would merge it later and then we can make improvements later. (Not having it is probably more of a problem than having it, but not "perfect" for some definition of perfect)

AGENTS.md
reschke: The key word is "automatically".

This also misses information about how internal APIs are labeled, and what changing version numbers actually means in terms of what breaks.

AGENTS.md
reschke: That does not work. Even trivial dependency changes can break OSGI-IT tests.

Changes to oak-commons (for example) can break *any* module that uses it (even transitively).

Test coverage needs to stay the same.

AGENTS.md
reschke: *all*

In particular, it needs to understand which fixtures to use, and what they require (eg MongoDB or Docker)

reschke (changes requested):
I believe most of this belongs into the actual Oak documention, in particular https://jackrabbit.apache.org/oak/docs/participating.html and https://jackrabbit.apache.org/oak/docs/testing.html.

In any case, we should not repeat ourselves -> inconsistencies.

As a general remark: contributors sure can use AI as help, but that  **_MUST NOT_** change the review process. In particular, if the author can not explain how the code works, that's IMHO a showstopper.

AGENTS.md
thomasmueller: It doesn't sounds like we can ever have a perfect solution that everyone agrees to.
I'll merge this now: we can always improve later.

AGENTS.md
thomasmueller: > That does not work.

OK, feel free to propose a solution that
* doesn't run all the tests locally for each trivial change, because that would be wasting time. The tests are anyway run on Jenkins before we merge.
* and at the same time ensures that no test ever breaks.

In my view, this is just impossible. So I'm now merging what we have. If you find a way to solve this riddle, that would be great!

AGENTS.md
thomasmueller: No. It doesn't make sense to run all tests each time.

thomasmueller:
> I believe most of this belongs into the actual Oak documention

I consider this to be part of the oak documentation, specifically on how to contribute.

So far nobody ever complained that this was missing... I wonder why now that we add this documentation you seem to have such a strong opinion on what exactly it should contain and where it should be?

AGENTS.md
reschke: 👎

AGENTS.md
reschke: Whether something is "trivial" is not trivial to decide. In the past, I've seen problems with "trivial" changes more than once (including changes done by me).

Examples:

- changing/adding APIs without considering exports
- changing package versions while ignoring baseline checks
- making changes to store implementations without running oak-jcr tests
- making changes to indexing without running oak-jcr tests
- adding APIs (packages) without considering whether internal or not
- adding APIs that rely on APIs of other libs (Guava...)

AGENTS.md
reschke: The result of a statement like this is that people will miss to run important tests, because that's what it says here.

reschke:
> I consider this to be part of the oak documentation, specifically on how to contribute.

**We already have information about how to contribute**. Please make sure that it's consistent and will stay consistent with existing docs.

And, by all means, put it into README.md (!!!!)

AGENTS.md
```diff
@@ -0,0 +1,248 @@
...
+- **Main branch:** `trunk` (not master/main). Never commit directly to trunk
+- **Issue tracker:** Apache Jira, project key `OAK` (e.g., OAK-12345)
+- **Branch naming:** `issue/OAK-<issue_number>` (e.g., `issue/OAK-12345`). If no Jira
+  issue is specified, ask the user what the branch name should be
+- **Commit message format:** Start with Jira issue key: `OAK-XXXXX: Description of change`
+- **PR target:** PRs should target `trunk`
+- All changes must be committed to the issue branch, never directly to trunk
+- **Code review:** After pushing changes, remind the user to request a review from
+  committers who have previously contributed to the affected modules
+
```
reschke, line 231: How does the reviewer ensure that the contribution is compatible with Apache IP requirements?

PR #2747, opened 2026-02-18
OAK-12101 - Skip indexing of very long similarity tags

antonhosgood:
Very long similarity tag names inflate index sizes and probably defeat the purpose of tags. This PR introduces an optional configurable value per index to configure how similarity tags should be filtered in `indexSimilarityTag` and `indexDynamicBoost`.

### DIFF
--- a/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
+++ b/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
@@ -749,12 +749,13 @@ public abstract class FulltextDocumentMaker<D> implements DocumentMaker<D> {
     }
 
     private boolean isTagWithinLengthLimit(String value, String pname) {
-        if (value.length() <= definition.getSimilarityTagMaxLength()) {
+        int maxLength = definition.getSimilarityTagMaxLength();
+        if (maxLength < 0 || value.length() <= maxLength) {
             return true;
         }
         if (!LOG_SILENCER.silence(LOG_KEY_SIMILARITY_TAG_SKIPPED)) {
             log.warn("[{}] Skipping similarity tag for property {}. Value length {} exceeds maximum allowed length {}",
-                    getIndexName(), pname, value.length(), definition.getSimilarityTagMaxLength());
+                    getIndexName(), pname, value.length(), maxLength);
         }
         return false;
     }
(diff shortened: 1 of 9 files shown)


oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -736,6 +748,18 @@ protected String getIndexName() {
...
     }
 
+    private boolean isTagWithinLengthLimit(String value, String pname) {
+        int maxLength = definition.getSimilarityTagMaxLength();
+        if (maxLength < 0 || value.length() <= maxLength) {
+            return true;
+        }
+        if (!LOG_SILENCER.silence(LOG_KEY_SIMILARITY_TAG_SKIPPED)) {
+            log.warn("[{}] Skipping similarity tag for property {}. Value length {} exceeds maximum allowed length {}",
+                    getIndexName(), pname, value.length(), maxLength);
```
thomasmueller: (Just my view) I think it's less surprising if the logging itself is done outside of this method, that is in the caller. If it is needed in multiple places, that would mean code duplication or another method, which is also bad. An alternative would be to change the method name to eg. "isTagWithinLengthLimitOrLogWarning". Yes it's a long name... but it's less surprising.

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -58,6 +60,10 @@
 public abstract class FulltextDocumentMaker<D> implements DocumentMaker<D> {
 
     private final Logger log = LoggerFactory.getLogger(getClass());
+
+    private static final LogSilencer LOG_SILENCER = new LogSilencer(Duration.ofMinutes(1).toMillis(), 4);
```
thomasmueller: That looks good... alternatively, we could use the default silencer settings (15 min, 64 entries cache)

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextDocumentMaker.java
```diff
@@ -736,6 +748,18 @@ protected String getIndexName() {
         return definition.getIndexName();
     }
 
+    private boolean isTagWithinLengthLimit(String value, String pname) {
+        int maxLength = definition.getSimilarityTagMaxLength();
+        if (maxLength < 0 || value.length() <= maxLength) {
+            return true;
+        }
+        if (!LOG_SILENCER.silence(LOG_KEY_SIMILARITY_TAG_SKIPPED)) {
```
thomasmueller: This means only one key. I think it would be better to use the property name as the key:
```suggestion
        if (!LOG_SILENCER.silence(pname)) {
```

If there are multiple properties indexed like that, each property name will have a separate silence period. I'm not sure if this happens in reality thought.

reschke (commented):
Please do not merge until the Sonar checks have been successful.

See https://jackrabbit.apache.org/oak/docs/participating.html#pull-requests-prs

thomasmueller:
Could you open a new PR against the (new) OAK-12101 branch please, so that we can run Sonar tests?

PR #2762, opened 2026-02-24
OAK-12104 : handled Bson payload exception while doing bulk create/up…

rishabhdaim:
…dates


reschke:
I think this approach is way more complex than it needs to be for an edge case.

Also, size estimation either should go through the existing utils or the BSON APIs that could work here.

rishabhdaim:
Based on the discussion with @reschke, I am closing this issue.

As of now, we don't need this feature, and there is no evidence that this has actually happened in any production environment.

PR #2764, opened 2026-02-24
OAK-12109: MongoDocumentStore: improve diagnostics for too large docs in bulk request payloads

reschke:
This change is intentionally minimal:

1. It does not change exception processing; it just adds more log information
2. It only estimates update changes, but that should be sufficient to find the biggest one

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
@@ -1672,7 +1672,8 @@ public class MongoDocumentStore implements DocumentStore {
                 }
             }
             LOG.error("bulkUpdate of size {} failed with: {}; biggest update was for i={} with approximate size of {}",
-                    updateOps.size(), bsonException.getMessage(), idOfbiggestUpdate, estimatedSizeOfBiggestUpdate, bsonException);
+                    updateOps.size(), bsonException.getMessage(), idOfbiggestUpdate, estimatedSizeOfBiggestUpdate,
+                    bsonException);
             // rethrow
             throw bsonException;
         }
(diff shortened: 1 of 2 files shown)


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1651,7 +1651,32 @@ private <T extends Document> BulkRequestResult sendBulkRequest(final Collection<
...
+                // conditions and updates
+                int estimatedSize = updateOp.toString().length();
+                LOG.debug("after bulk write: string serialization of changes for id={} had an approximate size of {}",
+                        id, estimatedSize);
+                if (estimatedSize > estimatedSizeOfBiggestUpdate) {
+                    idOfbiggestUpdate = id;
+                    estimatedSizeOfBiggestUpdate = estimatedSize;
+                }
+            }
+            LOG.error("bulkUpdate failure: biggest update was for i={} with approximate size of {}",
```
joerghoh: 2 log statements? Can't we fit that into a single one?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
```diff
@@ -1651,7 +1651,32 @@ private <T extends Document> BulkRequestResult sendBulkRequest(final Collection<
             for (BulkWriteError err : e.getWriteErrors()) {
                 failedUpdates.add(bulkIds[err.getIndex()]);
             }
+        } catch (BSONException bsonException) {
```
rishabhdaim, line 1654: the exception here could be due to combined payload as well.
So logging the biggest document might not yield anything.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
reschke: do you mean the two statement on ERROR level?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/mongo/MongoDocumentStore.java
reschke: If the biggest document is not *that* bug, that will be good to know as well - in that case a smaller batch would indeed help. Let's get to that later.

reschke:
will work on test coverage later.

reschke:
(will merge when Sonar is happy)

PR #2768, opened 2026-02-26
OAK-12101 - Skip indexing of very long tags

antonhosgood:
Very long tag names inflate index sizes and probably defeat the purpose of tags. This PR introduces optional configurable values per index to configure how tags should be filtered in `indexSimilarityTag` and `indexDynamicBoost`.


fabriziofortino (approved):

reschke:
Please do not merge until Sonar has run, which implies merging into an Oak branch.

PR #2776, opened 2026-03-02
OAK-12121 : add regression tests for offline compaction not persisting gc.log

rishabhdaim:
## Summary

- Adds `OfflineCompactionGcLogTest` — three regression tests that directly expose the bug where offline compaction (`compactFull()` + `cleanup()` as separate calls) never writes a gc.log entry
- All three tests fail against the unfixed code and pass once the fix retaining `CompactionResult` across the two calls is applied

## Changes

- `oak-segment-tar/src/test/java/.../file/OfflineCompactionGcLogTest.java` (new file)
  - `testOfflineCompactionPersistsGcLog` — basic offline compaction scenario: verifies gc.log has a valid entry with non-empty root id and incremented GC generation
  - `testCleanupFirstCompactOkCleanupGcLogWritten` — cleanup → compact → cleanup: pre-compaction cleanup must not write gc.log; only the post-compaction cleanup should
  - `testCompactOkCleanupCleanupGcLogHasExactlyOneEntry` — compact → cleanup → cleanup: first cleanup writes, second must not write a duplicate

## Test Plan

- [ ] All three new tests fail on unfixed code (regression tests)
- [ ] All three new tests pass once the fix is applied

## Links

- JIRA: https://issues.apache.org/jira/browse/OAK-12121

### DIFF
--- a/oak-segment-tar/src/test/java/org/apache/jackrabbit/oak/segment/file/OfflineCompactionGcLogTest.java
+++ b/oak-segment-tar/src/test/java/org/apache/jackrabbit/oak/segment/file/OfflineCompactionGcLogTest.java
@@ -1,0 +1,192 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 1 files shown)


rishabhdaim:
@lweitzendorf could you please review these test cases.

reschke (approved):
...grom the perspective that it reproduces the issue, so we can proceed with the fix.

lweitzendorf (approved):
The test cases seem reasonable. We might want to rethink whether GCJournal persistance should be handled as part of the clean up longer term, but that doesn't need to be part of the more immediate fix and these tests by extension.

PR #2777, opened 2026-03-03
OAK-12122

thomasmueller:
- Use Random(42) in prepareData for reproducible test data
- Add logFilesInDump when expected file is missing, listing all files with sizes


AGENTS.md
```diff
@@ -0,0 +1,248 @@
+# AGENTS.md - AI Agent Instructions for Apache Jackrabbit Oak
```
bhabegger, line 1: Maybe we should have an explicit PR for this ? Making it explicit that we are getting agents to work on the code base ?

thomasmueller: Sorry this was by mistakes... I'll remove AGENTS.md etc from the PR.

PR #2780, opened 2026-03-04
OAK-12108 : remove usage of Guava's Monitor from SegmentBufferWriterPool

rishabhdaim:
## Summary

- Removes the dependency on Guava's `Monitor` class from `SegmentBufferWriterPool` as part of the broader effort to reduce Guava usage in the project (OAK-10685)
- Replaces `Monitor` and `Monitor.Guard` with standard Java `ReentrantLock` and `Condition`

## Changes

- `SegmentBufferWriterPool.java` — replace `Monitor`/`Monitor.Guard` with `ReentrantLock` + `Condition`

## Test Plan

- [ ] Existing `SegmentBufferWriterPoolTest` passes
- [ ] Related unit tests in OAK-12115 cover the new locking behaviour

## Links

- JIRA: https://issues.apache.org/jira/browse/OAK-12108
- Related: https://issues.apache.org/jira/browse/OAK-12115

### DIFF
--- a/oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/SegmentBufferWriterPool.java
+++ b/oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/SegmentBufferWriterPool.java
@@ -32,11 +32,12 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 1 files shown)


oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/SegmentBufferWriterPool.java
```diff
@@ -225,20 +232,29 @@ public void flush(@NotNull SegmentStore store) throws IOException {
...
-                poolMonitor.leave();
+                lock.unlock();
             }
 
             // Wait for the return of the borrowed writers. This is the
             // case once all of them appear in the disposed set.
-            if (safeEnterWhen(poolMonitor, allReturned(toReturn))) {
-                try {
+            lock.lock();
+            try {
```
reschke, line 241: Not sure: is this loop guaranteed to terminate?

rishabhdaim: this has to block till are writer have been returned. i.e. what `safeEnterWhen` was doing internally

PR #2784, opened 2026-03-06
OAK-12125: Use sentinel when propertyNames is null

tmvlad:
- A property index definition with \`type=property\` but missing the required \`propertyNames\` property caused a \`NullPointerException\` in \`PropertyIndexEditor\` that was not caught by \`IndexUpdate\`, breaking the entire async indexing cycle for all indexes
- Fix: use a default when encountering a 'null' value for propertyNames, log a WARN and continue processing without throwing.

## Links

- JIRA: https://issues.apache.org/jira/browse/OAK-12125

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexEditor.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexEditor.java
@@ -34,6 +34,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 
@@ -42,6 +43,7 @@ import javax.jcr.PropertyType;
 import org.apache.jackrabbit.oak.api.CommitFailedException;
 import org.apache.jackrabbit.oak.api.PropertyState;
 import org.apache.jackrabbit.oak.commons.collections.SetUtils;
+import org.apache.jackrabbit.oak.plugins.index.ContextAwareCallback;
 import org.apache.jackrabbit.oak.plugins.index.IndexConstants;
 import org.apache.jackrabbit.oak.plugins.index.IndexEditor;
 import org.apache.jackrabbit.oak.plugins.index.IndexUpdateCallback;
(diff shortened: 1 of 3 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexEditor.java
```diff
@@ -132,7 +132,10 @@ public PropertyIndexEditor(NodeBuilder definition, NodeState root,
 
         // get property names
         PropertyState names = definition.getProperty(PROPERTY_NAMES);
-        if (names.count() == 1) { 
+        if (names == null) {
+            throw new IllegalStateException("Missing required property '" + PROPERTY_NAMES + "' in property index definition");
```
thomasmueller, line 136: So instead of a NPE we now get an IllegalStateException... It might currently resolve the problem, but I think it is not a good solution, because still an exception is thrown. There is a performance overhead in throwing an exception, but more importantly exception handing is not part of the API, and so changes in the future might break this logic. Also, there is no logging.

An alternative solution would be:
* Use a default value if no propertyNames are set, eg. use the value "propertyNamesIsMissing". A property name that is so unlikely to occur in practice that there are no repositories with this property. (And if there are, then actually it wouldn't be a problem.)
* Log a warning, but only once. Only once such that the log file is not filled.

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexEditor.java
```diff
@@ -132,7 +138,18 @@ public PropertyIndexEditor(NodeBuilder definition, NodeState root,
 
         // get property names
         PropertyState names = definition.getProperty(PROPERTY_NAMES);
-        if (names.count() == 1) { 
+        if (names == null && Boolean.parseBoolean(System.getProperty(FT_GRANITE_63829))) {
```
thomasmueller: So far we throw a NullPointerException. We could argue a feature toggle is not needed in this case. But let's add one anyway (possibly NPE is better than not throwing a NPE, for some weird reason).

Please use a feature toggle like other we use, see oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/toggle/FeatureToggle.java

thomasmueller (changes requested):

thomasmueller (approved):

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexTest.java
```diff
@@ -1187,4 +1190,156 @@ private String getCommitPath(String changeNodeName) {
...
+        PropertyIndexEditorProvider provider = new PropertyIndexEditorProvider();
+        injectFeature(provider, feature);
+        return new EditorHook(new IndexUpdateProvider(provider));
+    }
+
+    private static void injectFeature(PropertyIndexEditorProvider provider, Feature feature) {
+        try {
+            Field f = PropertyIndexEditorProvider.class.getDeclaredField("feature");
+            f.setAccessible(true);
+            f.set(provider, feature);
```
thomasmueller: I think we should not use setAccessible

thomasmueller (changes requested):

PR #2789, opened 2026-03-06
OAK-12117 - Limit the maximum number of indexed tags per document

antonhosgood:
Similarly to https://github.com/apache/jackrabbit-oak/pull/2768, this PR introduces optional configurable index properties to limit the maximum number of indexed similarity and dynamic boost tags. The first N similarity tags and the top N dynamic boost tags by confidence are indexed.

### DIFF
--- a/oak-doc/src/site/markdown/query/lucene.md
+++ b/oak-doc/src/site/markdown/query/lucene.md
@@ -241,6 +241,20 @@ selectionPolicy
   Set to -1 to disable the length check entirely.
   See [Dynamic Boost](#dynamic-boost) and [Search by similar feature vectors](#similar-fv) for details.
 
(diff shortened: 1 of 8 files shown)


thomasmueller:
Some remarks:

* Existing indexed documents will have the old behavior (first N tags), while newly indexed documents will have new behavior (top N by confidence). I assume that's fine.
* How many tags did we see currently? I assume memory usage (if there are many tags) is not very high normally; but possibly we could truncate if the number of in-memory tags exceeds eg. 10x the configured limit (example: if 50 is configured, then once we have more than 500 in memory, truncate the in-memory buffer to the top 250 entries).

thomasmueller:
Merged to the branch only, so far.

PR #2792, opened 2026-03-09
OAK-12128: Fail earlier (and add more diags) when detecting late writes

reschke:
test failures: see https://issues.apache.org/jira/browse/OAK-12128?focusedCommentId=18063476&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-18063476

### DIFF
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/LeaseCheckDocumentStoreWrapper.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/LeaseCheckDocumentStoreWrapper.java
@@ -27,7 +27,6 @@ import org.apache.jackrabbit.oak.plugins.document.ClusterNodeInfo;
 import org.apache.jackrabbit.oak.plugins.document.Collection;
 import org.apache.jackrabbit.oak.plugins.document.Document;
 import org.apache.jackrabbit.oak.plugins.document.DocumentStore;
-import org.apache.jackrabbit.oak.plugins.document.DocumentStoreException;
 import org.apache.jackrabbit.oak.plugins.document.UpdateOp;
 import org.apache.jackrabbit.oak.plugins.document.Throttler;
 import org.apache.jackrabbit.oak.plugins.document.cache.CacheInvalidationStats;
@@ -119,10 +118,9 @@ public final class LeaseCheckDocumentStoreWrapper implements DocumentStore {
 
     @Override
     public <T extends Document> int remove(Collection<T> collection,
-                                           String indexedProperty, long startValue, long endValue)
-            throws DocumentStoreException {
+                                           String indexedProperty, long startValue, long endValue) {
         return leaseChecking(() ->
-            delegate.remove(collection, indexedProperty, startValue, endValue));
+                delegate.remove(collection, indexedProperty, startValue, endValue));
     }
 
     @Override
@@ -258,20 +256,18 @@ public final class LeaseCheckDocumentStoreWrapper implements DocumentStore {
         return delegate.throttler();
     }
 
+    // invoke operation with lease check before/after
     private <T> T leaseChecking(Supplier<T> operation) {
         performLeaseCheck();
-        try {
-            return operation.get();
-        } finally {
-            performLeaseCheck();
-        }
+        T result = operation.get();
+        performLeaseCheck();
+        return result;
     }
+
+    // invoke operation with lease check before/after
     private void leaseChecking(Runnable operation) {
         performLeaseCheck();
-        try {
-            operation.run();
-        } finally {
-            performLeaseCheck();
-        }
+        operation.run();
+        performLeaseCheck();
     }
 }
--- a/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/LeaseCheckDocumentStoreWrapper.java
+++ b/oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/LeaseCheckDocumentStoreWrapper.java
@@ -32,7 +32,6 @@ import org.apache.jackrabbit.oak.plugins.document.UpdateOp;
 import org.apache.jackrabbit.oak.plugins.document.Throttler;
 import org.apache.jackrabbit.oak.plugins.document.cache.CacheInvalidationStats;
 import org.jetbrains.annotations.NotNull;
-import org.jspecify.annotations.NonNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/LeaseCheckDocumentStoreWrapper.java
```diff
@@ -52,133 +53,134 @@ public LeaseCheckDocumentStoreWrapper(final DocumentStore delegate, final Cluste
...
         }
     }
 
     @Override
-    public final <T extends Document> T find(Collection<T> collection, String key) {
-        performLeaseCheck();
-        return delegate.find(collection, key);
+    public <T extends Document> T find(Collection<T> collection, String key) {
+        return leaseChecking(() ->
+                delegate.find(collection, key));
```
rishabhdaim, line 75: Please remove the unintended spaces.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/LeaseCheckDocumentStoreWrapper.java
```diff
@@ -52,133 +53,134 @@ public LeaseCheckDocumentStoreWrapper(final DocumentStore delegate, final Cluste
...
+
     }
 
     @Override
-    public final <T extends Document> List<T> query(Collection<T> collection,
-            String fromKey, String toKey, int limit) {
-        performLeaseCheck();
-        return delegate.query(collection, fromKey, toKey, limit);
+    public <T extends Document> List<T> query(Collection<T> collection,
+                                              String fromKey, String toKey, int limit) {
```
rishabhdaim: same as above

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/LeaseCheckDocumentStoreWrapper.java
```diff
@@ -253,6 +256,24 @@ public int getNodeNameLimit() {
      */
     @Override
     public Throttler throttler() {
-        return delegate.throttler();
+        return leaseChecking(() ->
```
rishabhdaim: why we need to check the lease timeout here ?

cc @stefan-egli

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/LeaseCheckDocumentStoreWrapper.java
```diff
@@ -242,7 +244,8 @@ public <T extends Document> void prefetch(Collection<T> collection,
      */
     @Override
     public int getNodeNameLimit() {
-        return delegate.getNodeNameLimit();
+        return leaseChecking(() ->
```
rishabhdaim: same as `throttler` method.

rishabhdaim (changes requested):

Joscorbe:
I wonder if we should add a distinct log message when the lease check fails after the operation has already run (in the finally block). That would make it easier to find and analyze cases where the lease was lost during the call.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/LeaseCheckDocumentStoreWrapper.java
```diff
@@ -253,6 +256,24 @@ public int getNodeNameLimit() {
...
+        } finally {
+            performLeaseCheck();
+        }
+    }
+    private void leaseChecking(Runnable operation) {
+        performLeaseCheck();
+        try {
+            operation.run();
+        } finally {
+            performLeaseCheck();
```
Joscorbe: Haven't tested this, but isn't it possible that the operation throws an exception and the lease check in **finally** also fails, so the DocumentStoreException from the lease check would mask the exception thrown by the operation?

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/LeaseCheckDocumentStoreWrapper.java
reschke: We don't need to, indeed.

reschke:
> I wonder if we should add a distinct log message when the lease check fails after the operation has already run (in the finally block). That would make it easier to find and analyze cases where the lease was lost during the call.

@Joscorbe - https://issues.apache.org/jira/browse/OAK-12128?focusedCommentId=18064356&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-18064356

Well, help in analyzing is the whole point of this PR. The question is whether it actually adds anything, because next a DSException will be thrown anyway.

Adding this requires minor changes in `ClusterNode` - API to check without throwing; or potentially to `performLeaseCheck()`.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/LeaseCheckDocumentStoreWrapper.java
reschke: Yes, I think. Note that this case apparently has no test coverage (or it is the reason for the tests complaining about the incorrect exception message). Let me try that.

EDIT: yes, that's the reason. Good catch.

reschke:
Note: I could minimize the diff by re-adding the `final`s.

reschke:
also:

https://issues.apache.org/jira/browse/OAK-12128?focusedCommentId=18064369&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-18064369

reschke:
The test fails for me as well. Occasionally.

reschke:
Changes:

1. Late writes are now ERROR-logged (code change minimal so that we dnot have to touch `ClusterNodeInfo`.
2. Test assertion changed so that 'both' exception messages are tolerated.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/util/LeaseCheckDocumentStoreWrapper.java
```diff
@@ -52,133 +56,140 @@ public LeaseCheckDocumentStoreWrapper(final DocumentStore delegate, final Cluste
...
 
-    private final void performLeaseCheck() {
+    private void performLeaseCheck(boolean after) {
         if (clusterNodeInfo != null) {
-            clusterNodeInfo.performLeaseCheck();
+            try {
+                clusterNodeInfo.performLeaseCheck();
+            } catch (DocumentStoreException ex) {
+                if (after) {
+                    LOG.error("Potential late write operation detected", new Exception("call stack"));
```
stefan-egli: wondering, in case of a late write we might be under memory pressure. in which case even `new Exception` might not work. what about doing 2 logs, first one without the exception, followed by a second one with the exception?

Joscorbe (approved):

stefan-egli (commented):
+1 :)

PR #2795, opened 2026-03-11
OAK-12133 With the segment store, binary properties can not be aggregated

thomasmueller:
This fixes the problem for Lucene and Elastic indexes.

It also improves the error message and logs an error in the segment store.

### DIFF
--- a/oak-run/src/test/java/org/apache/jackrabbit/oak/run/query/IndexRegexPropertyWithBinaryExcludedTest.java
+++ b/oak-run/src/test/java/org/apache/jackrabbit/oak/run/query/IndexRegexPropertyWithBinaryExcludedTest.java
@@ -1,0 +1,228 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.jackrabbit.oak.run.query;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+import javax.jcr.Node;
+import javax.jcr.PropertyType;
+import javax.jcr.Repository;
+import javax.jcr.RepositoryException;
+import javax.jcr.Session;
+import javax.jcr.SimpleCredentials;
+import javax.jcr.query.Query;
+import javax.jcr.query.QueryManager;
+import javax.jcr.query.QueryResult;
+import javax.jcr.query.Row;
+import javax.jcr.query.RowIterator;
+
+import org.apache.jackrabbit.api.JackrabbitRepository;
+import org.apache.jackrabbit.oak.Oak;
+import org.apache.jackrabbit.oak.jcr.Jcr;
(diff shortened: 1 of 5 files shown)


fabriziofortino (approved):
The indexing related changes look good to me.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/SegmentPropertyState.java
```diff
@@ -190,7 +197,33 @@ private <T> T getValue(RecordId id, Type<T> type) {
...
 
-        String value = reader.readString(id);
+        // OAK-12133: Detect when trying to read a binary property as a string
+        Type<?> actualType = getType();
+        if (actualType.isArray()) {
+            actualType = actualType.getBaseType();
+        }
+        String message = null;
+        if (actualType == BINARY && type != BINARY) {
+            message = String.format(
```
thomasmueller, line 207: > this expression which always evaluates to "true"

It does not.

oak-run/src/test/java/org/apache/jackrabbit/oak/run/query/IndexRegexPropertyWithBinaryExcludedTest.java
```diff
@@ -0,0 +1,228 @@
...
+                assertEquals("Expected 2 results", 2, results.size());
+                assertTrue("Expected /content/one/child in results", results.contains("/content/one/child"));
+                assertTrue("Expected /content/two/child in results", results.contains("/content/two/child"));
+
+                // Success - exit retry loop
+                return;
+            } catch (AssertionError e) {
+                lastError = e;
+                Thread.sleep(100);
+            }
```
thomasmueller, line 199: Thread.sleep is needed here

ahanikel:
@thomasmueller How often do you expect these warnings to occur? As this is quite performance critical, can we perhaps move the warning and log generation in the exception handler, doing it only if `readString?` fails?
E.g. as in:
```
      String value;
      try {
          value = reader.readString(id);
      } catch (IllegalStateException e) {
          // OAK-12133: Detect when trying to read a binary property as a string
          if (e.getMessage() != null &&
                  e.getMessage().contains("possibly trying to read a BLOB using getString")) {
              Type<?> actualType = getType();
              if (actualType.isArray()) {
                  actualType = actualType.getBaseType();
              }
              if (actualType == BINARY) {
                  String message = String.format(
                      "Attempting to read binary property '%s' as %s. " +
                      "This can fail if the binary is stored externally. " +
                      "Binary properties should be checked before getValue() is called.",
                      name, type);
                  if (!LOG_SILENCER.silence(name)) {
                      LOG.warn(message, new Exception("Stack trace"));
                  }
                  throw new IllegalStateException(message, e);
              }
          }
          throw e;
      }
 ```

thomasmueller:
> @thomasmueller How often do you expect these warnings to occur? 

The warning is very rare. Hopefully, it never occurs:

`if (actualType == BINARY && type != BINARY) {`

(Sonar for some reason thinks this is always true, but in reality it is almost never true. The test here is very fast.)

If it does occur, we are already in "behavior is undefined" territory. And that's why logging a warning is important: if the binary is large, an exception is thrown (the one caught below). However, there might be some cases that do not trigger the exception, but because we are lucky and the binary is small enough to be inlined. But this is just pure luck. That's why I added this logic: so that we have warnings in the log file and don't rely on this pure luck.

The reason why I added this logic is so that we can detect such incorrect behavior (possibly in the indexing code, but possibly in other code in Oak.)

> As this is quite performance critical, can we perhaps move the warning and log generation in the exception handler

That would defeat what I have done, mostly. I want that we detect cases where we call getString() on a binary, even if it does _not_ throw an exception.

ahanikel:
I see, +1 from me then.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/SegmentPropertyState.java
reschke: This:

~~~
type != BINARY
~~~

indeed always evaluates to true

thomasmueller: Ah right! Now I understand. 

But "actualType == BINARY" doesn't always evaluate to true or false.

We could simplify this to "if (actualType == BINARY) { " - that would work, UNTIL someone removes the earlier lines

        if (type == BINARY) {
            return (T) reader.readBlob(id); // load binaries lazily
        }

... and then it would no longer be correct...

PR #2800, opened 2026-03-16
OAK-12137: Create aggregate jacoco report

reschke:


rishabhdaim (approved):

pom.xml
```diff
@@ -35,6 +35,8 @@
     <skip.deployment>true</skip.deployment>
     <sonar.organization>apache</sonar.organization>
     <sonar.host.url>https://sonarcloud.io</sonar.host.url>
+    <!-- this project uses cross-module reports with one aggregate, https://docs.sonarsource.com/sonarcloud/enriching/test-coverage/java-test-coverage/#add-coverage-in-a-multi-module-maven-project -->
+    <sonar.coverage.jacoco.xmlReportPaths>${project.basedir}/target/site/jacoco-aggregate/jacoco.xml</sonar.coverage.jacoco.xmlReportPaths>
```
kwin: This needs to be set for each module individually otherwise the non-aggregate report is used (https://docs.sonarsource.com/sonarqube-server/analyzing-source-code/test-coverage/java-test-coverage#multi-module-maven-project). Our hierarchy is different from the example in the Sonar documentation that the submodules don't inherit from the root POM but from the parent, therefore move the property there similar to https://github.com/apache/jackrabbit/blob/024bca3806907cff6abcf2f922149728b1cd08c0/jackrabbit-parent/pom.xml#L84.

kwin (changes requested):

pom.xml
kwin: Also you can't use project.basedir because the level differs per project, so you have to rely on `${maven.multiModuleProjectDirectory}` together with a marker `.mvn` directory.

reschke: Hmmmm. Why does it work then?

I verified with the coverage report generated in `target/site`.

reschke: oh, wait: *this* PR only repeats what we did in https://issues.apache.org/jira/browse/JCR-5173 - making Sonar aware of it will be a distinct ticket and PR.

kwin: Why? Let's do it right from the beginning. The property I was commenting on is only for Sonar and does not work as intended

oak-parent/pom.xml
```diff
@@ -361,6 +361,30 @@
           <artifactId>jacoco-maven-plugin</artifactId>
           <version>0.8.13</version>
         </plugin>
+        <!-- generates aggregate (cross-module) coverage reports, https://github.com/olamy/jacoco-aggregator-maven-plugin
+            only enable this plugin (without inheritance) in the root module, as this is an aggregator plugin
+         -->
+        <plugin>
+          <groupId>io.github.olamy.maven.plugins</groupId>
+          <artifactId>jacoco-aggregator-maven-plugin</artifactId>
+          <version>1.0.3</version>
```
kwin: 1.0.4?

pom.xml
reschke: These are IMHO separate tasks; connecting this to Sonar truly is a different step. (this step is useful independant of whether we use Sonar (or something else in the future).

kwin: Also https://github.com/apache/jackrabbit-oak/blob/b96b73b09313bcf082d617cc6e3d015be6ababf2/oak-parent/pom.xml#L93 leads to not leveraging the aggregate.

kwin (changes requested):

reschke:
@kwin - thanks for pointing out the issues with this PR. I'll work on the Sonar wiring in a separate ticket.

kwin (approved):

PR #2808, opened 2026-03-21
OAK-12134 - compaction with concurrent writes can increase segmentstore size

jsedding:
- fix test cleanup; data files should be deleted after FileStore is closed


jsedding:
@reschke I verified via debugging that this change fixes the order so the FileStore is closed first and the file deleted after. Could you please verify on Windows? Thank you.

reschke (approved):
It works!

PR #2812, opened 2026-03-23
OAK-12143 - Fix AsyncLaneSwitcher not switching lanes when async-previous pre-set

antonhosgood:
Currently, `switchLane` returns early whenever `async-previous` is set, assuming the lane has already been switched. However, `async-previous` can legitimately appear in user-provided index definitions (e.g. copied from a running system), causing the switch to be skipped incorrectly and the index to be silently dropped from reindexing.

The fix tightens the early-return check: only skip if `async-previous` is present and `async` already equals the target lane. If `async-previous` is present but `async` does not match the target lane, the property is treated as stale, discarded, and the switch proceeds normally.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/AsyncLaneSwitcher.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/AsyncLaneSwitcher.java
@@ -64,6 +64,8 @@ public class AsyncLaneSwitcher {
      * {@code laneName}, the property is treated as stale (e.g. copied from a running system
      * into a user-provided index definition). In that case the stale value is discarded and
      * the switch proceeds normally so the index is not silently skipped during reindexing.
+     * If {@code async-previous} is missing, then the switch to the temporary lane required for
+     * offline reindexing has not yet happened and will be carried out.
      */
     public static void switchLane(NodeBuilder idxBuilder, String laneName) {
         PropertyState currentAsyncState = idxBuilder.getProperty(ASYNC_PROPERTY_NAME);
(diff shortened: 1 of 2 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/AsyncLaneSwitcher.java
```diff
@@ -52,17 +56,33 @@ public class AsyncLaneSwitcher {
...
 
-        if (idxBuilder.hasProperty(ASYNC_PREVIOUS)){
-            //Lane already switched
-            return;
+        if (idxBuilder.hasProperty(ASYNC_PREVIOUS)) {
+            if (currentAsyncState != null
+                    && !currentAsyncState.isArray()
+                    && laneName.equals(currentAsyncState.getValue(Type.STRING))) {
+                // Lane already switched to the target — no-op
+                return;
```
thomasmueller, line 81:
```suggestion
                // Lane already switched to the target — no-op
                LOG.info("Index definition contains stale '{}' property (async='{}', target lane='{}') - " +
                    "retaining the old value because it matches the value", ASYNC_PREVIOUS, currentAsyncState, laneName);
                return;
```

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/importer/AsyncLaneSwitcher.java
antonhosgood: Added a similar comment - in this case, the property is not necessarily stale.

PR #2813, opened 2026-03-23
OAK-12146 Use the latest elastic OR lucene index, if there are multiple

thomasmueller:


oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/IndexNameAdditionalTest.java
```diff
@@ -0,0 +1,90 @@
...
+        // no competing paths: all candidates pass through
+        Collection<String> result = IndexName.filterGloballySuperseded(
+                List.of("/oak:index/lucene-2"),
+                List.of());
+        assertEquals(List.of("/oak:index/lucene-2"), List.copyOf(result));
+    }
+
+    @Test
+    public void filterGloballySuperseded_olderCompetitorKept() {
+        // lucene-2 vs. elasticsearch-1: lucene is newer, so lucene passes
```
thomasmueller:
```suggestion
        // lucene-2 is newer than /oak:index/lucene-1-custom-1, so it passes
```

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/IndexNameAdditionalTest.java
```diff
@@ -0,0 +1,90 @@
...
+        Collection<String> result = IndexName.filterGloballySuperseded(
+                List.of("/oak:index/lucene-2"),
+                List.of("/oak:index/lucene-2", "/oak:index/lucene-1-custom-1"));
+        assertEquals(List.of("/oak:index/lucene-2"), List.copyOf(result));
+    }
+
+    @Test
+    public void filterGloballySuperseded_newerCompetitorFilters() {
+        // lucene-1 vs. elasticsearch-2 (same base): lucene-1 is superseded
```
thomasmueller:
```suggestion
        // lucene-1 vs. lucene-2 (same base): lucene-1 is superseded
```

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/IndexNameAdditionalTest.java
```diff
@@ -0,0 +1,90 @@
...
+        List<String> candidates = List.of(
+                "/oak:index/lucene-1",
+                "/oak:index/other-1");
+        List<String> competing = List.of(
+                "/oak:index/lucene-2",
+                "/oak:index/other-1");
+        Collection<String> result = IndexName.filterGloballySuperseded(candidates, competing);
+        assertEquals(List.of("/oak:index/other-1"), List.copyOf(result));
+    }
+
```
thomasmueller:
```suggestion
```

oak-it-osgi/src/test/java/org/apache/jackrabbit/oak/osgi/IndexVersionSelectionTest.java
```diff
@@ -0,0 +1,222 @@
...
+        // planner *selects* (via EXPLAIN). For plan generation, ElasticIndexStatistics.numDocs()
+        // is called to estimate the entry count; it issues a COUNT request to Elasticsearch and
+        // throws index_not_found_exception if the index does not exist. That exception propagates
+        // as UncheckedExecutionException and is caught in FulltextIndex.getPlans(), which silently
+        // skips the index — causing the test to fail with "traverse allNodes".
+        //
+        // The fix: after committing the Oak index definitions (which registers the definition in
+        // the ElasticIndexTracker via the Observer callback), we create an empty Elasticsearch
+        // index directly via the REST client. numDocs() then returns 0, the planner can generate
+        // a plan, and version selection is exercised correctly.
```
fabriziofortino, line 185: I am wondering if we could avoid this. If the issue is just `ElasticIndexStatistics.numDocs()`, perhaps we can add a `@TestOnly` constructor in `ElasticIndexTracker` to pass a NoOp `ElasticIndexStatistics` instance. This will be used by `ElasticIndexNodeManager`/`ElasticIndexNode`. If this works, there would be no need to include `oak-search-elastic` and `testcontainers` dependencies along with `ElasticConnectionRule` that makes the test slow since it has to start an ES container.

thomasmueller: It's an IT test test, so I think it's fine if it's slower than other tests. I think it makes sense that we test with the "correct" setup, and don't mock it. And for this, I think we do _need_ to include both oak-search-elastic and oak-lucene, because this is what we test here: they need to reply with a plan for a given query. 

But I'm still trying to find is a way to make it more real-world: that is, I think it should be possible to configure things so that we don't risk NoClassDefFoundError, and we don't need to create a elastic index from within the test case...

thomasmueller: I now added a new test, but I think I'll remove it again: it has hardcoded versions, and requires to change the exported versions.

oak-search-elastic/pom.xml
```diff
@@ -44,6 +44,12 @@
         <artifactId>maven-bundle-plugin</artifactId>
         <configuration>
           <instructions combine.self="override">
+            <Export-Package>
+              org.apache.jackrabbit.oak.plugins.index.elastic,
```
reschke, line 48: These shouldn't have been added without adding package info files so that sematic versioning works.

PR #2817, opened 2026-03-26
OAK-12089: Add Lucene 9 index provider (oak-search-lucene-ng)

bhabegger:
## Summary

Introduces `oak-search-lucene-ng`, a new Oak module that provides a Lucene 9 based index engine. Indexes opt in explicitly via `type=lucene9`; all existing indexes are unaffected.

- New `oak-search-luceneNg` module: index editor, query index, tracker, index node, storage (Oak JCR node-based directory), and OSGi wiring
- Full query parity with the legacy `lucene` engine: property restrictions, path/type filters, fulltext, sorting, and excerpts
- Facet parity for all three ACL modes: insecure, statistical (TapeSampling), and secure (per-document access check) — Lucene 9 API adaptations and null-safe `MatchingDocs.bits` handling
- `LuceneNgFacetCommonTest` extends the shared `FacetCommonTest` suite for end-to-end JCR-level facet coverage
- `AbstractIndexComparisonTest` inlined into `oak-search` test-jar; `oak-search-test` module removed
- README documents feature parity vs legacy Lucene and Elastic


oak-search-luceneNg/README.md
```diff
@@ -0,0 +1,23 @@
...
+| Ordering / sorting | ✓ | ✓ | ✓ |
+| Suggestions | ✓ | ✓ | ✗ |
+| Spellcheck | ✓ | ✓ | ✗ |
+| Similarity / More Like This | ✓ | ✓ (+ KNN) | ✗ |
+| Native queries | ✓ | ✓ | ✗ |
+| Index statistics / JMX | ✓ | ✓ | ✗ |
+| Index augmentors | ✓ | ✗ | ✗ |
+| NRT / hybrid indexing | ✓ | ✗ | ✗ |
+| Index copier (CopyOnRead/Write) | ✓ | ✗ | ✗ |
+| Multi-index queries | ✓ | ✗ | ✗ |
```
thomasmueller: All indexes support "multi-index queries": union queries and joins can use multiple indexes.

I think this entry refers to the "composite node store", where eg two repositories are used (one writeable, and one read-only, eg. only /apps + /libs). The regular Lucene index runs basically two queries and combines the result. If that's what is mean there, then I would call it "Composite queries", and add a remark what exactly this is.

oak-search-luceneNg/README.md
```diff
@@ -0,0 +1,23 @@
...
+| Fulltext search | ✓ | ✓ | ✓ |
+| Facets (insecure / statistical / secure) | ✓ | ✓ | ✓ |
+| Excerpts | ✓ | ✓ | ✓ |
+| Ordering / sorting | ✓ | ✓ | ✓ |
+| Suggestions | ✓ | ✓ | ✗ |
+| Spellcheck | ✓ | ✓ | ✗ |
+| Similarity / More Like This | ✓ | ✓ (+ KNN) | ✗ |
+| Native queries | ✓ | ✓ | ✗ |
+| Index statistics / JMX | ✓ | ✓ | ✗ |
+| Index augmentors | ✓ | ✗ | ✗ |
```
thomasmueller: Hm, what is this? Could you add a remark.

oak-search-luceneNg/README.md
bhabegger: [AI] Good point — added a footnote explaining that these are OSGi extension points (IndexFieldProvider / FulltextQueryTermsProvider) that allow third-party code to inject additional fields into indexed documents or expand fulltext queries.

oak-search-luceneNg/README.md
bhabegger: [AI] Thanks for the clarification! Renamed to "Composite node store queries" and added a footnote describing the scenario (multi-mount repository, one query per mount, results merged).

pom.xml
```diff
@@ -78,6 +78,7 @@
     <module>oak-segment-azure</module>
     <module>oak-benchmarks</module>
     <module>oak-search-elastic</module>
+    <module>oak-search-luceneNg</module>
```
thomasmueller: hm, the upper-case "N" breaks the naming convention... what about "oak-search-lucene-ng" instead?

oak-search-lucene-ng/README.md
```diff
@@ -0,0 +1,26 @@
+# oak-search-luceneNg
+
+Lucene 9 index provider for Oak (`type="lucene9"`).
```
thomasmueller, line 3: What about:

```suggestion
Lucene NG index provider for Oak (`type="lucene-ng"`).
```
Hoping we can upgrade to Lucene 10 without having to change the type. I do assume the index storage version won't change, or that there is an option to add compatibility.

If not, and we do need to reindex for lucene 10, we can still add "lucene-ng-10" if that should be needed.

thomasmueller (commented):
Just some bikeshedding: I wonder if we should use the term "Lucene 9" at all.

oak-search-lucene-ng/README.md
bhabegger: So the idea is that the code should be less sensitive to Lucene upgrades. However, I wasn't really sure about having a fixed name and the in some future having to have a lucene-ng-ng ;) Here we could interpret it as lucene *since* 9 and still use that for lucene 10, 11 up to and imaginary lucene 12 that breaks compatibility again (which might never happen as likely the APIs are more stable now).

Personally, I fine with lucene-ng as I myself had doubts.

bhabegger: Also the NG in the code for now is fine since we do have 2 versions, but at some point my expectation is that the legacy code be removed and the NG no longer be new generation and be refactored with proper non NG names. So in the code I'm fine, in the type however, this sticks more. That's mostly what bother me with ng in the type.

bhabegger: @thomasmueller What about lucene2026 ? Less tied to 9 doesn't have explicit ng ?

reschke:
```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[FATAL] Non-resolvable parent POM for org.apache.jackrabbit:oak-search-lucene-ng:1.93-SNAPSHOT: The following artifacts could not be resolved: org.apache.jackrabbit:oak-parent:pom:1.93-SNAPSHOT (absent): Could not find artifact org.apache.jackrabbit:oak-parent:pom:1.93-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ line 23, column 13
```

oak-search-lucene-ng/src/main/java/org/apache/jackrabbit/oak/plugins/index/luceneNg/LuceneNgIndexEditor.java
```diff
@@ -0,0 +1,809 @@
...
+/**
+ * IndexEditor for Lucene 9.
+ *
+ * <p>Only indexes properties that are explicitly declared in the index definition's
+ * {@code indexRules}. This mirrors the behaviour of the legacy {@code oak-lucene}
+ * module and avoids the Lucene doc-values type-consistency constraint: since the
+ * declared type for a property is fixed at index-definition time, every document
+ * that contributes a doc-values field for that property will use the same type.</p>
+ */
+public class LuceneNgIndexEditor implements Editor {
```
amit-jain: This should extend from [FulltextIndexEditor](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java) as is done for [LuceneIndexEditor](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexEditor.java#L30)

The main things that come along with the implementation is:
* Aggregation support
* and some writing some status markers which the shared Tracker extension can then reuse
* any fixes that go into that, looks like https://github.com/apache/jackrabbit-oak/blob/trunk/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java#L305 is also missing from here

oak-search-lucene-ng/src/main/java/org/apache/jackrabbit/oak/plugins/index/luceneNg/LuceneNgIndexTracker.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * Tracks Lucene 9 indexes and provides access to index nodes.
+ * Scans the repository for lucene9 type indexes and maintains a cache.
+ */
+public class LuceneNgIndexTracker {
```
amit-jain: Again this should extend from the shared [FulltextIndexTracker](https://github.com/apache/jackrabbit-oak/blob/trunk/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndexTracker.java#L55)
This is critical as the shared tracker has a lot of concurrency issues which have been resolved (actually LuceneIndexTracker does not extend this even though the code is almost the same)
* https://github.com/bhabegger/jackrabbit-oak/blob/c335365941593ac2651d82f07c07443adefd9e6e/oak-search-lucene-ng/src/main/java/org/apache/jackrabbit/oak/plugins/index/luceneNg/LuceneNgIndexTracker.java#L118 - uses index state which could be expensive
The code can reuse the metadata written by the FulltextIndexEditor - https://github.com/bhabegger/jackrabbit-oak/blob/c335365941593ac2651d82f07c07443adefd9e6e/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndexTracker.java#L133
* Nested index changes discovery  (https://github.com/bhabegger/jackrabbit-oak/blob/c335365941593ac2651d82f07c07443adefd9e6e/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/query/FulltextIndexTracker.java#L222-L243)

oak-search-lucene-ng/src/main/java/org/apache/jackrabbit/oak/plugins/index/luceneNg/internal/IndexSearcherHolder.java
```diff
@@ -0,0 +1,105 @@
...
+            new ConcurrentHashMap<>();
+
+    /**
+     * @param storageState {@link LuceneNgIndexStorage#storageState(NodeState)} for the index definition
+     * @param indexName    the index name, used only for logging/error messages
+     */
+
+    public IndexSearcherHolder(NodeState storageState, String indexName) throws IOException {
+        this.indexName = indexName;
+        this.directory = new OakDirectory(storageState.builder(), indexName, true);
```
amit-jain, line 56: NRT handling is an important part ans is also missing i think

amit-jain (changes requested):
I think there are some architectural gaps which we should address, some are even mentioned in the readme as well and could be addressed with these changes. 

Mainly we should reuse the FulltextIndexEditor and FulltextIndexTracker from the oak-search module which the elastic module also reuses and extends.

oak-search-lucene-ng/src/main/java/org/apache/jackrabbit/oak/plugins/index/luceneNg/LuceneNgIndexEditor.java
bhabegger: Will do. Thanks for the heads up.

oak-search-lucene-ng/src/main/java/org/apache/jackrabbit/oak/plugins/index/luceneNg/LuceneNgIndexTracker.java
bhabegger: Same here. Will do even if the legacy doesn't use it.

oak-search-lucene-ng/src/main/java/org/apache/jackrabbit/oak/plugins/index/luceneNg/internal/IndexSearcherHolder.java
bhabegger: I heard a lot about NRT being mostly abused and I think we could have a first go without it. I would like to test an already big change. Let's make sure everything works without NRT and then add it back in a second step.

amit-jain: > I heard a lot about NRT being mostly abused 

Ok I am not aware of the abuse, can you share more details.
Also, no problem in doing as a second step but I just highlighted things missing.

amit-jain:
Another important missing piece that is needed is the local fs index file support with the IndexCopier etc class which then get called with https://github.com/apache/jackrabbit-oak/blob/trunk/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/reader/DefaultIndexReaderFactory.java#L97

oak-search-lucene-ng/README.md
bhabegger: Let's vote here for this: https://github.com/orgs/oak-indexing/discussions/1289

oak-search-lucene-ng/src/main/java/org/apache/jackrabbit/oak/plugins/index/luceneNg/internal/IndexSearcherHolder.java
bhabegger: Some users put nrt just because they can, not because they need it.

bhabegger: Let's keep this like this for now. I would really like to test this first "core" implementation on indices where it already has sufficient features and then extend the features to cover the other cases.

Limitations are anyway documented here in the README

bhabegger:
> Another important missing piece that is needed is the local fs index file support with the IndexCopier etc class which then get called with https://github.com/apache/jackrabbit-oak/blob/trunk/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/reader/DefaultIndexReaderFactory.java#L97

Can you remind me what this is for ? Is this a must for this PR or can it be moved to later ?

oak-search-lucene-ng/src/main/java/org/apache/jackrabbit/oak/plugins/index/luceneNg/internal/IndexSearcherHolder.java
amit-jain: > Let's keep this like this for now. I would really like to test this first "core" implementation on indices where it already has sufficient features and then extend the features to cover the other cases.

This functionality was added afair as users complained adding a doc and not being able to search on it. 

> Some users put nrt just because they can, not because they need it.

And does it create a problem? But for users requiring it, not having this it creates a problem

Also, adding this with the extended common oak-search extension might have some quirks (new overloaded methods etc) which might be cleaner in a new impl as this.
BTW i am just looking at this from a parity pov and do not have a personal view on it. @thomasmueller can best suggest if we are ok not adding it initially.

amit-jain:
> > Another important missing piece that is needed is the local fs index file support with the IndexCopier etc class which then get called with https://github.com/apache/jackrabbit-oak/blob/trunk/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/reader/DefaultIndexReaderFactory.java#L97
> 
> Can you remind me what this is for ? Is this a must for this PR or can it be moved to later ?

https://jackrabbit.apache.org/oak/docs/query/lucene.html#copyonread

oak-search-lucene-ng/src/main/java/org/apache/jackrabbit/oak/plugins/index/luceneNg/internal/IndexSearcherHolder.java
bhabegger: > And does it create a problem? But for users requiring it, not having this it creates a problem
That's very general question that fits better around a coffee or a beer ;) 

But understood that there are cases where this current version will not work. And it never has been the intent of this first PR to be feature complete, only to be a first step towards it.

bhabegger: > Also, adding this with the extended common oak-search extension might have some quirks (new overloaded methods etc) which might be cleaner in a new impl as this.

I'm not sure what you meant exactly by this. If it is to generalize and simplify the code, I'm all in. Just note that, this PR was voluntarily built in a way to keep running production code not using this feature completely untouched. (Notice only tests have been adjusted to validate feature iso-behavior).

bhabegger:
> > > Another important missing piece that is needed is the local fs index file support with the IndexCopier etc class which then get called with https://github.com/apache/jackrabbit-oak/blob/trunk/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/reader/DefaultIndexReaderFactory.java#L97
> > 
> > 
> > Can you remind me what this is for ? Is this a must for this PR or can it be moved to later ?
> 
> https://jackrabbit.apache.org/oak/docs/query/lucene.html#copyonread

Ok, so this is in fact a performance optimization in the case the NodeStore is remote (many cases I agree). This doesn't block functional testing though (unless performance is so degraded that this becomes just not usable at all even restricted to testing). I would accept this for now, and leave open to solve the performance issue in maybe some way different than how it was managed in the legacy.

Ok, with you to differ ?

oak-search-lucene-ng/README.md
```diff
@@ -0,0 +1,109 @@
...
+| Excerpts | ✓ | ✓ | ✓ |
+| Ordering / sorting | ✓ | ✓ | ✓ |
+| Suggestions | ✓ | ✓ | ✗ |
+| Spellcheck | ✓ | ✓ | ✗ |
+| Similarity / More Like This | ✓ | ✓ (+ KNN) | ✗ |
+| Native queries | ✓ | ✓ | ✗ |
+| Index statistics / JMX | ✓ | ✓ | ✗ |
+| Index augmentors [^1] | ✓ | ✗ | ✗ |
+| NRT / hybrid indexing | ✓ | ✗ | ✗ |
+| Index copier (CopyOnRead/Write) | ✓ | ✗ | ✗ |
```
amit-jain, line 22: This shuold be a required feature to add and not just good to have. Without this, the luceneNG index will not work in downstream applications. The reads from remote blob stores will trigger the readiness probes (on k8s specifically) and likely take a very long time for pods to start.

This was added after oberserving such problems in production and nothing in lucene9  imo ameliorates that problem.

bhabegger:
> Another important missing piece that is needed is the local fs index file support with the IndexCopier etc class which then get called with https://github.com/apache/jackrabbit-oak/blob/trunk/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/reader/DefaultIndexReaderFactory.java#L97

FYI, last commit adds CopyOnRead

PR #2825, opened 2026-03-31
OAK-12171 Cost estimation for is (not) null checks are wrong

thomasmueller:


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexProviderService.java
```diff
@@ -369,6 +371,9 @@ private void activate(BundleContext bundleContext, Configuration config) throws
         }
 
         whiteboard = new OsgiWhiteboard(bundleContext);
+        oakRegs.add(whiteboard.register(FeatureToggle.class,
+                new FeatureToggle(FulltextIndexPlanner.FT_OAK_12171, FulltextIndexPlanner.FT_OAK_12171_DISABLE),
+                emptyMap()));
```
fabriziofortino, line 376: do we need to register the FT also in the Elastic counterpart?

fabriziofortino (approved):

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexProviderService.java
thomasmueller: Good point! Even thought it's a global variable (final static), there might be two instances because of OSGi (different classloaders).

fabriziofortino (approved):

PR #2831, opened 2026-04-08
OAK-12057: Add feature toggle for LIMIT-aware index selection

bhabegger:
## Summary

Adds a feature toggle (`FT_IGNORE_LIMIT_IN_INDEX_SELECTION`) to enable gradual rollout of OAK-12057 (wrong index selection when using LIMIT).

- **Default (ON)**: Uses the optimized behavior from PR #2724 - ignores LIMIT/OFFSET when calculating index costs
- **Toggle OFF**: Reverts to legacy behavior - caps entry counts by offset+limit

## Changes

- Added toggle constant and infrastructure in `QueryEngineSettings`
- Registered feature in `Oak.java`
- Wrapped index entry count capping logic in `QueryImpl.java` with toggle check

Allows safe gradual deployment without requiring immediate rollout to all environments.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java
@@ -597,6 +597,10 @@ public class Oak {
             LOG.info("Registered optimize XPath union feature: " + QueryEngineSettings.FT_OPTIMIZE_XPATH_UNION);
             closer.register(optimizeXPathUnion);
             queryEngineSettings.setOptimizeXPathUnion(optimizeXPathUnion);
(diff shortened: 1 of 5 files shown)


reschke (changes requested):
There's an API change that requires updating the package version. Please see the build log.

thomasmueller (approved):

thomasmueller:
So, that means 

* If the feature toggle is unavailable or returning "true", then we use the new behavior. 
* if the feature toggle is available and returning "false", then we switch back to the old (legacy) behavior.

Right?

bhabegger:
@thomasmueller Yes, exactly ;) But default we want to ignore the LIMIT (which was the legacy behavior).

thomasmueller:
Merged into the branch

PR #2842, opened 2026-04-11
OAK-12152 : replace guava cache with caffeine in oak-search-elastic module

rishabhdaim:


oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexStatistics.java
```diff
@@ -208,25 +191,13 @@ private Long getCacheRefreshSeconds() {
...
-            CompletableFuture<Integer> task = CompletableFuture.supplyAsync(() -> {
-                try {
-                    return count(crd);
-                } catch (IOException e) {
-                    throw new CompletionException(e);
-                }
-            }, REFRESH_EXECUTOR);
-            return FutureConverter.toListenableFuture(task);
-        }
-
```
rishabhdaim, line 229: no need for reload for caffeince cache, it handles that part internally.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexStatistics.java
```diff
@@ -241,25 +212,13 @@ private int count(StatsRequestDescriptor crd) throws IOException {
...
-            CompletableFuture<StatsResponse> task = CompletableFuture.supplyAsync(() -> {
-                try {
-                    return stats(crd);
-                } catch (IOException e) {
-                    throw new CompletionException(e);
-                }
-            }, REFRESH_EXECUTOR);
-            return FutureConverter.toListenableFuture(task);
-        }
-
```
rishabhdaim, line 262: no need for caffeine cache.

thomasmueller (approved):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexStatistics.java
jsedding: Do we (currently) allow toggling between the Caffeine and Guava implementations?

If that's the case, we would need to somehow keep the "reload" functionality for Guava, right?

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/api/CacheBuilder.java
```diff
@@ -171,6 +173,19 @@ public CacheBuilder<K, V> refreshAfterWrite(@NotNull Duration duration) {
...
 
+    /**
+     * Sets the nanosecond ticker used to measure time for expiry and refresh.
+     * Intended for testing with a controllable clock.
+     *
+     * @param ticker a supplier returning the current time in nanoseconds (must not be null)
+     * @return this builder
+     */
+    @NotNull
+    public CacheBuilder<K, V> ticker(@NotNull Supplier<Long> ticker) {
```
jsedding, line 184: I like this signature, as it is very simple and versatile.

For convenience, should we add a second signature accepting a `Clock`? It could have a default implementation, which would be the same implementation you used in `ElasticIndexStatistics`. I expect we will have clock instances around in quite a few other places? This is just a suggestion and 100% optional. 🙂

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexStatistics.java
```diff
@@ -100,7 +89,7 @@ public class ElasticIndexStatistics implements IndexStatistics {
      */
     @Override
     public int numDocs() {
-        return countCache.getUnchecked(new StatsRequestDescriptor(elasticConnection, indexDefinition.getIndexAlias()));
+        return countCache.get(new StatsRequestDescriptor(elasticConnection, indexDefinition.getIndexAlias()));
```
jsedding, line 92: I don't know if `new StatsRequestDescriptor(elasticConnection, indexDefinition.getIndexAlias())` is constant, i.e. if the return value of `getIndexAlias()` ever changes?

If it is constant, wouldn't it make sense to create a single instance in the constructor and store it in a field? This would avoid creating the instance over and over. Of course, I don't know how frequently this happens, so it may make no practical difference.

jsedding (approved):
Looks good! I left a few questions and suggestions.

fabriziofortino (approved):

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/api/CacheBuilder.java
rishabhdaim: will add this in next PR.

PR #2847, opened 2026-04-13
OAK-12180 - AzureRepositoryLock renewal thread should not terminate

ahanikel:
There are still situations where an exception occurs in lease renewal and the thread exits, meaning it will block writes forever and never recover, even if the cause of the exception is temporary.


reschke:
@ahanikel - is this something that needs to go into the next release? (2.0.0)

ahanikel:
@reschke If possible, yes: if any connectivity issues / DNS failures happen, the lease renewal thread currently terminates, making it unable to recover if the condition goes away. With this change we tell the renewal thread to keep on trying.

ahanikel:
SonarQube fails because ```Error:  Project not found. Please check the 'sonar.projectKey' and 'sonar.organization' properties, the 'SONAR_TOKEN' environment variable, or contact the project administrator to check the permissions of the user the token belongs to``` so unrelated.

reschke:
@ahanikel - Sonar runs only on branches in the jackrabbit-oak repo.

PR #2850, opened 2026-04-14
OAK-12181 - add metric when repository lock could not be renewed

smiroslav:
https://issues.apache.org/jira/browse/OAK-12181

### DIFF
--- a/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzurePersistence.java
+++ b/oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzurePersistence.java
@@ -113,13 +113,16 @@ public class AzurePersistence implements SegmentNodeStorePersistence {
         BlockBlobClient blockBlobClient = getBlockBlob("repo.lock");
         BlockBlobClient noRetryBlockBlobClient = getNoRetryBlockBlob("repo.lock");
         BlobLeaseClient blobLeaseClient = new BlobLeaseClientBuilder().blobClient(noRetryBlockBlobClient).buildClient();
-        return new AzureRepositoryLock(blockBlobClient, blobLeaseClient, () -> {
-            log.error("Lost connection to Azure. The repository lock lease could not be renewed.");
-            RemoteStoreMonitor monitor = remoteStoreMonitor.get();
-            if (monitor != null) {
-                monitor.repositoryLockLost();
-            }
-        }, writeAccessController).lock();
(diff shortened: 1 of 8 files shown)


oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/v8/AzurePersistenceV8.java
```diff
@@ -122,7 +131,8 @@ private CloudAppendBlob getAppendBlob(String path) throws IOException {
         }
     }
 
-    private static void attachRemoteStoreMonitor(RemoteStoreMonitor remoteStoreMonitor) {
+    private void attachRemoteStoreMonitor(RemoteStoreMonitor remoteStoreMonitor) {
+        this.remoteStoreMonitor.set(remoteStoreMonitor);
```
jsedding, line 135: Would it make sense to only allow `RemoteStoreMonitor` to be set once?

```suggestion
        this.remoteStoreMonitor.getAndUpdate(oldValue -> {
            if (oldValue == null) {
                return remoteStoreMonitor;
            }
            throw new IllegalStateException("attachRemoteStoreMonitor may only be called once");
        });
```

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/AzurePersistence.java
```diff
@@ -141,6 +150,7 @@ private AppendBlobClient getAppendBlob(String path) throws IOException {
     }
 
     private void attachRemoteStoreMonitor(RemoteStoreMonitor remoteStoreMonitor) {
+        this.remoteStoreMonitor.set(remoteStoreMonitor);
```
jsedding, line 153: See comment for V8.

jsedding (approved):
Looks good @smiroslav. Just one question/suggestion. Up to you if it makes sense or not.

oak-segment-azure/src/main/java/org/apache/jackrabbit/oak/segment/azure/v8/AzurePersistenceV8.java
smiroslav: Thanks @jsedding for the suggestion. 
I am thinking since the method is public, theoretically it could be called more than once for the same persistence, not that I see how it would make sense. I am a bit reluctant to add that set-once guard since we do not document this behaviour in the interface.

PR #2854, opened 2026-04-14
OAK-12156 : removed guava cache from oak-store-document caches

rishabhdaim:


thomasmueller (commented):
I didn't find an obvious problem, but I just wonder if it would have been saver to retain the old behavior (and old code), eg. the change for getCause(). The idea is that if the behavior is the same, then the PR would be smaller, and so easier to analyze.

thomasmueller (approved):

PR #2864, opened 2026-04-16
OAK-12161 : removed guava's cachestats and did the final cleanup

rishabhdaim:
Renamed CacheStatsAdaptor. -> CacheStats


oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/AbstractCacheStats.java
```diff
@@ -38,8 +38,8 @@ public abstract class AbstractCacheStats extends AnnotatedStandardMBean implemen
     @NotNull
     private final String name;
 
-    private CacheStats lastSnapshot =
-            new CacheStats(0, 0, 0, 0, 0, 0);
+    private CacheStatsSnapshot lastSnapshot =
+            new CacheStatsSnapshot(0, 0, 0, 0, 0, 0);
```
jsedding, line 42: `lastSnapshot` should be `volatile`. It isn't properly `synchronized` anyway, and synchronization is not nessesary.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/AbstractCacheStats.java
```diff
@@ -51,19 +51,17 @@ protected AbstractCacheStats(@NotNull String name) {
...
+     * Call back invoked to retrieve the most recent {@link CacheStatsSnapshot} of the
      * underlying cache.
      */
-    protected abstract CacheStats getCurrentStats();
+    protected abstract CacheStatsSnapshot getCurrentStats();
 
-    private CacheStats stats() {
+    private CacheStatsSnapshot stats() {
         return getCurrentStats().minus(lastSnapshot);
     }
```
jsedding, line 65: Together with the proposal to make `lastSnapshot` volatile, we should make sure to first get the value of `lastSnapshot` before calling `getCurrentStats(). Otherwise it is possible that a `resetStats()` call happens after `getCurrentStats()` and reading `lastSnapshot`, leading to negative values.

```suggestion
    private CacheStatsSnapshot stats() {
        // get baseline before calling getCurrentStats() to guard against race with resetStats()
        CacheStatsSnapshot baseline = lastSnapshot;
        return getCurrentStats().minus(baseline);
    }
```

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/AbstractCacheStats.java
jsedding, line 19: This probably should move to the `org.apache.jackrabbit.oak.cache.api`, I think. The class is used by other bundles, right? So it needs to be exported, and probably annotated `@ProviderType`.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/AbstractCacheStats.java
```diff
@@ -51,19 +51,17 @@ protected AbstractCacheStats(@NotNull String name) {
...
 
     @Override
     public synchronized void resetStats() {
```
jsedding, line 68: If we switch `lastSnapshot` to volatile, this no longer needs to be synchronized.

```suggestion
    public void resetStats() {
```

oak-core-spi/src/test/java/org/apache/jackrabbit/oak/cache/api/CacheStatsTest.java
```diff
@@ -20,7 +20,7 @@
 import org.junit.Test;
 
 /** Tests for {@link CacheStatsSnapshot}. */
-public class CacheStatsSnapshotTest {
+public class CacheStatsTest {
```
jsedding: The name should be `CacheStatsSnapshotTest`.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/WriterCacheManager.java
```diff
@@ -351,17 +351,20 @@ public CacheStatsMBean getTemplateCacheStats() {
...
+            return () -> {
+                long hits = 0, misses = 0, loads = 0, failures = 0, loadTime = 0, evictions = 0;
+                for (RecordCache<?> cache : caches) {
+                    CacheStatsSnapshot s = cache.getStats();
+                    hits += s.hitCount();
+                    misses += s.missCount();
+                    loads += s.loadSuccessCount();
+                    failures += s.loadFailureCount();
+                    loadTime += s.totalLoadTime();
+                    evictions += s.evictionCount();
```
jsedding: We could consider implementing `CacheStatsSnapshot#plus(CacheStatsSnapshot)`. It already implements `#minus(CacheStatsSnapshot)`.

jsedding (changes requested):
Please take a look at my comments. Overall it looks good to me.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/WriterCacheManager.java
rishabhdaim: added in https://github.com/apache/jackrabbit-oak/pull/2864/commits/809e98790f5821f7245237d050505a510a366796

oak-core-spi/src/test/java/org/apache/jackrabbit/oak/cache/api/CacheStatsTest.java
rishabhdaim: added in https://github.com/apache/jackrabbit-oak/pull/2864/commits/809e98790f5821f7245237d050505a510a366796

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/AbstractCacheStats.java
rishabhdaim: As discussed offline, we have moved `CacheStats` and `CacheStatsSnapshot` into same package as that of `AbstractCacheStats` in https://github.com/apache/jackrabbit-oak/pull/2864/commits/809e98790f5821f7245237d050505a510a366796

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/api/CacheStatsSnapshot.java
```diff
@@ -89,4 +130,23 @@ public CacheStatsSnapshot minus(@NotNull CacheStatsSnapshot other) {
...
+     */
+    @NotNull
+    public CacheStatsSnapshot plus(@NotNull CacheStatsSnapshot other) {
+        return new CacheStatsSnapshot(
+                Long.sum(hitCount, other.hitCount),
+                Long.sum(missCount, other.missCount),
+                Long.sum(loadSuccessCount, other.loadSuccessCount),
+                Long.sum(loadFailureCount, other.loadFailureCount),
+                Long.sum(totalLoadTime, other.totalLoadTime),
+                Long.sum(evictionCount, other.evictionCount)
```
jsedding: Very unlikely to happen in practice. But `Math.addExact()` throws if a long overflow happens. `Long.sum)()` quietly overflows.

```suggestion
                Math.addExact(hitCount, other.hitCount),
                Math.addExact(missCount, other.missCount),
                Math.addExact(loadSuccessCount, other.loadSuccessCount),
                Math.addExact(loadFailureCount, other.loadFailureCount),
                Math.addExact(totalLoadTime, other.totalLoadTime),
                Math.addExact(evictionCount, other.evictionCount)
```

jsedding (approved):
Looks good. I have one more suggestion, but I don't expect this to cause issues in practice. Thanks @rishabhdaim!

PR #2866, opened 2026-04-17
OAK-12144: oak-lucene: remove embed of Cache classes

mbaedke:
Added test case.


reschke (commented):
This test actually does not fail when the cache embeds are removed.

That said, the new tests in IndexVersionSelectionIT. do in setUp().

PR #2868, opened 2026-04-20
OAK-12190 Add createUserWithAbsolutePath and createGroupWithAbsolutePath to UserManager API

nscendoni:


oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/security/user/UserManager.java
```diff
@@ -213,6 +213,34 @@ User createUser(@NotNull String userID, @Nullable String password, @NotNull Prin
...
+     * @throws AuthorizableExistsException              if an authorizable with the given
+     *                                                  userID or principal already exists.
+     * @throws UnsupportedRepositoryOperationException  if the implementation does not
+     *                                                  support creation at an absolute path.
+     * @throws RepositoryException                      If another error occurs.
+     */
+    @NotNull
+    default User createUserWithAbsolutePath(@NotNull String userID, @Nullable String password,
+                                            @NotNull Principal principal, @NotNull String absoluteOakPath)
+            throws AuthorizableExistsException, UnsupportedRepositoryOperationException, RepositoryException {
```
anchela, line 240: this should be an absolute JCR path that is then converted to an oak path (there are utilities for that). the reason: user management API is on the same level and an extension to JCR API. oak API is one layer below.
see JCR specification for the namespace remapping and the various representations of a JCR path. while i doubt that namespace remappings and the different variants of paths are widely use, it's still important to be aware of the distictions to keep the API contract consistent.

paramName: absolutePath... not mentioning oak

oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/security/user/UserManager.java
```diff
@@ -213,6 +213,34 @@ User createUser(@NotNull String userID, @Nullable String password, @NotNull Prin
...
+        throw new UnsupportedRepositoryOperationException("createUserWithAbsolutePath is not supported by this implementation");
```
anchela: ideally we would have a default implementation that avoids having to touch the old jackrabbit code.

oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/security/user/UserManager.java
```diff
@@ -305,6 +333,36 @@ User createUser(@NotNull String userID, @Nullable String password, @NotNull Prin
...
+     * @return The new {@code Group}.
+     * @throws AuthorizableExistsException             if an authorizable with the given
+     *                                                 groupID or principal already exists.
+     * @throws UnsupportedRepositoryOperationException if the implementation does not
+     *                                                 support creation at an absolute path.
+     * @throws RepositoryException                     If another error occurs.
+     */
+    @NotNull
+    default Group createGroupWithAbsolutePath(@NotNull String groupID, @NotNull Principal principal,
+                                              @NotNull String oakPath)
```
anchela: see above it should be 'absolutePath' and the description should make it clear that this is a JCR path not an oak path.
naming should be consistent with the user method.

oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/security/user/UserManager.java
```diff
@@ -305,6 +333,36 @@ User createUser(@NotNull String userID, @Nullable String password, @NotNull Prin
...
+            throws AuthorizableExistsException, UnsupportedRepositoryOperationException, RepositoryException {
+        throw new UnsupportedRepositoryOperationException("createGroupWithAbsolutePath is not supported by this implementation");
```
anchela: see above. ideally we have a default implementation

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/UserManagerDelegator.java
```diff
@@ -168,6 +168,21 @@ public User perform() throws RepositoryException {
         });
     }
 
+    @NotNull
+    @Override
+    public User createUserWithAbsolutePath(@NotNull final String userID, @Nullable final String password,
+                                           @NotNull final Principal principal, @NotNull final String absoluteOakPath)
```
anchela: rename param here as well

oak-jcr/src/main/java/org/apache/jackrabbit/oak/jcr/delegate/UserManagerDelegator.java
```diff
@@ -233,6 +248,21 @@ public Group perform() throws RepositoryException {
         });
     }
 
+    @NotNull
+    @Override
+    public Group createGroupWithAbsolutePath(@NotNull final String groupID, @NotNull final Principal principal,
+                                             @NotNull final String oakPath)
```
anchela: rename param here as well

oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/security/user/UserManager.java
```diff
@@ -305,6 +333,36 @@ User createUser(@NotNull String userID, @Nullable String password, @NotNull Prin
     @NotNull
     Group createGroup(@NotNull String groupID, @NotNull Principal principal, @Nullable String intermediatePath) throws AuthorizableExistsException, RepositoryException;
 
+    /**
+     * Creates a new {@code Group} at the specified absolute Oak repository path.
```
anchela: the javadoc should describe what happens in the following cases:
- the path collides with an existing node
- the last segment (name) violates the contract of the configured authorizableNodeName

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserManagerImpl.java
```diff
@@ -251,6 +251,45 @@ public Group createGroup(@NotNull String groupID, @NotNull Principal principal,
...
 
+    @NotNull
+    @Override
+    public User createUserWithAbsolutePath(@NotNull String userID, @Nullable String password,
+                                           @NotNull Principal principal, @NotNull String oakPath)
+            throws AuthorizableExistsException, UnsupportedRepositoryOperationException, RepositoryException {
+        checkValidId(userID);
+        checkValidPrincipal(principal, false);
+
+        Tree userTree = userProvider.createUser(userID, oakPath);
```
anchela, line 266: before this call the jcr path needs to be converted to an oak path

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserManagerImpl.java
```diff
@@ -251,6 +251,45 @@ public Group createGroup(@NotNull String groupID, @NotNull Principal principal,
...
+
+    @NotNull
+    @Override
+    public Group createGroupWithAbsolutePath(@NotNull String groupID, @NotNull Principal principal,
+                                             @NotNull String oakPath)
+            throws AuthorizableExistsException, UnsupportedRepositoryOperationException, RepositoryException {
+        checkValidId(groupID);
+        checkValidPrincipal(principal, true);
+
+        Tree groupTree = userProvider.createGroup(groupID, oakPath);
```
anchela, line 291: before calling createGroup the path needs to be converted from jcr path to oak path

anchela (changes requested):
first bunch of comments..... will take another look later today

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserManagerImpl.java
```diff
@@ -251,6 +251,53 @@ public Group createGroup(@NotNull String groupID, @NotNull Principal principal,
...
+    @NotNull
+    @Override
+    public User createUserWithAbsolutePath(@NotNull String userID, @Nullable String password,
+                                           @NotNull Principal principal, @NotNull String absolutePath)
+            throws AuthorizableExistsException, UnsupportedRepositoryOperationException, RepositoryException {
+        checkValidId(userID);
+        checkValidPrincipal(principal, false);
+
+        String oakPath = namePathMapper.getOakPath(absolutePath);
+        if (oakPath == null) {
```
anchela, line 263: i would move those duplicate lines from here and the group method into a private method getOakPath

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserManagerImpl.java
```diff
@@ -251,6 +251,53 @@ public Group createGroup(@NotNull String groupID, @NotNull Principal principal,
...
+            throw new RepositoryException("Invalid path: " + absolutePath);
+        }
+        Tree userTree = userProvider.createUser(userID, oakPath);
```
anchela, line 266: the userProvider.createUser call will add an additional element for the user node.
this will not result in the absolute path being uses as it....

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserManagerImpl.java
```diff
@@ -251,6 +251,53 @@ public Group createGroup(@NotNull String groupID, @NotNull Principal principal,
...
+        setPrincipal(userTree, principal);
```
anchela, line 267: lines 267-276 will be duplicated from the regular create user call, no?
please refactor to avoid duplciation

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserManagerImpl.java
```diff
@@ -251,6 +251,53 @@ public Group createGroup(@NotNull String groupID, @NotNull Principal principal,
...
+    @NotNull
+    @Override
+    public Group createGroupWithAbsolutePath(@NotNull String groupID, @NotNull Principal principal,
+                                             @NotNull String absolutePath)
+            throws AuthorizableExistsException, UnsupportedRepositoryOperationException, RepositoryException {
+        checkValidId(groupID);
+        checkValidPrincipal(principal, true);
+
+        String oakPath = namePathMapper.getOakPath(absolutePath);
+        if (oakPath == null) {
```
anchela, line 288: see above

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserManagerImpl.java
```diff
@@ -251,6 +251,53 @@ public Group createGroup(@NotNull String groupID, @NotNull Principal principal,
...
+            throw new RepositoryException("Invalid path: " + absolutePath);
+        }
+        Tree groupTree = userProvider.createGroup(groupID, oakPath);
```
anchela, line 291: same as above... this will not use the absolute path as it is but will add a node-name segment generated with the available authorizablenodename implentation.

anchela (changes requested):
found new stuff.... UserProvider.createX would add an additional segment for the node name

also: do you want to offer this method also for service-users? they come with a separate create method.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserProvider.java
```diff
@@ -361,4 +371,39 @@ private String getNodeName(@NotNull String authorizableId) {
...
+            throw new ConstraintViolationException("Attempt to create authorizable at '" + absoluteOakPath
+                    + "' outside of the configured root '" + authRoot + "'");
+        }
+
+        Tree tree = root.getTree(parentPath);
+        while (!tree.isRoot() && !tree.exists()) {
+            tree = tree.getParent();
+        }
+        if (!tree.exists()) {
+            throw new AccessDeniedException("Missing permission to create intermediate authorizable folders.");
```
anchela: the exception message may be a bit misleading. it's actually that read permission is lacking to read an existing parent path, no? ((note: i am seeing this now but i may be consistent with the existing code)). maybe still worth doing better here: "Unable to retrieve authorizable parent folder" (or something like that)

oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/security/user/UserManager.java
```diff
@@ -305,6 +333,43 @@ User createUser(@NotNull String userID, @Nullable String password, @NotNull Prin
...
+     * Unlike {@link #createGroup(String, Principal, String)} where the
+     * {@code intermediatePath} is a relative hint that implementations may ignore,
+     * this method requires the caller to supply the exact absolute JCR path at which
+     * the group node must be created. The path must start with {@code /} and must
+     * be within the configured group root.
+     * <p>
+     * If a node already exists at the resolved location, the implementation may
+     * append a numeric suffix to the node name to avoid the collision.
+     * <p>
+     * If the last segment of the resolved node name violates the contract of the
```
anchela: the current implementation does not do that. so, i would recommend to state "an ConstraintViolationException may be thrown".
or if we want to keep that as a mandate then the impl should verify... but there is currently no API in AuthorizableNodeName to verify a given name... this would need to be added.
either way fine with me.

oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/security/user/UserManager.java
```diff
@@ -305,6 +333,43 @@ User createUser(@NotNull String userID, @Nullable String password, @NotNull Prin
...
+     *                                                 groupID or principal already exists.
+     * @throws UnsupportedRepositoryOperationException if the implementation does not
+     *                                                 support creation at an absolute path.
+     * @throws RepositoryException                     If another error occurs.
+     */
+    @NotNull
+    default Group createGroupWithAbsolutePath(@NotNull String groupID, @NotNull Principal principal,
+                                              @NotNull String absolutePath)
+            throws AuthorizableExistsException, UnsupportedRepositoryOperationException, RepositoryException {
+        return createGroup(groupID, principal, absolutePath);
```
anchela: calling the method that takes an intermediate path will not yield the desired result, no? maybe throwing "unsupportedrepositoryoperationexception" would be more aligned with the api contract. wdyt?

oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/security/user/UserManager.java
```diff
@@ -305,6 +333,43 @@ User createUser(@NotNull String userID, @Nullable String password, @NotNull Prin
...
+    /**
+     * Creates a new {@code Group} at the specified absolute JCR repository path.
+     * <p>
+     * Unlike {@link #createGroup(String, Principal, String)} where the
+     * {@code intermediatePath} is a relative hint that implementations may ignore,
+     * this method requires the caller to supply the exact absolute JCR path at which
+     * the group node must be created. The path must start with {@code /} and must
+     * be within the configured group root.
+     * <p>
+     * If a node already exists at the resolved location, the implementation may
```
anchela: i am not 100% convinced about this option to add a suffix. this reads like a contradiction to the mandate that the resulting group node has indeed the specified path. wdyt?

oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/security/user/UserManager.java
```diff
@@ -213,6 +213,34 @@ User createUser(@NotNull String userID, @Nullable String password, @NotNull Prin
...
+     *                                                 userID or principal already exists.
+     * @throws UnsupportedRepositoryOperationException if the implementation does not
+     *                                                 support creation at an absolute path.
+     * @throws RepositoryException                     If another error occurs.
+     */
+    @NotNull
+    default User createUserWithAbsolutePath(@NotNull String userID, @Nullable String password,
+                                            @NotNull Principal principal, @NotNull String absolutePath)
+            throws AuthorizableExistsException, UnsupportedRepositoryOperationException, RepositoryException {
+        return createUser(userID, password, principal, absolutePath);
```
anchela: see comment with createGroupWithAbsolute path below. same here

anchela (changes requested):
just minor nitpicking left. the rest all looks good now.
@nscendoni let me know what you think about my comments.

anchela (approved):
@nscendoni the changes look good to me now. in addition i asked claude to create test cases for renamed namespaces -> additions to RemappingTest.java.

@Amoratinos do you want to take a look as well before we merge?

anchela:
@reschke , we would appreciate if you could also take a look. it was reviewed by anchela, the ai-review tool from @Joscorbe . but we may still have missed something.

oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/security/user/UserManager.java
reschke: actually, that is really really important.

reschke (approved):

reschke (commented):
Here's a question: how does FileVault do this if not using an existing API?

Amoratinos (approved):
Looks good to me

nscendoni:
> Here's a question: how does FileVault do this if not using an existing API?

I'm not expert or contributor in FileVault, but my understanding is that it is importing them via [JCR SysView XML](https://developer.adobe.com/experience-manager/reference-materials/spec/jcr/1.0/6.4.1_System_View_XML_Mapping.html) import using `session.importXML(path, xml, IMPORT_UUID_COLLISION_*) ` 

We initially tried to use it to import users, but performance was very poor. Using the modified UserManager API, performance improved by several orders of magnitude.

reschke:
FWFW, we should get Sonar to run over this; thus it needs to be in a branch on the Apache repo.

reschke (changes requested):
FWFW, we should get Sonar to run over this; thus it needs to be in a branch on the Apache repo.

reschke:
@kwin - would you be able to use that in FileVault?

anchela:
@reschke , i reviewed and approved the latest changes. good from my pov

anchela (approved):
see previous comments

kwin:
> @kwin - would you be able to use that in FileVault?

As documented in https://jackrabbit.apache.org/filevault/authorizables.html: authorizables are updated/created via JCR SysView import. No dedicated JR API necessary.

PR #2879, opened 2026-04-29
OAK-12197 : make caffeine evictions always synchronous to aling with …

rishabhdaim:
…guava

### DIFF
--- a/oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/api/CacheBuilder.java
+++ b/oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/api/CacheBuilder.java
@@ -254,7 +254,8 @@ public final class CacheBuilder<K, V> {
 
     @SuppressWarnings({"unchecked", "rawtypes"})
     private Caffeine<K, V> configureCaffeineBuilder() {
-        Caffeine caffeineBuilder = Caffeine.newBuilder();
+        Caffeine caffeineBuilder = Caffeine.newBuilder()
+                .executor(Runnable::run);
         if (initialCapacity >= 0) {
             caffeineBuilder = caffeineBuilder.initialCapacity(initialCapacity);
         }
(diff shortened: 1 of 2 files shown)


oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/api/CacheBuilder.java
```diff
@@ -271,9 +272,6 @@ private Caffeine<K, V> configureCaffeineBuilder() {
             caffeineBuilder = caffeineBuilder.recordStats();
         }
         if (evictionListener != null) {
```
reschke, line 274: So that "just" makes setting the executor independant of the evictionListener?

oak-core-spi/src/test/java/org/apache/jackrabbit/oak/cache/impl/CacheBuilderTest.java
```diff
@@ -178,6 +178,33 @@ public void statsReturnsCorrectHitMissCounts() {
         Assert.assertEquals(1, stats.missCount());
     }
```
reschke, line 179: I assume these tests fail without the change?

oak-core-spi/src/test/java/org/apache/jackrabbit/oak/cache/impl/CacheBuilderTest.java
```diff
@@ -178,6 +178,33 @@ public void statsReturnsCorrectHitMissCounts() {
...
 
+    /** Zero-size caches evict written entries before they are visible to later reads. */
+    @Test
+    public void zeroMaximumSizeEvictsImmediately() {
+        Cache<String, String> cache = CacheBuilder.<String, String>newBuilder()
+                .maximumSize(0)
```
thomasmueller, line 185: A zero-size cache won't have any entries, always, right? If so, why add this test?

rishabhdaim: The caffeine can have a entry even with 0 size cache momentarily, since its eviction logic is async.

thomasmueller (approved):

PR #2886, opened 2026-05-05
OAK-12201 : added missing IterableUtils import

rishabhdaim:


reschke (commented):
How can it be missing if the source compiled before.....?

reschke (approved):

rishabhdaim:
From Codex: 

```
Strictly: yes, it can compile because CdCommand is Groovy, not Java, and it is not statically compiled. Groovy can compile an unresolved IterableUtils reference as a dynamic property/expression.

Practically: no, you should not remove it. The source uses IterableUtils.limit(...) at CdCommand.groovy (line 42). Without the import at CdCommand.groovy (line 23), it would likely fail at runtime with MissingPropertyException: No such property: IterableUtils.

I also checked the generated Java stub: it has empty method bodies, so its lack of IterableUtils does not prove the real Groovy source is safe without the import. Current oak-run compile succeeds with the import.
```

PR #2902, opened 2026-05-15
OAK-12213 Tree store: child node list maybe be incorrect if an intermediate node is missing

thomasmueller:


oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/tree/TreeStoreNodeState.java
```diff
@@ -201,7 +205,15 @@ private void fetch() {
                         if (index < 0) {
                             throw new IllegalArgumentException(key);
                         }
-                        current = key.substring(index + 1);
+                        if (!key.startsWith(path + "\t")) {
```
bhabegger, line 252: If I understand correctly this is a stop condition because we've reached the sibling following the current node ? So `treeStore.getSession().iterator(path)` returns all the paths starting a `path` depth first ? So the method should really be named `getDescendantNodeNamesIterator(String path)`, or do I miss understand something ?

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/tree/TreeStoreNodeState.java
bhabegger, line 233: To follow the sonar point, this could indeed moved to an explicit TreeStorePathIterator class. that you could just instanciate here.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/tree/TreeStoreNodeState.java
thomasmueller: > If I understand correctly this is a stop condition because we've reached the sibling following the current node ?

Yes. The bug is that a node was returned that is not a child node.

> So treeStore.getSession().iterator(path) returns all the paths starting a path depth first ?

It doesn't use depth-first; it returns the entries in key order. I have now updated the documentation.

>  So the method should really be named getDescendantNodeNamesIterator(String path)

The methods is _supposed_ to return only the direct children. But it returned non-children in some specific cases. This PR is supposed to fix that, so it only returns direct children.

oak-run-commons/src/main/java/org/apache/jackrabbit/oak/index/indexer/document/tree/TreeStoreNodeState.java
thomasmueller: Sure, this could be done, but I don't think it would make the code easier to understand. I think whether or not a separate class should be used here is an issue that is unrelated to his PR.

> Cognitive Complexity

For this case, I don't agree with Sonar that this method is too complex and should be split into multiple methods.

fabriziofortino (approved):

oak-run-commons/src/test/java/org/apache/jackrabbit/oak/index/indexer/document/tree/TreeStoreIterateTest.java
```diff
@@ -0,0 +1,74 @@
...
+            store.putNode("/test-node", "{}");
+            store.putNode("/test-node/child/node", "{}");
+            store.putNode("/test-node/child/node/test", "{}");
+            store.putNode("/test/child", "{}");
+            Iterator<NodeStateEntry> it = store.iterator();
+            NodeStateEntry e = it.next();
+            assertEquals("/test", e.getPath());
+            e = it.next();
+            assertEquals("/test-node", e.getPath());
+            Iterator<String> it2 = e.getNodeState().getChildNodeNames().iterator();
```
bhabegger, line 58: With these 2 iterators this test is hard to follow. Would there be any way to make it clearer to the reader ? I am missing the intent of the test, to be able to validate that the test is correctly checking that intent (which I am sure it is, knowing the author ;) )

thomasmueller: Ah sorry I see this comment now. So the first iterator is over all nodes, and we get /test and then /test-node. 

The second iterator is the children of /test-node: we expect no children. The bug was here.

But from the next node, /test-node/child/node, there is one child.

Yes I understand it's now too late to change the PR.

PR #2909, opened 2026-05-19
OAK-12215 Add failover mode configuration in oak-segment-azure

ierandra:
OAK-12215 Add failover mode configuration in oak-segment-azure


reschke:
does this have test coverage?

smiroslav:
Looking at the proposed code changes, the intent is not to have two storage accounts that will be used simultaneously during the Oak runtime. 

With that, I do not think the configuration for the second storage account is needed, since Oak only ever connects to one account at activation. 

The same effect can be achieved by having a deployment layer mount either the primary or the secondary storage account credentials into the existing config fields. Oak doesn't need to know which one it is.

Duplicating the credential surface in Configuration  adds maintenance cost without enabling new runtime behavior.

PR #2910, opened 2026-05-19
OAK-12190 Add createUserWithAbsolutePath and createGroupWithAbsolutePath to UserManager API

nscendoni:


anchela:
as discussed with @reschke and @nscendoni i am going to merge the PR into the branch such that we can create a PR to trunk that goes through sonar analysis

PR #2918, opened 2026-05-26
OAK-12181 - add metric when repository lock could not be renewed

smiroslav:
https://issues.apache.org/jira/browse/OAK-12181


smiroslav:
@reschke I run new tests on my machine with the Docker agent stopped. Build does not fail in that case for me.

reschke:
ok, will try here and then merge myself

reschke (approved):

PR #2921, opened 2026-05-27
OAK-12229 Rethrow IOException in JournalReader#computeNext

jelmini:
Rethrow IOException in JournalReader#computeNext instead of silently swallowing it.

Currently, an IOException while reading the journal is swallowed and is
interpreted as no/empty journal, which can lead to data loss. This
behavior was contradicted by JournalReader#computeNext own Javadoc.


thomasmueller (commented):
Please add a feature toggle (possibly combined with a system property)

PR #2926, opened 2026-06-02
OAK-12125: gracefully handle property index missing propertyNames (#2924)

thomasmueller:
A missing 'propertyNames' caused an NPE that escaped IndexUpdate's IllegalStateException catch and broke the async indexing cycle. There is a feature toggle to switch to the old behavior.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexEditorProvider.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexEditorProvider.java
@@ -51,7 +51,7 @@ public class PropertyIndexEditorProvider implements IndexEditorProvider {
 
     public static final String TYPE = "property";
 
-    public static final String FT_GRANITE_63829 = "FT_GRANITE-63829";
+    public static final String FT_OAK_12125 = "FT_OAK-12125";
 
     @Reference
     private MountInfoProvider mountInfoProvider = Mounts.defaultMountInfoProvider();
@@ -62,7 +62,7 @@ public class PropertyIndexEditorProvider implements IndexEditorProvider {
     @Activate
     private void activate(BundleContext bundleContext) {
         Whiteboard whiteboard = new OsgiWhiteboard(bundleContext);
-        this.feature = newFeature(FT_GRANITE_63829, whiteboard);
+        this.feature = newFeature(FT_OAK_12125, whiteboard);
     }
 
     @Deactivate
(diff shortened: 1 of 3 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexEditorProvider.java
```diff
@@ -42,14 +51,33 @@ public class PropertyIndexEditorProvider implements IndexEditorProvider {
...
     @Reference
     private MountInfoProvider mountInfoProvider = Mounts.defaultMountInfoProvider();
 
+    @Nullable
+    private Feature feature;
+
+    @Activate
+    private void activate(BundleContext bundleContext) {
+        Whiteboard whiteboard = new OsgiWhiteboard(bundleContext);
+        this.feature = newFeature(FT_GRANITE_63829, whiteboard);
```
reschke: Toggle name?

reschke (changes requested):

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexTest.java
```diff
@@ -1187,4 +1194,146 @@ private String getCommitPath(String changeNodeName) {
...
+                    .count();
+            assertEquals("Expected exactly one WARN for invalidIndex_dedup across two cycles, got: "
+                    + customLogs.getLogs(), 1, warnsForThisIndex);
+        } finally {
+            customLogs.finished();
+        }
+    }
+
+    private EditorHook hookWithFeatureEnabled(boolean enabled) {
+        PropertyIndexEditorProvider provider = new PropertyIndexEditorProvider();
```
reschke, line 1333: Apparently this works but why is the name of the toggle not needed here? (just asking)

tmvlad: The provider class doesn't get the toggle from the constructor (only the editor does). It has the toggle name hardcoded in the `activate` method, where only the osgiContext is needed, for the creating the whiteboard with the shared state.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexEditorProvider.java
tmvlad: I've seen a similar naming scheme used in other parts of the codebase and went along with it. Thanks for the catch, I'll revise it.

PR #2934, opened 2026-06-05
OAK-12239: Remove FT_PREFETCH_OAK-9780 and remove 'no prefetch' code path

reschke:


oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
```diff
@@ -466,11 +465,6 @@ public boolean getReadOnlyMode() {
         return isReadOnlyMode;
     }
 
-    public T setPrefetchFeature(@Nullable Feature prefetch) {
-        this.prefetchFeature = prefetch;
-        return thisBuilder();
-    }
-
     @Nullable
     public Feature getPrefetchFeature() {
```
rishabhdaim: `setPrefetchFeature()` was removed but `getPrefetchFeature()` and the `prefetchFeature` field remain (always null). Consider removing the getter and dead fields on the builder, store, and service as part of this cleanup.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
```diff
@@ -4046,18 +4043,11 @@ boolean isReadOnlyMode() {
     @Override
     public void prefetch(java.util.Collection<String> paths, NodeState rootState) {
         if (paths != null
-                && rootState instanceof DocumentNodeState
-                && isPrefetchEnabled()) {
+                && rootState instanceof DocumentNodeState) {
```
rishabhdaim: With the guard gone, queries that already use `option(prefetches N)` will always warm the document cache now, even when `FT_PREFETCH_OAK-9780` was off before. Worth a release-note callout and a post-upgrade check of `DOCUMENT_NODES_PREFETCH` metrics.

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/prefetch/CacheWarmingTest.java
```diff
@@ -78,11 +72,6 @@ public class CacheWarmingTest {
 
     private CountingMongoDatabase db;
 
-    @Before
-    public void enablePrefetch() {
-        System.setProperty(SYS_PROP_PREFETCH, "true");
-    }
-
     @AfterClass
```
rishabhdaim, line 75: After dropping sys-prop setup, consider adding a test that calls `DocumentNodeStore.prefetch()` (not `CacheWarming` directly) with default builder setup and asserts cache warming — that would lock in the always-on store path this PR introduces.

rishabhdaim:
1. `oak-store-document/AGENTS.md` still lists `FT_PREFETCH_OAK-9780` and `setPrefetchFeature()` — both removed here. 
2. Store-level prefetch is always on when `prefetch()` is called; `PrefetchDispatcher` is unrelated and still gated by `prefetchExternalChanges`. 
3. Please update the Prefetch section and drop that row from the toggles table.

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStoreBuilder.java
reschke: indeed

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
reschke: Not sure. We only checked that prefetch as already enabled "everywhere", no?

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/prefetch/CacheWarmingTest.java
reschke: I don't quite understand; is this about adding a test that we should have had before?

reschke:
@rishabhdaim - thx for the review, and please see my questions (PR minimally updated).

(this shows that even removing a toggle is not totally trivial)

oak-store-document/src/main/java/org/apache/jackrabbit/oak/plugins/document/DocumentNodeStore.java
rishabhdaim: In that case, we can ignore this comment.

rishabhdaim:
Outside this diff — `CacheWarmingTest.prefetch()` (lines ~138–169):

`prefetch()` still exercises `CacheWarming` directly, so it won't catch a regression in `DocumentNodeStore.prefetch()` now that prefetch is always-on at the store layer. Call the store API instead (same assertions):

```diff
-        CacheWarming cw = new CacheWarming(ds);
         DocumentNodeStore store = builderProvider.newBuilder().setAsyncDelay(0)
                 .setDocumentStore(ds).getNodeStore();
         ...
-        cw.prefetch(children, store.getRoot());
+        store.prefetch(children, store.getRoot());
         ...
-        cw.prefetch(paths, store.getRoot());
+        store.prefetch(paths, store.getRoot());
```

oak-store-document/src/test/java/org/apache/jackrabbit/oak/plugins/document/prefetch/CacheWarmingTest.java
rishabhdaim: see https://github.com/apache/jackrabbit-oak/pull/2934#issuecomment-4634141983

rishabhdaim (approved):

PR #2938, opened 2026-06-08
OAK-12244: Adding/removing mixin type not reflected in index

bhabegger:
index nodes that gain a mixin rule, delete stale docs when mixin rule is lost

When an existing node's applicable indexing rule changes at runtime (e.g. jcr:mixinTypes added or removed), FulltextIndexEditor did not update the index because propertiesChanged was never set — jcr:mixinTypes is not normally listed in a rule's property definitions.

### DIFF
--- a/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java
+++ b/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java
@@ -69,6 +69,23 @@ public class FulltextIndexEditor<D> implements IndexEditor, Aggregate.AggregateR
      */
     public static final AtomicBoolean FT_OAK_12193_DISABLE = new AtomicBoolean(false);
 
+    /**
+     * Feature toggle name for OAK-12244.
+     * When active (default), nodes that gain or lose a mixin type at runtime are
+     * correctly added to or removed from the fulltext index, even when
+     * {@code jcr:mixinTypes} is not listed in the rule's property definitions.
+     */
+    public static final String FT_OAK_12244 = "FT_OAK-12244";
+
+    /**
+     * Kill switch for the OAK-12244 mixin-transition tracking. Set to {@code true} to
+     * revert to the pre-OAK-12244 behavior where mixin additions and removals do not
+     * trigger index updates unless an indexed property also changed.
+     * Default is {@code false} (mixin-transition tracking active). Wired to the
+     * {@link #FT_OAK_12244} feature toggle at runtime.
+     */
+    public static final AtomicBoolean FT_OAK_12244_DISABLE = new AtomicBoolean(false);
+
     private static final List<Aggregate.Matcher> EMPTY_AGGREGATE_MATCHER_LIST = List.of();
 
     private final FulltextIndexEditorContext<D> context;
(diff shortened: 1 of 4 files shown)


oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java
```diff
@@ -158,6 +164,19 @@ public void leave(NodeState before, NodeState after)
             }
         }
 
+        if (wasIndexable && !isIndexable()) {
```
thomasmueller: I _think_ there might be some edge case where a node was changed (first triggering an update), _and_ at the same time the mixin was changed so it is deleted. I don't know if there is a test case for this, and if this would fail.

To cover this case, I think it would be better to use a "staged" logic:

toBeAdded = ... || ... || ... || ...
toBeDeleted = ... || ... || ... || ...
toBeUpdated = ... || ... || ... || ...

if (toBeDeleted) {
    // delete logic
} else if (toBeAdded) {
    // add logic
} else if (toBeUpdated) {
    // update logic
}

thomasmueller (changes requested):

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditor.java
bhabegger: Normally, here, we are working at commit level not at individual change level.

So, if, in the code a mixin type was added to node A and then A was removed. We would not see the first operation as the `after` NodeState would be null. So we should be fine.

However, your proposal could make the code easier to read ;) So, let me see if we can adjust that.

bhabegger: [AI] Done — I've restructured `leave()` to use staged logic as suggested:

```java
boolean toBeDeleted = wasIndexable && !isIndexable();
boolean toBeAdded   = !wasIndexable && isIndexable();
boolean toBeUpdated = wasIndexable && isIndexable() && propertiesChanged;

if (toBeDeleted) {
    // deleteDocuments
} else if (toBeAdded || toBeUpdated) {
    // addOrUpdate
}
```

The `else if` makes the mutual exclusivity explicit. I combined `toBeAdded` and `toBeUpdated` into a single branch since the action (`addOrUpdate`) is identical for both — the `before.exists()` flag already encodes the add-vs-update distinction passed into the writer.

All 1242 oak-lucene tests still pass.

bhabegger:
[AI] Added feature toggle `FT_OAK-12244` (kill switch `FT_OAK_12244_DISABLE`, default `false`) following the same pattern as `FT_OAK-12193`:

- **`FulltextIndexEditor`**: declares the toggle name and `AtomicBoolean` kill switch; the mixin-transition logic in `leave()` is now guarded by `!FT_OAK_12244_DISABLE.get()`; when disabled, pre-OAK-12244 behavior is used.
- **`LuceneIndexProviderService`**: registers the toggle with the OSGi whiteboard.
- **`LuceneIndexEditor2Test`**: two new tests verify that mixin-add and mixin-remove do **not** trigger index changes when the toggle is disabled.

thomasmueller (approved):

PR #2943, opened 2026-06-09
OAK-12247: Keep track of total indexed documents

bhabegger:
Persists \`totalIndexedNodes\` to \`:status\` after each indexing cycle so planners can know whether an index is empty.

**What it does:**

- Adds \`getTotalDocCount()\` to \`FulltextIndexWriter\` (default \`-1\` = not tracked)
- \`DefaultIndexWriter\`: calls \`commit()\` + \`numDocs()\` before \`close()\`
- \`MultiplexingIndexWriter\` and \`PooledLuceneIndexWriter\`: delegate to sub-writers
- \`ElasticIndexWriter\`: uses a \`LongAdder\` incremented from bulk response results and \`deleteByQuery\`; initialised to the previous total for incremental cycles
- \`LocalIndexWriter\` (NRT): returns \`-1\` — NRT docs are in-memory only, count is updated by the next async cycle
- \`FulltextIndexEditorContext.closeWriter()\`: reads \`getTotalDocCount()\` and writes it to \`:status/totalIndexedNodes\`; also fixes an Elastic gap where an empty reindex returned \`indexUpdated=false\` and \`REINDEX_COMPLETION_TIMESTAMP\` was never written
- Kill switch \`FT_OAK-12247\` (default off = tracking enabled) registered in \`LuceneIndexProviderService\`


bhabegger:
NOTE: The tracking issue was also noted (and accepted) in https://github.com/apache/jackrabbit-oak/pull/3085 (Maybe better to merge that one first and then revisit)

PR #2947, opened 2026-06-10
feat/OAK-12253 : add a mechanism to override FT defaults via system property

reschke:


oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/toggle/Feature.java
```diff
@@ -22,8 +22,11 @@
 import java.util.Collections;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import org.apache.jackrabbit.oak.commons.properties.SystemPropertySupplier;
 import org.apache.jackrabbit.oak.spi.whiteboard.Registration;
 import org.apache.jackrabbit.oak.spi.whiteboard.Whiteboard;
+import org.slf4j.Logger;
```
rishabhdaim, line 28: These `Logger` and `LoggerFactory` imports are not used anywhere in `Feature.java`.

Oak usually keeps imports clean; leaving them in may trip checkstyle or confuse readers. Drop both imports.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/toggle/Feature.java
```diff
@@ -60,7 +63,9 @@ private Feature(AtomicBoolean value, Registration registration) {
      * @return the feature toggle.
      */
     public static Feature newFeature(String name, Whiteboard whiteboard) {
-        AtomicBoolean value = new AtomicBoolean();
+        // by default the initial value is false, but it can be overridden by a system property
+        AtomicBoolean value = new AtomicBoolean(
+                SystemPropertySupplier.create("oak-feature." + name, false).get());
```
rishabhdaim: `newFeature` now honors `-Doak-feature.<name>` via `SystemPropertySupplier`, but that contract is only described in an inline comment.

`org.apache.jackrabbit.oak.spi.toggle` is exported — operators and integrators read the Javadoc, not comments. Document the property name pattern (`oak-feature.<name>`), that it is read once at creation, and that the default stays `false` when unset (class-level Javadoc and on `newFeature`).

oak-core-spi/src/test/resources/logback-test.xml
```diff
@@ -27,7 +27,7 @@
     </encoder>
 </appender>
 
-<root level="ERROR">
+<root level="INFO">
```
rishabhdaim: Raising the module test root logger from `ERROR` to `INFO` affects every test in `oak-core-spi`, not just `FeatureToggleTest`.

That is likely only needed because `SystemPropertySupplier` logs at INFO when a property differs from default. Prefer a targeted logger instead:

```xml
<logger name="org.apache.jackrabbit.oak.commons.properties.SystemPropertySupplier" level="INFO"/>
```

…and keep `<root level="ERROR">`, or call `logSuccessAs("TRACE")` on the supplier in production code.

oak-core-spi/src/test/java/org/apache/jackrabbit/oak/spi/toggle/FeatureToggleTest.java
```diff
@@ -50,6 +50,17 @@ public void disabledByDefault() {
         }
     }
 
+    @Test
+    public void defaultOverridden() {
```
rishabhdaim: `defaultOverridden` covers the happy path (`oak-feature.my.toggle=true`).

Optional: add cases for an explicit `false` override and for sysprop `true` followed by `FeatureToggle.setEnabled(false)` on the same toggle name — the existing `register()` test uses a different name and does not cover that interaction.

rishabhdaim:
Findings outside this diff:

`AGENTS.md` (~Feature Toggles): the new `-Doak-feature.<toggleName>=true` JVM property is not mentioned. Root `AGENTS.md` still describes whiteboard-only enablement. Worth a bullet noting that the initial state can be set at `Feature.newFeature` time via `-Doak-feature.<toggleName>=true`.

reschke:
> AGENTS.md (~Feature Toggles): the new -Doak-feature.<toggleName>=true JVM property is not mentioned. Root AGENTS.md still describes whiteboard-only enablement. Worth a bullet noting that the initial state can be set at Feature.newFeature time via -Doak-feature.<toggleName>=true.

I have a problem with that. We should not duplicate informations. This needs to go into oak-doc somewhere (as everything else we have to say about feature toggles).

rishabhdaim (approved):

rishabhdaim:
> > AGENTS.md (~Feature Toggles): the new -Doak-feature.=true JVM property is not mentioned. Root AGENTS.md still describes whiteboard-only enablement. Worth a bullet noting that the initial state can be set at Feature.newFeature time via -Doak-feature.=true.
> 
> I have a problem with that. We should not duplicate information. This needs to go into [oak-doc](https://issues.apache.org/jira/browse/OAK-doc) somewhere (as everything else we have to say about feature toggles).

Then, this would defeat the very purpose of adding the AGENTS.md file. AI wouldn't be able to read this information from [oak-doc](https://issues.apache.org/jira/browse/OAK-doc)

jsedding (approved):
Makes sense.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/toggle/Feature.java
```diff
@@ -60,7 +70,10 @@ private Feature(AtomicBoolean value, Registration registration) {
      * @return the feature toggle.
      */
     public static Feature newFeature(String name, Whiteboard whiteboard) {
-        AtomicBoolean value = new AtomicBoolean();
+        // by default the initial value is false, but it can be overridden by a system property
+        AtomicBoolean value = new AtomicBoolean(
+                SystemPropertySupplier.create("oak-feature." + name, false).
```
mreutegg: I'm not too happy with this change. One of the goals of the toggle SPI is to provide an abstraction for feature toggles. The SPI is currently independent of any source where the toggle information comes from. Controlling the toggle through system properties is just one option. I'm also a bit concerned there is no way to opt out. This change introduces hard-coded link between system properties and feature toggles. I would prefer an implementation that is opt in if a user wants to control feature toggles through system properties and then maintain that implementation in a differnt package / bundle.

reschke: That's why I waited for your feedback :-)

Hmmm. The initial idea was to make this consistent.

Would it be ok to let the code creating the feature to opt-in using a parameter in `newFeature`? That way, existing code would remain unaffected. OTOH, letting the toggle to be defaulted from a system property would be made consistent in naming and behavior.

EDIT: hopefully fixed re FeatureToggle vs Feature.

reschke:
> Then, this would defeat the very purpose of adding the AGENTS.md file. AI wouldn't be able to read this information from [oak-doc](https://issues.apache.org/jira/browse/OAK-doc)

Well, then our current approach is in total violation of the DRY rule. We can't maintain both. Either let the documentation link to the AGENTS.md file (hoping it is accurate), or find a way to refresh the file when the documentation changes.

reschke:
@mreutegg - something like this...?
```

    private static Feature internalNewFeatureWithSystemPropertyDefault(String name, Whiteboard whiteboard, boolean withSysPropDefault) {
        // by default the initial value is false, but it can be overridden by a system property
        AtomicBoolean value = withSysPropDefault ? new AtomicBoolean(
                SystemPropertySupplier.create("oak-feature." + name, false).
                        loggingTo(LOG).get()) : new AtomicBoolean();
        FeatureToggle adapter = new FeatureToggle(name, value);
        return new Feature(value, whiteboard.register(
                FeatureToggle.class, adapter, Collections.emptyMap()));
    }

    /**
     * Creates a new {@link Feature} with the given name and registers the
     * corresponding {@link FeatureToggle} on the {@link Whiteboard}.
     * Client code must call {@link Feature#close()} when the toggle is
     * not used anymore.
     *
     * @param name the name of the feature toggle.
     * @param whiteboard the whiteboard where to register the feature toggle.
     * @return the feature toggle.
     */
    public static Feature newFeature(String name, Whiteboard whiteboard) {
        return internalNewFeatureWithSystemPropertyDefault(name, whiteboard, false);
    }

    /**
     * Same as {@linkplain #newFeature(String, Whiteboard)}m but with the initial state provided
     * by a system property, named based on the feature's name.
     */
    public static Feature newFeatureWithSystemPropertyDefault(String name, Whiteboard whiteboard) {
        return internalNewFeatureWithSystemPropertyDefault(name, whiteboard, true);
    }

```

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/toggle/Feature.java
mreutegg: I should have been more precise. When I wrote "user", I meant someone deploying an application that contains Oak and not Oak code using the toggle SPI.

To me code in Oak using the toggle SPI should not have to know how a `Feature` is controlled. This is a deployment choice. Some may hook it up to LaunchDarkly or something similar. This requires additional code that interacts with `FeatureToggle`.

reschke: ok, what this intended to help with is this use case:
new feature toggle added, quickly run tests (locally) without any container

I also have seen cases where in the code, a sysprop *is* used to initialize the feature toggle. (would need to cehck whether this was a singular case). If that make sense, it would be good to make that pattern consistent.

oak-core-spi/src/test/java/org/apache/jackrabbit/oak/spi/toggle/FeatureToggleTest.java
```diff
@@ -50,6 +51,28 @@ public void disabledByDefault() {
         }
     }
 
+    @Test
```
rishabhdaim, line 54: Optional (for local consideration only — not pushing code): `defaultOverriddenAsTrue` / `defaultOverriddenAsFalse` already cover explicit boolean overrides. For other `oak-feature.*` values, `SystemPropertySupplier` uses `Boolean.valueOf(String)` semantics: only `"true"` (case-insensitive) enables the toggle; strings like `"yes"`, `"garbage"`, `"1"`, etc. are treated as `false`, not as a parse error with log-and-fallback. If you want to document that contract in tests, something like:

```java
    /**
     * SystemPropertySupplier uses {@link Boolean#valueOf(String)} for boolean props: only
     * {@code "true"} (case-insensitive) enables the toggle; values like {@code "yes"} or
     * {@code "garbage"} are treated as {@code false}, not as a parse error.
     */
    @Test
    public void defaultOverriddenWithNonTrueValueStaysDisabled() {
        String sysPropName = "oak-feature.my.toggle";
        for (String value : new String[] {"garbage", "yes", "1", "on", ""}) {
            System.setProperty(sysPropName, value);
            try (Feature feature = newFeatureWithSystemPropertyDefault("my.toggle", whiteboard)) {
                assertFalse("property value '" + value + "' must not enable the toggle", feature.isEnabled());
            } finally {
                System.clearProperty(sysPropName);
            }
        }
    }
```

rishabhdaim:
**Findings outside the diff (impact review follow-up)**

**Documentation (`newFeatureWithSystemPropertyDefault`)** — The PR adds an opt-in path for JVM overrides, but user-facing docs still do not describe `newFeatureWithSystemPropertyDefault` or the `oak-feature.<toggleName>` naming. Thread consensus here is to put that in **oak-doc** (feature toggles section) rather than duplicating prose in `AGENTS.md`; a short link from agent guidance is enough if we want discoverability.

**Commit-Check / branch policy** — CI still rejects this PR: branch name `OAK-12253` does not match the expected `issue/OAK-12253` pattern, and several commits use non–Conventional Commits subjects (some over the 80-character limit). Squash/reword to conventional messages and rename the branch before merge.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/toggle/Feature.java
```diff
@@ -34,12 +37,20 @@
...
  * potentially comes with some overhead (e.g. when the whiteboard is based on
  * OSGi). Therefore, client code should not create a new feature, check the
- * state and then immediately release/close it again. Instead a feature should
+ * state and then immediately release/close it again. Instead, a feature should
  * be acquired initially, checked at runtime whenever needed and finally
  * released when the client component is destroyed.
+ * <p>
+ * The default state of {@code false} can be overridden by a system property
+ * using {@linkplain #newFeatureWithSystemPropertyDefault(String, Whiteboard)},
+ * in which case The property name is derived from the toggle name. This helps to quickly verify
```
rishabhdaim, line 46:
```suggestion
 * in which case the property name is derived from the toggle name. This helps to quickly verify
```

PR #2949, opened 2026-06-11
OAK-12244: index nodes that gain a mixin rule, delete stale docs when…

thomasmueller:
… mixin rule is lost (#2938)

TODO: right now the PR does a descendent-document delete if the primary type or mixin is changed / removed. This is incorrect and needs to be fixed.

When an existing node's applicable indexing rule changes at runtime (e.g. jcr:mixinTypes added or removed), FulltextIndexEditor did not update the index because propertiesChanged was never set — jcr:mixinTypes is not normally listed in a rule's property definitions.

Track wasIndexable (rule matched before) alongside isIndexable() (rule matches after). In leave(), act on transitions:
- !wasIndexable && isIndexable(): node gained a rule → addOrUpdate
- wasIndexable && !isIndexable(): node lost a rule → deleteDocuments

Tests added:
- PropertyIndexCommonTest: two end-to-end integration tests (all backends)
- LuceneIndexEditor2Test: two unit tests verifying writer.docs / writer.deletedPaths

### DIFF
--- a/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditorContext.java
+++ b/oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditorContext.java
@@ -89,6 +89,8 @@ public abstract class FulltextIndexEditorContext<D> {
 
     private final boolean indexDefnRewritten;
 
+    private final boolean typeChangeTrackingEnabled;
+
     private FulltextBinaryTextExtractor textExtractor;
 
     private PropertyUpdateCallback propertyUpdateCallback;
@@ -109,6 +111,7 @@ public abstract class FulltextIndexEditorContext<D> {
         this.updateCallback = updateCallback;
         this.extractedTextCache = extractedTextCache;
         this.asyncIndexing = asyncIndexing;
+        this.typeChangeTrackingEnabled = !FulltextIndexEditor.FT_OAK_12244_DISABLE.get();
         if (this.definition.isOfOldFormat()) {
             indexDefnRewritten = true;
             IndexDefinition.updateDefinition(definition, indexingContext.getIndexPath());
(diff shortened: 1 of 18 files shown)


oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditorContext.java
```diff
@@ -121,6 +124,10 @@ protected FulltextIndexEditorContext(NodeState root, NodeBuilder definition,
 
     public abstract DocumentMaker<D> newDocumentMaker(IndexDefinition.IndexingRule rule, String path);
 
+    public boolean isTypeChangeTrackingEnabled() {
+        return typeChangeTrackingEnabled;
+    }
+
```
fabriziofortino, line 130: I would remove this and rollback all the changes in this class. The `typeChangeTrackingEnabled` is always based on the FT flag and `isTypeChangeTrackingEnabled` is only called in `FulltextIndexEditor` that defines the toggle. Since the toggle should be removed after a while, I propose rollback the changes in this class and explicitly check the FT flag instead of calling `context.isTypeChangeTrackingEnabled()`.

fabriziofortino (approved):
The logic looks good to me. I just added a potential improvement.

Re feature toggles: the number of toggles are increasing. The code is getting more complex because of them (a lot of if/else). It's okay to have them in the short term, but I am concerned they won't be removed. I suggest adding a time-bombed test (eg https://github.com/apache/jackrabbit-oak/pull/2925/changes#diff-7f701488919abf1ac0a96ed15d558fc9615f2dc3a420f5ec58545fca43ba7990R38-R44) so that we won't forget to remove them.

reschke:
@fabriziofortino -> https://issues.apache.org/jira/browse/OAK-12233

oak-search/src/main/java/org/apache/jackrabbit/oak/plugins/index/search/spi/editor/FulltextIndexEditorContext.java
thomasmueller: Sure, we could move the code, but I think it wouldn't change all that much or make it much easier to maintain. I'll keep it as is. (Too many things going on right now sorry)

PR #2950, opened 2026-06-11
OAK-12248: graceful 404 when ES alias is not yet provisioned

bhabegger:
## Summary

- Adds feature toggle `FT_OAK-12248` (off by default) to `ElasticIndexStatistics`
- When enabled, a 404 `ElasticsearchException` from an ES `count` request during query planning is caught in `getCountOrZeroOn404()`, logged at INFO, and treated as 0 estimated documents — the planner stays on the ES path instead of falling back to traversal with an ERROR log
- `ElasticResultRowAsyncIterator.onFailure()` / `hasNext()` also check the toggle to suppress ERROR at execution time if the alias is still missing
- Toggle registered in `ElasticIndexProviderService` alongside the existing `FT_OAK-12206` and `FT_OAK-12234` toggles

## Root cause note

The 404 first surfaces during **planning** (not execution): `ElasticIndexStatistics.getDocCountFor()` fires an ES `count` request whose `ElasticsearchException` propagates uncaught to `FulltextIndex.getPlans():176` and is logged as ERROR there. Oak's cache re-throws `RuntimeException` causes **unwrapped** (only checked exceptions are wrapped in `CompletionException`), so the catch must target `ElasticsearchException` directly.

## Test

`ElasticGraceful404QueryTest` (new):
- `queryOnMissingAlias_withToggleOff_logsError` — seeds data, provisions index, deletes alias, asserts ERROR log (confirms existing behaviour)
- `queryOnMissingAlias_withToggleOn_returnsEmptyWithoutError` — same setup, toggle on, asserts 0 results + no ERROR + INFO "alias not found"

## Relation to other tickets

Builds on OAK-12247 (track `totalIndexedNodes`). Prerequisite for OAK-12249 (lazy ES provisioning): without this toggle, a non-existent alias after an empty reindex would ERROR on every query.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexStatistics.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexStatistics.java
@@ -22,7 +22,9 @@ import java.time.Duration;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
+import co.elastic.clients.elasticsearch._types.ElasticsearchException;
 import co.elastic.clients.elasticsearch._types.query_dsl.Query;
 
 import org.apache.jackrabbit.oak.cache.api.CacheBuilder;
@@ -33,6 +35,8 @@ import org.apache.jackrabbit.oak.plugins.index.search.IndexStatistics;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.jetbrains.annotations.TestOnly;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import co.elastic.clients.elasticsearch._types.Bytes;
 import co.elastic.clients.elasticsearch.cat.indices.IndicesRecord;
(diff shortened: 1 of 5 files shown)


nit0906:
@bhabegger  - do we need a toggle for this ? 

The change seems very limited.

bhabegger:
> @bhabegger - do we need a toggle for this ?
> 
> The change seems very limited.

Yes, that's why it is on by default. This is just a precaution because it does change the behavior (maybe in some situations we would want an error if the index doesn't exist ?).

nit0906 (approved):

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexStatistics.java
```diff
@@ -195,7 +202,15 @@ static class CountCacheLoader implements CacheLoader<StatsRequestDescriptor, Int
 
         @Override
         public @NotNull Integer load(@NotNull StatsRequestDescriptor countRequestDescriptor) throws IOException {
-            return count(countRequestDescriptor);
+            try {
+                return count(countRequestDescriptor);
+            } catch (ElasticsearchException ee) {
+                if (ee.status() == 404 && FT_OAK_12248_ENABLE.get()) {
+                    LOG.debug("ES alias not found for index {} — treating as empty (OAK-12248)", countRequestDescriptor.index);
```
fabriziofortino: what about printing this at `INFO`? The value is actually cached so the number of logs coming from here will be limited. We can then move it to `DEBUG` if/when the toggle will be removed and this will be the default.

oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexStatistics.java
```diff
@@ -216,7 +231,15 @@ static class StatsCacheLoader implements CacheLoader<StatsRequestDescriptor, Sta
 
         @Override
         public @NotNull StatsResponse load(@NotNull StatsRequestDescriptor countRequestDescriptor) throws IOException {
-            return stats(countRequestDescriptor);
+            try {
+                return stats(countRequestDescriptor);
+            } catch (ElasticsearchException ee) {
+                if (ee.status() == 404 && FT_OAK_12248_ENABLE.get()) {
+                    LOG.debug("ES alias not found for index {} — returning empty stats (OAK-12248)", countRequestDescriptor.index);
```
fabriziofortino: same comment as above

fabriziofortino (approved):

PR #2955, opened 2026-06-12
OAK-12249: lazy ES index provisioning — skip creation for empty reindex

bhabegger:
See  https://github.com/apache/jackrabbit-oak/pull/3085 instead

## Summary

- Defers `provisionIndex()` from the `ElasticIndexWriter` constructor to the first `updateDocument()`, `deleteDocumentTree()`, or `deleteDocument()` call when both `FT_OAK-12249` and `FT_OAK-12248` are enabled
- A reindex that produces zero documents never creates an Elasticsearch index or alias, eliminating the empty-index problem described in OAK-12249
- `ensureProvisioned()` handles the incremental-write-after-empty-reindex case: if an alias does not exist when the first document arrives, it creates a new backing index with a fresh seed and points the alias at it

## Dependency on OAK-12248

OAK-12248 (#2950, graceful 404 handling) has merged to trunk. This branch is rebased on top of it — the diff now shows only the OAK-12249 commits.

The runtime dependency remains: `isLazyProvisioningActive()` returns `true` only when **both** `FT_OAK-12249` and `FT_OAK-12248` are enabled. Enabling `FT_OAK-12249` alone falls back to eager provisioning and logs a `WARN`.

## Tests

Three new unit tests in `ElasticIndexWriterTest`:

- `lazyProvisioning_requiresGraceful404Toggle` — asserts lazy provisioning is inactive when OAK-12248 toggle is off
- `emptyReindex_doesNotCreateEsIndex` — verifies no ES index is created during construction when no documents are written
- `nonEmptyReindex_provisionsOnFirstDocument` — verifies `provisionIndex()` is called on the first `updateDocument()` and not before

All 11 tests in `ElasticIndexWriterTest` pass; full `oak-search-elastic` suite (514 tests) passes.

## Jira

https://issues.apache.org/jira/browse/OAK-12249

### DIFF
--- a/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
+++ b/oak-search-elastic/src/main/java/org/apache/jackrabbit/oak/plugins/index/elastic/ElasticIndexDefinition.java
@@ -88,6 +88,13 @@ public class ElasticIndexDefinition extends IndexDefinition {
     public static final String PROP_INDEX_NAME_SEED = ":nameSeed";
 
     /**
+     * Hidden property written when a lazy reindex (OAK-12249) produced zero documents and left no
+     * ES index or alias. Signals the next incremental-write cycle to provision the index on demand.
+     * Cleared once provisioning completes.
+     */
+    public static final String PROP_REQUIRES_PROVISIONING = ":requiresProvisioning";
+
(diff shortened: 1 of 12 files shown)


bhabegger:
Replaced by https://github.com/apache/jackrabbit-oak/pull/3085

PR #2957, opened 2026-06-14
OAK-12259: oak-http: fix HTTP Basic credential parsing in OakServlet

ciechanowiec:
Fixes [OAK-12259](https://issues.apache.org/jira/browse/OAK-12259).

### DIFF
--- a/oak-http/src/main/java/org/apache/jackrabbit/oak/http/OakServlet.java
+++ b/oak-http/src/main/java/org/apache/jackrabbit/oak/http/OakServlet.java
@@ -68,16 +68,8 @@ public class OakServlet extends HttpServlet {
             HttpServletRequest request, HttpServletResponse response)
             throws ServletException, IOException {
         try {
-            Credentials credentials = null;
-
-            String authorization = request.getHeader("Authorization");
-            if (authorization != null && authorization.startsWith("Basic ")) {
-                String[] basic =
-                        Base64.decode(authorization.substring("Basic ".length())).split(":");
-                credentials = new SimpleCredentials(basic[0], basic[1].toCharArray());
-            } else {
-                throw new LoginException();
-            }
+            Credentials credentials =
+                    parseBasicCredentials(request.getHeader("Authorization"));
 
             ContentSession session = repository.login(credentials, null);
             try {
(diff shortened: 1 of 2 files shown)


oak-http/src/test/java/org/apache/jackrabbit/oak/http/OakServletTest.java
```diff
@@ -0,0 +1,142 @@
...
+        HttpServletRequest request = mock(HttpServletRequest.class);
+        when(request.getHeader("Authorization")).thenReturn(authorization);
+        HttpServletResponse response = mock(HttpServletResponse.class);
+
+        new OakServlet(repository).service(request, response);
+
+        verify(response).setHeader("WWW-Authenticate", "Basic realm=\"Oak\"");
+        verify(response).sendError(HttpServletResponse.SC_UNAUTHORIZED);
+        verify(repository, never()).login(any(), any());
+    }
```
rishabhdaim, line 88: Nits : Move the helper method to the end of Test file.

oak-http/src/main/java/org/apache/jackrabbit/oak/http/OakServlet.java
```diff
@@ -115,6 +107,30 @@ protected void service(
...
+            throws LoginException {
+        if (authorization == null || !authorization.startsWith("Basic ")) {
+            throw new LoginException("Missing or unsupported Authorization header; expected Basic scheme");
+        }
+        String decoded = Base64.decode(authorization.substring("Basic ".length()));
+        int colon = decoded.indexOf(':');
+        if (colon < 0) {
+            throw new LoginException("Malformed Basic credentials: missing ':' separator");
+        }
+        String userId = decoded.substring(0, colon);
```
rishabhdaim, line 129: Shouldn't we explicitly check for whether the username has a `:` or not, and throw an exception if it has ?

cc @anchela @Amoratinos @reschke

rishabhdaim:
If we are indeed following https://www.rfc-editor.org/info/rfc7617/, we should document that.

@Amoratinos @anchela wdyt ?

reschke:
While at it, we should also handle the whitespace after "Basic" correctly (https://www.greenbytes.de/tech/specs/rfc7235.html#challenge.and.response), case sensitivity, decoding non-ASCII, invalid username/passwords.

For obvious reasons, I'll take this.

reschke (changes requested):
While at it, we should also handle the whitespace after "Basic" correctly (https://www.greenbytes.de/tech/specs/rfc7235.html#challenge.and.response), case sensitivity, decoding non-ASCII, invalid username/passwords.

For obvious reasons, I'll take this.

reschke:
@ciechanowiec - idea:

1. Refactor basic auth support into until class so it's easier to test
2. Write tests (as mentioned above), of which many will be failing right now
3. Fix bugs.

Do you want to give that a try, or should I take over (of course acking you)?

ciechanowiec:
@reschke , feel free to take over. I doubt I'll have the time/capacity to work on this issue in the foreseeable future.

PR #2958, opened 2026-06-15
OAK-12260 improve cleanup of expired tokens

joerghoh:
* Only delete a subset of all expired tokens to get have a smaller commit.
* WARN if a large amount of login-tokens are present (expired or not)


joerghoh:
Working on the failing Commit Message Validation in parallel in #2960

PR #2960, opened 2026-06-16
OAK-12261: update commit-check plugin to 2.8

joerghoh:


.github/workflows/commit-check.yml
```diff
@@ -47,7 +47,5 @@ jobs:
           branch: true
```
kwin: We can set this to `false` (https://github.com/commit-check/commit-check-action#branch).

.github/cchk.toml
```diff
@@ -15,18 +15,11 @@
 
 # format as outlined in https://github.com/commit-check/commit-check/blob/main/.commit-check.yml
```
kwin: This link is no longer valid, let us use https://commit-check.github.io/commit-check/configuration.html#options-table-description.

kwin (changes requested):
Thanks, looks good in general. Some nitpicks in the comment. Please also squash and reference https://issues.apache.org/jira/browse/OAK-12257 in the commit message instead.

joerghoh:
I will reference both OAK-12261 and OAK-12257, as I also have created a new Jira for it

kwin (approved):

PR #2961, opened 2026-06-17
OAK-12262 Simplified index mgmt: support indexes without version number

thomasmueller:


oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexTest.java
```diff
@@ -259,6 +260,112 @@ public void testDiffIndexUpdate() throws Exception {
...
+
+        // seed the store with an unversioned OOTB index /oak:index/test (type lucene)
+        NodeBuilder builder = store.getRoot().builder();
+        builder.child(INDEX_DEFINITIONS_NAME).child("test")
+                .setProperty("jcr:primaryType", IndexConstants.INDEX_DEFINITIONS_NODE_TYPE, Type.NAME)
+                .setProperty(TYPE_PROPERTY_NAME, "lucene");
+        store.merge(builder, hook, CommitInfo.EMPTY);
+
+        // diff adds indexRules to the product index
+        storeDiff(store, "2026-01-01T00:00:00.000Z",
```
bhabegger, line 280: Multi-line strings for readability ?

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexTest.java
```diff
@@ -259,6 +260,112 @@ public void testDiffIndexUpdate() throws Exception {
         assertSameJson("{}", repositoryDefinitions.toString());
     }
 
+    @Test
+    public void testDiffIndexUpdateUnversionedOotbExplicit() throws Exception {
+        NodeStore store = new MemoryNodeStore(INITIAL_CONTENT);
+
+        List<IndexEditorProvider> indexEditors = List.of(
+                new ReferenceEditorProvider(), new PropertyIndexEditorProvider(), new NodeCounterEditorProvider());
```
bhabegger, line 268: Do we need all of these ?

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexTest.java
```diff
@@ -259,6 +260,112 @@ public void testDiffIndexUpdate() throws Exception {
...
+                + "        \"prop1\": { \"name\": \"prop1\", \"propertyIndex\": true }"
+                + "      }"
+                + "    }"
+                + "  }"
+                + "} }");
+
+        // expect test-custom-1 with:
+        // - type "lucene" from the product index
+        // - indexRules/nt:base/properties/prop1 from the diff
+        assertSameJson("{\n"
```
bhabegger, line 294: As `test-custom-1` is what we are testing maybe we can content ourselves of asserting that ? 

Couldn't we have an existence test by checking a path from `RootIndexesListService.getRootIndexDefinitions(store, "lucene")` ?

bhabegger: Similar to something below. Why is it different here ?

bhabegger (approved):

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexTest.java
thomasmueller: I think it is not strictly necessary, but it is closer to reality.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexTest.java
thomasmueller: Just checking for existence is not enough in my view; we need to be sure the merging is correct as well.

PR #2964, opened 2026-06-18
OAK-12241 : support for S3 tests for oak-blob-cloud to run on S3Mock

rishabhdaim:


rishabhdaim:
@reschke @amit-jain All tests are passing for actual AWS as well with access/secret key.

rishabhdaim:
This change is mainly intended to make the `oak-blob-cloud` S3 tests runnable in CI without requiring real AWS credentials, by using S3Mock/Testcontainers.

That said, the PR is not test-only. We also had to update a small amount of production S3 code so the same Oak S3 client/presigner paths work correctly with S3-compatible emulator endpoints used by the tests.

## What changed

- Added S3Mock/Testcontainers-based support for `oak-blob-cloud` tests
- Updated test utilities and selected tests to use emulator-backed config when real S3 credentials are not available
- Skipped or adjusted a few tests for known S3Mock incompatibilities
- Updated production presigner/client configuration so presigned URLs work correctly with custom S3-compatible endpoints such as S3Mock
- Added `pathStyleAccess` support so local emulator endpoints can be addressed correctly
- Updated related OSGi config documentation

## Why production code changed

To keep the tests meaningful, we wanted the emulator-backed tests to go through the same production S3 code paths instead of introducing separate test-only logic.

Because of that, a few production-path updates were needed so Oak can work correctly with both:

- real AWS S3
- S3-compatible local/mock endpoints used in CI

## Validation

Verified targeted utility/configuration coverage locally with:

`mvn test -pl oak-blob-cloud`

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
```diff
@@ -481,6 +497,10 @@ public void close() {
             if (s3PresignService != null) {
                 s3PresignService.close();
             }
+            closePresignClient();
+            if (sdkTimeoutExecutor != null) {
+                sdkTimeoutExecutor.shutdownNow();
```
thomasmueller: Why not just shutdown()? shutdownNow() calls Thread.interrupt, which is dangerous because it can close files, including Lucene files and persistent cache files, when called at the wrong time.

thomasmueller (changes requested):

oak-blob-cloud/src/main/java/org/apache/jackrabbit/oak/blob/cloud/s3/S3Backend.java
rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/2964/commits/7839ec53c8f7021d496ca1ed8cf07f4eee9923db

thomasmueller (approved):

PR #2966, opened 2026-06-23
OAK-12273 : added https link for fetching java-docs

rishabhdaim:


oak-parent/pom.xml
```diff
@@ -129,9 +129,9 @@
             <doclint>none</doclint>
             <detectJavaApiLink>true</detectJavaApiLink>
             <links>
-              <link>https://s.apache.org/jcr-2.0-javadoc/</link>
+              <link>https://developer.adobe.com/experience-manager/reference-materials/spec/javax.jcr/javadocs/jcr-2.0/</link>
```
kwin: Please leave the short link in place. The Adobe one changed too often in the past. Let's use this indirection via ASF

kwin (changes requested):
Looks good for SLF4J but let's not modify the ASF shortlink

reschke (changes requested):
As @kwin said.

oak-parent/pom.xml
rishabhdaim: reverted in https://github.com/apache/jackrabbit-oak/pull/2966/commits/b17272eeea2d896a6d6a1da8e01580ae3d10279f

kwin (approved):

PR #2970, opened 2026-06-24
OAK-12275: NameValidator -> improve logging for unmapped prefixes

reschke:


rishabhdaim (commented):
I would add unit cases for the change too.

reschke:
for re-arranging the contents on a log statement????

mbaedke (approved):

rishabhdaim:
Sorry, for not being explicit.

My rational was. that we should add a log so that someone don't accidentally change it.
On a second thought, I comment can do that as well.

PR #2972, opened 2026-06-24
OAK-12276: Upgrade jackson-databind dependency to 2.22.1

reschke:


oak-parent/pom.xml
```diff
@@ -64,7 +64,8 @@
     <h2.version>2.1.214</h2.version>
     <tika.version>1.28.5</tika.version>
     <derby.version>10.16.1.1</derby.version>
-    <jackson.version>2.19.4</jackson.version>
+    <jackson.version>2.22.0</jackson.version>
+    <jackson.annotations.version>2.22</jackson.annotations.version> <!-- not aligned with jackson anymore -->
```
rishabhdaim: just as a cautionary step, can we jump one minor version at one time.

Apart from that PR looks okk.

reschke: That would be three steps for changes since last August.

reschke: @rishabhdaim - what's your preference: just separate PRs, or seperate tickets?

rishabhdaim: I would prefer 3 separate tickets, 1 for each and we should do 1 per release (if possible).

reschke: With one per release we may not be able to keep up :-)

rishabhdaim: okk, for this case, lets jump twice in next release.

PR #2973, opened 2026-06-25
OAK-12269 : Add index tag support to user query (QueryBuilder#setIndexTag)

anchela:
Allow user/group queries to force a specific index by exposing a new QueryBuilder#setIndexTag(String) method. When set, the generated XPath statement is augmented with an "option(index tag <tag>)" clause (see the Oak query-engine index-tag query option).

- oak-jackrabbit-api: add default no-op setIndexTag(String) to the QueryBuilder interface so existing implementors remain source/binary compatible; bump the exported package version (minor) accordingly.
- oak-core: XPathQueryBuilder stores the tag; UserQueryManager appends the option clause to the statement and validates the tag against [a-zA-Z0-9_]+ to prevent query injection.
- Tests for the new builder accessor and for the generated statement (valid tag executes, invalid tag is rejected).

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/query/UserQueryManager.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/query/UserQueryManager.java
@@ -22,6 +22,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Set;
 import java.util.function.Predicate;
+import java.util.regex.Pattern;
 
 import javax.jcr.RepositoryException;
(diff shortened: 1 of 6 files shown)


anchela:
@Amoratinos , there is a major version increase associated with the PR from 2.5 to 3.0. imho we should avoid that and given that you have a default implementation i am not sure where there is the major increase. wdyt?

anchela:
maybe also adding a test to the API bundle for the QueryBuilder method that comes with a default impl to verify that the default implementation does not hit the querymanager. the rest looks good to me.

PR #2975, opened 2026-06-25
OAK-12269: Allow forcing an index for user/group queries via QueryBuilder index tag

Amoratinos:
## OAK-12269

Allow user/group queries to force a specific index by exposing a new `QueryBuilder#setIndexTag(String)` method. When set, the generated XPath statement is augmented with an `option(index tag <tag>)` clause (see the Oak [query option index tag](https://jackrabbit.apache.org/oak/docs/query/query-engine.html#query-option-index-tag)).

### Changes
- **`oak-jackrabbit-api`**: add a `default` no-op `setIndexTag(String)` to the `QueryBuilder` interface, so existing implementations remain source/binary compatible. The exported package version is bumped (minor) accordingly.
- **`oak-core`**: `XPathQueryBuilder` stores the tag; `UserQueryManager#buildXPathStatement` appends the `option(index tag <tag>)` clause and validates the tag against `[a-zA-Z0-9_]+` to prevent query injection.

### Tests
- `XPathQueryBuilderTest`: getter/setter for the index tag.
- `UserQueryManagerTest`: a query with a valid tag executes successfully; an invalid tag is rejected.
- `QueryBuilderTest` (oak-jackrabbit-api): a `QueryBuilder` implementation that does not override `setIndexTag` is never called back into when the default method is invoked, verifying the default is a genuine no-op.

### Compatibility
The new method is a `default` method; consumers and existing providers are unaffected. When a tag is set but no index carries it, the query falls back to normal index selection / traversal.

### Related
- JCR-5248 (consumer side): expose an `indexHint` field in `AuthorizableQueryManager`'s JSON query format. That change depends on this one being released and `oak-jackrabbit-api.version.used` being bumped in Jackrabbit.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/query/UserQueryManager.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/query/UserQueryManager.java
@@ -22,6 +22,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Set;
 import java.util.function.Predicate;
+import java.util.regex.Pattern;
 
 import javax.jcr.RepositoryException;
 import javax.jcr.Value;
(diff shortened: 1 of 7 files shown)


anchela (approved):
finding mentioned earlier are addressed. thanks @Amoratinos

PR #2977, opened 2026-06-25
OAK-12280: Provide a way to reconcile the offline GC result back 

royteeuwen:
There is a blobid tracker used by the online GC, which should be reconciled by the offline GC run command. Make sure that if the file is not reconciled, the online GC still cleans up the blobs that no longer exist from the blobid tracker


rishabhdaim:
@royteeuwen The blobId Inline tracker was added as an experimental feature.

The feature only solved the purpose if very few repos shared the data store and consistently enabled DSGC. Long-pending DSGC tasks or irregular schedules worsen the problem.

So disabling this feature is the correct way to proceed here.

In future releases, we are even planning to remove this feature.


cc @amit-jain

royteeuwen:
OK clear, lets close this PR and the jira ticket

PR #2978, opened 2026-06-25
OAK-12282 : defining a fixed bound for the AbstractDiskCache

pat-lego:
The AbstractPersistentCache is unbound leading to OOM exceptions.

### DIFF
--- a/oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
+++ b/oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
@@ -111,8 +111,8 @@ public class PersistentDiskCache extends AbstractPersistentCache {
                 () -> maxCacheSizeBytes,
                 () -> Long.valueOf(directory.listFiles().length),
                 () -> FileUtils.sizeOfDirectory(directory),
-                () -> evictionCount.get());
-        segmentCacheStats.setWriteDiscardCountSupplier(() -> discardCount.get());
+                () -> evictionCount.get(),
+                () -> discardCount.get());
     }
 
     @Override
--- a/oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
+++ b/oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
@@ -112,6 +112,7 @@ public class PersistentDiskCache extends AbstractPersistentCache {
                 () -> Long.valueOf(directory.listFiles().length),
                 () -> FileUtils.sizeOfDirectory(directory),
                 () -> evictionCount.get());
+        segmentCacheStats.setWriteDiscardCountSupplier(() -> discardCount.get());
     }
 
     @Override
--- a/oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/RemotePersistentCacheService.java
+++ b/oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/RemotePersistentCacheService.java
@@ -24,6 +24,7 @@ import org.apache.jackrabbit.oak.cache.CacheStats;
 import org.apache.jackrabbit.oak.commons.pio.Closer;
 import org.apache.jackrabbit.oak.osgi.OsgiWhiteboard;
 import org.apache.jackrabbit.oak.segment.spi.monitor.RoleStatisticsProvider;
+import org.apache.jackrabbit.oak.segment.spi.persistence.persistentcache.AbstractPersistentCache;
 import org.apache.jackrabbit.oak.segment.spi.persistence.persistentcache.PersistentCache;
 import org.apache.jackrabbit.oak.spi.toggle.FeatureToggle;
 import org.apache.jackrabbit.oak.spi.whiteboard.Registration;
@@ -101,6 +102,12 @@ public class RemotePersistentCacheService {
                     new FeatureToggle(PersistentDiskCache.FT_OAK_12212, PersistentDiskCache.FT_OAK_12212_SKIP_MISSING_FILE_CHECK),
                     Collections.emptyMap()));
 
+            // OAK-12282: expose the bounded write-queue kill switch. Changing this
+            // flag requires a process restart since the executor is created at startup.
+            registerCloseable(osgiWhiteboard.register(FeatureToggle.class,
+                    new FeatureToggle(AbstractPersistentCache.FT_OAK_12282, AbstractPersistentCache.FT_OAK_12282_BOUNDED_WRITE_QUEUE_ENABLED),
+                    Collections.emptyMap()));
+
             CacheStatsMBean diskCacheStatsMBean = persistentDiskCache.getCacheStats();
             registerCloseable(registerMBean(CacheStatsMBean.class, diskCacheStatsMBean, CacheStats.TYPE, diskCacheStatsMBean.getName()));
 
@@ -119,6 +126,11 @@ public class RemotePersistentCacheService {
             PersistentRedisCache redisCache = new PersistentRedisCache(configuration.redisCacheHost(), configuration.redisCachePort(), configuration.redisCacheExpireSeconds(), configuration.redisSocketTimeout(), configuration.redisConnectionTimeout(),
                     configuration.redisMinConnections(), configuration.redisMaxConnections(), configuration.redisMaxTotalConnections(), configuration.redisDBIndex(), redisCacheIOMonitor);
             closer.register(redisCache);
+
+            // OAK-12282: expose the bounded write-queue kill switch. Requires a restart to take effect.
+            registerCloseable(osgiWhiteboard.register(FeatureToggle.class,
+                    new FeatureToggle(AbstractPersistentCache.FT_OAK_12282, AbstractPersistentCache.FT_OAK_12282_BOUNDED_WRITE_QUEUE_ENABLED),
+                    Collections.emptyMap()));
 
             CacheStatsMBean redisCacheStatsMBean = redisCache.getCacheStats();
             registerCloseable(registerMBean(CacheStatsMBean.class, redisCacheStatsMBean, CacheStats.TYPE, redisCacheStatsMBean.getName()));
--- a/oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/RemotePersistentCacheService.java
+++ b/oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/RemotePersistentCacheService.java
@@ -24,7 +24,6 @@ import org.apache.jackrabbit.oak.cache.CacheStats;
 import org.apache.jackrabbit.oak.commons.pio.Closer;
 import org.apache.jackrabbit.oak.osgi.OsgiWhiteboard;
 import org.apache.jackrabbit.oak.segment.spi.monitor.RoleStatisticsProvider;
-import org.apache.jackrabbit.oak.segment.spi.persistence.persistentcache.AbstractPersistentCache;
 import org.apache.jackrabbit.oak.segment.spi.persistence.persistentcache.PersistentCache;
 import org.apache.jackrabbit.oak.spi.toggle.FeatureToggle;
 import org.apache.jackrabbit.oak.spi.whiteboard.Registration;
@@ -102,12 +101,6 @@ public class RemotePersistentCacheService {
                     new FeatureToggle(PersistentDiskCache.FT_OAK_12212, PersistentDiskCache.FT_OAK_12212_SKIP_MISSING_FILE_CHECK),
                     Collections.emptyMap()));
 
-            // OAK-12282: expose the bounded write-queue kill switch. Changing this
-            // flag requires a process restart since the executor is created at startup.
-            registerCloseable(osgiWhiteboard.register(FeatureToggle.class,
-                    new FeatureToggle(AbstractPersistentCache.FT_OAK_12282, AbstractPersistentCache.FT_OAK_12282_BOUNDED_WRITE_QUEUE_ENABLED),
-                    Collections.emptyMap()));
-
             CacheStatsMBean diskCacheStatsMBean = persistentDiskCache.getCacheStats();
             registerCloseable(registerMBean(CacheStatsMBean.class, diskCacheStatsMBean, CacheStats.TYPE, diskCacheStatsMBean.getName()));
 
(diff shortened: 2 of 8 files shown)


oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/RemotePersistentCacheService.java
```diff
@@ -120,6 +127,11 @@ private PersistentCache createPersistentCache(Configuration configuration, Close
                     configuration.redisMinConnections(), configuration.redisMaxConnections(), configuration.redisMaxTotalConnections(), configuration.redisDBIndex(), redisCacheIOMonitor);
             closer.register(redisCache);
 
+            // OAK-12282: expose the bounded write-queue kill switch. Requires a restart to take effect.
```
joerghoh:
```suggestion
            // OAK-12282: expose the bounded write-queue kill switch. 
```

That 2nd setence is a bit misleading

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/AbstractPersistentCache.java
```diff
@@ -31,27 +31,68 @@
...
+     * {@link #WRITE_QUEUE_SIZE} and silently discards write tasks when full,
+     * preventing OOM under high write load. This is safe because the disk
+     * cache is an optimisation only — a dropped write means the segment is
+     * fetched from remote storage on the next read.
+     * <p>
+     * Set to {@code false} via the {@link org.apache.jackrabbit.oak.spi.toggle.FeatureToggle}
+     * registered with the Whiteboard to revert to the pre-fix unbounded queue.
+     * <strong>Note:</strong> changing this flag requires a process restart, as
+     * the executor is created at startup.
+     */
```
joerghoh: I would remove this for consistency reasons (or we would need to annotate every configuration parameter for the cache with the same).

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
```diff
@@ -112,6 +112,7 @@ public PersistentDiskCache(File directory, int cacheMaxSizeMB, DiskCacheIOMonito
                 () -> Long.valueOf(directory.listFiles().length),
                 () -> FileUtils.sizeOfDirectory(directory),
                 () -> evictionCount.get());
+        segmentCacheStats.setWriteDiscardCountSupplier(() -> discardCount.get());
```
joerghoh: Very inconsistent. I would extend the c'tor of ``SegmentCacheStats`` by one more parameter.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/AbstractPersistentCache.java
```diff
@@ -31,27 +31,68 @@
...
+    public static final AtomicBoolean FT_OAK_12282_BOUNDED_WRITE_QUEUE_ENABLED = new AtomicBoolean(true);
+
     public AbstractPersistentCache() {
-        executor = Executors.newFixedThreadPool(THREADS);
+        // Formerly Executors.newFixedThreadPool() was used here, which creates an unbounded
+        // LinkedBlockingQueue — allowing unlimited segment buffers to pile up in memory under
+        // high write load. The bounded queue (gated by FT_OAK_12282) prevents OOM by dropping
+        // write tasks when full; this is safe because the disk cache is an optimisation only.
```
joerghoh: This comment is typical AI, but it won't help if in a year someone is looking at this code (If I want to understand what was here before, I can checkout the right git commit ...)



```suggestion
        // Limit the amount of Runnable in the queue to avoid memory issues. Discarding elements from
        // the queue is not a problem, as caching is just an optimization; if it's dropped, it will be downloaded
        // again and tried again.
```

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/AbstractPersistentCache.java
```diff
@@ -31,27 +31,68 @@
...
+        BlockingQueue<Runnable> writeQueue = FT_OAK_12282_BOUNDED_WRITE_QUEUE_ENABLED.get()
+                ? new LinkedBlockingQueue<>(WRITE_QUEUE_SIZE)
+                : new LinkedBlockingQueue<>();
+        executor = new ThreadPoolExecutor(
```
joerghoh, line 70: right now these 10 threads are named quite randomly, and therefor they are hard to detect; can you use a ThreadFactory to name them accordingly?

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/AbstractPersistentCache.java
```diff
@@ -31,27 +31,68 @@
...
+                THREADS, THREADS,
+                0L, TimeUnit.MILLISECONDS,
+                writeQueue,
+                (r, e) -> {
+                    discardCount.incrementAndGet();
+                    logger.debug("Segment write task discarded: write queue full (capacity={})", WRITE_QUEUE_SIZE);
```
joerghoh: I would also log the number of discarded items from the queue

```suggestion
                    long discarded = discardCount.incrementAndGet();
                    logger.debug("Segment write task discarded: write queue full (capacity={}, already discarded={})", WRITE_QUEUE_SIZE, discarded);
```

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/SegmentCacheStats.java
```diff
@@ -78,6 +81,14 @@ protected CacheStats getCurrentStats() {
         );
     }
 
+    public void setWriteDiscardCountSupplier(@NotNull Supplier<Long> supplier) {
```
joerghoh: as mentioned above, I would add this to the constructor for consistency reasons

joerghoh (changes requested):
a lot of minor things:
* Comments: AI writes too many comments which are 100% valid and helpful in the context of this PR, but not really helpful in a year in the context of a different problem.
* some consistency

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/AbstractPersistentCache.java
```diff
@@ -31,27 +31,68 @@
...
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 
 public abstract class AbstractPersistentCache implements PersistentCache, Closeable {
     private static final Logger logger = LoggerFactory.getLogger(AbstractPersistentCache.class);
 
     public static final int THREADS = Integer.getInteger("oak.segment.cache.threads", 10);
+    public static final int WRITE_QUEUE_SIZE = Integer.getInteger("oak.segment.cache.writeQueueSize", THREADS * 100);
```
nfsantos: Writing to disk will be limited to the performance of the IO system and does not scale much with the number of threads that are writing. In fact, a single thread writing efficiently to disk will likely saturate the disk. The max queue size should be a function of the write IO of the disk and of the memory that we want to devote to buffering pending writes, it should not depend on the number of write threads. And we also don't want to cause an OOME only by increasing the number of threads. The number of writers and the size of the queue should be separate settings.

Since segments are around 256KB, a queue size of 100 would take up 25MB when full. I'd say this is enough, as writes to a local SSD should be faster than reading from the network, so a few write threads should be able to keep up with whatever we can get from Azure.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/AbstractPersistentCache.java
nfsantos: This log should be throttled, as otherwise we risk flooding the logs if for some reason the write threads cannot keep up with the reads from Azure. See the LogSilencer class.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/AbstractPersistentCache.java
```diff
@@ -31,27 +31,68 @@
...
     protected final Set<String> writesPending;
+    protected AtomicLong discardCount = new AtomicLong();
 
     protected SegmentCacheStats segmentCacheStats;
 
+    /**
+     * Name of the feature toggle for the OAK-12282 bounded write-queue fix.
+     * See {@link #FT_OAK_12282_BOUNDED_WRITE_QUEUE_ENABLED}.
+     */
+    public static final String FT_OAK_12282 = "FT_OAK-12282";
```
nfsantos: Personally, I think this is unnecessary. The change is quite safe. And if for some reason having a bound on the queue causes trouble, we can increase it's size to be effectively unlimited.

nfsantos (changes requested):

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/AbstractPersistentCache.java
joerghoh: > Writing to disk will be limited to the performance of the IO system and does not scale much with the number of threads that are writing

This is definitely true for the classic hard drives (aka "spinning rust"), but not anymore for NVME-SSDs. You get the full IOPS only when you execute operations at a massive scale (IIRC 256 concurrent threads). And this is applicable for both read and write. Assuming a modern operating system also the filesystem should be able to handle that.
For that I strongly recommend to keep these 10 as default.

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
```diff
@@ -111,7 +111,8 @@ public PersistentDiskCache(File directory, int cacheMaxSizeMB, DiskCacheIOMonito
                 () -> maxCacheSizeBytes,
                 () -> Long.valueOf(directory.listFiles().length),
                 () -> FileUtils.sizeOfDirectory(directory),
```
nfsantos, line 113: Just a side comment, unrelated to the changes in this PR: `sizeOfDirectory` does an `fstat` for every file in the cache, which is likely too slow to be done every time the metrics are polled. This should also be kept as in-memory counter. I have addressed this issue in our internal fork, because we are exposing these metrics in Grafana. If there is an interest, I can contribute the patch upstream.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/AbstractPersistentCache.java
nfsantos: Yes, my claim that read/write does not scale much with number of threads is wrong for modern SSDs. They do scale. 
In this particular case, the maximum download rate we can achieve from Azure will likely be far below what we can write to an SSD, so a small number of writers may be enough to keep up. But I was not advocating to decrease the number of writer threads. 10 is probably more than needed but the cost of having these threads around is minimal, so it's not an issue.

nfsantos (approved):
LGTM, but please get a few more approvals from people who are more involved with the usage of this cache.

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
joerghoh: I would love to have that patch upstream (in a dedicated PR)

nfsantos: Here: https://github.com/apache/jackrabbit-oak/pull/3004

pat-lego:
@joerghoh any additional requests for this PR or can we merge it?

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
joerghoh: thanks!

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/RemotePersistentCacheService.java
pat-lego: I cleared these comments as part of the fix since we can simply increase the list size no need for the toggle

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/AbstractPersistentCache.java
pat-lego: I cleared these comments as part of the fix since we can simply increase the list size no need for the toggle

oak-segment-remote/src/main/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCache.java
pat-lego: Exposed the SegmentCacheStats by one parameter to make it more consistent

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/AbstractPersistentCache.java
pat-lego: Removed this since we dont need the toggle.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/AbstractPersistentCache.java
pat-lego: Gave the thread factory a proper name for easier debugging.

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/spi/persistence/persistentcache/SegmentCacheStats.java
pat-lego: This has been applied.

joerghoh (approved):
lgtm

PR #2981, opened 2026-06-26
oak-http: extract Authorization Field parsing and add test coverage

reschke:
Reviewers: not that this code is a demo project / POC only.


rishabhdaim:
@reschke so the scope of this PR is just to refactor the code and handle a case where some might have added more values in `Authorization` header, correct ?

We are not fixing the original issue reported in OAK-12259.

oak-http/src/test/java/org/apache/jackrabbit/oak/http/AuthorizationFieldTest.java
```diff
@@ -0,0 +1,72 @@
...
+        assertEquals("foo", credentials.getUserID());
+        assertEquals("bar", new String(credentials.getPassword()));
+    }
+
+    @Test(expected = ArrayIndexOutOfBoundsException.class) // BUG: OAK-12259
+    public void testInvalidBase64() throws LoginException {
+        String b64 = Base64.getEncoder().encodeToString("foo:bar".getBytes(StandardCharsets.UTF_8));
+        SimpleCredentials credentials = AuthorizationField.valueOf(Collections.enumeration(List.of("Basic dksjdkj" + b64)));
+        assertEquals("foo", credentials.getUserID());
+        assertEquals("bar", new String(credentials.getPassword()));
```
rishabhdaim, line 55: exception would be thrown from above line.

These 2 are not reachable.

how about we add `Assert.fail("shouldn't reach here")` here.

rishabhdaim (commented):
Just a on comment on a test case.

Rest is fine with me.

reschke:
yes, this just adds test coverage for the existing behavior; bugs will be fixed separately.

oak-http/src/test/java/org/apache/jackrabbit/oak/http/AuthorizationFieldTest.java
reschke: they may become reachable later on.

rishabhdaim: fine +1

rishabhdaim (approved):

PR #2983, opened 2026-06-26
OAK-12259: oak-http: OakServlet mis-parses HTTP Basic credentials - m…

reschke:
…any many many tests.

EDIT: hint: this is a demo project. The PR fixes a ton of issues. The fixes have test coverage. There's a point where nitpicking on the test cases becomes a waste of time. Please only report major problems.

### DIFF
--- a/oak-http/src/main/java/org/apache/jackrabbit/oak/http/AuthorizationField.java
+++ b/oak-http/src/main/java/org/apache/jackrabbit/oak/http/AuthorizationField.java
@@ -21,6 +21,7 @@ import org.apache.jackrabbit.util.Base64;
 import javax.jcr.SimpleCredentials;
 import javax.security.auth.login.LoginException;
 import java.util.Enumeration;
+import java.util.Locale;
 import java.util.NoSuchElementException;
 
 public class AuthorizationField {
@@ -44,10 +45,18 @@ public class AuthorizationField {
         return parseCredentials(field);
     }
 
-    private static SimpleCredentials parseCredentials(String fieldValue) throws LoginException {
-        if (fieldValue.startsWith("Basic ")) {
+    private static SimpleCredentials parseCredentials(String rawFieldValue) throws LoginException {
+        boolean hasControls = rawFieldValue.chars().anyMatch(c -> c < ' ');
+        if (hasControls) {
+            throw new LoginException("Control characters are not allowed");
+        }
+
+        String fieldValue = rawFieldValue.trim();
+
+        if (fieldValue.toLowerCase(Locale.ENGLISH).startsWith("basic ")) {
+            String token68 = fieldValue.substring("basic ".length());
             String[] basic =
-                    Base64.decode(fieldValue.substring("Basic ".length())).split(":");
+                    Base64.decode(token68).split(":");
             return new SimpleCredentials(basic[0], basic[1].toCharArray());
         } else {
             throw new LoginException("Only Basic Authentication supported");
--- a/oak-http/src/main/java/org/apache/jackrabbit/oak/http/AuthorizationField.java
+++ b/oak-http/src/main/java/org/apache/jackrabbit/oak/http/AuthorizationField.java
@@ -16,6 +16,7 @@
  */
 package org.apache.jackrabbit.oak.http;
 
+import com.fasterxml.jackson.databind.ser.Serializers;
 import org.apache.jackrabbit.util.Base64;
 
 import javax.jcr.SimpleCredentials;
@@ -55,9 +56,16 @@ public class AuthorizationField {
 
         if (fieldValue.toLowerCase(Locale.ENGLISH).startsWith("basic ")) {
             String token68 = fieldValue.substring("basic ".length());
-            String[] basic =
-                    Base64.decode(token68).split(":");
-            return new SimpleCredentials(basic[0], basic[1].toCharArray());
+            String decoded  = Base64.decode(token68);
+            int colon = decoded.indexOf(':');
+            if (colon < 0) {
+                throw new LoginException(
+                        "Malformed Basic credentials: missing ':' separator");
+            }
+            String userId = decoded.substring(0, colon);
+            String password = decoded.substring(colon + 1);
+
+            return new SimpleCredentials(userId, password.toCharArray());
         } else {
             throw new LoginException("Only Basic Authentication supported");
         }
--- a/oak-http/src/main/java/org/apache/jackrabbit/oak/http/AuthorizationField.java
+++ b/oak-http/src/main/java/org/apache/jackrabbit/oak/http/AuthorizationField.java
@@ -16,11 +16,10 @@
  */
 package org.apache.jackrabbit.oak.http;
 
-import com.fasterxml.jackson.databind.ser.Serializers;
-import org.apache.jackrabbit.util.Base64;
-
 import javax.jcr.SimpleCredentials;
 import javax.security.auth.login.LoginException;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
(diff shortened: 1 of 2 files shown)


oak-http/src/test/java/org/apache/jackrabbit/oak/http/AuthorizationFieldTest.java
```diff
@@ -64,9 +64,72 @@ public void testNoScheme() throws LoginException {
...
+        SimpleCredentials credentials = AuthorizationField.valueOf(Collections.enumeration(List.of("Basic \t " + b64)));
+        assertEquals("foo", credentials.getUserID());
+        assertEquals("bar", new String(credentials.getPassword()));
+    }
+
+    @Test(expected = LoginException.class)
+    public void testBrokenBase64() throws LoginException {
+        String b64 = Base64.getEncoder().encodeToString("foo:bar".getBytes(StandardCharsets.UTF_8));
+        b64 = b64.substring(0,5) + " " + b64.substring(5);
+        SimpleCredentials credentials = AuthorizationField.valueOf(Collections.enumeration(List.of("Basic \t " + b64)));
```
mbaedke: Sollte das nicht `"Basic " + b64`  sein?

mbaedke (changes requested):

oak-http/src/test/java/org/apache/jackrabbit/oak/http/AuthorizationFieldTest.java
reschke: No, it puts a single SP into the base64 string. But yes, that's far from obvious.

oak-http/src/test/java/org/apache/jackrabbit/oak/http/AuthorizationFieldTest.java
```diff
@@ -64,9 +64,73 @@ public void testNoScheme() throws LoginException {
...
+        assertEquals("foo", credentials.getUserID());
+        assertEquals("bar", new String(credentials.getPassword()));
+    }
+
+    @Test(expected = LoginException.class)
+    public void testBrokenBase64() throws LoginException {
+        String b64 = Base64.getEncoder().encodeToString("foo:bar".getBytes(StandardCharsets.UTF_8));
+        // insert a single SP into the base64 sequence
+        b64 = b64.substring(0,5) + " " + b64.substring(5);
+        SimpleCredentials credentials = AuthorizationField.valueOf(Collections.enumeration(List.of("Basic \t " + b64)));
```
rishabhdaim: `"Basic \t " + b64` — the `\t` (0x09 < 0x20) fires the control-char guard before the Base64 decoder is reached. This test exercises the same path as `testNonSpWhitespace`, not broken base64. The `catch (IllegalArgumentException)` branch in `AuthorizationField.java:69` is left untested.

Fix: change prefix to `"Basic "`.

oak-http/src/test/java/org/apache/jackrabbit/oak/http/AuthorizationFieldTest.java
```diff
@@ -64,9 +64,73 @@ public void testNoScheme() throws LoginException {
...
+        assertEquals("foo", credentials.getUserID());
```
rishabhdaim: Dead code — `@Test(expected=LoginException.class)` means JUnit exits on the throw; these assertions are never reached. Same issue in `testNonSpWhitespace` (lines 98–99), and pre-existing in `testInvalidBase64` (lines 54–55) and `testMoreBrokenBase64` (lines 117–118).

oak-http/src/test/java/org/apache/jackrabbit/oak/http/AuthorizationFieldTest.java
```diff
@@ -64,9 +64,73 @@ public void testNoScheme() throws LoginException {
...
+    @Test
+    public void testMoreNonAscii() throws LoginException {
+        String b64 = Base64.getEncoder().encodeToString("test:123\u00a3".getBytes(StandardCharsets.UTF_8));
+        SimpleCredentials credentials = AuthorizationField.valueOf(Collections.enumeration(List.of("Basic " + b64)));
+        assertEquals("test", credentials.getUserID());
+        assertEquals("123\u00a3", new String(credentials.getPassword()));
+    }
+
+    @Test
+    public void Basic64NoPadding() throws LoginException {
```
rishabhdaim: nit: should be `testBase64NoPadding` — all other methods use the `testXxx` naming convention.

oak-http/src/main/java/org/apache/jackrabbit/oak/http/AuthorizationField.java
```diff
@@ -44,11 +45,30 @@ public static SimpleCredentials valueOf(Enumeration<String> values) throws Login
...
+                if (colon < 0) {
+                    throw new LoginException(
+                            "Malformed Basic credentials: missing ':' separator");
+                }
+                String userId = decoded.substring(0, colon);
+                String password = decoded.substring(colon + 1);
+
+                return new SimpleCredentials(userId, password.toCharArray());
+            } catch (IllegalArgumentException ex) {
+                throw new LoginException(ex.getMessage());
```
rishabhdaim, line 70: `ex.getMessage()` may include an attacker-controlled byte offset in server logs (CWE-532). Use a static message:
```java
throw new LoginException("Invalid Base64 encoding in Basic credentials");
```

rishabhdaim (changes requested):
**AI Review:**

The production logic is correct — `indexOf(':')` properly handles colons in passwords, case-insensitive scheme match is right, and the AIOOBE→LoginException shift is the intended fix. Three test issues before merge.

nit: `AuthorizationField.java:33` — new public method, AGENTS.md requires `@NotNull`/`@Nullable` annotations:
```java
public static @NotNull SimpleCredentials valueOf(@NotNull Enumeration<String> values) throws LoginException
```

oak-http/src/main/java/org/apache/jackrabbit/oak/http/AuthorizationField.java
reschke: So the attacker sends a field value with broken base 64, and the message contains the offset of the part that is broken. Where's the issue here?

(halluscinating AI maybe?)

(and no, attackers having access to our log files would be a huge ops problem, and we would have to change many many pieces in Oak would that be a problem)

oak-http/src/test/java/org/apache/jackrabbit/oak/http/AuthorizationFieldTest.java
reschke: Indeed. Copy & Paste mistake.

oak-http/src/test/java/org/apache/jackrabbit/oak/http/AuthorizationFieldTest.java
reschke: Same.

oak-http/src/test/java/org/apache/jackrabbit/oak/http/AuthorizationFieldTest.java
rishabhdaim: un-reachable code.

oak-http/src/test/java/org/apache/jackrabbit/oak/http/AuthorizationFieldTest.java
```diff
@@ -64,9 +64,71 @@ public void testNoScheme() throws LoginException {
...
```
rishabhdaim:
```suggestion
    public void testBasic64NoPadding() throws LoginException {
```

PR #2984, opened 2026-06-29
OAK-12287: Update to Apache Parent POM to version 39

reschke:


oak-parent/pom.xml
```diff
@@ -140,7 +140,6 @@
         <plugin>
           <groupId>org.apache.felix</groupId>
           <artifactId>maven-bundle-plugin</artifactId>
-          <version>5.1.9</version>
```
kwin, line 143: Where is that version managed if not here?

kwin (changes requested):

oak-parent/pom.xml
reschke: In the parent pom (I thought).

(Good find)

rishabhdaim (approved):

PR #2989, opened 2026-06-30
OAK-12219: Add support for Azure SDK V12 in oak-blob-azure 

reschke:


oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AbstractAzureDataStoreService.java
```diff
@@ -32,6 +32,12 @@
 import org.osgi.framework.ServiceRegistration;
 import org.osgi.service.component.ComponentContext;
 
+/**
+ * Kept for binary compatibility with existing callers. Use {@link AzureDataStoreWrapper} instead.
+ *
+ * @deprecated
+ */
+@Deprecated
```
reschke: forRemoval/since?

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AbstractAzureDataStoreService.java
```diff
@@ -32,6 +32,12 @@
 import org.osgi.framework.ServiceRegistration;
 import org.osgi.service.component.ComponentContext;
 
+/**
+ * Kept for binary compatibility with existing callers. Use {@link AzureDataStoreWrapper} instead.
+ *
+ * @deprecated
```
reschke: add reason to javadoc tag

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreWrapper.java
```diff
@@ -0,0 +1,392 @@
...
+import org.osgi.service.component.annotations.ConfigurationPolicy;
+import org.osgi.service.component.annotations.Deactivate;
+import org.osgi.service.component.annotations.Reference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.InputStream;
+import java.net.URI;
+import java.util.*;
```
reschke: no wildcard imports, please.

.gitignore
```diff
@@ -14,3 +14,4 @@ derby.log
 oak-shaded-guava/dependency-reduced-pom.xml
 .DS_Store
 .claude/settings.local.json
+pom.xml.versionsBackup
```
reschke, line 15: where does this come from?

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreRegistrar.java
```diff
@@ -0,0 +1,392 @@
...
+        return context.getBundleContext().registerService(
+                AbstractSharedCachingDataStore.class, service, delegateProps);
+    }
+
+    /**
+     * Priority: JVM property (test/local override) > env var (fleet-wide container config) > OSGi config (normal production path).
+     * Higher-authority sources win so operators can override without touching OSGi config.
+     */
+    static boolean getUseV12Value(Map<String, Object> config) {
+        if (System.getProperty(JVM_PROPERTY_V12_ENABLED) != null) {
```
reschke, line 111: SystemPropertySupplier

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreWrapper.java
```diff
@@ -0,0 +1,392 @@
...
+        if (useV12) {
+            log.info("Starting blob store using Azure SDK v12");
+            activeImpl = createV12Store(toProperties(config));
+        } else {
+            log.info("Starting blob store using Azure SDK v8");
+            activeImpl = createV8Store(toProperties(config));
+        }
+        activeImpl.setStatisticsProvider(getStatisticsProvider());
+        // Registers activeImpl separately as AbstractSharedCachingDataStore so consumers
+        // bound to that type (e.g. oak-repository-service) get the concrete store directly,
```
reschke: what's oak-repository-service?

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreRegistrar.java
```diff
@@ -0,0 +1,392 @@
...
+            log.info("Azure SDK v12 flag: JVM property {}={}", JVM_PROPERTY_V12_ENABLED, useV12);
+            return useV12;
+        }
+        String envVar = System.getenv(ENV_VAR_V12_ENABLED);
+        if (StringUtils.isNotBlank(envVar)) {
+            boolean useV12 = Boolean.parseBoolean(envVar);
+            log.info("Azure SDK v12 flag: environment variable {}={}", ENV_VAR_V12_ENABLED, useV12);
+            return useV12;
+        }
+        if (config.containsKey(OSGI_CONFIG_V12_ENABLED)) {
```
reschke, line 123: so we have an OSGi switch, not a feature toggle, right?

reschke (commented):
Left a few nit comments.

The wrapper class should be reviewed by somebody with more OSGi knowledge.

reschke (commented):
left a few nit comments

OSGi wiring should be checked by somebody with more knowledge in that area.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreWrapper.java
seropian: ah, IDE settings. will change

.gitignore
seropian: it is from maven release, it makes a pom backup and sometimes it fails to delete it

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreRegistrar.java
seropian: SystemPropertySupplier  It is not properly implemented to detect missing values, it always returns value or defaut value. We need to know if the var was supplied or not

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreWrapper.java
seropian: internal client, will remove comment

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreRegistrar.java
seropian: we have env var  and OSGI switch, priority in this order

seropian: FT is not suited because it is only for runtime switches while this is one time choice  fixed from startup

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AbstractAzureDataStoreService.java
seropian: not sure, I thought at it before committing but did not have a clear idea

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreRegistrar.java
reschke:
~~~
    Boolean result = SystemPropertySupplier.create("foo", (Boolean)null).
    if (result != null) {
      ....
~~~ 

should work. I have a test, and will add documentation.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreRegistrar.java
reschke: But where's the FT? That's OSGi config, right?

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreRegistrar.java
reschke: Actually, that does not work, but I'll make it work.

reschke:
-> https://github.com/apache/jackrabbit-oak/pull/3003

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AbstractAzureDataStoreService.java
seropian: added some info about what to use.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreWrapper.java
seropian: fixed

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreWrapper.java
seropian: fixed

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreRegistrar.java
seropian: It is not changed via FT, but only via OSGI admin config. FT is not suitable because of it's lifecycle that does not have persisted values pushed into the component, but only  allows runtime polling for changes from the already initialized comonent.

amit-jain:
Reviewed (by claude) the v12 migration focusing on correctness, resource lifecycle, and listing/iteration memory behavior. The core rewrite looks careful — `getAllIdentifiers`/`getAllRecords` stream lazily via `PagedIterable` (no full-listing materialization), client/credential/delegation-key caching avoids per-call Netty allocation and OAuth round-trips, the `sig=` redaction in the logging policy is correct, and the fixed 64 MiB upload block size is a sensible OOM fix over v8's `min(len, 4GB)`. Two things worth addressing before merge:

## 1. Secondary-location failover is silently disabled unless `maxErrorRetry` is also set (behavior regression vs v8)

In `AzureBlobStoreBackendV12.initAzureDSConfig()`, `computeSecondaryLocationEndpoint()` correctly resolves the secondary endpoint when `enableSecondaryLocation=true`, but it is only passed through as an argument to `UtilsV12.getRetryOptions(...)`:

```java
retryOptions = UtilsV12.getRetryOptions(
        properties.getProperty(AzureConstantsV12.AZURE_BLOB_MAX_REQUEST_RETRY),
        requestTimeout,
        computeSecondaryLocationEndpoint());
```

`getRetryOptions()` returns `null` *before* it ever inspects `secondaryLocation`, whenever `maxErrorRetry` is unset:

```java
public static RequestRetryOptions getRetryOptions(final String maxRequestRetryCount, Integer requestTimeout, String secondaryLocation) {
    int retries = PropertiesUtil.toInteger(maxRequestRetryCount, -1);
    if (retries < 0) {
        return null;   // secondaryLocation discarded here
    }
    ...
}
```

and `AzureBlobContainerProviderV12` only wires the secondary host into the client when `retryOptions != null`:

```java
if (retryOptions != null) {
    builder.retryOptions(retryOptions);
}
```

**Net effect:** an operator who sets `enableSecondaryLocation=true` without also setting `maxErrorRetry` gets no secondary-region failover at all, silently. On the v8 backend (`AzureBlobStoreBackend.getBlobRequestOptions()`), `enableSecondaryLocation` and the retry policy were independent `if` blocks on the same `BlobRequestOptions`, so secondary-location worked regardless of retry config — this is a regression for anyone relying on that independence, and there's no test covering the combination.

**Suggested fix** — build retry options whenever *either* retries or a secondary location is configured, so a secondary-location-only config still gets a `RequestRetryOptions` (with SDK-default retry count) carrying the `secondaryHost`:

```java
public static RequestRetryOptions getRetryOptions(final String maxRequestRetryCount, Integer requestTimeout, String secondaryLocation) {
    int retries = PropertiesUtil.toInteger(maxRequestRetryCount, -1);
    if (retries < 0 && secondaryLocation == null) {
        return null;
    }
    RetryPolicyType type = retries > 0 ? RetryPolicyType.EXPONENTIAL : RetryPolicyType.FIXED;
    int maxTries = retries >= 0 ? Math.max(retries, 1) : /* SDK default, e.g. 4 */ 4;
    return new RequestRetryOptions(type, maxTries, requestTimeout, null, null, secondaryLocation);
}
```

(Exact default retry count is the author's call — the key point is that the `retries < 0` short-circuit must also account for `secondaryLocation != null`.) Worth adding a test asserting `getRetryOptions(null, timeout, "https://foo-secondary...")` returns non-null with the secondary host set.

## 2. (Minor / hardening) Presigned-URI expiry above 7 days silently breaks under service-principal auth

`getOrRefreshDelegationKey()` always requests the user delegation key for a fixed 7-day `DELEGATION_KEY_LIFETIME`. If `presignedHttpDownloadURIExpirySeconds` / `...UploadURIExpirySeconds` is configured above 7 days, then `sasExpiry > keyExpiry`, which has two silent consequences:

1. The fast-path cache check `cached.expiry.isAfter(sasExpiry.plus(RENEWAL_BUFFER))` can never be satisfied, so every presigned-URI call makes a fresh `getUserDelegationKey` round-trip — defeating the cache the class was built around (the O(N)-per-part path it explicitly avoids).
2. The SAS is signed by a key that expires before the SAS's own stated expiry, so the URI stops working after 7 days regardless of the configured expiry.

Presigned URIs are disabled by default (expiry `0`), so this only affects operators who deliberately configure a >7-day expiry — but it fails with no warning. Consider validating/capping the configured expiry at `DELEGATION_KEY_LIFETIME` (with a log warning) so the misconfiguration is visible.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobContainerProviderV12.java
```diff
@@ -0,0 +1,432 @@
...
+    // signed with a user delegation key (and are therefore bounded by DELEGATION_KEY_LIFETIME).
+    boolean authenticateViaServicePrincipal() {
+        return StringUtils.isBlank(azureConnectionString) &&
+                StringUtils.isNoneBlank(accountName, tenantId, clientId, clientSecret);
+    }
+
+    private BlobServiceClient getOrCreateBlobServiceClient() {
+        BlobServiceClient client = cachedBlobServiceClient.get();
+        if (client == null) {
+            synchronized (this) {
```
jsedding, line 301: Synchronization can be narrower.

```suggestion
            synchronized (cachedBlobServiceClient) {
```

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobContainerProviderV12.java
```diff
@@ -0,0 +1,432 @@
...
+            // BlobServiceClient.getBlobContainerClient() shares the existing pipeline — no new Netty client.
+            return getOrCreateBlobServiceClient()
+                    .getBlobContainerClient(containerName)
+                    .getBlobClient(key)
+                    .getBlockBlobClient();
+        }
+        // Non-SP auth: cache one container client per activation (signing never makes HTTP calls).
+        BlobContainerClient container = cachedContainerForSigning.get();
+        if (container == null) {
+            synchronized (this) {
```
jsedding, line 267: Synchronization can be narrower.

```suggestion
            synchronized (cachedContainerForSigning) {
```

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobContainerProviderV12.java
```diff
@@ -0,0 +1,432 @@
...
+     * requested for {@link #DELEGATION_KEY_LIFETIME} (7 days) so it covers any SAS expiry we
+     * would generate without frequent round-trips to Azure.
+     */
+    UserDelegationKey getOrRefreshDelegationKey(BlobServiceClient blobServiceClient, OffsetDateTime sasExpiry) {
+        // Fast path: cached key still covers sasExpiry with headroom.
+        CachedDelegationKey cached = cachedDelegationKey.get();
+        if (cached != null && cached.expiry.isAfter(sasExpiry.plus(DELEGATION_KEY_RENEWAL_BUFFER))) {
+            return cached.key;
+        }
+        synchronized (this) {
```
jsedding, line 235: Synchronization can be narrower.

```suggestion
        synchronized (cachedDelegationKey) {
```

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobContainerProviderV12.java
```diff
@@ -0,0 +1,432 @@
...
+    /**
+     * Releases cached Azure clients. The underlying Netty event loops are not eagerly shut down
+     * (the Azure SDK {@link com.azure.core.http.HttpClient} interface has no close contract), but
+     * clearing the references allows GC to reclaim them, preventing accumulation across OSGi
+     * restart cycles.
+     */
+    public void close() {
+        cachedBlobServiceClient.set(null);
+        cachedContainerForSigning.set(null);
+        cachedDelegationKey.set(null);
```
jsedding, line 287: Setting the AtomicReferences is `synchronized` where the actual values are set.

How do we prevent a race, where the AtomicReference is set while the provider is being closed?

I think we need synchronization here, but we also need to have a "is closed?" type check in the places where the refs are set.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobStoreBackendV12.java
```diff
@@ -0,0 +1,1255 @@
...
+    // Lazy: retryOptions and azureBlobContainerProvider aren't set until initContainerConnection() runs.
+    protected BlobContainerClient getAzureContainer() throws DataStoreException {
+        BlobContainerClient existing = azureContainerReference.get();
+        if (existing != null) {
+            return existing;
+        }
+        // Synchronize so getBlobContainer() (which allocates a Netty event loop) is called
+        // at most once — the previous non-synchronized compareAndSet could lose a race and
+        // silently discard a fully initialised client including its event loop group.
+        synchronized (this) {
```
jsedding, line 190: Synchronization can be narrower.

```suggestion
        synchronized (azureContainerReference) {
```

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobStoreBackendV12.java
```diff
@@ -0,0 +1,1255 @@
...
+                    concurrentRequestCount,
+                    AZURE_BLOB_MAX_CONCURRENT_REQUEST_COUNT);
+            concurrentRequestCount = AZURE_BLOB_MAX_CONCURRENT_REQUEST_COUNT;
+        }
+        LOG.info("Using concurrentRequestsPerOperation={}", concurrentRequestCount);
+
+        presignedDownloadURIVerifyExists = PropertiesUtil.toBoolean(
+                emptyToNull(properties.getProperty(AzureConstantsV12.PRESIGNED_HTTP_DOWNLOAD_URI_VERIFY_EXISTS)), true);
+
+        BlobContainerClient azureContainer = getAzureContainer();
```
jsedding, line 272: Why do we not just initialize the `BlobContainerClient` here, and set it to a field in the class? This would be much simpler than "pseudo-lazily" instantiating it in `getAzureContainer()`.

To avoid null checks, I would suggest throwing an `IllegalStateException` with a message like e.g. "not initialized, call init() before use" from `getAzureContainer()` if it is not yet set.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobContainerProviderV12.java
```diff
@@ -0,0 +1,432 @@
...
+    public String getContainerName() {
+        return containerName;
+    }
+
+    public String getAzureConnectionString() {
+        return azureConnectionString;
+    }
+
+    @NotNull
+    public BlobContainerClient getBlobContainer() throws DataStoreException {
```
jsedding, line 143: This looks more like a `createBlobContainer()` or `getOrCreateBlobContainer()` method.

However, the usages look like it _should_ be `getBlobContainer()`. Do I understand correctly, that a new BlobContainerClient is created on each call, unless service-principal authentication is used?

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobContainerProviderV12.java
```diff
@@ -0,0 +1,432 @@
...
+        this.tenantId = builder.tenantId;
+        this.clientId = builder.clientId;
+        this.clientSecret = builder.clientSecret;
+        this.clientSecretCredential = StringUtils.isNoneBlank(builder.clientId, builder.clientSecret, builder.tenantId)
+                ? new ClientSecretCredentialBuilder()
+                .clientId(builder.clientId)
+                .clientSecret(builder.clientSecret)
+                .tenantId(builder.tenantId)
+                .build()
+                : null;
```
jsedding, line 103:
```suggestion
        this.clientSecretCredential = authenticateViaServicePrincipal()
                ? new ClientSecretCredentialBuilder()
                  .clientId(clientId)
                  .clientSecret(clientSecret)
                  .tenantId(tenantId)
                  .build()
                : null;
```

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobContainerProviderV12.java
```diff
@@ -0,0 +1,432 @@
...
+            return generateUserDelegationKeySignedSas(blob, serviceSasSignatureValues, expiry, properties);
+        }
+        return generateSas(blob, serviceSasSignatureValues);
+    }
+
+    @NotNull
+    public String generateUserDelegationKeySignedSas(BlockBlobClient blobClient,
+                                                     BlobServiceSasSignatureValues serviceSasSignatureValues,
+                                                     OffsetDateTime expiryTime,
+                                                     Properties properties) {
```
jsedding, line 216: `properties` is unused.

```suggestion
    public String generateUserDelegationKeySignedSas(BlockBlobClient blobClient,
                                                     BlobServiceSasSignatureValues serviceSasSignatureValues,
                                                     OffsetDateTime expiryTime) {
```

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobContainerProviderV12.java
```diff
@@ -0,0 +1,432 @@
...
+    // endpoint (N = number of parts). Azure allows keys valid up to 7 days.
+    // Package-private for test injection.
+    final AtomicReference<CachedDelegationKey> cachedDelegationKey = new AtomicReference<>();
+
+    // Request keys for the full 7-day window so they cover any SAS expiry we'd generate.
+    // Also the hard upper bound Azure allows for a user delegation key's lifetime — package-private
+    // so callers can validate configured presigned-URI expiries against it (see AzureBlobStoreBackendV12).
+    static final Duration DELEGATION_KEY_LIFETIME = Duration.ofDays(7);
+    // Renew early enough to cover clock skew between this host and Azure.
+    private static final Duration DELEGATION_KEY_RENEWAL_BUFFER = Duration.ofSeconds(60);
```
jsedding, line 85: Most of these AI comments sound terribly clever, but don't actually tell me anything. That may be, because I don't understand or because they don't add any value.

Can you please human review and edit them, so they make sense? If they add no valuable insights, they are just noise; delete them.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobContainerProviderV12.java
```diff
@@ -0,0 +1,432 @@
...
+
+import java.net.URISyntaxException;
+import java.security.InvalidKeyException;
+import java.time.Duration;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.util.Properties;
+import java.util.concurrent.atomic.AtomicReference;
+
+class AzureBlobContainerProviderV12 {
```
jsedding, line 48: I am not convinced that lazily setting `cachedBlobServiceClient` and `cachedContainerForSigning` makes sense. The code would be simplified by eager initialization.

Can you please explain what the value of lazy initialization is?

Note: before writing this comment, I made some other comments rgd. synchronization. They become a mute point if you opt for eager initialization.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobContainerProviderV12.java
```diff
@@ -0,0 +1,432 @@
...
+        cachedDelegationKey.set(null);
+        log.debug("AzureBlobContainerProviderV12 closed; cached Azure clients released");
+    }
+
+    // Package-private: AzureBlobStoreBackendV12 needs this to know whether presigned URIs will be
+    // signed with a user delegation key (and are therefore bounded by DELEGATION_KEY_LIFETIME).
+    boolean authenticateViaServicePrincipal() {
+        return StringUtils.isBlank(azureConnectionString) &&
+                StringUtils.isNoneBlank(accountName, tenantId, clientId, clientSecret);
+    }
```
jsedding, line 296: Several code-paths in this class differ significantly depending on this boolean method.

Would it make sense to make this class abstract and have too concrete implementations (one for service-principal auth, one for the other cases). Then the distinction could be made in the builder, and the code would be simplified.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobStoreBackendV12.java
```diff
@@ -0,0 +1,1255 @@
...
+    }
+
+    private void initPresignedURIConfig() {
+        String putExpiry = properties.getProperty(AzureConstantsV12.PRESIGNED_HTTP_UPLOAD_URI_EXPIRY_SECONDS);
+        if (putExpiry != null) {
+            this.setHttpUploadURIExpirySeconds(capToDelegationKeyLifetime(Integer.parseInt(putExpiry)));
+        }
+        String getExpiry = properties.getProperty(AzureConstantsV12.PRESIGNED_HTTP_DOWNLOAD_URI_EXPIRY_SECONDS);
+        if (getExpiry != null) {
+            this.setHttpDownloadURIExpirySeconds(capToDelegationKeyLifetime(Integer.parseInt(getExpiry)));
```
jsedding, line 294: IMHO `getExpiry` sounds too much like a java method name. Make it easier to read.

```suggestion
        String uploadExpiry = properties.getProperty(AzureConstantsV12.PRESIGNED_HTTP_UPLOAD_URI_EXPIRY_SECONDS);
        if (uploadExpiry != null) {
            this.setHttpUploadURIExpirySeconds(capToDelegationKeyLifetime(Integer.parseInt(uploadExpiry)));
        }
        String downloadExpiry = properties.getProperty(AzureConstantsV12.PRESIGNED_HTTP_DOWNLOAD_URI_EXPIRY_SECONDS);
        if (downloadExpiry != null) {
            this.setHttpDownloadURIExpirySeconds(capToDelegationKeyLifetime(Integer.parseInt(downloadExpiry)));
```

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobStoreBackendV12.java
```diff
@@ -0,0 +1,1255 @@
...
+        if (existing != null) {
+            return existing;
+        }
+        // Synchronize so getBlobContainer() (which allocates a Netty event loop) is called
+        // at most once — the previous non-synchronized compareAndSet could lose a race and
+        // silently discard a fully initialised client including its event loop group.
+        synchronized (this) {
+            existing = azureContainerReference.get();
+            if (existing == null) {
+                existing = azureBlobContainerProvider.getBlobContainer();
```
jsedding, line 193: At this point the name `existing` becomes a lie. Maybe just rename to `client` or `container`?

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobStoreBackendV12.java
```diff
@@ -0,0 +1,1255 @@
...
+            Thread.currentThread().setContextClassLoader(saved);
+        }
+    }
+
+    private void withBundleContextClassLoaderVoid(AzureSDKCallVoid call) throws DataStoreException {
+        withBundleContextClassLoader(() -> {
+            call.execute();
+            return null;
+        });
+    }
```
jsedding, line 221: Making these static signals that they are pure utilities.

```suggestion
    private static <T> T withBundleContextClassLoader(AzureSDKCall<T> call) throws DataStoreException {
        ClassLoader saved = Thread.currentThread().getContextClassLoader();
        try {
            Thread.currentThread().setContextClassLoader(AzureBlobStoreBackendV12.class.getClassLoader());
            return call.execute();
        } catch (DataStoreException | RuntimeException e) {
            throw e;
        } catch (Exception e) {
            throw new DataStoreException(e);
        } finally {
            Thread.currentThread().setContextClassLoader(saved);
        }
    }

    private static void withBundleContextClassLoaderVoid(AzureSDKCallVoid call) throws DataStoreException {
        withBundleContextClassLoader(() -> {
            call.execute();
            return null;
        });
    }
```

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobContainerProviderV12.java
```diff
@@ -0,0 +1,432 @@
...
+            withClientId(properties.getProperty(AzureConstantsV12.AZURE_CLIENT_ID, ""));
+            withClientSecret(properties.getProperty(AzureConstantsV12.AZURE_CLIENT_SECRET, ""));
+            withProxyHost(properties.getProperty(AzureConstantsV12.PROXY_HOST, ""));
+            withProxyPort(properties.getProperty(AzureConstantsV12.PROXY_PORT, ""));
+            return this;
+        }
+
+        public AzureBlobContainerProviderV12 build() {
+            return new AzureBlobContainerProviderV12(this);
+        }
```
jsedding, line 430: Class visibility is broken here.

The class `AzureBlobContainerProviderV12` is package-private, but `Builder` is public. However, the public `build()` method returns package-private type `AzureBlobContainerProviderV12`.

Options:
- make `build()` package-private; however, that makes the public `Builder` useless
- make `Builder` package-private (preferred IMHO)
- make `AzureBlobContainerProviderV12` public

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobStoreBackendV12.java
```diff
@@ -0,0 +1,1255 @@
...
+
+        BlobContainerClient azureContainer = getAzureContainer();
+
+        try {
+            if (createBlobContainer && Boolean.FALSE.equals(azureContainer.exists())) {
+                azureContainer.create();
+                LOG.info("New container created. containerName={}", getContainerName());
+            } else {
+                LOG.info("Reusing existing container. containerName={}", getContainerName());
+            }
```
jsedding, line 280: Simplify and reduce from max 2 calls to max 1 call.

Unless we need to check for existence also when `createBlobContainer = false`?! In that case, I expect we fail later due to a non-existent container? It might be nicer to fail early with a clear message.

```suggestion
            if (createBlobContainer && azureContainer.createIfNotExists()) {
                LOG.info("New container created. containerName={}", getContainerName());
            } else {
                LOG.info("Reusing existing container. containerName={}", getContainerName());
            }
```

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobStoreBackendV12.java
```diff
@@ -0,0 +1,1255 @@
...
+            if (existing == null) {
+                existing = azureBlobContainerProvider.getBlobContainer();
+                azureContainerReference.set(existing);
+            }
+            return existing;
+        }
+    }
+
+    // Swaps Thread Class Context Loader to this bundle's classloader so Azure SDK's ServiceLoader-based SPI discovery works in OSGi.
+    // RuntimeExceptions (including BlobStorageException) propagate as-is; other checked exceptions are wrapped.
```
jsedding, line 201: It seems odd that the context-classloader should be required so often.

Where exactly does the Azure SDK use a `ServiceLoader`? Or what breaks without changing the class loader?

I would like to understand this better, because swapping the class loader to work around OSGi issues should be a last resort. And I would like to understand if we can find a better, more OSGi-friendly solution to the problem.

Given that the Azure SDK is embedded in this bundle, it isn't intuitive to me why we need to swap class loaders.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreWrapper.java
```diff
@@ -0,0 +1,407 @@
...
+ *
+ * <p>Replaces the old dual-service architecture (AzureDataStoreService + AzureDataStoreServiceV12
+ * + AzureSDKConditionGate) that caused deadlocks during runtime OSGi service swap.
+ */
+@Component(
+        name = AzureDataStoreWrapper.NAME,
+        configurationPid = AzureDataStoreWrapper.NAME,
+        configurationPolicy = ConfigurationPolicy.REQUIRE
+)
+public class AzureDataStoreWrapper extends AbstractDataStoreService {
```
jsedding: Could we use a more descriptive name? This does not exactly "wrap" an AzureDataStore, in the commonly used sense of delegation.

```suggestion
public class AzureDataStoreRegistrar extends AbstractDataStoreService {
```

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/AzureDataStoreRegistrar.java
```diff
@@ -0,0 +1,407 @@
...
+                    .loggingTo(log)
+                    .formatSetMessage((name, useV12) -> "Azure SDK v12 flag: JVM property " + name + "=" + useV12)
+                    .get();
+        }
+        String envVar = System.getenv(ENV_VAR_V12_ENABLED);
+        if (StringUtils.isNotBlank(envVar)) {
+            boolean useV12 = Boolean.parseBoolean(envVar);
+            log.info("Azure SDK v12 flag: environment variable {}={}", ENV_VAR_V12_ENABLED, useV12);
+            return useV12;
+        }
```
jsedding, line 122: Is the env-var handling really necessary? It should be trivial to inject an env-var value into the OSGi configuration if needed.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureDataStoreV12.java
```diff
@@ -0,0 +1,131 @@
...
+        azureBlobStoreBackend = new AzureBlobStoreBackendV12();
+        if (properties != null) {
+            azureBlobStoreBackend.setProperties(properties);
+        }
+        return azureBlobStoreBackend;
+    }
+
+    public void setProperties(final Properties properties) {
+        this.properties = properties;
+    }
```
jsedding, line 54: It would be really nice to stop perpetuating this pattern using a properties setter. Why not pass it to the constructor? You could even have a default constructor that passes in `null` or an empty properties object.

This should then propagate to `AzureBlobStoreBackendV12` etc. Or maybe it would be even better to "extract" all relevant properties in the `createBackend() method and pass the actual values to the `AzureBlobStoreBackendV12` constructor. That way, all the supported properties would be visible in one place. And at the same time, `AzureBlobStoreBackendV12` would use regular fields to access this configuration, making the code simpler.

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/UtilsV12.java
```diff
@@ -0,0 +1,164 @@
...
+    public static final String DASH = "-";
+    public static final String DEFAULT_CONFIG_FILE = "azurev12.properties";
+
+    private UtilsV12() {
+    }
+
+    /**
+     * Returns proxy options when both host and port are non-blank, null otherwise.
+     */
+    static ProxyOptions computeProxyOptions(@Nullable String proxyHost, @Nullable String proxyPort) {
```
jsedding, line 50:
```suggestion
    static ProxyOptions createProxyOptions(@Nullable String proxyHost, @Nullable String proxyPort) {
```

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/UtilsV12.java
```diff
@@ -0,0 +1,164 @@
...
+        if (StringUtils.isNoneBlank(proxyHost, proxyPort)) {
+            return new ProxyOptions(ProxyOptions.Type.HTTP,
+                    new InetSocketAddress(proxyHost, Integer.parseInt(proxyPort)));
+        }
+        return null;
+    }
+
+    public static RequestRetryOptions getRetryOptions(final String maxRequestRetryCount, Integer requestTimeout, String secondaryLocation) {
```
jsedding, line 58:
```suggestion
    public static RequestRetryOptions createRetryOptions(final String maxRequestRetryCount, Integer requestTimeout, String secondaryLocation) {
```

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/UtilsV12.java
```diff
@@ -0,0 +1,164 @@
...
+            return new RequestRetryOptions(RetryPolicyType.FIXED, 1,
+                    requestTimeout, null, null,
+                    secondaryLocation);
+        }
+        return new RequestRetryOptions(RetryPolicyType.EXPONENTIAL, retries,
+                requestTimeout, null, null,
+                secondaryLocation);
+    }
+
+    public static String getConnectionStringFromProperties(Properties properties) {
```
jsedding, line 83:
```suggestion
    public static String createConnectionStringFromProperties(Properties properties) {
```

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/UtilsV12.java
```diff
@@ -0,0 +1,164 @@
...
+            return getConnectionStringForSas(sasUri, blobEndpoint, accountName);
+        }
+
+        return getConnectionString(
+                accountName,
+                accountKey,
+                blobEndpoint);
+    }
+
+    public static String getConnectionStringForSas(String sasUri, String blobEndpoint, String accountName) {
```
jsedding, line 104:
```suggestion
    public static String createConnectionStringForSas(String sasUri, String blobEndpoint, String accountName) {
```

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/UtilsV12.java
```diff
@@ -0,0 +1,164 @@
...
+        if (StringUtils.isEmpty(blobEndpoint)) {
+            return String.format("AccountName=%s;SharedAccessSignature=%s", accountName, sasUri);
+        } else {
+            return String.format("BlobEndpoint=%s;SharedAccessSignature=%s", blobEndpoint, sasUri);
+        }
+    }
+
+    public static String getConnectionString(final String accountName, final String accountKey, String blobEndpoint) {
```
jsedding, line 112:
```suggestion
    public static String createConnectionString(final String accountName, final String accountKey, String blobEndpoint) {
```

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/UtilsV12.java
```diff
@@ -0,0 +1,164 @@
...
+            connString.append(";BlobEndpoint=").append(blobEndpoint);
+        }
+        return connString.toString();
+    }
+
+    /**
+     * Returns a {@link BlobContainerClient} from a connection string. The caller supplies a shared
+     * {@link HttpClient}; pass {@code null} to use the SDK default.
+     */
+    public static BlobContainerClient getBlobContainerFromConnectionString(final String azureConnectionString,
```
jsedding, line 126:
```suggestion
    public static BlobContainerClient createBlobContainerFromConnectionString(final String azureConnectionString,
```

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/UtilsV12.java
```diff
@@ -0,0 +1,164 @@
...
+    public static Properties readConfig(String fileName) throws IOException {
+        if (!new File(fileName).exists()) {
+            throw new IOException("Config file not found. fileName=" + fileName);
+        }
+        Properties prop = new Properties();
+        try (InputStream in = new FileInputStream(fileName)) {
+            prop.load(in);
+        }
+        return prop;
+    }
```
jsedding, line 163: Modern `Path` API, and less code. Drawback: slightly less specific `IOException`. But probably good enough.

```suggestion
    public static Properties readConfig(String fileName) throws IOException {
        Properties prop = new Properties();
        try (InputStream in = Files.newInputStream(Path.of(fileName))) {
            prop.load(in);
        }
        return prop;
    }
```

jsedding (changes requested):

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobContainerProviderV12.java
```diff
@@ -0,0 +1,432 @@
...
+            final String connectionStringWithSasToken = UtilsV12.getConnectionStringForSas(sasToken, blobEndpoint, accountName);
+            return UtilsV12.getBlobContainerFromConnectionString(connectionStringWithSasToken, containerName, retryOptions, httpClient);
+        }
+        log.debug("connecting to azure blob storage via access key");
+        final String connectionStringWithAccountKey = UtilsV12.getConnectionString(accountName, accountKey, blobEndpoint);
+        return UtilsV12.getBlobContainerFromConnectionString(connectionStringWithAccountKey, containerName, retryOptions, httpClient);
+    }
+
+    @NotNull
+    public String generateSharedAccessSignature(RequestRetryOptions retryOptions,
```
smiroslav, line 163: This method is unused.

seropian (commented):
Items are being addressed

oak-blob-cloud-azure/src/main/java/org/apache/jackrabbit/oak/blob/cloud/azure/blobstorage/v12/AzureBlobContainerProviderV12.java
seropian: will be removed.

PR #3003, opened 2026-07-06
OAK-12298: SystemPropertySupplier: allow to signal that property is not present (implies default value)

reschke:
…t present (implies default value)

### DIFF
--- a/oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/properties/SystemPropertySupplier.java
+++ b/oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/properties/SystemPropertySupplier.java
@@ -23,6 +23,7 @@ import java.util.function.Predicate;
 import java.util.function.Supplier;
 
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


rishabhdaim:
@reschke regarding the commit-check failure due to 120 characters, should we increase it further to 200 ?

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/properties/SystemPropertySupplier.java
```diff
@@ -100,18 +109,10 @@ public SystemPropertySupplier<T> formatSetMessage(@NotNull BiFunction<String, T,
...
-            case "TRACE":
-            case "WARN":
-                newLevel = successLogLevel;
-                break;
-            default:
-                throw new IllegalArgumentException("unsupported log level: " + successLogLevel);
-        }
+        String newLevel = switch (Objects.requireNonNull(successLogLevel)) {
+            case "DEBUG", "ERROR", "INFO", "TRACE", "WARN" -> successLogLevel;
+            default -> throw new IllegalArgumentException("unsupported log level: " + successLogLevel);
```
rishabhdaim: In https://github.com/apache/jackrabbit-oak/pull/2995, we are removing this exception, so I would avoid changing this method.

oak-commons/src/test/java/org/apache/jackrabbit/oak/commons/properties/SystemPropertySupplierTest.java
```diff
@@ -137,4 +138,38 @@ public void testUnsupportedType() {
         } catch (IllegalArgumentException expected) {
         }
     }
+
+    @Test
+    public void testCheckNullDefault() {
+        try {
+            assertNull(SystemPropertySupplier.create("foo", Boolean.class).
+                    usingSystemPropertyReader((n) -> null).get());
+        } catch (IllegalArgumentException expected) {
```
rishabhdaim, line 157: I would re-write this test to use `@Test(expected=....)` pattern rather than try/catch.
and add `Assert.fail()` after the test so that it does test that exception is thrown.

mbaedke: After your previous comment will be addressed, which I support, this test will drop anyway.

mbaedke (commented):
See comments inline.

oak-commons/src/main/java/org/apache/jackrabbit/oak/commons/properties/SystemPropertySupplier.java
reschke: indeed

reschke: actually, this is because of two changes that are indeed not indepentant. Will rebease.

reschke:
Closed in favor of https://github.com/apache/jackrabbit-oak/pull/3044

PR #3004, opened 2026-07-06
OAK-12299 : precompute elementCount and currentWeight in PersistentDiskCache

nfsantos:
Track cache size and element count incrementally in AtomicLong fields instead of recomputing them on every stats read via `directory.listFiles()` and `FileUtils.sizeOfDirectory()`, which walked the whole cache directory. Counters are seeded from the on-disk state at startup.


oak-segment-remote/src/test/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCacheCounterTest.java
```diff
@@ -0,0 +1,185 @@
...
+    public void testStartupExcludesTempFiles() throws Exception {
+        File cacheFolder = temporaryFolder.newFolder();
+
+        File segment = new File(cacheFolder, UUID.randomUUID().toString());
+        try (FileOutputStream fos = new FileOutputStream(segment)) {
+            fos.write(new byte[100]);
+        }
+
+        // One temp file (.part) that must NOT be counted
+        File temp = new File(cacheFolder, UUID.randomUUID().toString() + System.nanoTime() + ".part");
```
joerghoh: can you reuse the constant ``TEMP_FILE_SUFFIX``  in PersistentDiskCache?

joerghoh (approved):
looks good, just a minor nit in the tests

reschke (commented):
Would be nice to park this until after 2.4.0.

oak-segment-remote/src/test/java/org/apache/jackrabbit/oak/segment/remote/persistentcache/PersistentDiskCacheCounterTest.java
nfsantos: Done in b9c4d24758b435a68abd325eed12d415d1518a3c

nfsantos:
> Would be nice to park this until after 2.4.0.

I'm neutral on when to merge this. But maybe @joerghoh can comment on whether this change is something that would be useful to get in quickly or if it can wait. I guess this can reduce by a noticeable margin how long it takes to collect metrics from the cache.

BTW, this patch is already live in our Oak fork, so it has been running in production for a couple of months.

nfsantos:
> Would be nice to park this until after 2.4.0.

Can we proceed with merging now?

PR #3011, opened 2026-07-08
OAK-12295 : fixed version range of mongo driver version

rishabhdaim:


reschke (changes requested):
Fiven the fact that we we wan't to start the release process for Oak 2.4.0, I#d really prefer to back out the version change for now.

PR #3022, opened 2026-07-10
OAK-12309: DocumentNodeStoreService may fail to register or double register the NodeStore due to race conditions

mbaedke:
Added synchronization.


reschke (commented):
I would get rid of the comments.

reschke (approved):
I would recommend to get rid of the comments; it feels weird to just have them for this minor code change.

PR #3054, opened 2026-07-23
OAK-12327: PropertyValueImpl with type restriction lost when "is not null"

jsedding:
- support property type information for "is not null" and "is null" constraints
- unit tests

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/AndImpl.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/AndImpl.java
@@ -99,7 +99,7 @@ public class AndImpl extends ConstraintImpl {
 
     @Override
     public Set<PropertyExistenceImpl> getPropertyExistenceConditions() {
-        Set<PropertyExistenceImpl> result = new HashSet<>();
+        Set<PropertyExistenceImpl> result = new LinkedHashSet<>();
         for (ConstraintImpl constraint : constraints) {
             result.addAll(constraint.getPropertyExistenceConditions());
         }
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/SQL2Parser.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/SQL2Parser.java
@@ -543,11 +543,11 @@ public class SQL2Parser {
     }
 
     private PropertyExistenceImpl getPropertyExistence(PropertyValueImpl p) throws ParseException {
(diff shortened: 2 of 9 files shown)


jsedding:
@joerghoh FYI

oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/AndImpl.java
```diff
@@ -99,7 +99,7 @@ ConstraintImpl not() {
 
     @Override
     public Set<PropertyExistenceImpl> getPropertyExistenceConditions() {
-        Set<PropertyExistenceImpl> result = new HashSet<>();
+        Set<PropertyExistenceImpl> result = new LinkedHashSet<>();
```
jsedding, line 102: TL;DR - if you prefer, I can remove this change.

This change is only present to make `org.apache.jackrabbit.oak.query.ast.PropertyExistenceTest#andXY` fully deterministic.

With the other changes present, the test seems to pass reliably even without this change. Presumably this is due to the `#hashCode()` implementations. However during development, before I settled on the current `PropertyExistenceImpl#hashCode()` implementation, the test failed for a while.

jsedding:
Any objections to merging this today?

jsedding:
@reschke alerted me to the following risk. Fixing this properly would require more changes, for which I currently don't have the time. I'll therefore abandon this PR.

I think that `PropertyExistenceImpl.getPropertyExistenceConditions()` would need to return two instances if the propertyType is restricted: the one restricted by `propertyType` and one with `propertyType=UNDEFINED`. Otherwise a valid existence condition could be dropped due to an `equals()` mismatch.

**What Changed:**
`PropertyExistenceImpl.hashCode()` and `equals() `now delegate to `PropertyValueImpl.hashCode()/equals()`, which includes `propertyType` and `selector` in the comparison. Previously, equality was based only on `selectorName` and `normalizePropertyName(propertyName)`.

**Why It Is Risky:**
`OrImpl.getPropertyExistenceConditions()` uses `Set.retainAll()` to find common property existence conditions across OR branches. If one branch uses `property([test], 'String') = 'x' `and another uses `[test] = 'y'`, the derived PropertyExistenceImpl objects will have different propertyType values (STRING vs UNDEFINED) and will NOT be considered equal. Previously, both would have produced equal `PropertyExistenceImpl` objects (same selector, same property name), allowing the optimizer to derive a `[test] IS NOT NULL` restriction for index selection.

PR #3058, opened 2026-07-30
OAK-12331: document the audit SPI

dulvac:
Documentation for the audit SPI added under OAK-12331.

I'm opening this ahead of the implementation on purpose. The whole change is
large, and it's much easier to review in two chunks than in one, so the docs
come first and the code is ready to follow straight after. Reviewing this
first also means any disagreement about the SPI's shape or its contracts
surfaces here, in prose, rather than after the implementation is written.

Two new pages under `oak-doc/src/site/markdown/security/`:

- **`audit.md`**, the consumer guide: event model, the two producer paths
  (commit-attached and fire-and-forget), the `commit.*` metadata keys,
  configuring and probing the pipeline, emitting events, implementing a
  listener, and the trust model.
- **`audit-design.md`**, the design document: pipeline internals, SPI layout
  across `oak-core-spi` / `oak-security-spi` / `oak-core`, OSGi and embedded
  wiring, and the threading and ordering rules.

Both are linked from `security/overview.md` and nested under Security in
`site.xml`.

**Docs only, so please don't merge ahead of the implementation**, or the site
will describe an SPI that isn't in the release.

### Notable design choices

- The observer drains the per-session buffer *before* checking the feature
  toggle. Deliberate: draining unconditionally stops a mid-flight toggle flip
  from stranding staged events for a later commit to misattribute.
- The commit-attached path depends on the observer running on the thread that
  called `Root.commit()`, before `merge` returns, because the event buffer is
  a `ThreadLocal`. It does not depend on every store dispatching through
  `ChangeDispatcher`: `MemoryNodeStore` notifies its observers directly from
  `setRoot`.
- Two segment configurations silently produce no commit-attached events: a
  cold-standby instance, and a store configured through
  `SegmentNodeStoreFactory` without `dispatchChanges` set.
- The per-session buffer cap means a persisted write can leave no audit event
  behind. Documented as a gap in the trail, since it matters for anyone
  building a compliance record.

`mvn site -Pdoc` builds clean and RAT passes. The embedded wiring snippet has
been compiled and run.


oak-doc/src/site/markdown/security/audit-design.md
```diff
@@ -0,0 +1,435 @@
...
+<a name="overview"></a>
+### Overview
+
+The audit pipeline transports structured `AuditEvent`s from producers to
+bundle-registered `AuditEventListener` consumers, gated by a feature toggle
+and a per-domain listener registry.
+
+There are two delivery paths:
+
+- **Commit-attached.** Oak-internal capture sites (e.g. `UserManagerImpl`)
```
joerghoh, line 43: You should explain the term ``site`` , at least it is not known to me.

oak-doc/src/site/markdown/security/audit-design.md
```diff
@@ -0,0 +1,435 @@
...
+  call `AuditEvents.record(root, event)`. Events land in a per-session
```
joerghoh: What about operations which do not require an explicit ``session.save()``?

oak-doc/src/site/markdown/security/audit-design.md
```diff
@@ -0,0 +1,435 @@
...
+The observer fires synchronously on the commit thread, after durable
+persistence and before `store.merge` returns. That is the property the
+per-session buffer depends on: the drain has to happen on the thread that
+filled it. Every production `NodeStore` notifies observers that way, though
+not all by the same route. The document and composite stores dispatch
+through `ChangeDispatcher`; `MemoryNodeStore` iterates its registered
+observers directly from `setRoot`. Either way the notification completes
+before `merge` returns.
+
+Two segment-store configurations are worth knowing about, because in both the
```
joerghoh, line 117: I am not on top of the Oak development, but are nowadays all different persistence options a segment-store? What about DocumentNodeStore implementations? And in that case: What about clustering?

oak-doc/src/site/markdown/security/audit-design.md
```diff
@@ -0,0 +1,435 @@
...
+
+Producer-side factories are deliberately not part of the SPI. They live as
+package-private classes next to their only callers, e.g.
+`UserAuditEvents` next to `UserManagerImpl` in `oak-core`. The asymmetry
+(read-side vocabulary public, write-side factories impl-private) raises the
+bar for casually forging Oak-attested events, but it is not a hard boundary:
+any bundle can call `AuditEvent.of(domain, type, payload)` directly.
+Listeners that need to distinguish Oak-attested commit-attached events from
+fire-and-forget emissions check for the three reserved `commit.*` payload
+keys, per the trust contract on `AuditEvent#getPayload()`.
```
joerghoh: That will lead to a of redundant code (and relying on the knowledge that these 3 properties only exist when the event is initially created by Oak, which is IMO too much implementation knowledge); can you provide a static method in the SPI to check for this particular case?

oak-doc/src/site/markdown/security/audit-design.md
```diff
@@ -0,0 +1,435 @@
...
+bypassing `MutableRoot` and the capture sites. Migration tools do not set the
+pipeline up, so when one runs standalone there is no observer attached and no
+audit machinery on the path at all. Run inside a container where audit is
+deployed, the observer does fire on migration commits, finds an empty buffer
+for the session id, and returns. Either way migration mutations are not
+audited; if that is ever wanted, it is a capture-site addition, not a
+pipeline change.
+
+<a name="design_rules"></a>
+### Design rules
```
joerghoh: I would make this the first section in the ``Implementation`` chapter.

joerghoh:
For the implementation I am missing some basic metrics in the implementation, for example:
* number of events per domain (a counter would be sufficient)
* duration of the ``AuditEventListener.onEvents()`` per listener (a constantly increasing value in ms?), which allows to detect and identify slow AuditEventListeners.
* and maybe a few more

oak-doc/src/site/markdown/security/audit-design.md
dulvac: that's a term I made up and never defined. A "capture site" is a place in Oak's own code that records an audit event. Right now the only ones are in `UserManagerImpl`, on group membership add/remove. I'll define it on first use in both pages.

oak-doc/src/site/markdown/security/audit-design.md
dulvac: Writing "the surrounding `Root.commit()`" made this sound like i meant `session.save()`. I'll reword to say "the commit that follows, explicit or implicit".
The one case with no drain is a write that never goes through `MutableRoo`t at all. e.g. oak-upgrade calling `NodeStore.merge` directly. That's the Migration commits section.

oak-doc/src/site/markdown/security/audit-design.md
dulvac: Document and composite are the paragraph just above: both dispatch through ChangeDispatcher, and `DocumentNodeStore.addObserver` always returns a real handle. The segment paragraph is only about segment, where `addObserver` hands back a no-op unless change dispatch is on.

I'll add "in the segment store" to the opening sentence so it doesn't seem like a general claim.

 Clustering I don't cover at all. Events are node-local: the observer short-circuits on CommitInfo.isExternal(), so a write on instance A produces events on A only, and a listener deployed cluster-wide sees each event once, on the instance that did the write.  I'll add a short clustering section to audit.md.

oak-doc/src/site/markdown/security/audit-design.md
dulvac: fair. the key constants are package-private in [oak-core](https://issues.apache.org/jira/browse/OAK-core), so a listener has no choice but to hardcode the string. I'll add a static predicate plus public key constants to the SPI, and cut the doc down to naming it instead of describing the check.
But the predicate can only ever be "these keys are present", since Oak doesn't sign anything so I'll name it for what it checks rather than implying a stronger guarantee. And it belongs in oak-core-spi next to AuditEvent, so listeners keep their single-module dependency. The API change lands in the implementation PR (https://github.com/apache/jackrabbit-oak/pull/3059) and I'll update this page to match.

oak-doc/src/site/markdown/security/audit-design.md
dulvac: Makes sens. reading the invariants first and the components second is the better order. It's written as a recap right now, so a few back-references (rule 9's "described above") turn into forward pointers. I'll restate those inline so each rule stands on its own and move the section to the top of `Implementation`.

dulvac:
Good catch, there are no metrics at all right now. A slow listener runs synchronously on the commit thread, so it directly costs commit latency, and there's currently no way to see that from outside. I'll wire a `StatisticsProvider` into the pipeline for a per-domain event counter and a per-listener timer around `onEvents` + a counter for events dropped at the per-session cap. This last one is a gap in the audit trail and today only shows up as a WARN.

Both will go in the implementation PR with a monitoring section here. If you have other metric counters we should add, let me know

dulvac:
@joerghoh thanks again for the review. I addressed your feedback in the latest changes. please have a look when you can.

PR #3059, opened 2026-07-30
OAK-12331: add an audit event SPI with commit-attached capture and observer drain

dulvac:
Implementation half of OAK-12331. Pairs with #3058, which carries the
documentation; this PR touches no files under `oak-doc/`.

Draft for now: opening it early for feedback on the SPI shape before the
capture-site coverage grows.

## What this adds

A pluggable audit SPI, so a consumer bundle can observe repository
operations without patching Oak internals. Off by default behind the
`FT_OAK-12331` toggle. With the toggle off, or with no audit module
deployed, capture short-circuits through a NOOP sink.

The SPI lives in `oak-core-spi`, in the newly exported package
`org.apache.jackrabbit.oak.spi.audit`:

- `AuditEvent` / `AuditEventImpl`: immutable event carrying domain, type,
  timestamp and payload
- `AuditEvents`: static facade over a pluggable `Sink`, NOOP until an audit
  module installs one
- `AuditEventListener`: the consumer contract, selecting by domain and
  ordering by rank
- `AuditEventEmitter`: OSGi-facing emitter for the fire-and-forget path
- `AuditBufferLifecycle`: session lifecycle callouts used by `MutableRoot`

`oak-security-spi` contributes the security domain name and the
user-management type vocabulary, in the newly exported
`org.apache.jackrabbit.oak.spi.security.audit` plus `UserAuditTypes`, so
listener bundles compile against a stable set of strings rather than the
producer-side helpers. The pipeline itself sits in `oak-core` under
`org.apache.jackrabbit.oak.security.audit` and stays bundle-internal.

## How it works

Capture sites buffer events per session in a `ThreadLocal`.
`AuditDrainObserver` runs as an `Observer` on the root NodeStore. Observers
fire after durable persistence and never for a failed merge, so a
dispatched event always corresponds to a persisted write. The drain is
destructive, so a store that invokes the observer twice for one merge
dedupes itself.

`CommitMetadataDecorator` stamps the three reserved `commit.*` keys at drain
time and overwrites anything a caller supplied. The fire-and-forget path
strips those keys instead. Their presence is therefore an attestation that
an event came from the commit-attached pipeline, which is what the tests in
`CommitMetadataDecoratorTest` pin.

An outer `Throwable` barrier in `contentChanged` keeps a failing listener
from surfacing as a commit failure to the merge caller. `CompositeObserver`
gives no per-observer isolation, so without the barrier a bad listener would
also break peer observers such as JCR observation.

`MutableRoot` drains on `refresh()` and on a failed merge. It deliberately
does not drain on `rebase()`, which keeps transient changes and replays them
on the new base; audit events captured alongside those surviving changes
have to survive with them.

`AuditConfigurationImpl` owns the lifecycle. It registers as
`AuditConfiguration` only, contributes no commit hooks, and is not a
`SecurityConfiguration`, so it stays out of
`SecurityProvider.getConfiguration()`. Under OSGi it publishes the observer
as a service for `ObserverTracker`; embedded callers get it from
`getDrainObserver()` and attach it themselves.

## Capture sites

Only the two `UserManagerImpl.onGroupUpdate` overloads so far, covering
single and bulk group membership changes. Path resolution failures there
drop the event and log rather than failing the surrounding update, with
repeat warnings demoted to DEBUG. More capture sites are meant to follow in
separate issues.

## Testing

| Module | Tests |
|---|---|
| `oak-core-spi` | 243, all passing |
| `oak-security-spi` | 1110, all passing |
| `oak-core` | 4873, all passing |
| `oak-run-commons` | 250, all passing |

`oak-benchmarks` compiles; it gains two runs measuring capture-site and
empty-commit overhead. Also adds an audit-enabled in-memory fixture to
`oak-run-commons`.

The `SecurityProviderRegistrationTest` change is a comment plus keeping the
existing count at 6, documenting that audit is deliberately not a
`SecurityConfiguration`. No assertion was weakened.

## Review notes

Both newly exported packages pass the `maven-bundle-plugin` baseline check.
Worth a look from anyone who has worked on `MutableRoot` or the observer
chain, since the commit-attached path depends on observer timing
guarantees.

### DIFF
--- a/oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEvent.java
+++ b/oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEvent.java
@@ -30,44 +30,66 @@ import org.osgi.annotation.versioning.ProviderType;
  * <ul>
  *   <li>Oak-internal capture sites tied to a successful
  *       {@code Root.commit()}. The commit-attached drain step decorates
- *       payload with {@code commit.sessionId}, {@code commit.userId},
- *       and {@code commit.timestamp} entries before dispatch.</li>
+ *       payload with the three reserved attestation entries before
+ *       dispatch; use {@link #isCommitAttested(AuditEvent)} to test for
+ *       them rather than checking the keys by hand.</li>
  *   <li>Any bundle calling {@link AuditEventEmitter#emit(AuditEvent)}.
  *       Such events carry the payload provided by the caller, except that
  *       Oak strips caller-supplied values for the three reserved
- *       {@code commit.*} attestation keys before dispatch — see the trust
- *       contract on {@link #getPayload()}.</li>
+ *       attestation keys before dispatch — see the trust contract on
+ *       {@link #getPayload()}.</li>
  * </ul>
  * The {@link #getDomain()} value selects the listeners that receive this
  * event.
  * <p>
  * Most callers do not implement this interface directly: use the static
- * factory {@link #of(String, String, Map)} (or the no-payload overload
- * {@link #of(String, String)}) to construct an immutable event with the
- * current wall-clock timestamp. The package-private {@code AuditEventImpl}
- * backs these factories.
+ * factory {@link #of(AuditDomain, AuditType, Map)} (or the no-payload
+ * overload {@link #of(AuditDomain, AuditType)}) to construct an immutable
+ * event with the current wall-clock timestamp. The package-private
+ * {@code AuditEventImpl} backs these factories.
  */
 @ProviderType
 public interface AuditEvent {
+
+    /**
+     * Payload key carrying the id of the session whose commit produced this
+     * event. One of the three reserved attestation keys; see the trust
+     * contract on {@link #getPayload()} and {@link #isCommitAttested}.
+     */
+    String COMMIT_SESSION_ID = "oak.commit.sessionId";
+
+    /**
+     * Payload key carrying the user id of the commit that produced this
+     * event. {@code "oak:unknown"} for system commits, which listeners
+     * <strong>MUST NOT</strong> try to resolve to a real identity.
+     */
+    String COMMIT_USER_ID = "oak.commit.userId";
+
+    /**
+     * Payload key carrying the commit timestamp (millis since epoch), i.e.
+     * when the change became durable — as opposed to
+     * {@link #getTimestamp()}, which is when the event was captured.
+     */
+    String COMMIT_TIMESTAMP = "oak.commit.timestamp";
 
     /**
      * Returns the domain that owns this event. Listeners are domain-scoped:
      * an {@link AuditEventListener} only receives events whose
      * {@code getDomain()} matches its own {@link AuditEventListener#getDomain()}.
      *
-     * @return non-null domain name (e.g. {@code "oak.security"}, {@code "aem.content"}).
+     * @return non-null domain (e.g. {@code "oak.security"}).
      */
     @NotNull
-    String getDomain();
+    AuditDomain getDomain();
 
     /**
-     * Returns the event type identifier within the domain. Type strings are
-     * stable across releases for any given domain.
+     * Returns the event type within the domain. Types are stable across
+     * releases for any given domain.
      *
-     * @return non-null type identifier.
+     * @return non-null type.
      */
     @NotNull
-    String getType();
+    AuditType getType();
 
     /**
      * Returns the wall-clock timestamp (millis since epoch) captured at
@@ -94,15 +116,16 @@ public interface AuditEvent {
      * this to expose typed accessors and include their fields here.
      * <p>
      * For commit-attached events, Oak's drain path adds entries with the
-     * keys {@code commit.sessionId}, {@code commit.userId}, and
-     * {@code commit.timestamp} when the buffer is drained on commit
+     * keys {@code oak.commit.sessionId}, {@code oak.commit.userId}, and
+     * {@code oak.commit.timestamp} when the buffer is drained on commit
      * success. Oak does not <em>add</em> these entries on the fire-and-forget
      * path (see the trust contract below).
      * <p>
      * <strong>Trust contract</strong> (normative — other audit SPI and
      * implementation docs defer to this paragraph). For events delivered
-     * through Oak dispatch, the three reserved keys {@code commit.sessionId},
-     * {@code commit.userId} and {@code commit.timestamp} are Oak-attested:
+     * through Oak dispatch, the three reserved keys
+     * {@code oak.commit.sessionId}, {@code oak.commit.userId} and
+     * {@code oak.commit.timestamp} are Oak-attested:
      * <ul>
      *   <li>On the <em>commit-attached</em> path Oak <em>unconditionally
      *       overwrites</em> the three keys with the values from
@@ -113,12 +136,13 @@ public interface AuditEvent {
      *       {@code AuditEvents.dispatch}) Oak <em>strips</em> caller-supplied
      *       values for the same three keys before delivery.</li>
      * </ul>
-     * A listener may therefore treat the presence of any of the three keys
-     * in a dispatched payload as "commit-attached event, values supplied by
-     * Oak". Every other entry — including other {@code commit.*}-prefixed
-     * keys — is forwarded verbatim from the caller-supplied payload on both
-     * paths and is untrusted: anchor trust on the three reserved keys, never
-     * on the {@code commit.} prefix in general.
+     * A listener may therefore treat the presence of the three keys in a
+     * dispatched payload as "commit-attached event, values supplied by
+     * Oak". Use {@link #isCommitAttested(AuditEvent)} for that test rather
+     * than reading the keys directly. Every other entry — including other
+     * {@code oak.commit.*}-prefixed keys — is forwarded verbatim from the
+     * caller-supplied payload on both paths and is untrusted: anchor trust
+     * on the three reserved keys, never on the prefix in general.
      * <p>
      * Boundaries of the attestation:
      * <ul>
@@ -183,28 +207,43 @@ public interface AuditEvent {
      * for the producer-side responsibility under the open trust model.
      */
     @NotNull
-    static AuditEvent of(@NotNull String domain,
-                         @NotNull String type,
+    static AuditEvent of(@NotNull AuditDomain domain,
+                         @NotNull AuditType type,
                          @NotNull Map<String, Object> payload) {
-        if (domain.isBlank()) {
-            throw new IllegalArgumentException("domain must not be blank");
-        }
-        if (type.isBlank()) {
-            throw new IllegalArgumentException("type must not be blank");
-        }
         return new AuditEventImpl(domain, type, System.currentTimeMillis(), Map.copyOf(payload));
     }
 
     /**
      * Convenience overload for events with no payload.
      *
-     * @param domain non-blank domain identifier.
-     * @param type   non-blank event type identifier within {@code domain}.
+     * @param domain the event domain.
+     * @param type   the event type within {@code domain}.
      * @return non-null event instance with an empty payload.
-     * @throws IllegalArgumentException if {@code domain} or {@code type} is blank.
      */
     @NotNull
-    static AuditEvent of(@NotNull String domain, @NotNull String type) {
+    static AuditEvent of(@NotNull AuditDomain domain, @NotNull AuditType type) {
         return of(domain, type, Map.of());
+    }
+
+    /**
+     * Returns {@code true} when the event carries Oak's commit attestation,
+     * i.e. when it reached the listener through the commit-attached path
+     * after a successful {@code Root.commit()}.
+     * <p>
+     * This is the supported way to make that distinction. The attestation
+     * is carried by three reserved payload keys, but their names are an
+     * implementation detail: read them directly and the check silently
+     * breaks if they are ever renamed. The semantics — what the attestation
+     * does and does not guarantee — are documented on {@link #getPayload()}.
+     *
+     * @param event the event to test, non-null.
+     * @return {@code true} when all three reserved attestation entries are
+     *         present.
+     */
+    static boolean isCommitAttested(@NotNull AuditEvent event) {
+        Map<String, Object> payload = event.getPayload();
+        return payload.containsKey(COMMIT_SESSION_ID)
+                && payload.containsKey(COMMIT_USER_ID)
+                && payload.containsKey(COMMIT_TIMESTAMP);
     }
 }
--- a/oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEvent.java
+++ b/oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEvent.java
@@ -1,0 +1,210 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.jackrabbit.oak.spi.audit;
+
+import java.util.Collections;
+import java.util.Map;
+
+import org.jetbrains.annotations.NotNull;
+import org.osgi.annotation.versioning.ProviderType;
+
+/**
+ * Structured audit event. Implementations are expected to be immutable
+ * value types.
+ * <p>
+ * Events may originate from two pipelines:
+ * <ul>
+ *   <li>Oak-internal capture sites tied to a successful
+ *       {@code Root.commit()}. The commit-attached drain step decorates
+ *       payload with {@code commit.sessionId}, {@code commit.userId},
+ *       and {@code commit.timestamp} entries before dispatch.</li>
+ *   <li>Any bundle calling {@link AuditEventEmitter#emit(AuditEvent)}.
+ *       Such events carry the payload provided by the caller, except that
+ *       Oak strips caller-supplied values for the three reserved
+ *       {@code commit.*} attestation keys before dispatch — see the trust
+ *       contract on {@link #getPayload()}.</li>
+ * </ul>
+ * The {@link #getDomain()} value selects the listeners that receive this
+ * event.
+ * <p>
+ * Most callers do not implement this interface directly: use the static
+ * factory {@link #of(String, String, Map)} (or the no-payload overload
+ * {@link #of(String, String)}) to construct an immutable event with the
+ * current wall-clock timestamp. The package-private {@code AuditEventImpl}
+ * backs these factories.
+ */
+@ProviderType
+public interface AuditEvent {
+
+    /**
+     * Returns the domain that owns this event. Listeners are domain-scoped:
+     * an {@link AuditEventListener} only receives events whose
+     * {@code getDomain()} matches its own {@link AuditEventListener#getDomain()}.
+     *
+     * @return non-null domain name (e.g. {@code "oak.security"}, {@code "aem.content"}).
+     */
+    @NotNull
+    String getDomain();
+
+    /**
+     * Returns the event type identifier within the domain. Type strings are
+     * stable across releases for any given domain.
+     *
+     * @return non-null type identifier.
+     */
+    @NotNull
+    String getType();
+
+    /**
+     * Returns the wall-clock timestamp (millis since epoch) captured at
+     * the API call site when the event was constructed — i.e. the
+     * <em>capture</em> timestamp.
+     * <p>
+     * For commit-attached events this can differ from the
+     * <em>commit</em> timestamp ({@code commit.timestamp} in
+     * {@link #getPayload()}): the capture timestamp is taken when the
+     * Oak API call ran; the commit timestamp is taken when the surrounding
+     * {@code Root.commit()} actually merged. The two can diverge when the
+     * surrounding operation takes a long time between capture and commit.
+     * Listeners that want "when did the change become visible?" should
+     * read {@code commit.timestamp}; listeners that want "when did the
+     * API call ran?" should read this value.
+     *
+     * @return event capture timestamp in milliseconds since epoch.
+     */
+    long getTimestamp();
+
+    /**
+     * Returns the structured payload for this event. The default
+     * implementation returns an empty map; concrete event types override
+     * this to expose typed accessors and include their fields here.
+     * <p>
+     * For commit-attached events, Oak's drain path adds entries with the
+     * keys {@code commit.sessionId}, {@code commit.userId}, and
+     * {@code commit.timestamp} when the buffer is drained on commit
+     * success. Oak does not <em>add</em> these entries on the fire-and-forget
+     * path (see the trust contract below).
+     * <p>
+     * <strong>Trust contract</strong> (normative — other audit SPI and
+     * implementation docs defer to this paragraph). For events delivered
+     * through Oak dispatch, the three reserved keys {@code commit.sessionId},
+     * {@code commit.userId} and {@code commit.timestamp} are Oak-attested:
+     * <ul>
+     *   <li>On the <em>commit-attached</em> path Oak <em>unconditionally
+     *       overwrites</em> the three keys with the values from
+     *       {@code CommitInfo} at drain time (via
+     *       {@code CommitMetadataDecorator}).</li>
+     *   <li>On the <em>fire-and-forget</em> path
+     *       ({@link AuditEventEmitter#emit(AuditEvent)} /
+     *       {@code AuditEvents.dispatch}) Oak <em>strips</em> caller-supplied
+     *       values for the same three keys before delivery.</li>
+     * </ul>
+     * A listener may therefore treat the presence of any of the three keys
+     * in a dispatched payload as "commit-attached event, values supplied by
+     * Oak". Every other entry — including other {@code commit.*}-prefixed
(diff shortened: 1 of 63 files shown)


dulvac:
implements https://github.com/apache/jackrabbit-oak/pull/3058

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEvent.java
```diff
@@ -0,0 +1,210 @@
...
+     * Oak API call ran; the commit timestamp is taken when the surrounding
+     * {@code Root.commit()} actually merged. The two can diverge when the
+     * surrounding operation takes a long time between capture and commit.
+     * Listeners that want "when did the change become visible?" should
+     * read {@code commit.timestamp}; listeners that want "when did the
+     * API call ran?" should read this value.
+     *
+     * @return event capture timestamp in milliseconds since epoch.
+     */
+    long getTimestamp();
```
joerghoh, line 112: Why not using an ``Instant``?

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEventEmitter.java
```diff
@@ -0,0 +1,76 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.osgi.annotation.versioning.ProviderType;
+
+/**
+ * OSGi service for emitting audit events. Consumed via {@code @Reference}:
+ * <pre>{@code
+ * @Reference private AuditEventEmitter audit;
+ *
+ * void onSomething() {
+ *     if (audit.isEnabledFor("aem.content")) {
```
joerghoh: I would remove the term "aem" in this context, even if it's just javadoc.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEventEmitter.java
```diff
@@ -0,0 +1,76 @@
...
+ *         audit.emit(new MyEvent(...));
+ *     }
+ * }
+ * }</pre>
+ * <p>
+ * The emitter dispatches synchronously on the calling thread to all
+ * listeners registered for the event's domain. Not tied to any commit;
+ * not buffered; not rolled back on failure.
+ * <p>
+ * Listeners are invoked under per-listener try/catch isolation — covering
```
joerghoh: Is this the mandated behavior for classes implementhis interface? If yes, the javadoc should state that this as required implementation details.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEventImpl.java
```diff
@@ -0,0 +1,76 @@
...
+ * limitations under the License.
+ */
+package org.apache.jackrabbit.oak.spi.audit;
+
+import java.util.Map;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Package-private immutable {@link AuditEvent} backing the
```
joerghoh: If this not part of the official API contract, you should not place into the same package as API classes, but rather store it a package which is not exported.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEventImpl.java
```diff
@@ -0,0 +1,76 @@
...
+final class AuditEventImpl implements AuditEvent {
+
+    private final String domain;
+    private final String type;
+    private final long timestamp;
+    private final Map<String, Object> payload;
+
+    AuditEventImpl(@NotNull String domain,
+                   @NotNull String type,
+                   long timestamp,
```
joerghoh: as above: Why not an ``Instant``?

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEventImpl.java
```diff
@@ -0,0 +1,76 @@
...
+ * stores it by reference.
+ */
+final class AuditEventImpl implements AuditEvent {
+
+    private final String domain;
+    private final String type;
+    private final long timestamp;
+    private final Map<String, Object> payload;
+
+    AuditEventImpl(@NotNull String domain,
```
joerghoh: I am unsure of we should use a dedicated ``ÀuditDomain`` type for this (even if it's implemented as a simple String in the first place); that would allow us to apply some constraints to it (for example, it must not contain any ``/`` character, which would make it easy to use this as a node name when persisting such things in the JCR itself).

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEventImpl.java
```diff
@@ -0,0 +1,76 @@
...
+                   @NotNull String type,
```
joerghoh: Same here: Maybe an dedicated ``AuditType`` type?

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditBuffer.java
```diff
@@ -0,0 +1,200 @@
...
+        drain(sessionId);
+    }
+
+    /**
+     * Per-session staging holder: the captured events plus a one-shot
+     * flag that ensures the overflow WARN is logged at most once per
+     * session slot (re-armed when the slot is recreated after a drain).
+     */
+    private static final class SessionBuffer {
+        final List<AuditEvent> events = new ArrayList<>(4);
```
joerghoh: Nit: really an ArrayList? Re-allocating that that array if the capacity must be increased (and 4 is smaller than the default of 10) can be expensive.

joerghoh (commented):
A first brief review, will continue tomorrow.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditConfigurationImpl.java
```diff
@@ -0,0 +1,493 @@
...
+ * via {@code SecurityProvider.getConfiguration(AuditConfiguration.class)}.
+ * Embedded callers obtain the drain observer via {@link #getDrainObserver()}.
+ * <p>
+ * When the feature toggle is disabled, capture is a no-op and the observer
+ * short-circuits — see {@link AuditEvents#isEnabled()} and
+ * {@link AuditDrainObserver#contentChanged}.
+ */
+@Component(service = AuditConfiguration.class)
+@Designate(ocd = AuditConfigurationImpl.Configuration.class)
+public class AuditConfigurationImpl implements AuditConfiguration {
```
joerghoh: In my opinion the class name should be different; its main purpose is not to be the implementation class of an AuditConfiguration interface, but rather the implementation of a pipeline (see the javadoc above).

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditConfigurationImpl.java
```diff
@@ -0,0 +1,493 @@
...
+
+    @ObjectClassDefinition(name = "Apache Jackrabbit Oak AuditConfiguration",
+            description = "Audit event pipeline. Capture and dispatch are " +
+                    "gated by the '" + FEATURE_TOGGLE_NAME + "' feature toggle " +
+                    "(disabled by default).")
+    @interface Configuration {
+        // Configuration is currently empty by design: capture/dispatch behavior
+        // is controlled exclusively by the feature toggle. Listeners are
+        // contributed via OSGi services / the Whiteboard.
+    }
```
joerghoh: This is redundant, and you can always add such a configuration later if you really need it. I would remove it.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditPipeline.java
```diff
@@ -0,0 +1,493 @@
...
+     * NOOP sink (no buffer write) rather than through a live {@code BufferSink}
+     * with an orphaned lifecycle handle. The inverse ordering would minimize
+     * lifecycle bypass but maximize silent capture loss; we prefer the former.
+     *
+     * @param whiteboard the whiteboard to register the {@code Feature}
+     *                   toggle and {@code AuditEventListener} tracker on;
+     *                   non-null.
+     */
+    public void initialize(@NotNull Whiteboard whiteboard) {
+        featureToggle = Feature.newFeature(FEATURE_TOGGLE_NAME, whiteboard);
```
joerghoh, line 210: Above it's stated that this feature is disabled by default. Where exactly is this default value set? if it's not explicitly set (but the default behavior of a ``Feature``) please add a brief comment here.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditConfigurationImpl.java
```diff
@@ -0,0 +1,493 @@
...
+
+        AuditEvents.install(new BufferSink(featureToggle, registry, buffer));
+
+        // Construct the singleton AuditDrainObserver LAST in initialize().
+        // This closes a potential TOCTOU window in the OSGi @Activate flow:
+        // ObserverTracker (which subscribes to Observer services on a
+        // background thread) might fire the bootstrap CommitInfo.EMPTY_EXTERNAL
+        // invocation before @Activate returns. With the singleton constructed
+        // here (inside initialize, which @Activate calls FIRST), getDrainObserver()
+        // is safe to call from anywhere in @Activate after this point.
```
joerghoh: Is this deep explanation really necessary here?

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditConfigurationImpl.java
```diff
@@ -0,0 +1,493 @@
...
+        drainObserver = new AuditDrainObserver(featureToggle, buffer, registry);
+
+        log.info("Audit pipeline activated. Toggle '{}' = {}.",
+                FEATURE_TOGGLE_NAME, featureToggle.isEnabled());
```
joerghoh: That's a bit confusing here. If the feature toggle is disabled, does an activated pipeline even matter?

In other words: Instead of just printing ``true`` or ``false`` we should be more explicit what this enabled/disabled feature toggle means, especially if the audit pipeline will always be active.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditConfigurationImpl.java
```diff
@@ -0,0 +1,493 @@
...
+    public @NotNull Observer getDrainObserver() {
+        if (drainObserver == null) {
+            throw new IllegalStateException(
+                    "AuditConfigurationImpl.initialize(...) must be called first" +
+                            " (or dispose() has already run)");
+        }
+        return drainObserver;
+    }
+
+    // Package-private (was private) so the test suite can invoke the
```
joerghoh:
```suggestion
    // Package-private so the test suite can invoke the
```

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/CommitMetadataDecorator.java
```diff
@@ -0,0 +1,217 @@
...
+ * not the decorator's. Adding per-entry null checks here would impose
+ * hot-path cost for what is an SPI-contract violation.
+ */
+final class CommitMetadataDecorator {
+
+    private static final Logger log = LoggerFactory.getLogger(CommitMetadataDecorator.class);
+
+    static final String KEY_SESSION_ID = "commit.sessionId";
+    static final String KEY_USER_ID = "commit.userId";
+    static final String KEY_TIMESTAMP = "commit.timestamp";
```
joerghoh: would it make sense to prefix these values with something like ``oak.``? ``commit`` feels a bit generic and not necessarily tied to to oak.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserManagerImpl.java
```diff
@@ -407,6 +422,68 @@ void onGroupUpdate(@NotNull Group group, boolean isRemove, boolean isContentId,
...
+            String memberId = member.getID();
+            String memberPath = member.getPath();
+            AuditEvents.record(root, isRemove
+                    ? UserAuditEvents.memberRemoved(groupPath, memberId, memberPath)
+                    : UserAuditEvents.memberAdded(groupPath, memberId, memberPath));
+        } catch (RepositoryException e) {
+            // Path resolution failed — drop the event rather than fail the
+            // surrounding group update. It is an audit-completeness signal:
+            // a successful membership change produced no audit event.
+            warnAuditPathResolutionFailed("failed to resolve path for group membership update", e);
```
joerghoh:
```suggestion
            warnAuditPathResolutionFailed("failed to resolve path for group membership update, not recording the audit event", e);
```

oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/UserManagerImpl.java
```diff
@@ -407,6 +422,68 @@ void onGroupUpdate(@NotNull Group group, boolean isRemove, boolean isContentId,
...
+        if (memberIds.isEmpty()) {
+            return;
+        }
+        try {
+            String groupPath = group.getPath();
+            AuditEvents.record(root, isRemove
+                    ? UserAuditEvents.membersRemovedBulk(groupPath, memberIds, isContentId, failedIds)
+                    : UserAuditEvents.membersAddedBulk(groupPath, memberIds, isContentId, failedIds));
+        } catch (RepositoryException e) {
+            warnAuditPathResolutionFailed("failed to resolve group path for bulk membership update", e);
```
joerghoh:
```suggestion
            warnAuditPathResolutionFailed("failed to resolve group path for bulk membership update, not recording audit event", e);
```

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEvent.java
dulvac: I used long to match Oak's existing convention. `CommitInfo.getInfo()` timestamps and the `commit.timestamp` payload key are both epoch millis. If you prefer `Instant` on the interface I'll switch it and convert once at the decorator.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEventEmitter.java
dulvac: agree. will fix.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEventEmitter.java
dulvac: The guarantee is in oak. `@ProviderType` means Oak supplies the emitter, and the `try/catch` lives in our dispatch code, so a consumer can't opt out. I'll reword the javadoc as a guarantee the caller can rely on rather than a description of behaviour.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEventImpl.java
dulvac: same comment.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEventImpl.java
dulvac: It's already package-private. final class `AuditEventImpl`, not public. So it never reaches the exported surface even though it sits in an exported package. Consumers only ever see the `AuditEvent` interface, and they get instances from `AuditEvent.of(...)`. 

It has to be in this package precisely because that factory is on the interface. If we move it to a different package, it would have to become public. wdyt?

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditBuffer.java
dulvac: The common case is that there's one or two events per oak session. Would you prefer I leave the default of 10?

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditConfigurationImpl.java
dulvac: Makes sense. I'll rename it to `AuditPipeline`

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditConfigurationImpl.java
dulvac: will remove, thanks.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditPipeline.java
dulvac: It's the `Feature default`, not set explicitly — `Feature.newFeature` creates an AtomicBoolean with no initial value, so it starts as `false`. Nothing in the audit code sets it. Will add a comment.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditConfigurationImpl.java
dulvac: yeah, too much. I'll cut it to a sentence saying the observer must exist before `@Activate` publishes the service

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditConfigurationImpl.java
dulvac: I think you've caught a genuine inconsistency, noy just a confusing message. The pipeline is always wired at activation; the toggle only affectscapture and dispatch. So "activated" and the toggle state are two different things, and printing the toggle as a bare true/false right after "activated" reads as contradictory. I'll make the message say explicitly whether events will be captured, and fix the wording with `isActive()`

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/CommitMetadataDecorator.java
dulvac: makes sense. I'll rename to `oak.commit.*` unless you'd prefer a different prefix.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEventImpl.java
joerghoh: I am not sure how OSGI plays along with Java visibility, but I would move that class into a dedicated impl package (in the same bundle). IIRC BND calculates the package version based on all changes in all classes in that package. In that case a change in the ``AuditEventImpl`` could cause a package version bump even if the public API remains unchanged.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditBuffer.java
joerghoh: yes, I would leave it with the default.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditBuffer.java
```diff
@@ -0,0 +1,200 @@
...
+                        "This usually indicates a very large transaction or a session that records " +
+                        "audit events without committing.", sessionId, MAX_EVENTS_PER_SESSION);
+            }
+            return;
+        }
+        sb.events.add(event);
+    }
+
+    /**
+     * Test-only inspector. Returns a <strong>defensive copy</strong> of the
```
joerghoh: What is a ``defensive copy``? I think that you provide a ``deep copy`` in this case.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditBuffer.java
```diff
@@ -0,0 +1,200 @@
...
+     * events staged for {@code sessionId} <strong>without</strong> removing
+     * them. Mutating the returned list does not affect the buffer. Production
+     * drain goes through {@link #drain(String)}.
+     *
+     * @param sessionId session id, non-null.
+     * @return an immutable copy of the staged events, or {@code null} when
+     *         nothing was staged for the session on the current thread.
+     */
+    @Nullable
```
joerghoh: Why ``Nullable``? This would a perfect case to return an empty list; which eliminates the need for explicit null checks.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditBuffer.java
```diff
@@ -0,0 +1,200 @@
...
+
+    /**
+     * Detaches and returns the staged events for {@code sessionId},
+     * leaving the buffer empty for that session.
+     *
+     * @param sessionId session id, non-null.
+     * @return the staged events, or {@code null} when nothing was
+     *         staged for the session on the current thread.
+     */
+    @Nullable
```
joerghoh: Same as above: why ``Nullable``?

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEventImpl.java
joerghoh: The more I think about this, the more sense does it make to me. It can be a very simple wrapper around a String plus a static Builder (including any validation) for it.

dulvac:
@joerghoh thanks for the review. I addressed your feedback, minus two open questions - the package scope and arraylist default size.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEvents.java
```diff
@@ -0,0 +1,115 @@
...
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Static façade used by Oak-internal code (commit-attached capture sites,
+ * the OSGi {@link AuditEventEmitter} impl) to talk to the audit pipeline.
+ * The façade is wired to a {@link Sink} on activation of the audit module;
+ * when no module is deployed (or the feature toggle is off) it short-circuits
+ * with zero allocation.
+ */
+public final class AuditEvents {
```
joerghoh: Note the class ``AuditEvent`` in the same package which the singular version of this class name. 

I would rename ``AuditEvents`` to something else, because in my understanding it has nothing to do with multiple ``AuditEvent`` instances, which the name could suggest.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEventImpl.java
dulvac: You're right, and my earlier answer missed the point. Package-private keeps consumers out, but BND versions the whole package, so editing the class could bump the exported version with no API change. Moved it to `spi.audit.impl` and left that package out of `Export-Package`. It has to be public now, because `AuditEvent.of` in the parent package constructs it. Checked the generated manifest: still only `spi.audit` exported.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditBuffer.java
dulvac: Done, using the default.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditBuffer.java
dulvac: Reworded, though not to "deep copy". `List.copyOf` returns a new list holding the same `AuditEvent` references, so it is a shallow copy. It now says "unmodifiable shallow copy" and notes that sharing the instances is safe because events are immutable.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditBuffer.java
dulvac: Agreed. Both `peek` and `drain` return `List.of()` now. The only production caller had a redundant `events == null ||` in front of its `isEmpty()`, and that is gone.

oak-core/src/main/java/org/apache/jackrabbit/oak/security/audit/AuditBuffer.java
dulvac: Covered with `peek` above.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEventImpl.java
dulvac: Landed in c955987672, in the shape you describe: a final wrapper over a String with validation in a static `of(...)`. It rejects blank names, colons, whitespace, and anything `JcrNameParser` refuses, so a domain is safe to use as a node name.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/spi/audit/AuditEvents.java
dulvac: Agreed, the plural reads as a factory for `AuditEvent` and it is not one. Renamed to `AuditDispatch`, since installing the sink and routing is what the class does. Happy to use a different name if you have a preference.

dulvac:
Second round is in.

- `AuditEventImpl` moved to `spi.audit.impl`, which is not exported.
- `AuditEvents` renamed to `AuditDispatch`.
- `AuditBuffer.peek` and `drain` return empty lists instead of null.
- `SessionBuffer` uses the default ArrayList capacity.
- The "defensive copy" wording is fixed, though to shallow rather than deep. Details in the thread.

`AuditDomain` and `AuditType` went in earlier, in c955987672.

2 things left so far, @joerghoh: whether you want `Instant` instead of `long` on `getTimestamp()`. And an ack if you agree with the  `AuditDispatch` naming.

joerghoh:
@dulvac if Oak prefers ``long`` for timestamps, then let's use that. +1 to AuditDispatch

PR #3064, opened 2026-08-05
OAK-12336 : update Postgres database server to 14

rishabhdaim:


rishabhdaim:
ITs have passed locally.

cc @reschke

rishabhdaim:
## Review

**P2** · `oak-doc/src/site/markdown/testing.md:68` · Testing doc's PostgreSQL Docker example still used `postgres:13-alpine`, below the new minimum.

`RDBDocumentStoreDB.POSTGRES.checkVersion` now requires PostgreSQL >= 14.0 (this PR). A developer following the doc's Docker command to stand up a local test database would pull `postgres:13-alpine`, below the new minimum, and hit the unsupported-version diagnostic at connection time.

Fixed in c6f80699: bumped the Docker image tag/prose to `postgres:14-alpine`, and added `RDBJDBCToolsTest` covering `RDBJDBCTools.versionCheck` and `RDBDocumentStoreDB.POSTGRES.checkVersion` (including a regression guard for the 13→14 threshold, which had no prior test coverage).

No correctness issues found in the version-check change itself — it only builds a logged diagnostic string and never throws or blocks the connection.

rishabhdaim:
Two more findings surfaced by a second, independent PR analysis were verified and addressed in 0d883791:

- `RDBBlobStoreDB.POSTGRES.checkVersion` still required only PostgreSQL >= 9.5 while this PR raised the document store minimum to 14.0 — bumped the blob store minimum to 14.0 as well for consistency (driver-version threshold left unchanged).
- `RDBDocumentStore.java` logs `versionDiags` at `LOG.error` on every startup when the check fails. Raising the minimum here means PostgreSQL 13.x users (previously fine) will now hit that ERROR-level log on every init — advisory only, no functional break, but can trigger monitoring alerts. Added a code comment documenting this so it is not mistaken for a new bug when users start reporting it.

reschke (commented):
I believe the new tests are really a distraction. RDB support is dormant.

reschke (approved):

PR #3068, opened 2026-08-07
OAK-12340 Simplified index management improvements

thomasmueller:


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
```diff
@@ -278,6 +336,15 @@ private static void extractExistingMergedIndexes(JsonObject indexDefs, HashMap<S
...
                 continue;
             }
+            String type = JsonNodeUpdater.oakStringValue(value, "type");
+            if (type != null && "disabled".equals(type)) {
+                // ignore disabled indexes:
+                // only if the version that is in use contains the "mergeInfo" property,
+                // then simplified index management is currently used
+                // (otherwise we create a new merged info even for environments that no longer use it,
+                // if there are indexes that are not yet deleted)
+                continue;
```
bhabegger, line 346: The comment is a bit opaque to me. 
- mergeInfo presence on a an index contains -custom- in it's name indicates a version comming from diff.json in a previous deploy. 
- here we are in the presence of an index definition that was previously stored as disabled

So if I understand correctly here, we "pretend" it didn't exist in the previous merge... ?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
```diff
@@ -424,7 +490,10 @@ public boolean processMerge(String indexName, JsonObject indexDiff, JsonObject n
         } else {
             String latestMergeChecksum = JsonNodeUpdater.oakStringValue(latestIndexVersion, "mergeChecksum");
             JsonObject latestDef = cleanedAndNormalized(switchToLuceneIfNeeded(latestIndexVersion));
-            if (isSameIgnorePropertyOrder(mergedDef, latestDef)) {
+            if (latestMergeChecksum == null && latestCustomized != null) {
+                log("mergeChecksum is missing in the old customized version, so this is a migration: {}", indexName);
+                // continue
```
bhabegger, line 495: Any reason to keep the commented `// continue` ?

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMergerTest.java
```diff
@@ -0,0 +1,1405 @@
...
+                        "jcr:primaryType": "nt:unstructured",
+                        "type": "lucene", "includedPaths": "/a", "queryPaths": "/a"
+                    }
+                }
+                """, true);
+
+        getMerger().merge(newImageLuceneDefinitions, repositoryDefinitions, null);
+        assertEquals("""
+                {
+                  "/oak:index/damAssetLucene-12-custom-2": {
```
bhabegger, line 142: Why do we then keep any custom at all ? Shouldn't we be moving back to just the OOTB ?

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMergerTest.java
```diff
@@ -0,0 +1,1405 @@
...
+import org.apache.jackrabbit.oak.api.CommitFailedException;
+import org.apache.jackrabbit.oak.commons.json.JsonObject;
+import org.apache.jackrabbit.oak.plugins.memory.MemoryNodeStore;
+import org.apache.jackrabbit.oak.spi.commit.CommitInfo;
+import org.apache.jackrabbit.oak.spi.commit.EmptyHook;
+import org.apache.jackrabbit.oak.spi.state.NodeBuilder;
+import org.apache.jackrabbit.oak.spi.state.NodeStore;
+import org.junit.Test;
+
+public class DiffIndexMergerTest {
```
bhabegger, line 34: I like these tests :)

These allow for a clear understanding of what is expected to go on !

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMergerTest.java
```diff
@@ -0,0 +1,1405 @@
...
+                          }
+                        }
+                      }
+                    }
+                  }
+                }""", newImageLuceneDefinitions.toString());
+    }
+
+    @Test
+    public void firstCustomizationAsJson() {
```
bhabegger: firstOptimizerCustomization ?

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMergerTest.java
```diff
@@ -0,0 +1,1405 @@
...
+                        "properties": {
+                          "x": {
+                            "name": "x",
+                            "propertyIndex": true
+                          }
+                        }
+                      }
+                    }
+                  },
+                  "/oak:index/diff.index": {
```
bhabegger, line 426: In the previous tests, the diff.index doesn't appear in the newImageLuceneDefinitions. Why would this case be different ?

bhabegger (approved):
No blockers ;) Mostly curiosity

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
thomasmueller: Yes, in a way... disabled indexes (type=disabled) might as well not exist, in relation to "merging" indexes.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMerger.java
thomasmueller: What I mean is that we do not return from the method in this case (no "return false" or "return true"). I want to explicitly mention this.

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMergerTest.java
thomasmueller: Because it is already stored in the repo (not part of newImageLuceneDefinitions).

But good point, I'll add comments

oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/diff/DiffIndexMergerTest.java
thomasmueller: // we can not just move back to /oak:index/damAssetLucene-12, because
        // that index has a lower version and so is then not use for queries.
        // we need to explicitly create a new version.
        // once damAssetLucene-13 is rolled out, then no "-custom-" is created however

PR #3069, opened 2026-08-07
OAK-12341 adding cache for service lookup

pat-lego:
The use of a service cache will prevent the need to perform service tracker requests lowering the chances of a blocking permission lookup

### DIFF
--- a/oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/SyncConfigTrackerConcurrencyTest.java
+++ b/oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/SyncConfigTrackerConcurrencyTest.java
@@ -16,33 +16,44 @@
  */
 package org.apache.jackrabbit.oak.spi.security.authentication.external.impl.principal;
 
+import org.apache.jackrabbit.api.security.user.UserManager;
+import org.apache.jackrabbit.oak.spi.security.authentication.external.ExternalIdentityProvider;
+import org.apache.jackrabbit.oak.spi.security.authentication.external.SyncContext;
 import org.apache.jackrabbit.oak.spi.security.authentication.external.SyncHandler;
+import org.apache.jackrabbit.oak.spi.security.authentication.external.SyncedIdentity;
 import org.apache.sling.testing.mock.osgi.junit.OsgiContext;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import javax.jcr.RepositoryException;
+import javax.jcr.ValueFactory;
 import java.lang.management.ManagementFactory;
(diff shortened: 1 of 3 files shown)


anchela:
hi @pat-lego , thanks a lot for the contribution. would you be able to also add one or multiple test cases that illustrates the issue? that's needed for us to merge the PR.

Amoratinos (approved):
Looks good to me 👍 

Regarding the test it helps to understand the topic but I do not see too much value to add it to the test set as it will add more delay on test execution and not sure if it could be a bit flaky depending on the environment is executed. wdyt @anchela

jsedding:
For the stacktrace shared in the JIRA ticket, I believe it would be sufficient to change the `isEnabled()` method to the following. The rationale being that `ServiceTracker.getServiceReference()` reads the cached value from a volatile field, i.e. no synchronization overhead.

```
    boolean isEnabled() {
        return getServiceReference() != null;
    }
```

Now, the question is whether we see contention with different stack traces as well.

jsedding:
Note, PR #3072 proposes the alternative fix mentioned above.

jsedding:
I have closed #3072 because it looks like we need to speed up not only the `isEnabled()` code-path, but also the retrieval of references.

@pat-lego could you consider taking e7ab1d6105fcff155ce0b88ef30d646dac649d69 from my PR over? It would speed up the unit-tests. Feel free to adjust, based on @rishabhdaim 's comments over there.

pat-lego:
@jsedding  added the test with 50 ms polling let me know if there are any other points on this if not perhaps we can move towards resolving this issue.

oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/SyncConfigTrackerConcurrencyTest.java
```diff
@@ -123,51 +138,56 @@ public void concurrentGetServiceReferencesBlocksReaders() throws Exception {
...
+        assumeTrue("JVM supporting thread contention monitoring required",
+                threadMXBean.isThreadContentionMonitoringSupported());
+        threadMXBean.setThreadContentionMonitoringEnabled(true);
+
         ExecutorService pool = Executors.newFixedThreadPool(THREAD_COUNT);
-        CountDownLatch started = new CountDownLatch(THREAD_COUNT);
+        CountDownLatch allThreadsStarted = new CountDownLatch(THREAD_COUNT);
         AtomicBoolean stop = new AtomicBoolean(false);
-        List<Long> threadIds = new CopyOnWriteArrayList<>();
+        long[] threadIds = new long[THREAD_COUNT];
```
jsedding, line 163: While it is safe  to _write_ to the array from multiple threads, as long as each thread only writes to a given array index (what you are doing here), there is no guarantee that these updates are visible to other threads. I.e. the call to `threadMXBean.getThreadInfo(threadIds)` _may_ not see the thread ID written into the array (and would presumably see `null`).

See https://stackoverflow.com/a/8978397

jsedding (approved):
Looks good to me. Please check if you want to improve on the (possibly theoretical) concurrency issue I mentioned.

jsedding:
@Amoratinos could you merge this PR, please? It's security related, which is more your field. I only reviewed because it's also OSGi related.

PR #3071, opened 2026-08-10
OAK-12290 : run caffeine cache maintenance asynchronously

rishabhdaim:
## Summary

`CacheBuilder` ran all Caffeine cache maintenance inline on the calling thread, holding
Caffeine's eviction lock for the duration - causing lock contention and, when a lock
holder thread died, a JVM wedge (OAK-12290).

The fix dispatches maintenance on a dedicated executor instead, gated by a bug-fix
feature toggle (`FT_OAK-12290`, default on).

## Changes

- `CacheBuilder`/`CacheMaintenanceExecutor` dispatch maintenance off the calling thread,
  with rate-limited logging when the pool falls back to inline execution
  (`CallerRunsPolicy`); `SegmentNodeStoreRegistrar` registers the toggle on the Whiteboard.
- Tests updated/added since eviction callbacks are now asynchronous.

## Test Plan

- [x] Affected module test suites and OSGi baseline check pass

## Links

- JIRA: https://issues.apache.org/jira/browse/OAK-12290
- Related: https://issues.apache.org/jira/browse/OAK-12214, https://issues.apache.org/jira/browse/OAK-12197

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/api/CacheBuilder.java
```diff
@@ -44,6 +49,52 @@
...
+     * <p>
+     * The toggle is registered on the OSGi Whiteboard under {@link #FT_OAK_12290}. Its value is
+     * read when a cache is built, so flipping it only affects caches built afterwards - Oak's
+     * long-lived caches are built during startup and keep the setting they were built with.
+     * <p>
+     * Caches configured with {@link #refreshAfterWrite(Duration)} ignore the toggle and always use
+     * the maintenance executor, because Caffeine shares one executor between maintenance and
+     * refresh and a synchronous refresh would run the loader (potentially a remote call) on the
+     * calling thread.
+     */
```
joerghoh: The comment is way too verbose
```suggestion
     * Whether Caffeine runs cache maintenance (eviction, removal notification, buffer drains)
     * on Oak's maintenance executor instead of the calling thread. Defaults to {@code true}.
     */
```

(mind, that I also removed the reference to the SKYOPS ticket!)

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/api/CacheBuilder.java
```diff
@@ -277,6 +337,12 @@ private Caffeine<K, V> configureCaffeineBuilder() {
         }
         if (evictionListener != null) {
             EvictionListener<? super K, ? super V> listener = evictionListener;
+            // Deliberately removalListener and not evictionListener: Caffeine invokes the latter
+            // inside the map's atomic removal, holding the bin lock for the key. Oak's listeners do
+            // real work there - NodeCache.evicted() enqueues a persistent-cache write under its own
+            // monitor - so that would reintroduce exactly the kind of lock coupling OAK-12290 is
+            // about. removalListener runs on the maintenance executor instead; listeners must
+            // therefore tolerate lagging behind the write that caused the removal.
```
joerghoh: I don't understand this comment at all ... do you want to explain why you use ``removalListener`` instead of ``evictionListeners``? In that case I would appreciate a simpler and more high-level explanation (and more brief would be good as well); the current form just raises much more question marks with me.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/api/CacheBuilder.java
```diff
@@ -296,6 +362,60 @@ private Caffeine<K, V> configureCaffeineBuilder() {
...
+     *       eventually, degrading to the pre-OAK-12290 inline behaviour rather than stalling.</li>
+     *   <li><em>Isolation.</em> The common pool is shared with every {@code parallelStream()} in
+     *       the JVM, including the hosting application's. A saturated common pool would delay
+     *       segment-cache weight accounting and, once Caffeine's write buffer fills, push
+     *       maintenance back onto request threads - reintroducing the very lock contention
+     *       OAK-12290 is about.</li>
+     *   <li><em>Diagnosability.</em> Named daemon threads make cache maintenance identifiable in a
+     *       thread dump, which is how OAK-12290 and SKYOPS-149400 were diagnosed in the first
+     *       place. This also matches Oak's existing convention of never using the common pool
+     *       (see {@code ForkJoinUtils#submitInCustomPool}).</li>
```
joerghoh: This is an explanation I would expect either in Jira or the PR description, but not in the code. For that just remove it.
```suggestion
```

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/api/CacheBuilder.java
```diff
@@ -296,6 +362,60 @@ private Caffeine<K, V> configureCaffeineBuilder() {
...
+        private static final Executor EXECUTOR = newMaintenanceExecutor();
+
+        private static Executor newMaintenanceExecutor() {
+            AtomicInteger threadCounter = new AtomicInteger();
+            ThreadPoolExecutor executor = new ThreadPoolExecutor(
+                    MAINTENANCE_THREADS, MAINTENANCE_THREADS,
+                    60, TimeUnit.SECONDS,
+                    new LinkedBlockingQueue<>(MAINTENANCE_QUEUE_CAPACITY),
+                    runnable -> {
+                        Thread thread = new Thread(runnable,
```
joerghoh: You might want to add to call ``thread.setUncaughtExceptionHandler()`` as well to just log the exception.

joerghoh (changes requested):
We need to check that we don't add too much comments, which are useful only in the context of the current issue, but rarely later.

Also please pay attention that you only use Apache JIRA links, no references to Adobe tickets!

joerghoh:
When the cache maintenance runs in its own thread, what is the impact on the other threads (interactive requests with a focus on low latency) be? 

because this

> so HTTP request threads serving pages ran full cache maintenance mid-request and slowed all concurrent segment-cache operations on the pod.

makes me think, that no matter which thread performs this maintenance, there will always be an impact on other threads which are accessing that cache concurrently.  And that this change will just speed up the (request) thread, which ran all the maintenance and therefor incurred a massive latency on that particular Oak-repository operation.

Is my understanding correct?

rishabhdaim:
@joerghoh, thanks for the review. However, the PR is still not ready for review and has gaps that need to be addressed.


> When the cache maintenance runs in its own thread, what is the impact on the other threads (interactive requests with a focus on low latency) be?
> 
> because this
> 
> > so HTTP request threads serving pages ran full cache maintenance mid-request and slowed all concurrent segment-cache operations on the pod.
> 
> makes me think, that no matter which thread performs this maintenance, there will always be an impact on other threads which are accessing that cache concurrently. And that this change will just speed up the (request) thread, which ran all the maintenance and therefor incurred a massive latency on that particular [Oak-repository](https://issues.apache.org/jira/browse/OAK-repository) operation.
> 
> Is my understanding correct?


**AI (Claude Analysis)**

You're right that the work itself doesn't vanish; it just moves to the 2-4 dedicated maintenance threads. But it's not true that other threads feel the same pain either way.

Caffeine normally grabs its eviction lock with tryLock() — if it's busy, the caller just backs off instantly, no waiting. That lock is meant to be held for microseconds, just long enough to hand the cleanup work off to an executor. The bug here was that with executor(Runnable::run), "handing it off" actually meant running the whole cleanup — including our removal listeners, which do real work like persistent-cache writes — while still holding that lock.

If that drags on and Caffeine's write buffer fills up in the meantime, other writer threads stop being polite about it — they fall back to a blocking lock() call and just wait. That's exactly where the "excessive wait times acquiring the eviction lock" warning comes from: several threads queued up behind one slow one.

So this isn't about shifting cost from one thread to another — it's about no longer holding that lock long enough for a queue of other threads to build up behind it. That's what was actually causing the stalls and the wedge.

One caveat: if the maintenance pool itself gets swamped, we do fall back to running inline again (CallerRunsPolicy) — but only the thread that triggered it pays for that, nobody else gets stuck waiting on it.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/api/CacheBuilder.java
```diff
@@ -255,11 +285,19 @@ private LoadingCache<K, V> buildCaffeine(CacheLoader<K, V> loader) {
...
+        // remote call; running it inline would block every caller thread that triggers a refresh on
+        // that call, which defeats the point of refreshAfterWrite (return the stale value, reload in
+        // the background). So refreshing caches always run on Oak's maintenance executor, regardless
+        // of the toggle - the tradeoff is that a slow reload can occupy one of the pool's threads for
+        // longer, which is preferable to blocking callers. Zero-capacity caches are a "disable
+        // caching" idiom relied upon elsewhere for immediate eviction, so they always run inline -
+        // otherwise a read immediately following a write could still observe the entry before
+        // background maintenance evicts it.
+        boolean zeroCapacity = maximumWeight == 0 || maximumSize == 0;
+        boolean inline = zeroCapacity
```
joerghoh:
```suggestion
        boolean inSameThread = zeroCapacity
```

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/impl/CacheMaintenanceExecutor.java
```diff
@@ -0,0 +1,135 @@
...
+ * notification, buffer drains) on, instead of {@link java.util.concurrent.ForkJoinPool#commonPool()}.
+ */
+public final class CacheMaintenanceExecutor {
+
+    private static final Logger LOG = LoggerFactory.getLogger(CacheMaintenanceExecutor.class);
+
+    private static final Thread.UncaughtExceptionHandler UNCAUGHT_EXCEPTION_HANDLER = (t, e) ->
+            LOG.warn("Uncaught exception in thread {}", t.getName(), e);
+
+    private static final String THREAD_PREFIX = "oak-cache-maintenance-";
```
joerghoh: That feels a bit too generic, maybe more specific?
```suggestion
    private static final String THREAD_PREFIX = "oak-caffeine-maintenance-";
```

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/SegmentId.java
```diff
@@ -221,6 +233,19 @@ void unloaded() {
...
+     * {@code expected}. Use this for asynchronous removal notifications, where {@code expected}
+     * may already have been superseded by a concurrent {@link #loaded(Segment)}.
+     *
+     * @param expected the segment that was removed; the field is left untouched if it no longer
+     *                 holds this value
+     * @return {@code true} if {@code expected} was still memoised and has now been cleared
+     */
+    boolean unloadIfCurrent(@NotNull Segment expected) {
+        return SEGMENT.compareAndSet(this, expected, null);
+    }
```
jsedding: I would suggest to change the name to just `unload` or `compareAndUnload`.

For me, the current name was more confusing than helpful. Now I understand your intention, but it didn't immediately make sense.

```suggestion
    boolean compareAndUnload(@NotNull Segment expected) {
        return SEGMENT.compareAndSet(this, expected, null);
    }
```

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/impl/CacheMaintenanceExecutor.java
```diff
@@ -0,0 +1,135 @@
...
+        private final AtomicLong nextLogNanos = new AtomicLong();
+
+        @Override
+        public void rejectedExecution(@NotNull Runnable task, @NotNull ThreadPoolExecutor executor) {
+            long now = System.nanoTime();
+            long next = nextLogNanos.get();
+            if (now >= next && nextLogNanos.compareAndSet(next, now + LOG_INTERVAL_NANOS)) {
+                LOG.warn("Cache maintenance pool exhausted ({} threads, {}-deep queue full); running "
+                        + "maintenance inline on the calling thread instead. Expected under a burst; "
+                        + "sustained occurrence means the pool is undersized for the load.",
```
joerghoh, line 129: In this particular case I would naively search for a way to configure this limit in the pool size, but it's not possible.

Should we make ``QUEUE_CAPACITY`` configurable? Now?

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/SegmentNodeStoreRegistrar.java
```diff
@@ -282,6 +283,12 @@ private SegmentNodeStore register() throws IOException {
                 new FeatureToggle(SegmentCache.FT_OAK_12214, SegmentCache.FT_OAK_12214_PROPAGATE_L1_HITS_TO_L2_ENABLED),
                 Collections.emptyMap()));
 
+        // OAK-12290: bug-fix toggle (default on) so Caffeine maintenance never runs inline
+        // on request, indexer or writer threads while holding the eviction lock
```
jsedding: remove AI comment

oak-core-spi/src/test/java/org/apache/jackrabbit/oak/cache/impl/CacheBuilderMaintenanceTest.java
```diff
@@ -0,0 +1,369 @@
...
+                })
+                .build();
+
+        cache.put("k1", "v1");
+        cache.put("k2", "v2");
+
+        Assert.assertTrue("size-based eviction was never notified",
+                evicted.await(TIMEOUT_SECONDS, TimeUnit.SECONDS));
+        Assert.assertTrue("maintenance ran on an unexpected thread: " + threadName.get(),
+                threadName.get().startsWith("oak-cache-maintenance-"));
```
joerghoh: Isn't there a constant defined for the thread name prefix?

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/api/Cache.java
```diff
@@ -144,9 +144,10 @@ public interface Cache<K, V> {
     /**
      * Performs any pending maintenance operations needed by the cache.
      *
-     * <p><em>Note: no Oak module currently calls this method; the CacheLIRS
-     * implementation is a no-op. It may be removed from the interface in a
-     * future release if it remains unused.</em></p>
+     * <p>Applies pending evictions to the cache's internal state before returning, but does not
+     * wait for their {@link EvictionListener} callbacks, which still run asynchronously (unless
+     * async cache maintenance is disabled, in which case those callbacks already ran inline
+     * before this method was even called). The CacheLIRS implementation is a no-op.</p>
```
jsedding: This should describe the API contract rather than the various implementations. E.g.

```suggestion
     * <p>Applies pending evictions to the cache's internal state before returning, but
     * {@link EvictionListener} callbacks may run asynchronously. Thus there is no
     * guarantee that {@link EvictionListener} callbacks have been called or have
     * completed when this call returns.
```

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/api/EvictionListener.java
```diff
@@ -22,9 +22,13 @@
...
- * <p>Register via {@link CacheBuilder#evictionListener(EvictionListener)}.
- * The callback is invoked synchronously during cache operations that trigger
- * removal (eviction, invalidation, replacement).</p>
+ * <p>Register via {@link CacheBuilder#evictionListener(EvictionListener)}.</p>
+ *
+ * <p>The callback runs on the cache's maintenance executor, not on the triggering thread, so it
+ * may lag behind and observe a key that was already re-inserted. Use {@link Cache#cleanUp()} to
+ * drain pending callbacks before relying on external accounting they maintain. Exception: with
+ * {@link CacheBuilder#FT_OAK_12290_ASYNC_CACHE_MAINTENANCE_ENABLED} disabled, the callback runs
+ * synchronously on the triggering thread instead, as it always did before that toggle existed.</p>
```
jsedding: Again, I would leave implementation details out of API documentation.

```suggestion
 * <p>The callback <i>may</> run asynchronously. Implementations should expect that they may lag
 * behind and observe a key that was already re-inserted. Use {@link Cache#cleanUp()} to
 * drain pending callbacks before relying on external accounting they maintain.</p>
```

joerghoh (commented):
I don't understand why there were changes necessary in the ``oak-segment-tar`` module. Are these tied to the changes in the cache maintenance?

jsedding (approved):
The implementation makes sense to me, if I understand everything correctly.

At the core, you move `EvictionListener` callbacks from synchronous processing to asynchronous processing.
This could result in `SegmentId#segment` to get set to `null` even in situations where a newly loaded `Segment` has already been set. You guard against this with the atomic call `VarHandle#compareAndSet`.

I would like you to address some naming and javadoc comments, but don't see them as blockers.

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/impl/CacheMaintenanceExecutor.java
rishabhdaim: @joerghoh can we make that a separate ticket ?

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/api/EvictionListener.java
rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/3071/commits/9a664888e298537fe6feefb04fdca234204336a5

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/api/Cache.java
rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/3071/commits/9a664888e298537fe6feefb04fdca234204336a5

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/SegmentNodeStoreRegistrar.java
rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/3071/commits/9a664888e298537fe6feefb04fdca234204336a5

oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/SegmentId.java
rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/3071/commits/9a664888e298537fe6feefb04fdca234204336a5

oak-core-spi/src/test/java/org/apache/jackrabbit/oak/cache/impl/CacheBuilderMaintenanceTest.java
rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/3071/commits/9a664888e298537fe6feefb04fdca234204336a5

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/impl/CacheMaintenanceExecutor.java
rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/3071/commits/9a664888e298537fe6feefb04fdca234204336a5

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/api/CacheBuilder.java
rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/3071/commits/9a664888e298537fe6feefb04fdca234204336a5

rishabhdaim:
> I don't understand why there were changes necessary in the `oak-segment-tar` module. Are these tied to the changes in the cache maintenance?

These are required because the caffeine maintenance will now be done asynchronously, and some behaviours might become flaky.

Fixed such a case (a race condition).

oak-core-spi/src/main/java/org/apache/jackrabbit/oak/cache/impl/CacheMaintenanceExecutor.java
rishabhdaim: https://issues.apache.org/jira/browse/OAK-12363?focusedCommentId=18105178&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-18105178

protogenes:
Unfortunately this change has introduced a split package conflict between `oak-core` and `oak-core-spi` caused by `src/main/java/org/apache/jackrabbit/oak/cache/impl/CacheMaintenanceExecutor.java`

rishabhdaim:
> CacheMaintenanceExecutor

@protogenes Created https://github.com/apache/jackrabbit-oak/pull/3117 to fix this.

PR #3072, opened 2026-08-10
OAK-12341: Thread Contention authentication.external.impl.principal.SyncConfigTracker

jsedding:
I propose this fix as an alternative to PR #3069. I took the unit-test from the original PR, made it fast, and implemented a fix that relies on an internally cached field of `ServiceTracker`.

cc @pat-lego

### DIFF
--- a/oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/SyncConfigTracker.java
+++ b/oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/SyncConfigTracker.java
@@ -41,7 +41,7 @@ import java.util.Set;
  * {@code ServiceTracker} to detect any {@link SyncHandler} that has
  * dynamic membership enabled.
  */
-final class SyncConfigTracker extends ServiceTracker {
+final class SyncConfigTracker extends ServiceTracker<SyncHandler, SyncHandler> {
 
     private static final Logger log = LoggerFactory.getLogger(SyncConfigTracker.class);
 
@@ -56,7 +56,7 @@ final class SyncConfigTracker extends ServiceTracker {
      * @return {@code true} if dynamic membership is enabled for at least one registered sync-handler; {@code false} otherwise.
      */
     boolean isEnabled() {
-        return getReferences().length > 0;
+        return getServiceReference() != null;
     }
 
     /**
@@ -67,7 +67,7 @@ final class SyncConfigTracker extends ServiceTracker {
         if (!isEnabled()) {
             return false;
         }
-        for (ServiceReference ref : getReferences()) {
+        for (ServiceReference<SyncHandler> ref : getReferences()) {
             if (PropertiesUtil.toBoolean(ref.getProperty(DefaultSyncConfigImpl.PARAM_GROUP_DYNAMIC_GROUPS), DefaultSyncConfigImpl.PARAM_GROUP_DYNAMIC_GROUPS_DEFAULT)) {
                 return true;
             }
@@ -90,9 +90,9 @@ final class SyncConfigTracker extends ServiceTracker {
             return Collections.emptySet();
         }
 
-        ServiceReference[] serviceReferences = getServiceReferences();
+        ServiceReference<SyncHandler>[] serviceReferences = getServiceReferences();
         Set<String> idpNames = new HashSet<>(serviceReferences.length);
-        for (ServiceReference ref : serviceReferences) {
+        for (ServiceReference<SyncHandler> ref : serviceReferences) {
             if (PropertiesUtil.toBoolean(ref.getProperty(DefaultSyncConfigImpl.PARAM_GROUP_DYNAMIC_GROUPS), DefaultSyncConfigImpl.PARAM_GROUP_DYNAMIC_GROUPS_DEFAULT)) {
                 String syncHandlerName = PropertiesUtil.toString(ref.getProperty(DefaultSyncConfigImpl.PARAM_NAME), DefaultSyncConfigImpl.PARAM_NAME_DEFAULT);
                 for (String idpName : mappingTracker.getIdpNames(syncHandlerName)) {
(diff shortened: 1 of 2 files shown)


pat-lego:
Wouldnt the issue still be present in this scenario, we really dont want to call getServiceReferences, in this scenario we still call it and thats what leads to the blocking operation. In the case of a large burst of requests this lookup can cause hundreds of threads to become a FIFO like queue.

In my PR what I propose is what we let the OSGi framework push the services to us and we track them avoiding the lookup completely.

jsedding:
@pat-lego I am calling `getServiceReference()` (singular, not plural). This caches the first reference in a [field of the `ServiceTracker` class](https://github.com/osgi/osgi/blob/8c7184dad43779668052afe5b29a2a201e527b52/org.osgi.util.tracker/src/org/osgi/util/tracker/ServiceTracker.java#L561).

BTW, I included your test case and my change passes it. I like how you're checking for contention! 👍

pat-lego:
```
 private ServiceReference<SyncHandler>[] getReferences() {
        ServiceReference<SyncHandler>[] refs = getServiceReferences();
```

Yes sorry I should be more specific, is there ever a reason we should also fix `getReferences` to not call `getServiceReferences` to prevent this condition. I know in the thread dump its `isEnabled` -> `getReferences` which this PR fixes that flow but the question is should we fix `getReferences` as well.

oak-auth-external/src/main/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/SyncConfigTracker.java
```diff
@@ -56,7 +56,7 @@ final class SyncConfigTracker extends ServiceTracker {
      * @return {@code true} if dynamic membership is enabled for at least one registered sync-handler; {@code false} otherwise.
      */
     boolean isEnabled() {
-        return getReferences().length > 0;
+        return getServiceReference() != null;
```
rishabhdaim, line 59: This only makes the `isEnabled()` check itself lock-free. Every caller that actually gets past that check — `AutomembershipService`/`DynamicGroupMembershipService` building an `AutoMembershipProvider`/`ExternalGroupPrincipalProvider` — immediately calls `getAutoMembership()`, `hasDynamicGroupsEnabled()`, `getGroupAutoMembership()`, `getAutoMembershipConfig()`, `getIdpNamesWithDynamicGroups()`, all of which still call `getServiceReferences()` and still synchronize on the same `Tracked` monitor.

So for real requests (anywhere `isEnabled()` is true, which is the whole point of this code existing) the contention from the reported thread dump doesn't actually go away — it's only fixed for the isolated `isEnabled()`-only case this PR's test exercises. This is the same thing @pat-lego asked above : 
>should `getReferences()` get the same treatment?

oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/SyncConfigTrackerConcurrencyTest.java
```diff
@@ -0,0 +1,226 @@
...
+     */
+    @Test
+    public void concurrentGetServiceReferencesBlocksReaders() throws Exception {
+        long blockedMs = sampleForBlockingOnTrackedMonitor(tracker::getServiceReferences);
+        LOG.info("Contended threads were blocked for {}ms", blockedMs);
+        assertTrue("expected concurrent ServiceTracker.getServiceReferences() callers to contend on "
+                + "the shared Tracked monitor, as observed in the OAK-12341 thread dump", blockedMs > 0);
+    }
+
+    /**
```
rishabhdaim, line 132: This javadoc doesn't match what the fix actually does — there's no `CopyOnWriteArrayList` and no `addingService`/`removedService` override anywhere in `SyncConfigTracker`. The real mechanism is the inherited `ServiceTracker.getServiceReference()` (singular), which caches the resolved reference internally and only falls back to the synchronized `getServiceReferences()` scan on a cache miss — like @jsedding described above. Worth fixing so a future reader doesn't go looking for a cache field that isn't there.

oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/SyncConfigTrackerConcurrencyTest.java
```diff
@@ -0,0 +1,226 @@
...
+                        action.run();
+                    });
+                }
+            });
+
+            assertTrue("worker threads did not start in time", allThreadsStarted.await(10, TimeUnit.SECONDS));
+
+            long[] threads = threadIds.stream().mapToLong(Long::longValue).toArray();
+            long blockedMs = 0;
+            for (int i = 0; i < 1000; i++) {
```
rishabhdaim, line 184: This assertion (`blockedMs > 0`) depends on a single spinner thread resubmitting one-shot tasks through a bounded queue, rather than all `THREAD_COUNT` threads hammering concurrently — real overlap only happens once the queue fills, which isn't guaranteed on a fast or CPU-constrained CI box. Might be worth having each worker loop directly instead of resubmitting through `pool.execute()`, and polling against a wall-clock timeout instead of a fixed 1000 iterations, so this doesn't flake.

rishabhdaim:
`commit-check` is actually failing here — not for the "Conventional Commits" reason the bot claims, but because this repo's own regex (`.github/cchk.toml`) requires commit subjects to start with `OAK-<digits>`, and one of these three commits doesn't. Worth rewording that commit before merge.

oak-auth-external/src/test/java/org/apache/jackrabbit/oak/spi/security/authentication/external/impl/principal/SyncConfigTrackerConcurrencyTest.java
jsedding: Polling against a wall clock was in place before. One assertion checks that the threads are not blocked, so it always waits for the wall clock duration. If we can set the wall clock to 50ms or less, that's fine.

jsedding:
Closing in favour of #3069. Apparently, the contention does not only affect the `isEnabled()` code path.

PR #3076, opened 2026-08-12
OAK-12348: Property and node type indexes support costPerEntry/costPerExecution overrides

bhabegger:
## What

`PropertyIndex`/`NodeTypeIndex` (`oak-core`) compute query cost purely from
indexed entry counts plus a fixed overhead of 2, with no way to influence
the estimate from the index definition — unlike Lucene/Elastic indexes
(`oak-search`), which already read `costPerEntry`/`costPerExecution`.

In production this let the `nodeType`/property index win a cost comparison
against a more selective, purpose-built index for the same query, because
the built-in entry-count estimate can be significantly wrong at scale (seen
in three related incidents on `rep:User`/`rep:Group` queries against a
tagged `authorizables` Lucene index). The workaround each time was a
manual `costPerEntry`/`costPerExecution` override — but that only exists on
the Lucene/Elastic side.

## How

`PropertyIndexPlan` and `PropertyIndexLookup` each split into:
- `getCostLegacy(...)` — the original hardcoded formula, unconditionally.
- `getCostConfigurable(...)` — `cost = costPerExecution + costPerEntry * entryCount`,
  both optionally set on the property index definition (new `IndexConstants.COST_PER_ENTRY`/
  `COST_PER_EXECUTION`, same property names `oak-search`'s `FulltextIndexConstants`
  already uses for Lucene/Elastic).
- `getCost(...)` — dispatches between the two based on `FT_OAK-12348`
  (`PropertyIndexLookup.FT_OAK_12348_ENABLE`), **enabled by default**: with
  no properties set, `getCostConfigurable()` reproduces `getCostLegacy()`
  exactly, so this is behavior-preserving for every existing index
  definition — the toggle is an escape hatch, not an opt-in gate.

`NodeTypeIndex` needs no code changes at all — its cost is the sum of two
`PropertyIndexLookup.getCost()` calls (`jcr:primaryType`, `jcr:mixinTypes`),
so it picks up the override transitively (covered by a dedicated test).

`IndexUtils` gains a small `getOptionalValue(NodeState, String, double)`
helper (mirrors `oak-search`'s `IndexDefinition.getOptionalValue`, which
`oak-core` can't depend on directly) instead of duplicating the same
property read in both classes.

Out of scope: `resultCacheSize` (mentioned in the originating request) is a
Lucene/Elastic query-result caching concept with no equivalent in
`PropertyIndex`'s `ContentMirrorStoreStrategy`-based lookup.

## Testing

New/updated tests in `PropertyIndexTest`, `PropertyIndexLookupTest` (unchanged,
still green), and `NodeTypeIndexTest` cover: default behavior unchanged
regardless of toggle position, override taking effect by default, the
toggle correctly falling back to the legacy formula when disabled, the
`costPerEntry=0` infinity-guard (must not become `NaN`), the unique-index
short-circuit surviving an override, and `NodeTypeIndex` picking up the
override with zero code changes in its own package.

Full `oak-core` suite: 4779 tests, 0 failures.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/nodetype/NodeTypeIndex.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/nodetype/NodeTypeIndex.java
@@ -41,13 +41,25 @@ class NodeTypeIndex implements QueryIndex, JcrConstants {
 
     private final MountInfoProvider mountInfoProvider;
 
+    // OAK-12348: false (default) uses the configurable cost formula.
+    private final boolean disableConfigurableCosts;
+
+    // Value returned by getMinimumCost(), precomputed from disableConfigurableCosts.
+    private final double minimumCost;
+
     public NodeTypeIndex(MountInfoProvider mountInfoProvider) {
+        this(mountInfoProvider, false);
+    }
+
+    public NodeTypeIndex(MountInfoProvider mountInfoProvider, boolean disableConfigurableCosts) {
         this.mountInfoProvider = mountInfoProvider;
+        this.disableConfigurableCosts = disableConfigurableCosts;
+        this.minimumCost = disableConfigurableCosts ? NodeTypeIndexLookup.MINIMUM_COST : 0;
     }
 
     @Override
     public double getMinimumCost() {
-        return NodeTypeIndexLookup.MINIMUM_COST;
+        return minimumCost;
     }
 
     @Override
@@ -70,7 +82,7 @@ class NodeTypeIndex implements QueryIndex, JcrConstants {
             return Double.POSITIVE_INFINITY;
         }
         
-        NodeTypeIndexLookup lookup = new NodeTypeIndexLookup(root, mountInfoProvider);
+        NodeTypeIndexLookup lookup = new NodeTypeIndexLookup(root, mountInfoProvider, disableConfigurableCosts);
         if (lookup.isIndexed(filter.getPath(), filter)) {
             return lookup.getCost(filter);
         } else {
@@ -92,7 +104,7 @@ class NodeTypeIndex implements QueryIndex, JcrConstants {
 
     @Override
     public Cursor query(Filter filter, NodeState root) {
-        NodeTypeIndexLookup lookup = new NodeTypeIndexLookup(root, mountInfoProvider);
+        NodeTypeIndexLookup lookup = new NodeTypeIndexLookup(root, mountInfoProvider, disableConfigurableCosts);
         if (!hasNodeTypeRestriction(filter) || !lookup.isIndexed(filter.getPath(), filter)) {
             throw new IllegalStateException(
                     "NodeType index is used even when no index is available for filter " + filter);
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexLookup.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexLookup.java
@@ -35,6 +35,7 @@ import org.apache.jackrabbit.oak.api.Type;
 import org.apache.jackrabbit.oak.commons.PathUtils;
 import org.apache.jackrabbit.oak.commons.collections.IterableUtils;
 import org.apache.jackrabbit.oak.plugins.index.IndexConstants;
+import org.apache.jackrabbit.oak.plugins.index.IndexUtils;
 import org.apache.jackrabbit.oak.plugins.index.property.strategy.IndexStoreStrategy;
 import org.apache.jackrabbit.oak.spi.mount.MountInfoProvider;
 import org.apache.jackrabbit.oak.spi.mount.Mounts;
@@ -79,13 +80,21 @@ public class PropertyIndexLookup {
 
     private final MountInfoProvider mountInfoProvider;
 
+    // OAK-12348: false (default) uses the configurable cost formula.
+    private final boolean disableConfigurableCosts;
+
     public PropertyIndexLookup(NodeState root) {
         this(root, Mounts.defaultMountInfoProvider());
     }
 
     public PropertyIndexLookup(NodeState root, MountInfoProvider mountInfoProvider) {
+        this(root, mountInfoProvider, false);
+    }
+
+    public PropertyIndexLookup(NodeState root, MountInfoProvider mountInfoProvider, boolean disableConfigurableCosts) {
         this.root = root;
         this.mountInfoProvider = mountInfoProvider;
+        this.disableConfigurableCosts = disableConfigurableCosts;
     }
 
     /**
@@ -135,18 +144,63 @@ public class PropertyIndexLookup {
                 definition, INDEX_CONTENT_NODE_NAME);
     }
 
+    /**
+     * Dispatches to {@link #getCostConfigurable} or {@link #getCostLegacy}
+     * based on {@code disableConfigurableCosts}.
+     */
     public double getCost(Filter filter, String propertyName, PropertyValue value) {
+        return disableConfigurableCosts
+                ? getCostLegacy(filter, propertyName, value)
+                : getCostConfigurable(filter, propertyName, value);
+    }
+
+    /**
+     * Original cost formula: {@code COST_OVERHEAD + entryCount}. Ignores
+     * {@code costPerEntry}/{@code costPerExecution} even if set on the index
+     * definition.
+     */
+    public double getCostLegacy(Filter filter, String propertyName, PropertyValue value) {
         NodeState indexMeta = getIndexNode(root, propertyName, filter);
         if (indexMeta == null) {
             return Double.POSITIVE_INFINITY;
         }
-        Set<IndexStoreStrategy> strategies = getStrategies(indexMeta);
-        ValuePattern pattern = new ValuePattern(indexMeta);
-        double cost = strategies.isEmpty() ? MAX_COST : COST_OVERHEAD;
-        for (IndexStoreStrategy s : strategies) {
-            cost += s.count(filter, root, indexMeta, encode(value, pattern), MAX_COST);
+        return getCost(indexMeta, filter, value, COST_OVERHEAD, 1.0);
+    }
+
+    /**
+     * {@code cost = costPerExecution + costPerEntry * entryCount}, both
+     * optionally configured on the index definition (OAK-12348). Defaults
+     * ({@code costPerEntry=1.0}, {@code costPerExecution=COST_OVERHEAD})
+     * reproduce {@link #getCostLegacy} exactly.
+     */
+    public double getCostConfigurable(Filter filter, String propertyName, PropertyValue value) {
+        NodeState indexMeta = getIndexNode(root, propertyName, filter);
+        if (indexMeta == null) {
+            return Double.POSITIVE_INFINITY;
         }
-        return cost;
+        double costPerEntry = IndexUtils.getOptionalValue(indexMeta, IndexConstants.COST_PER_ENTRY, 1.0);
+        double costPerExecution = IndexUtils.getOptionalValue(indexMeta, IndexConstants.COST_PER_EXECUTION, COST_OVERHEAD);
+        return getCost(indexMeta, filter, value, costPerExecution, costPerEntry);
+    }
+
+    // Shared by getCostLegacy/getCostConfigurable, which only differ in
+    // costPerExecution/costPerEntry -- legacy hardcodes them, configurable
+    // reads them from indexMeta (which is why they can't be precomputed
+    // earlier than this: they depend on which property's index definition
+    // this call resolved, and the same PropertyIndexLookup is reused across
+    // different property names, e.g. by NodeTypeIndexLookup).
+    private double getCost(NodeState indexMeta, Filter filter, PropertyValue value,
+            double costPerExecution, double costPerEntry) {
+        Set<IndexStoreStrategy> strategies = getStrategies(indexMeta);
+        if (strategies.isEmpty()) {
+            return MAX_COST;
+        }
+        ValuePattern pattern = new ValuePattern(indexMeta);
+        double entryCount = 0;
+        for (IndexStoreStrategy s : strategies) {
+            entryCount += s.count(filter, root, indexMeta, encode(value, pattern), MAX_COST);
+        }
+        return costPerExecution + costPerEntry * entryCount;
     }
 
     /**
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
@@ -67,27 +67,18 @@ public class PropertyIndexPlan {
 
     private final String name;
 
-    private final Set<String> properties;
-
     private final Set<IndexStoreStrategy> strategies;
 
     private final Filter filter;
-
-    private boolean matchesAllTypes;
-
-    private boolean matchesNodeTypes;
-
-    private final double cost;
 
     private final Set<String> values;
 
     private final int depth;
 
-    private final PathFilter pathFilter;
-
-    private final boolean unique;
-
     private final boolean deprecated;
+
+    // Value returned by getCost().
+    private final double cost;
 
     PropertyIndexPlan(String name, NodeState root, NodeState definition,
                       Filter filter){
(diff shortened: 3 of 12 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexLookup.java
```diff
@@ -75,6 +77,24 @@ public class PropertyIndexLookup {
      */
     static final int MAX_COST = 100;
 
+    /**
+     * Feature toggle name for the configurable costPerEntry/costPerExecution
+     * cost formula (OAK-12348).
+     */
+    public static final String FT_OAK_12348 = "FT_OAK-12348";
```
ChlineSaurus: Is this toggle usable? Where is it wired into the whiteboard mechanism? 

In my opinion this kind of wiring is missing:
https://github.com/apache/jackrabbit-oak/blob/a3d05a49c0b622b0fea30f9d25c5a66a83797fe1/oak-core/src/main/java/org/apache/jackrabbit/oak/query/QueryEngineSettings.java#L68
https://github.com/apache/jackrabbit-oak/blob/a3d05a49c0b622b0fea30f9d25c5a66a83797fe1/oak-core/src/main/java/org/apache/jackrabbit/oak/Oak.java#L588

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexLookup.java
```diff
@@ -75,6 +77,24 @@ public class PropertyIndexLookup {
...
+    /**
+     * When {@code true} (the default), {@link #getCost} reads {@code costPerEntry}/
+     * {@code costPerExecution} from the index definition ({@link #getCostConfigurable}).
+     * When {@code false}, {@link #getCost} uses the original hardcoded formula
+     * ({@link #getCostLegacy}) unconditionally, ignoring those properties even if
+     * set. Enabled by default: the new formula reproduces the legacy one exactly
+     * whenever {@code costPerEntry}/{@code costPerExecution} are absent, so this is
+     * a behavior-preserving default for anyone not using the new properties -- the
+     * toggle exists as an escape hatch, not as an opt-in gate.
+     */
```
ChlineSaurus: According to AI: oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndex.java:137 has an early return, that get's around this PR. However, in my opinion it should not cause any issues, as it only happens for cost == 2.0 which is rare, and anyways a very fast case.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
```diff
@@ -194,15 +200,45 @@ public class PropertyIndexPlan {
...
 
+    /**
+     * Dispatches to {@link #getCostConfigurable} or {@link #getCostLegacy}
+     * depending on {@link PropertyIndexLookup#FT_OAK_12348_ENABLE}, evaluated
+     * once per plan (a new plan is built whenever the filter changes, so a
+     * toggle flip is picked up on the next query, not on this cached plan).
+     */
     double getCost() {
-        return cost;
+        return PropertyIndexLookup.FT_OAK_12348_ENABLE.get() ? getCostConfigurable() : getCostLegacy();
```
ChlineSaurus: Nit: Computed each time we call getCost() (called multiple times in createPlan), but probably works best with toggle this way.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexLookup.java
```diff
@@ -75,6 +77,24 @@ public class PropertyIndexLookup {
...
+    public static final AtomicBoolean FT_OAK_12348_ENABLE = new AtomicBoolean(true);
```
ChlineSaurus: Did we decide/discuss if we want _default=on_ for bug fixes, or improvements in this style. ? Would also be relevant for my other PR.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexLookup.java
ChlineSaurus: But there seem to be things no longer correct in PropertyIndex.java like getMinimumCost()

https://github.com/apache/jackrabbit-oak/blob/a3d05a49c0b622b0fea30f9d25c5a66a83797fe1/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndex.java#L223

Probably worth checking.

ChlineSaurus (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexLookup.java
bhabegger: My take on this one:
- if it's a bug for which we are confident on the fix and that it is a really issue -> default on, toggle back to fals for cases where the bug has become a "feature"
- if it's a new feature we would first like to test in some scenarios -> default off, toggle on in the specific cases where we want to test and once validated toggle externally for all cases. 


Here we are in the first case. Cost is wrong for property and node types so we should make it opt out.

ChlineSaurus: Yes, agree. 
I also think it's good like this, as it works much faster while still providing the safety of being able to turn off the behavior.

I'm just asking because after the index regression in Oak 2.0.0 we discussed that it is not optimal if too many changes arrive at once. If there are regressions it is more difficult to know where they are coming from.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexLookup.java
bhabegger: Nice catch :)

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexLookup.java
bhabegger: Indeed the whiteboard registration was missed.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexLookup.java
bhabegger: I would say that this is not really a toggle issue. In a PR you just can't know what will land in the next release. Maybe the defaults should be chosen at release time to see which fixes we want in, but then that means revisiting how we put toggles in.

ChlineSaurus (approved):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/nodetype/NodeTypeIndex.java
```diff
@@ -41,13 +41,25 @@ class NodeTypeIndex implements QueryIndex, JcrConstants {
...
+    public NodeTypeIndex(MountInfoProvider mountInfoProvider, boolean disableConfigurableCosts) {
         this.mountInfoProvider = mountInfoProvider;
+        this.disableConfigurableCosts = disableConfigurableCosts;
+        this.minimumCost = disableConfigurableCosts ? NodeTypeIndexLookup.MINIMUM_COST : 0;
     }
 
     @Override
     public double getMinimumCost() {
-        return NodeTypeIndexLookup.MINIMUM_COST;
+        return minimumCost;
```
thomasmueller, line 62: I think there is no need to change getMinimumCost()

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
```diff
@@ -67,53 +67,49 @@ public class PropertyIndexPlan {
...
     private final int depth;
 
-    private final PathFilter pathFilter;
-
-    private final boolean unique;
-
     private final boolean deprecated;
 
+    // Value returned by getCost().
+    private final double cost;
```
thomasmueller, line 81: I think this is not a useful comment 
```suggestion
    private final double cost;
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
```diff
@@ -67,53 +67,49 @@ public class PropertyIndexPlan {
 
     private final String name;
 
-    private final Set<String> properties;
-
```
thomasmueller, line 71: these changes seem unrelated

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
bhabegger: Yes, not directly, but as we are revisiting the constructor, I thought it made sense. Otherwise such things end up never getting cleaned up (a PR just for that is not worth it). Or how would you get such things cleaned up ?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
bhabegger: Agreed ;)

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
```diff
@@ -67,53 +67,49 @@ public class PropertyIndexPlan {
...
+        boolean matchesAllTypes = !definition.hasProperty(DECLARING_NODE_TYPES);
         this.deprecated = definition.getBoolean(IndexConstants.INDEX_DEPRECATED);
-        this.matchesNodeTypes =
+        boolean matchesNodeTypes =
                 matchesAllTypes || StreamUtils.toStream(types).anyMatch(filter.getSupertypes()::contains);
 
         ValuePattern valuePattern = new ValuePattern(definition);
 
-        double bestCost = Double.POSITIVE_INFINITY;
+        double bestCount = Double.POSITIVE_INFINITY;
```
thomasmueller: I don't think renaming cost to count makes sense in this case, for multiple reasons:
* cost is floating point; count typically isn't
* count = MAX_COST seems weird
* it is an unrelated change

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
bhabegger: To me the type mismatch and weird assignments are a symptom that something is not quite right (having a count as a double could just me more practical as further down we multiply with a cost per entry which legitimately is double).

The true question is, are we counting or not ?  And I think we are counting. Otherwise this formula:
```
return costPerExecution + costPerEntry * entryCount;
```

Would become:
```
return costPerExecution + costPerEntry * entryCost;
```

Which makes way less sense... So, if fact, the bug is not in the rename but in the name of MAX_COST which is in fact a MAX_COUNT. I just did a little local refactor to better understand the code.

So question is this a `bestCount` (raw count without factors taken into account) or a `bestCost` (count including factors)? Below we are comparing it with something that, functionally,  *is* a count (even is we count using doubles).

bhabegger: I think the true problem is that the original code is mixing both notions... Before it did not really matter because cost == count (more or less), but now that we are introducing costPerEntry... well we can't mix up both notions anymore. 

With this PR, we mostly have a count, but there are still some cost/count inconsistencies (e.g. cost overhead is included in the count because it is deep in the code, making bubbling it up challenging).

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexLookup.java
```diff
@@ -135,18 +144,63 @@ Set<IndexStoreStrategy> getStrategies(NodeState definition) {
...
+        }
         ValuePattern pattern = new ValuePattern(indexMeta);
-        double cost = strategies.isEmpty() ? MAX_COST : COST_OVERHEAD;
+        double entryCount = 0;
         for (IndexStoreStrategy s : strategies) {
-            cost += s.count(filter, root, indexMeta, encode(value, pattern), MAX_COST);
+            entryCount += s.count(filter, root, indexMeta, encode(value, pattern), MAX_COST);
         }
-        return cost;
+        return costPerExecution + costPerEntry * entryCount;
```
bhabegger, line 203: This is why I renamed all the misnamed `xxxCost` to `xxxCount`. We multiply the entry cost per count not per cost.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
bhabegger: Switched to long (but in fact could be int as it never goes above MAX_COUNT (originally MAX_COST) = 100.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
thomasmueller: OK I see! Yes it make sense then.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
thomasmueller: > but now that we are introducing costPerEntry.

OK, yes then it makes sense.

If we can switch from double to long then that's fine.

> but in fact could be int as it never goes above MAX_COUNT

It's probably better to use long... I don't think it has a memory or performance impact, and it's easy to overflow 2^31 when we multiply.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexLookup.java
```diff
@@ -135,18 +144,63 @@ Set<IndexStoreStrategy> getStrategies(NodeState definition) {
...
+    // different property names, e.g. by NodeTypeIndexLookup).
+    private double getCost(NodeState indexMeta, Filter filter, PropertyValue value,
+            double costPerExecution, double costPerEntry) {
         Set<IndexStoreStrategy> strategies = getStrategies(indexMeta);
+        if (strategies.isEmpty()) {
+            return MAX_COST;
+        }
         ValuePattern pattern = new ValuePattern(indexMeta);
-        double cost = strategies.isEmpty() ? MAX_COST : COST_OVERHEAD;
+        double entryCount = 0;
```
thomasmueller, line 199:
```suggestion
        long entryCount = 0;
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
```diff
@@ -205,6 +200,30 @@ String getName() {
         return cost;
     }
 
+    /**
+     * Original cost formula: {@code COST_OVERHEAD + bestCount}. Ignores
+     * {@code costPerEntry}/{@code costPerExecution} even if set on the index
+     * definition.
+     */
+    double getCostLegacy(double bestCount) {
+        return bestCount == Double.POSITIVE_INFINITY ? Double.POSITIVE_INFINITY : COST_OVERHEAD + bestCount;
```
thomasmueller:
```suggestion
        return bestCount == Long.MAX_VALUE ? Double.POSITIVE_INFINITY : COST_OVERHEAD + bestCount;
```

And it looks like there needs to be a test case?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
```diff
@@ -205,6 +200,30 @@ String getName() {
         return cost;
     }
 
+    /**
+     * Original cost formula: {@code COST_OVERHEAD + bestCount}. Ignores
+     * {@code costPerEntry}/{@code costPerExecution} even if set on the index
+     * definition.
+     */
+    double getCostLegacy(double bestCount) {
```
thomasmueller, line 208:
```suggestion
    double getCostLegacy(long bestCount) {
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
```diff
@@ -205,6 +200,30 @@ String getName() {
...
+        return bestCount == Double.POSITIVE_INFINITY ? Double.POSITIVE_INFINITY : COST_OVERHEAD + bestCount;
+    }
+
+    /**
+     * {@code cost = costPerExecution + costPerEntry * bestCount}, both
+     * optionally configured on the index definition (OAK-12348). Defaults
+     * ({@code costPerEntry=1.0}, {@code costPerExecution=COST_OVERHEAD})
+     * reproduce {@link #getCostLegacy} exactly.
+     */
+    double getCostConfigurable(double bestCount) {
```
thomasmueller:
```suggestion
    double getCostConfigurable(long bestCount) {
```

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
```diff
@@ -205,6 +200,30 @@ String getName() {
...
+        if (bestCount == Double.POSITIVE_INFINITY) {
```
thomasmueller: MAX_VALUE

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
```diff
@@ -168,22 +163,22 @@ public class PropertyIndexPlan {
                         }
                     }
                     values = PropertyIndexUtil.encode(values);
-                    double cost = strategies.isEmpty() ? MAX_COST : 0;
+                    long count = strategies.isEmpty() ? MAX_COUNT : 0;
```
thomasmueller, line 166: Now it's rather dangerous, because the switch from double to long also impacts the old (legacy) code path (as you see later with comparison against infinity / max value).

So it might be better to have a "legacy" method that is a copy of the old code, and a second method that is the new code, and then one central switch based on the feature toggle.

thomasmueller (changes requested):

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
bhabegger: Now it's a long, but MAX_COUNT is still and int as the interface method `IndexStoreStrategy.count`expects an int.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/property/PropertyIndexPlan.java
bhabegger: Indeed. Let me see what I can do. I would have preferred to avoid code duplication as much as possible.

bhabegger: In fact, that's likely why I kept it a double...

bhabegger: I think the risk is quite low. Let me ensure there are proper tests covering those cases and then I would accept the (low) risk.

PR #3077, opened 2026-08-12
OAK-12346: Oak should compile & test on Java 26

mbaedke:
Now using contemporary algorithms for tests in oak-auth-ldap (RSA-2048, SHA256).


oak-auth-ldap/src/test/java/org/apache/jackrabbit/oak/security/authentication/ldap/AbstractServer.java
```diff
@@ -449,5 +463,11 @@ protected void tearDown() throws Exception {
         if (cacheService != null) {
             cacheService.destroy();
         }
+        if (previousKeyStoreType == null) {
```
rishabhdaim, line 466: If this was initially empty, then don't set it.

oak-auth-ldap/src/test/java/org/apache/jackrabbit/oak/security/authentication/ldap/AbstractServer.java
```diff
@@ -230,7 +236,15 @@ private void startLdapServer() throws Exception {
...
-        
+
+        URL keyStoreUrl = AbstractServer.class.getClassLoader().getResource(TLS_KEYSTORE_RESOURCE);
+        if (keyStoreUrl == null) {
+            throw new IOException("Unable to locate test TLS keystore: " + TLS_KEYSTORE_RESOURCE);
+        }
+        previousKeyStoreType = Security.getProperty(KEYSTORE_TYPE_PROPERTY);
+        Security.setProperty(KEYSTORE_TYPE_PROPERTY, "jks");
+        ldapServer.setKeystoreFile(new File(keyStoreUrl.toURI()).getAbsolutePath());
+        ldapServer.setCertificatePassword(TLS_KEYSTORE_PASSWORD);
```
rishabhdaim, line 247: Run it behind an `enableSSL` flag. No need for non-SSL tests.

rishabhdaim:
Could we avoid changing the JVM-global `keystore.type` here and instead use a PKCS12 test fixture? ApacheDS 2.0.0-M24 loads file-based keystores using `KeyStore.getDefaultType()`, which is PKCS12 on the supported Java 11–26 runtimes, so matching the fixture to that default removes the process-wide mutation entirely. That avoids setup-failure leaks, concurrent test interference, and restoration bugs, while keeping non-SSL tests isolated; it is safer than synchronizing and restoring a global security property.

rishabhdaim (changes requested):

oak-auth-ldap/src/test/java/org/apache/jackrabbit/oak/security/authentication/ldap/AbstractServer.java
mbaedke: We reset to the default if it was empty, which restores the original config.

mbaedke:
Going to make a new PR because it's now a new issue: https://issues.apache.org/jira/browse/OAK-12354

PR #3078, opened 2026-08-12
OAK-12352: Include diff indexes in the index definition status printer

chibulcuteanu:
/oak:index/diff.index and /oak:index/diff.index.optimizer are not oak:QueryIndexDefinition nodes, so IndexPathService does not return them and they were missing from the oak-index-defn status printer output. Append them explicitly, inlining their diff.json payload as JSON instead of a base64 blob. All other properties and child nodes render as before.

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexDefinitionPrinter.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexDefinitionPrinter.java
@@ -23,10 +23,15 @@ import java.io.PrintWriter;
 
 import org.apache.felix.inventory.Format;
 import org.apache.felix.inventory.InventoryPrinter;
+import org.apache.jackrabbit.oak.api.PropertyState;
+import org.apache.jackrabbit.oak.commons.json.JsonObject;
 import org.apache.jackrabbit.oak.commons.json.JsopBuilder;
 import org.apache.jackrabbit.oak.json.Base64BlobSerializer;
 import org.apache.jackrabbit.oak.json.JsonSerializer;
 import org.apache.jackrabbit.oak.plugins.index.IndexPathService;
+import org.apache.jackrabbit.oak.plugins.index.diff.DiffIndex;
+import org.apache.jackrabbit.oak.plugins.index.diff.DiffIndexMerger;
+import org.apache.jackrabbit.oak.spi.state.ChildNodeEntry;
 import org.apache.jackrabbit.oak.spi.state.NodeState;
 import org.apache.jackrabbit.oak.spi.state.NodeStateUtils;
 import org.apache.jackrabbit.oak.spi.state.NodeStore;
@@ -69,6 +74,17 @@ public class IndexDefinitionPrinter implements InventoryPrinter {
                 NodeState idxState = NodeStateUtils.getNode(root, indexPath);
                 createSerializer(json).serialize(idxState);
             }
+            // The "diff" indexes (diff.index / diff.index.optimizer) are not oak:QueryIndexDefinition nodes, so they
+            // are not returned by the IndexPathService and would otherwise be missing from the output. Add them
+            // explicitly, rendering their diff.json payload as inline JSON so the pending diff is readable.
+            for (String name : new String[] {DiffIndexMerger.DIFF_INDEX, DiffIndexMerger.DIFF_INDEX_OPTIMIZER}) {
+                String diffPath = "/oak:index/" + name;
+                NodeState idxState = NodeStateUtils.getNode(root, diffPath);
+                if (idxState.exists()) {
+                    json.key(diffPath);
+                    serializeDiffIndex(json, idxState);
+                }
+            }
             json.endObject();
             printWriter.print(JsopBuilder.prettyPrint(json.toString()));
         }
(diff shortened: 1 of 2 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexDefinitionPrinter.java
```diff
@@ -69,6 +74,17 @@ public void print(PrintWriter printWriter, Format format, boolean isZip) {
                 NodeState idxState = NodeStateUtils.getNode(root, indexPath);
                 createSerializer(json).serialize(idxState);
             }
+            // The "diff" indexes (diff.index / diff.index.optimizer) are not oak:QueryIndexDefinition nodes, so they
```
ChlineSaurus, line 77: Detail (what confused Claude): In oak-doc/src/site/markdown/query/indexing.md we write:

The diff index consists of:

    /oak:index/diff.index
    - type: disabled
    - jcr:primaryType: oak:QueryIndexDefinition
    + diff.json (nt:file)

This contradicts the jcr:primaryType actually used and recommended in the customer facing doc, but if following the oak doc the primary type is oak:QueryIndexDefinition and therefore picked up in the index printer.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexDefinitionPrinter.java
```diff
@@ -81,4 +97,44 @@ public void setFilter(String filter) {
...
+        // definition properties (mirror the default filter, which drops :childOrder)
+        for (PropertyState p : idxState.getProperties()) {
+            if (":childOrder".equals(p.getName())) {
+                continue;
+            }
+            json.key(p.getName());
+            serializer.serialize(p);
+        }
+        // non-hidden child nodes other than diff.json, rendered normally
+        for (ChildNodeEntry child : idxState.getChildNodeEntries()) {
```
ChlineSaurus, line 117: Is this running with the right order of child nodes? Or does it even matter here?

(I struggle a bit to understand where OAK requires correct node ordering and where not)

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexDefinitionPrinter.java
ChlineSaurus: But if one uses _jcr:primaryType: oak:QueryIndexDefinition_, we now end up with a duplicate key _/oak:index/diff.index_, so maybe worth changing the doc.

ChlineSaurus (approved):
LGTM

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexDefinitionPrinter.java
chibulcuteanu: agree. @thomasmueller , I will also create a small PR to change the docs.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexDefinitionPrinter.java
```diff
@@ -81,4 +97,44 @@ public void setFilter(String filter) {
...
+            String childName = child.getName();
+            if (childName.startsWith(":") || "diff.json".equals(childName)) {
+                continue;
+            }
+            json.key(childName);
+            createSerializer(json).serialize(child.getNodeState());
```
bhabegger, line 123: Is the serializer so stateful that we need to recreate it ?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexDefinitionPrinter.java
```diff
@@ -69,6 +74,17 @@ public void print(PrintWriter printWriter, Format format, boolean isZip) {
                 NodeState idxState = NodeStateUtils.getNode(root, indexPath);
                 createSerializer(json).serialize(idxState);
             }
+            // The "diff" indexes (diff.index / diff.index.optimizer) are not oak:QueryIndexDefinition nodes, so they
+            // are not returned by the IndexPathService and would otherwise be missing from the output. Add them
+            // explicitly, rendering their diff.json payload as inline JSON so the pending diff is readable.
+            for (String name : new String[] {DiffIndexMerger.DIFF_INDEX, DiffIndexMerger.DIFF_INDEX_OPTIMIZER}) {
```
bhabegger, line 80: IMHO, there is a lot of *logic* for something that should be a *printer*

bhabegger: When I see this kind of "work around", it very often means there is a design flaw somewhere. It's likely not the issue of this PR, but something else (the complexity of this PR is a symptom). 

Before `/oak:index/` contained a single type of object `oak:QueryIndexDefinition`. Now it can contain 2 types of objects `oak:QueryIndexDefinition` and these `diff.index` or `diff.index.optimizer` for those 2 it's not the type that is important but their name. 

Before, printing was a natural walk through the tree outputting what ever was on it's path.

Maybe we can make this "printing" a bit cleaner by splitting this method up:
loop on the indexPathService.getIndexPaths() -> call a auxilary "printIndexDefinitionNode" method for this path 

Move the logic to determine which diffIndexPaths to the indexPathService (where we add a method .getDiffPaths())
loop on those results and internally call an auxilary "printDiffNode" method (which is basically serializeDiffIndex, see my comments there).

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/inventory/IndexDefinitionPrinter.java
```diff
@@ -81,4 +97,44 @@ public void setFilter(String filter) {
     private JsonSerializer createSerializer(JsopBuilder json) {
         return new JsonSerializer(json, filter, new Base64BlobSerializer());
     }
+
+    /**
+     * Serialize a diff index node, inlining its {@code diff.json} payload as JSON. All other file child nodes are
+     * rendered as base64 blobs, for backward compatibility.
+     */
+    private void serializeDiffIndex(JsopBuilder json, NodeState idxState) {
+        json.object();
```
bhabegger, line 106: I don't like so much that we have walking logic in this class and them more walking logic in the serializer class.

The role of this class is to delegate to JsonSerializer (which consists in walking a NodeState and outputting in JSON what it finds). So it does the walking... we shouldn't be doing it here. Why are we doing it here ? Because we want a special processing of a certain parts : the inlining of the diff.json file object. Why don't we just update the serializer (eventually with a config) to add inlining of certain nodes and the configure the serializer with that option here ?

Here we would not need the 2 extra loops as this would be taken care of by the serializer.

PR #3080, opened 2026-08-12
OAK-12354: Use contemporary crypto algorithms for testing in oak-auth-ldap

mbaedke:
Using PKCS12 keystore and RSA-2048/SHA256withRSA now.


oak-auth-ldap/src/test/java/org/apache/jackrabbit/oak/security/authentication/ldap/AbstractServer.java
```diff
@@ -230,7 +233,12 @@ private void startLdapServer() throws Exception {
     protected void setupLdapServer() throws Exception {
         TcpTransport transport = new TcpTransport((port));
         transport.enableSSL(enableSSL);
-        
+        URL keyStoreUrl = AbstractServer.class.getClassLoader().getResource(TLS_KEYSTORE_RESOURCE);
+        if (keyStoreUrl == null) {
+            throw new IOException("Unable to locate test TLS keystore: " + TLS_KEYSTORE_RESOURCE);
+        }
+        ldapServer.setKeystoreFile(new File(keyStoreUrl.toURI()).getAbsolutePath());
+        ldapServer.setCertificatePassword(TLS_KEYSTORE_PASSWORD);
```
rishabhdaim, line 241: Can we only do this when `enableSSL` is true ?

mbaedke: No, this is about TLS, not SSL.

rishabhdaim: What I meant is that we should set these values when running the server in secure mode by setting `enableSSL` to `true`.

IMO, these won't make much difference when security has been disabled.

mbaedke: I don't understand. These are completely different configuration settings (AFAIR one for LDAPS and the other for LDAP over TLS) and we shouldn't change the semantics of the test with this PR.

rishabhdaim (approved):

Amoratinos (approved):

reschke (approved):
...but optimally after the upcoming Oak release.

PR #3081, opened 2026-08-12
OAK-12355: Groovy 3.0.25 doesn't support Java 26.

mbaedke:
Updated to 4.0.33.


oak-run/src/main/groovy/org/apache/jackrabbit/oak/console/GroovyConsole.groovy
```diff
@@ -26,18 +26,19 @@ import jline.console.history.FileHistory
 import org.apache.jackrabbit.oak.run.commons.Utils
 import org.apache.jackrabbit.oak.console.commands.*
 import org.apache.jackrabbit.oak.plugins.document.DocumentNodeStore
+import org.apache.groovy.groovysh.*
```
rishabhdaim: Can we fix this as well to avoid using wildcard imports?

oak-run/pom.xml
```diff
@@ -34,6 +34,7 @@
     <jetty.version>9.4.53.v20231009</jetty.version>
     <!--
       Size History:
+      + 4 MB (Groovy 3.0.25 -> 4.0.33, OAK-xxxxx)
```
rishabhdaim: Actual `OAK-<ticket>`, like added for rest.

reschke (approved):

PR #3084, opened 2026-08-17
OAK-12364: XPath queries fail to parse if they contain invalid XML characters

ChlineSaurus:
XPath queries failed to parse when a path contained certain valid characters that XML allows in names but Java does not (such as the middle dot ·),  because the parser treated them as spaces and split the path into two pieces, leaving a leftover token it couldn't make sense of. This fix changes the parser to accept those XML name characters, such that those paths can be queried.

The bug fix is very simple, but in a critical path of the code (XPath to SQL converter). Therefore, I decided to put the fix behind a feature toggle (but since it's a bug fix I have it on by default, as documented in Agents.md). 

Issue: https://issues.apache.org/jira/browse/OAK-12364

Moved to: https://github.com/apache/jackrabbit-oak/pull/3121

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/query/xpath/XPathToSQL2Converter.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/query/xpath/XPathToSQL2Converter.java
@@ -28,11 +28,12 @@ import org.apache.jackrabbit.oak.query.QueryOptions;
 import org.apache.jackrabbit.oak.query.QueryOptions.Traversal;
 import org.apache.jackrabbit.oak.query.xpath.Statement.UnionStatement;
 import org.apache.jackrabbit.util.ISO9075;
+import org.apache.jackrabbit.util.XMLChar;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * This class can can convert a XPATH query to a SQL2 query.
+ * This class can convert a XPATH query to a SQL2 query.
  */
 public class XPathToSQL2Converter {
 
@@ -1005,6 +1006,9 @@ public class XPathToSQL2Converter {
                     type = CHAR_VALUE;
                 } else {
                     if (Character.isJavaIdentifierPart(c)) {
+                        type = CHAR_NAME;
+                    } else if ((settings == null || settings.isXmlNameCharsInPathEnabled()) && XMLChar.isName(c)) {
+                        // accept XML name characters that ISO9075 leaves unencoded, so they are not split off the name
                         type = CHAR_NAME;
                     }
                 }
--- a/oak-core/src/test/java/org/apache/jackrabbit/oak/query/index/TraversingIndexQueryTest.java
+++ b/oak-core/src/test/java/org/apache/jackrabbit/oak/query/index/TraversingIndexQueryTest.java
@@ -502,11 +502,11 @@ public class TraversingIndexQueryTest extends AbstractQueryTest {
 
     @Test
     public void testXmlNameCharsInPathAreQueryable() throws Exception {
-        root.getTree("/").addChild("content").addChild("m·d").addChild("child");
+        root.getTree("/").addChild("content").addChild("m\u00b7d").addChild("child");
         root.commit();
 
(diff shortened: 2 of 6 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/query/xpath/XPathToSQL2Converter.java
```diff
@@ -1006,6 +1007,9 @@ private void initialize(String query) throws ParseException {
                 } else {
                     if (Character.isJavaIdentifierPart(c)) {
                         type = CHAR_NAME;
+                    } else if ((settings == null || settings.isXmlNameCharsInPathEnabled()) && XMLChar.isName(c)) {
```
bhabegger, line 1010: Shouldn't we not enable this if settings is null (to keep the existing behavior in that case) ?

ChlineSaurus: I was not sure about this. I also had in mind that we decided that we want to keep the old behavior as the default behavior for Oak changes, but in the Agents.md we write that in case something is a bug fix, default should be on, and only new features should be default off. 
Further, it's not a very risky change, therefore I went with default enabled, but if we want to be more defensive I can change it. 😌

reschke: Is there a case where the new behavior would break something that previously "worked" (that is, didn't throw an exception)?

oak-core/src/test/java/org/apache/jackrabbit/oak/query/index/TraversingIndexQueryTest.java
```diff
@@ -500,6 +500,15 @@ public void testMultipleRelativeProperties() throws Exception {
         assertQuery("//*[*/@d < 3]", "xpath", Arrays.asList("/content"));
     }
 
+    @Test
+    public void testXmlNameCharsInPathAreQueryable() throws Exception {
+        root.getTree("/").addChild("content").addChild("m·d").addChild("child");
```
reschke: readabilty: use \u notation here,

oak-core/src/main/java/org/apache/jackrabbit/oak/query/xpath/XPathToSQL2Converter.java
ChlineSaurus: There is one case I found, but I'm not sure how realistic it is. The query so far interpreted XML characters as whitespaces. If this is in the middle, it failed, as described in the Issue. However, if the XML character is directly next to a separator, it is ignored, meaning the new behavior can change queries behavior for such queries instead of failing (See test below for an example.) Not sure the feature toggle is too defensive, but better safe than sorry. 

```
@Test
    public void testXmlNameCharsInPathFeatureChangesBehavior() throws ParseException {
        // a middle dot (\u00b7) attached to a name is dropped when the feature is disabled,
        // but kept as part of the name when enabled, resulting in a different query
        String xpath = "//*[@a\u00b7 = 1]";

        QueryEngineSettings disabled = new QueryEngineSettings();
        disabled.setXmlNameCharsInPathFeature(createFeature(false));
        String withoutFeature = new XPathToSQL2Converter(disabled).convert(xpath);

        QueryEngineSettings enabled = new QueryEngineSettings();
        enabled.setXmlNameCharsInPathFeature(createFeature(true));
        String withFeature = new XPathToSQL2Converter(enabled).convert(xpath);

        assertTrue(withoutFeature.contains("[a] = 1"));
        assertTrue(withFeature.contains("[a\u00b7] = 1"));
        assertNotEquals(withoutFeature, withFeature);
    }
```

oak-core/src/test/java/org/apache/jackrabbit/oak/query/index/TraversingIndexQueryTest.java
ChlineSaurus: Updated 😌
Btw. if I should first merge against an Oak branch, I'm happy to do so, but I can't create branches on my own.

thomasmueller (approved):

PR #3091, opened 2026-08-26
OAK-12372: ignore Lucene facet properties that are too long

mkapalka:
Lucene limits the lenght of the category path of a facet to 8191 characters. When Oak tries to index a property which name + value + 1 (path separator) exceeds this limit, Lucene throws an exception that aborts the indexing step. Since those kinds of facet properties are unlikely to be useful, it is better to ignore them and continue the indexing process. This patch does it but uses a feature flag to disable this new behaviour if ever needed (enabled by default).

### DIFF
--- a/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMaker.java
+++ b/oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMaker.java
@@ -203,32 +203,16 @@ public class LuceneDocumentMaker extends FulltextDocumentMaker<Document> {
                 getFacetsConfig().setMultiValued(pname, true);
                 Iterable<String> values = property.getValue(Type.STRINGS);
                 for (String value : values) {
-                    if (value != null && !value.isEmpty()) {
+                    if (isValidFacetProperty(pname, value)) {
                         doc.add(new SortedSetDocValuesFacetField(pname, value));
                     }
                 }
                 fieldAdded = true;
             } else if (tag == Type.STRING.tag()) {
                 String value = property.getValue(Type.STRING);
-                if (!value.isEmpty()) {
-                    if (FT_OAK_12372_DISABLE.get()) {
-                        // Legacy mode
-                        doc.add(new SortedSetDocValuesFacetField(pname, value));
-                        fieldAdded = true;
-                    } else {
-                        // Category path = pname + "/" + value --> cannot be longer than the Lucene limit of 8191
-                        int categoryPathLength = pname.length() + value.length() + 1;
-                        if (categoryPathLength > FacetLabel.MAX_CATEGORY_PATH_LENGTH) {
-                            if (!LOG_SILENCER.silence(LOG_KEY_IGNORING_LONG_FACET_PROPERTY)) {
-                                LOG.warn("[{}] Ignoring long facet property. Property {} is too long (name + value length: {})"
-                                                + " and cannot be used for facets",
-                                        getIndexName(), pname, categoryPathLength);
-                            }
-                        } else {
-                            doc.add(new SortedSetDocValuesFacetField(pname, value));
-                            fieldAdded = true;
-                        }
-                    }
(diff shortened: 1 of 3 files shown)


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMaker.java
ChlineSaurus, line 207: We don't protect the path for multi-values. I'm not sure, but it might be worth capturing both in this PR and with this feature toggle. 🤔

Test that still fails with the same exception:
```
    @Test
    public void facetPropertyMultiValuedExceedingMaxCategoryPathLengthStillThrows() throws IOException {
        LuceneIndexDefinitionBuilder builder = new LuceneIndexDefinitionBuilder();
        builder.indexRule("nt:base")
                .property("foo")
                .propertyIndex()
                .facets();

        LuceneIndexDefinition defn = LuceneIndexDefinition.newLuceneBuilder(root, builder.build(), "/foo").build();
        LuceneDocumentMaker docMaker = new LuceneDocumentMaker(null, FacetsConfig::new, null, defn,
                defn.getApplicableIndexingRule("nt:base"), "/x");

        NodeBuilder test = EMPTY_NODE.builder();
        // Multi-valued facet with a value well over Lucene's 8191 category-path limit
        test.setProperty("foo", List.of("a".repeat(10000)), Type.STRINGS);

        boolean originalFtValue = LuceneDocumentMaker.FT_OAK_12372_DISABLE.get();

        LuceneDocumentMaker.FT_OAK_12372_DISABLE.set(false); // fix enabled (default) -- still not guarded here
        docMaker.makeDocument(test.getNodeState());
    }
```

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMaker.java
```diff
@@ -199,8 +211,24 @@ protected boolean indexFacetProperty(Document doc, int tag, PropertyState proper
...
+                    if (FT_OAK_12372_DISABLE.get()) {
+                        // Legacy mode
+                        doc.add(new SortedSetDocValuesFacetField(pname, value));
+                        fieldAdded = true;
+                    } else {
+                        // Category path = pname + "/" + value --> cannot be longer than the Lucene limit of 8191
+                        int categoryPathLength = pname.length() + value.length() + 1;
+                        if (categoryPathLength > FacetLabel.MAX_CATEGORY_PATH_LENGTH) {
+                            if (!LOG_SILENCER.silence(LOG_KEY_IGNORING_LONG_FACET_PROPERTY)) {
+                                LOG.warn("[{}] Ignoring long facet property. Property {} is too long (name + value length: {})"
```
ChlineSaurus: Just a thought: would it be worth to log the node path? (to make it easier to identify where the issue occurred)?

ChlineSaurus (commented):
Overall looks good, but left two comments.

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMaker.java
mkapalka: Good point, I'll add that, thanks!

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneDocumentMaker.java
mkapalka: Indeed, even though only some paths will be logged (because of the log silencer).

ChlineSaurus (approved):

reschke:
So did this PR pass CI testing?

mkapalka:
> So did this PR pass CI testing?

At some point yes (except that there were some permission issues with Sonar), and the build of `trunk` (including this change) is green now, so everything seems OK.

PR #3096, opened 2026-09-01
OAK-12379 also output the stack when the session was closed

joerghoh:


pat-lego (approved):

reschke:
We might have some code in oak-commons that does some filtering in stack traces to remove the Oak part (need to change).

joerghoh:
I doubt that we have a large amount of Oak code in this particular stacktrace, the code path from ``session.logout()`` to ``sessionDelegate.logout()`` should be quite short. 

Also I hope that this exception is not thrown that frequently that it makes a difference. In the end this is just additional diagnostics.

reschke (approved):

PR #3100, opened 2026-09-02
OAK-12249: lazy ES index provisioning — skip creation for empty reindex

bhabegger:
When FT_OAK-12249 and FT_OAK-12248 are both enabled, ElasticIndexWriter defers provisionIndex() from the constructor to the first updateDocument() or deleteDocuments() call. A reindex that produces zero documents never creates an Elasticsearch index or alias, eliminating the empty-index problem described in OAK-12249.

Deployment order is enforced at runtime: isLazyProvisioningActive() returns true only when both toggles are on. Enabling FT_OAK-12249 alone logs a WARN and falls back to eager provisioning, preventing 404 errors on query paths that lack graceful 404 handling.

ensureProvisioned() handles the incremental-write-after-empty-reindex case: if an alias does not exist when the first document arrives, it creates a new backing index with a fresh seed and points the alias at it.


reschke:
By all means, could you *please* wait until https://github.com/apache/jackrabbit-oak/pull/3088 is merged?????????

reschke:
Unfortunately, I had to revert ths one due to the issue above,.

PR #3110, opened 2026-09-03
OAK-12173: Query traverses even with correct index, at startup

tihom88:


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
```diff
@@ -735,6 +735,13 @@ protected LuceneIndexNode acquireIndexNode(String indexPath) {
         if (NON_LAZY) {
             return tracker.acquireIndexNode(indexPath);
         }
+        if (!tracker.isIndexReady(indexPath)) {
```
thomasmueller, line 738: NON_LAZY is true in 99% of the cases, so this change, in reality, has no effect

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LuceneIndexProviderService.java
```diff
@@ -390,12 +395,20 @@ private void activate(BundleContext bundleContext, Configuration config) throws
...
         tracker = createTracker(bundleContext, config);
+
+        oakRegs.add(whiteboard.register(FeatureToggle.class,
+                new FeatureToggle("FT_OAK-12173", FT_OAK_12173),
+                emptyMap()));
+        if (FT_OAK_12173.get()) {
+            // OAK-12173: update the tracker now, so indexes built before this
+            // service started can be used right away. Without this, we'd have
+            // to wait for the Observer below, which can be slow to start.
+            tracker.update(nodeStore.getRoot());
```
thomasmueller, line 409: If I understand correctly, this is the main / only change in the behavior, right?

But it only covers indexes that were created during startup of this process? I don't think this would be an important use case...

thomasmueller: After discussion: no, it also affects old existing indexes.

thomasmueller (approved):

tihom88:
Closing it in favour of https://github.com/apache/jackrabbit-oak/pull/3119

PR #3112, opened 2026-09-08
OAK-12394: Async indexing LAST_INDEXED_TIME metric should reflect durable repo property

mihaibercea:
### Problem

The `<lane>.LAST_INDEXED_TIME` metric exposed by `AsyncIndexUpdate` does not reliably
reflect the true last-indexed time. Two independent sources of this information diverge
in practice:

| Source | Backing store | Behaviour |
|--------|--------------|-----------|
| `IndexStatsMBean.getLastIndexedTime()` | Durable repository property `:async/<lane>-LastIndexedTo` | Ground truth — survives restarts, always current |
| `<lane>.LAST_INDEXED_TIME` metric (`CounterStats`) | In-memory Dropwizard counter, updated in `ExecutionStats.doneOneCycle()` | Only updated on the **success path**; resets to `0` on restart |

`doneOneCycle()` is reachable only via `AsyncIndexStats.done()` ←
`postAsyncRunStatsStatus()`. Any run that takes the failure path — e.g. a lane stuck on
a lease / concurrent-update exception — calls `indexStats.failed()` instead and **never
reaches `done()`**. As a result:

- A lane that fails on every cycle keeps the counter at whatever value it held at the
  last successful cycle on *this process instance*.
- A process restart resets the counter to `0`. If the lane continues to fail, the
  metric stays at `0` indefinitely — not a stale-but-real timestamp, but a value
  indistinguishable from "never indexed".
- Standby cluster members never run async indexing, so their counters are permanently
  `0` regardless of what the repository state says.

The metric therefore cannot be used to reliably determine when a lane last indexed, which
is its entire purpose.

### Fix

Replace the `CounterStats` with a `GaugeStats<Long>` whose supplier is evaluated on
every metric scrape:

- **Feature enabled (default):** reads the durable `:async/<lane>-LastIndexedTo`
  property from the repository and returns epoch millis (`0` if absent or unparseable,
  e.g. before the first index cycle has completed).
- **Feature disabled:** returns the legacy in-memory value (wall-clock of the last
  successful cycle), preserving the previous behaviour for a clean revert.

Reading from the shared `NodeStore` means the value is consistent across all cluster
members (leader and standby), survives process restarts, and reflects the actual
repository state even while the lane is failing. The metric name and units (epoch millis)
are unchanged, so any tooling built on top of this metric continues to work; only the
metric type changes from counter to gauge.

### Feature toggle

`FT_OAK_ASYNC_METRIC_DURABLE`, enabled by default (bug-fix convention per Oak
conventions). Registered on the OSGi Whiteboard via `Feature.newFeature()` and
runtime-flippable. When no `Whiteboard` is provided (embedded/test use), the durable
behaviour applies unconditionally.

### Changes

**`AsyncIndexUpdate`**
- New constant `FT_OAK_ASYNC_METRIC_DURABLE` and an optional `Whiteboard` constructor
  parameter; existing constructors delegate with `null` — no call-site changes required.
- `ExecutionStats`: registers a `GaugeStats<Long>` instead of a `CounterStats`;
  adds `getLastIndexedTimeMetricValue()` (feature-gated dispatch),
  `readLastIndexedToMillis()` (reads and parses the durable property), and `close()`
  (releases the feature toggle).
- `AsyncIndexUpdate.close()` now calls `indexStats.close()` to release the feature.

**`AsyncIndexerService`**
- Passes the activation `Whiteboard` to the `AsyncIndexUpdate` constructor for all
  configured lanes and the async-reindex lane.

**`AsyncIndexUpdateTest`**
- New test `lastIndexedTimeGaugeReflectsDurableProperty()`: constructs a fresh indexer
  instance against a store that already has the `LastIndexedTo` property written (from a
  prior cycle), and asserts the gauge returns the persisted epoch millis without any
  cycle having run — directly testing the restart / in-memory-reset scenario.

### Testing

```bash
mvn test -pl oak-core -Dtest=AsyncIndexUpdateTest
```

### DIFF
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
@@ -1431,7 +1431,7 @@ public class AsyncIndexUpdate implements Runnable, Closeable {
                     Calendar cal = ISO8601.parse(ps.getValue(Type.STRING));
                     return cal != null ? cal.getTimeInMillis() : 0;
                 } catch (Exception e) {
-                    log.debug("[{}] Unable to read {} for LAST_INDEXED_TIME metric", name, lastIndexedTo, e);
+                    log.warn("[{}] Unable to read {} for LAST_INDEXED_TIME metric", name, lastIndexedTo, e);
                     return 0;
                 }
             }
--- a/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
+++ b/oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
@@ -1371,6 +1371,13 @@ public class AsyncIndexUpdate implements Runnable, Closeable {
             private final MeterStats indexedNodeCountMeter;
             private final TimerStats indexerTimer;
             private final HistogramStats indexedNodePerCycleHisto;
+            /**
+             * Counter storing the last-indexed time as epoch millis. Initialised at
+             * construction from the durable ":async/&lt;lane&gt;-LastIndexedTo" repository
+             * property so that a correct non-zero value is visible immediately after a
+             * restart, even if the lane has not yet completed a successful cycle.
+             * Updated on every successful cycle via {@link #doneOneCycle}. See OAK-12394.
+             */
             private final CounterStats lastIndexedTime;
             private StatisticsProvider statisticsProvider;
 
@@ -1387,6 +1394,10 @@ public class AsyncIndexUpdate implements Runnable, Closeable {
                 indexedNodePerCycleHisto = statsProvider.getHistogram(stats("INDEXER_NODE_COUNT_HISTO"), StatsOptions
                         .METRICS_ONLY);
                 lastIndexedTime = statsProvider.getCounterStats(stats("LAST_INDEXED_TIME"), StatsOptions.DEFAULT);
+                long initialMillis = readLastIndexedToMillis();
+                if (initialMillis > 0) {
+                    lastIndexedTime.inc(initialMillis);
+                }
                 try {
                     consolidatedType = new CompositeType("ConsolidatedStats",
                             "Consolidated stats", names,
@@ -1404,6 +1415,25 @@ public class AsyncIndexUpdate implements Runnable, Closeable {
                 indexedNodePerCycleHisto.update(updates);
                 long previousLastIndexedTime = lastIndexedTime.getCount();
                 lastIndexedTime.inc(System.currentTimeMillis() - previousLastIndexedTime);
+            }
+
+            /**
+             * Reads the persisted ":async/&lt;lane&gt;-LastIndexedTo" property as epoch millis,
+             * or 0 when it is absent or cannot be parsed (e.g. before the initial index).
+             * Used to initialise the {@code lastIndexedTime} counter at construction.
+             */
+            private long readLastIndexedToMillis() {
+                try {
+                    PropertyState ps = store.getRoot().getChildNode(ASYNC).getProperty(lastIndexedTo);
+                    if (ps == null) {
+                        return 0;
+                    }
+                    Calendar cal = ISO8601.parse(ps.getValue(Type.STRING));
+                    return cal != null ? cal.getTimeInMillis() : 0;
+                } catch (Exception e) {
+                    log.debug("[{}] Unable to read {} for LAST_INDEXED_TIME metric", name, lastIndexedTo, e);
+                    return 0;
+                }
             }
 
             public Counting getExecutionCounter() {
--- a/oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdateTest.java
+++ b/oak-core/src/test/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdateTest.java
@@ -1379,6 +1379,47 @@ public class AsyncIndexUpdateTest {
         assertEquals(0, lastExecutionStats(async.getIndexStats().getExecutionCount()));
     }
 
+    /**
+     * The LAST_INDEXED_TIME counter must be initialised from the durable
+     * {@code :async/<lane>-LastIndexedTo} repository property at construction time (OAK-12394).
+     * A fresh indexer instance simulating a process restart must report the persisted timestamp
+     * immediately, without needing to complete a successful cycle first.
+     */
+    @Test
+    public void lastIndexedTimeCounterInitialisedFromDurableProperty() throws Exception {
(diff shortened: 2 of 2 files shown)


oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -486,6 +511,7 @@ public void close() {
         } else {
             log.info("[{}] Closed", name);
         }
+        indexStats.close();
```
thomasmueller: This looks unrelated.... Maybe it's not strictly needed? What is the effect of removing it?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1364,29 +1390,50 @@ ExecutionStats getExecutionStats() {
             return execStats;
         }
 
+        void close() {
+            execStats.close();
```
thomasmueller: If we don't pass the whiteboard to the ExecutionStats constructor (see below), then this is not needed.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1364,29 +1390,50 @@ ExecutionStats getExecutionStats() {
...
             private final HistogramStats indexedNodePerCycleHisto;
-            private final CounterStats lastIndexedTime;
+            /**
+             * LAST_INDEXED_TIME exposed as a gauge evaluated on every scrape. When the
+             * durable-metric feature is enabled (default) it returns the persisted
+             * ":async/&lt;lane&gt;-LastIndexedTo" timestamp, so the metric is accurate for
+             * every lane and every cluster member, survives restarts and stays correct
+             * while a lane is failing. See OAK-10260.
+             */
+            private final GaugeStats<Long> lastIndexedTimeGauge;
```
thomasmueller: Is it possible to retain the CounterStats (_not_ switch to GaugeStats)? Switching could cause compatibility issues.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1364,29 +1390,50 @@ ExecutionStats getExecutionStats() {
...
+            @Nullable
+            private final Feature durableMetricFeature;
             private StatisticsProvider statisticsProvider;
 
             private final String[] names = {"Executions", "Nodes"};
             private final String name;
             private CompositeType consolidatedType;
 
-            public ExecutionStats(String name, StatisticsProvider statsProvider) {
+            public ExecutionStats(String name, StatisticsProvider statsProvider, @Nullable Whiteboard whiteboard) {
```
thomasmueller: Instead of adding a Whiteboard as a parameter (and then internally use the Feature at runtime), I think it's slightly better to just pass a boolean parameter "durableMetric". This mean when switching the feature toggle, there is no effect in the running system, but that's fine: the feature toggle is then only read at initialization, which is enough. 

The main advantage is that this approach avoids having to add "close()" everywhere.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
thomasmueller: This seems unrelated. Is it needed?

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
mihaibercea: I reviewed both keeping a counter and converting to a gauge.
The correct approach, for this particular metric, would be to change it to a gauge and to always read it from the property. This will make sure that the value is always accurate (and fresh).

The risk, in this situation, would be on JMX, that it might fail when reading the value, in the case where it is hardcoded to expect a counter.
- Doing an LLM review, I get this: "Every single consumer in this codebase reads this metric via Prometheus - zero JMX hardcoded attribute reads."
- Prometheus — "no compatibility risk"

Dropwizard's Counter and Gauge both export as Prometheus gauge type. The Dropwizard→Prometheus bridge treats a Dropwizard Counter as a gauge because Dropwizard counters can go up and down (they're general-purpose, not monotonically-increasing). So:

Before (CounterStats)
TYPE oak_fulltext_async_LAST_INDEXED_TIME gauge
oak_fulltext_async_LAST_INDEXED_TIME 1788789168378

After (GaugeStats)
TYPE oak_fulltext_async_LAST_INDEXED_TIME gauge
oak_fulltext_async_LAST_INDEXED_TIME 1788789168378

======

Implementing a fix to keep the counter, would involve extra work to always refresh the value:

  runWhenPermitted()   ← runs EVERY scheduled tick, on EVERY member
  ┌───────────────────────────────────────────────────────────┐
  │  ★ NEW: if (durableMetric)                                 │
  │        counter.set( read :async/async-LastIndexedTo )      │  ← push durable → counter
  │                                                            │
  │    if (paused) return;                                     │
  │    if (lease held by someone else) { failed(); return; }   │  ← standby bails HERE,
  │    ...                                                      │     but the push above
  │    updateIndex(...)  →  doneOneCycle()                     │     already ran ✓
  └───────────────────────────────────────────────────────────┘
  
  ====
  
  Should I keep the current proposal to change it to a gauge?

thomasmueller:
We should still use a feature toggle (now it's gone), but only read from it once (at init time).

The change from Counter to Gauge seems unnecessary.

oak-core/src/main/java/org/apache/jackrabbit/oak/plugins/index/AsyncIndexUpdate.java
```diff
@@ -1406,6 +1417,25 @@ public void doneOneCycle(long timeInMillis, long updates){
...
+            private long readLastIndexedToMillis() {
+                try {
+                    PropertyState ps = store.getRoot().getChildNode(ASYNC).getProperty(lastIndexedTo);
+                    if (ps == null) {
+                        return 0;
+                    }
+                    Calendar cal = ISO8601.parse(ps.getValue(Type.STRING));
+                    return cal != null ? cal.getTimeInMillis() : 0;
+                } catch (Exception e) {
+                    log.debug("[{}] Unable to read {} for LAST_INDEXED_TIME metric", name, lastIndexedTo, e);
```
thomasmueller:
```suggestion
                    log.warn("[{}] Unable to read {} for LAST_INDEXED_TIME metric", name, lastIndexedTo, e);
```

thomasmueller (approved):

thomasmueller (approved):

PR #3114, opened 2026-09-09
OAK-12396 : fix countDirectChildren depth

rishabhdaim:
## Summary

Fix `oak.countDirectChildren` so non-root paths count direct child documents instead of the next depth level.

## Changes

- Correct depth selection in `oak-run/src/main/js/oak-mongo.js`.
- Preserve root-path behavior.
- Correct the method return documentation.

## Test Plan

- Verified against local MongoDB database `cm-p55811-e576007-database`.
- Verified root and non-root counts match equivalent raw MongoDB queries.
- Passed `node --check oak-run/src/main/js/oak-mongo.js`.

## Links

- Jira: https://issues.apache.org/jira/browse/OAK-12396

Co-Authored-By: OpenAI Codex <noreply@openai.com>

### DIFF
--- a/oak-run/src/main/js/oak-mongo.js
+++ b/oak-run/src/main/js/oak-mongo.js
@@ -135,7 +135,7 @@ var oak = (function(global){
      * @memberof oak
      * @method countDirectChildren
      * @param {string} path the path of a node.
-     * @returns {number} the number of children, including all descendant nodes.
+     * @returns {number} the number of direct child nodes.
      */
     api.countDirectChildren = function(path){
         if (path === undefined) {
@@ -145,7 +145,8 @@ var oak = (function(global){
         }
         var depth = pathDepth(path);
         var totalCount = 0;
-        var count = db.nodes.count({_id: pathFilter(depth + 1, path)});
+        var childDepth = path == "/" ? depth + 1 : depth;
+        var count = db.nodes.count({_id: pathFilter(childDepth, path)});
         totalCount += count;
         return totalCount;
     };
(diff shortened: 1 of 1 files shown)


Joscorbe (approved):

oak-run/src/main/js/oak-mongo.js
```diff
@@ -145,7 +145,8 @@ var oak = (function(global){
         }
         var depth = pathDepth(path);
         var totalCount = 0;
-        var count = db.nodes.count({_id: pathFilter(depth + 1, path)});
+        var childDepth = path == "/" ? depth + 1 : depth;
```
Joscorbe: Are you sure this is correct? For root deptch should be 0, so we increase +1, ok.

But for the rest if I'm in /conf the depth is 1, and this will result in counting the nodes on the same level, not its children. This should count the number of nodes in depth 2, not 1, otherwise we would be counting siblings of the same path we are already in.

oak-run/src/main/js/oak-mongo.js
Joscorbe: This part tricked me, because the path is modified here before calculating the depth, so this is actually targeting the children already, no need to add a +1 then.

I wonder if we could change the naming or simplify this code to avoid confusion.

rishabhdaim: fixed in https://github.com/apache/jackrabbit-oak/pull/3114/commits/83c6a5e42c6e0682366c7f6a93efd1d242cbf481

PR #3124, opened 2026-09-15
OAK-12399: Lucene keeps jcr:score as a secondary sort key alongside an ordered property

djscherrer:
**Problem Description**
Ordering a full-text query by an indexed property and jcr:score
(e.g. ORDER BY [foo] DESC, [jcr:score] DESC) dropped jcr:score from the
Lucene sort and executed a field sort without score tracking, so the
secondary relevance ordering was lost and the projected jcr:score was NaN.

**Implemented Fix**
getSort now keeps jcr:score as a real relevance sort field
(SortField.Type.SCORE) at its position, pairing property entries with their
PropertyDefinition via a separate index (also avoiding the OAK-4602
IndexOutOfBoundsException), and runs the search with doDocScores=true when
the sort contains a score field.

Enabled by default and reversible at runtime via the FT_LEGACY_SORT_OAK-12399
kill-switch feature toggle (wired through LuceneIndexProvider /
LuceneIndexProviderService, mirroring FT_OAK-12146). Adds
LuceneScoreWithPropertyOrderByTest.


oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
```diff
@@ -373,14 +397,16 @@ private boolean loadDocs() {
                                 if (sort == null) {
                                     docs = searcher.searchAfter(lastDoc, query, nextBatchSize);
                                 } else {
-                                    docs = searcher.searchAfter(lastDoc, query, nextBatchSize, sort);
+                                    // OAK-12399: needsScores as doDocScores so a jcr:score sort field is populated.
+                                    docs = searcher.searchAfter(lastDoc, query, null, nextBatchSize, sort, needsScores, false);
```
thomasmueller, line 401: If the feature toggle is disabled, then the old code needs to be called. We could have an "if" condition here for the feature toggle, or maybe it's enough to have one for the needsScores

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
```diff
@@ -373,14 +397,16 @@ private boolean loadDocs() {
...
                                 }
                             } else {
                                 LOG.debug("loading the first {} entries for query {}", nextBatchSize, query);
                                 if (sort == null) {
                                     docs = searcher.search(query, nextBatchSize);
                                 } else {
-                                    docs = searcher.search(query, nextBatchSize, sort);
+                                    // OAK-12399: needsScores as doDocScores so a jcr:score sort field is populated.
+                                    docs = searcher.search(query, null, nextBatchSize, sort, needsScores, false);
```
thomasmueller, line 409: Same as above

oak-lucene/src/main/java/org/apache/jackrabbit/oak/plugins/index/lucene/LucenePropertyIndex.java
```diff
@@ -268,6 +289,9 @@ public Cursor query(final IndexPlan plan, NodeState rootState) {
         }
         final Filter filter = plan.getFilter();
         final Sort sort = getSort(plan);
+        // OAK-12399: when the sort includes the relevance score, scores must be tracked explicitly,
+        // otherwise Lucene returns NaN for a field sort.
+        final boolean needsScores = requiresScores(sort);
```
thomasmueller: We need to check the feature toggle, either here or in the requiresScores method.

thomasmueller (changes requested):

djscherrer:
With needsScores = !legacySortEnabled() && requiresScores(sort), the toggle in legacy mode forces needsScores = false, so the call is search(query, null, n, sort, false, false). In Lucene 4.7.2 the old search(query, n, sort) already delegates to exactly that (wrapFilter(query, null) == query, doDocScores/doMaxScore = false), so the legacy path is byte-for-byte the old behavior — no separate if needed. The two only diverge when needsScores == true (a jcr:score sort field is present), where doDocScores=true is what computes the scores.

djscherrer (commented):
should be resolved now

PR #3125, opened 2026-09-16
OAK-12400 : online revision cleanup can delete referenced segments on a store not written in compaction order

rishabhdaim:
## Summary

Online revision cleanup can delete still-referenced segments on a segment store that was **not written in compaction order** (e.g. a promoted cold standby).

`DefaultCleanupContext.isDanglingFutureSegment` decides which *compacted* segments are unused aborted-compaction leftovers by **physical scan position**: cleanup marks TAR entries newest-first and treats every compacted segment as reclaimable until the last compacted root recorded in `gc.log` is reached (the `aheadOfRoot` flag). This assumes segments are laid out in compaction write order (the compacted root written last).

That assumption does not hold for a store assembled out of compaction order — e.g. a cold standby, whose segments are written via `FileStore.writeSegment` in sync-arrival order. There the recorded root is not the newest segment, so the reverse scan reaches live, head-referenced compacted segments while `aheadOfRoot` is still `true` and marks them reclaimable. Online revision cleanup then removes referenced segments and a subsequent read/compaction fails with `SegmentNotFoundException`.

## Changes

- Add an order-independent, **generation-based** detection: a compacted segment is an unused future segment iff its generation is strictly newer than the head generation (`generation.compareWith(headGeneration) > 0`).
- Gated by system property `oak.segment.cleanup.danglingByGeneration`, **default `false`** (no behaviour change); opt in with `-Doak.segment.cleanup.danglingByGeneration=true`.
- Thread the head `GCGeneration` into `DefaultCleanupContext` (covers both the pre-compaction `CleanupFirstCompactionStrategy` and post-compaction `DefaultCleanupStrategy` paths).
- Add `DefaultCleanupContextTest`.

## Test plan

- [x] `mvn test -pl oak-segment-tar -Dtest=DefaultCleanupContextTest` — 13 tests, green (reproduces the positional over-reclaim by default; verifies the generational mode retains live compacted segments while still reclaiming genuine future leftovers).

## Links

- https://issues.apache.org/jira/browse/OAK-12400


ahanikel:
I'm not too deep into compaction and esp not online compaction but the reasoning makes sense to me and the change is opt-in.

jsedding (changes requested):
_Note: I'll use a shorthand for GCGenerations: (g = generation, f = fullGeneration, C=compacted/-=not compacted)._

This predicate does not hold for _partially compacted_ head states. Partially compacted head states can legitimately reference compacted segments of a strictly newer generation.

Example:
 - head = partial (g,f,C)
 - live referenced target segments = (g+1,f,C) or (g+1,f+1,C)

Thus the change is somewhat dangerous.

This also works in the opposite direction, which is arguably less problematic. Dangling compacted leftovers are not always "strictly newer" than the head.

 Example:
 - current head = (g,f,-)
 - aborted partial compaction left behind (g,f,C)

 Those are dangling, but not strictly newer.

That said: within Oak, partial compaction is currently unreachable because the GC layer never provides a real soft canceller or state-save trigger:
 AbstractGarbageCollectionStrategy#getSoftCanceller() returns Canceller.newCanceller() and getStateSaveTriggerSupplier() returns Canceller::newCanceller
 (oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/AbstractGarbageCollectionStrategy.java:245-251). As a result, in
 AbstractCompactionStrategy both stateSaveTrigger.isCancelable() and softCanceller.isCancelable() are false, so Oak always takes the compactUp(...) path
 instead of compactDown(...) with partial-state persistence
 (oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/AbstractCompactionStrategy.java:185-196).

We could consider removing support for partial-compaction in order to simplify the GC implementation altogether.

rishabhdaim:
Thanks @jsedding — you're right about the partially-compacted cases. I verified that partial compaction is currently unreachable: `AbstractCompactionStrategy` only takes the `compactDown` (partial-state) path when `stateSaveTrigger.isCancelable()` or `softCanceller.isCancelable()` is true, and both come from `AbstractGarbageCollectionStrategy#getSoftCanceller()` / `getStateSaveTriggerSupplier()`, which return the non-cancelable `Canceller.newCanceller()`. So Oak always takes `compactUp(...)`, the head is never a partially-compacted state, and a live head reference to a strictly-newer compacted segment can't occur in production today. The generation predicate is therefore safe as-is — but I agree it just swaps one fragile proxy (physical position) for another (generation), which is not ideal.

Alternative approach in #3135: instead of deciding *which* "future" compacted segments are dangling, skip the future-reclaim in pre-compaction cleanup entirely (behind a flag). `shouldReclaim` then relies only on generational (`isOldDataSegment`) and reference-based (`isUnreferencedBulkSegment`) reclamation. That:

- never reclaims a head-referenced (live) segment, so it fixes the promoted-standby SNFE regardless of physical/gc.log ordering, and is immune to both directions of your example;
- still collects genuine aborted-compaction leftovers — just via normal generational GC once the head advances past their generation, rather than eagerly.

The only trade-off is losing the *eager* reclamation the `aheadOfRoot` heuristic (OAK-3349) provided: on an instance where compaction repeatedly fails and the head never advances, those leftovers linger until a compaction succeeds. It's disk-space only, no correctness impact.

This also dovetails with your suggestion to drop partial-compaction support to simplify GC — with partial compaction gone, the "future compacted segment" notion during pre-compaction cleanup largely goes away. Happy to go whichever way you and the others prefer between this PR and #3135.

rishabhdaim:
Closing this PR, in favor of https://github.com/apache/jackrabbit-oak/pull/3135

PR #3129, opened 2026-09-16
OAK-12370: warn when a query uses a leading '/' in a property or column name (#3127)

thomasmueller:
A leading '/' on a property/column name is not a valid relative path: it silently matches nothing and defeats indexes. Emit a WARN (on by default, kill-switch FT_OAK-12370) so real usage surfaces in logs before any future hard rejection.


oak-core/src/main/java/org/apache/jackrabbit/oak/query/SQL2Parser.java
```diff
@@ -406,11 +451,11 @@ private JoinConditionImpl parseJoinCondition() throws ParseException {
         } else {
             String selector1 = name;
             read(".");
-            String property1 = readName();
+            String property1 = checkPropertyName(readName());
```
bhabegger, line 454: Question of style: rather than requiring to put everywhere.
```
checkPropertyName(readName())
```

Why not keep the split checkPropertyName/readName but slightly different ? :
```
readNameWithCheck()
```

```
String readNameWithCheck() {
   return checkPropertyName(readName())
}
```

Or even do the check inside.


(Hint: to do this I would refactor readName to readeNameWIthCheck and then rename just the readNameWithCheck back to readName)

bhabegger:
I'm not sure about this one. I agree we should log the issue. But a feature toggle and a time based check to not logg too often seams like a lot of non-functional code just for that.

PR #3134, opened 2026-09-18
OAK-12404 - CompactionWriter.flush() persists partial writer before target writer

jsedding:


oak-segment-tar/src/test/java/org/apache/jackrabbit/oak/segment/file/CompactionWriterFlushOrderTest.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.stream.Collectors;
+
+import org.apache.jackrabbit.oak.segment.ClassicCompactor;
+import org.apache.jackrabbit.oak.segment.Compactor;
+import org.apache.jackrabbit.oak.segment.Segment;
+import org.apache.jackrabbit.oak.segment.SegmentId;
+import org.apache.jackrabbit.oak.segment.SegmentNodeState;
+import org.apache.jackrabbit.oak.segment.SegmentNodeStoreBuilders;
+import org.apache.jackrabbit.oak.segment.SegmentWriterFactory;
+import org.apache.jackrabbit.oak.segment.file.cancel.Canceller;
```
rishabhdaim, line 41: nit: `oak-segment-tar` tests are JUnit 4 (≈174 files vs 3 on Jupiter), and AGENTS.md says not to introduce JUnit 5 unless explicitly requested. Consider rewriting on JUnit 4 (`org.junit.Before`/`org.junit.Test`, `org.junit.Assert.*` with the message first) to match the sibling tests — or note it here if Jupiter is intentional.

jsedding: I told my agent to ignore this rule and told it explicitly to use junit 5. It makes no sense to stick with junit 4.

rishabhdaim: in this case, I think, we should update the AGENTS.md as well.

@reschke wdyt ?

PR #3135, opened 2026-09-18
OAK-12400 : add opt-in flag to skip positional dangling-future-segment reclaim

rishabhdaim:
## Summary

Alternative approach to OAK-12400 (see #3125), addressing @jsedding's review feedback there.

The positional "dangling future segment" heuristic in `DefaultCleanupContext.isDanglingFutureSegment` reclaims a compacted segment once it is scanned (reverse order) before the gc.log compacted root (`aheadOfRoot`). This assumes the physical TAR order matches the gc.log root, which does **not** hold for a store assembled out of compaction order (e.g. a promoted cold standby) — there it can reclaim live, head-referenced compacted segments, later surfacing as `SegmentNotFoundException`.

Rather than replacing the positional proxy with a generational one (#3125), this change makes the future-reclaim **opt-out entirely** behind a flag: cleanup then relies only on generational (`isOldDataSegment`) and reference-based (`isUnreferencedBulkSegment`) reclamation, which never reclaim a live head-referenced segment.

- Fixes the promoted-standby SNFE regardless of physical/gc.log ordering.
- Immune to both directions of the partially-compacted-head edge cases raised in #3125 (no generation/position predicate is used at all).
- Trade-off: genuine aborted-compaction leftovers are no longer *eagerly* reclaimed; they are collected by ordinary generational GC once the head advances past their generation. Disk-space only, no correctness impact.

## Changes

- `DefaultCleanupContext`: new system property `oak.segment.cleanup.skipDanglingFutureReclaim` (default `false` — existing behaviour). When `true`, `isDanglingFutureSegment` returns `false`. Constructor signature unchanged.
- `DefaultCleanupContextTest`: reproduces the OAK-12400 over-reclaim with the default heuristic, and verifies the live segment is retained with the flag on, while generational and unreferenced-bulk reclamation still work.

## Test Plan

- [x] `DefaultCleanupContextTest` (8 tests) passes
- [ ] Existing oak-segment-tar cleanup/GC suite

## Links

- JIRA: https://issues.apache.org/jira/browse/OAK-12400
- Alternative: #3125


jsedding (approved):
Thanks @rishabhdaim, I think that looks safe.

PR #3137, opened 2026-09-22
OAK-12408 : Journal gc.log in CompactionResult.skipped() when current generation is already compacted

Shivam-Agg:
## Summary

`AbstractGarbageCollectionStrategy.cleanup(Context)`, when called without a preceding
`compactFull()`/`compactTail()` on the same instance, builds a `CompactionResult.skipped(...)`
whose `requiresGCJournalEntry()` is hard-coded to `false`. So `gc.log` is never updated for this
path — used by `StandbyClientSync` after every sync, and by `FileStoreBackupImpl`/
`FileStoreBackupRestoreMBean` — even though the reclaimer is actively removing segments from
older generations. `gc.log` keeps pointing at a stale, superseded root, and once such a standby
is promoted, later cleanup/compaction can throw `SegmentNotFoundException` for segments already
reclaimed.

This makes `skipped()` journal `gc.log` whenever the store's current generation is itself
already compacted, instead of unconditionally skipping it.

## Changes

- `CompactionResult`: `skipped(...)`'s `requiresGCJournalEntry()` now returns
  `currentGeneration.isCompacted()` instead of the inherited `false` default.
- `StandaloneCleanupGcLogTest`: new regression test simulating the standby/backup scenario
  (fresh `FileStore` instance, no local compaction, compacted head on disk) — verifies `gc.log`
  is journaled in that case, and stays empty for a never-compacted head.

## Why this is safe

- `GCJournal.persist(...)` already no-ops when the generation is unchanged, so no duplicate or
  bloated entries are written on repeated standalone cleanup calls.
- Fresh, never-compacted repositories are unaffected (`isCompacted()` is `false` at generation 0).
- `succeeded()`/`aborted()` paths are untouched.
- Verified experimentally: a cold-standby setup that reliably reproduced `SegmentNotFoundException`
  after promotion showed 0 occurrences across repeated cycles with this change, with `gc.log`
  correctly advancing after every sync + cleanup round instead of freezing.

## Test Plan

- [x] `StandaloneCleanupGcLogTest` (new, 2 tests) passes with the fix, fails without it
- [x] `OfflineCompactionGcLogTest`, `GcJournalTest` unaffected

## Links

- JIRA: https://issues.apache.org/jira/browse/OAK-12408


oak-segment-tar/src/main/java/org/apache/jackrabbit/oak/segment/file/GCJournal.java
```diff
@@ -68,7 +68,15 @@ public synchronized void persist(long reclaimedSize, long repoSize,
...
-        if (current.getGcGeneration().equals(gcGeneration)) {
+        GCGeneration currentGeneration = current.getGcGeneration();
+        // Compare generation and fullGeneration only, not isCompacted: the on-disk journal
+        // format never persists isCompacted (GCJournalEntry.fromString always deserializes it
+        // as false), so comparing the full GCGeneration would wrongly treat every persist() of
+        // an already-journaled, still-compacted generation as new after a restart (e.g. a
+        // standalone cleanup() on a store re-opened with an already-compacted head), causing a
+        // duplicate entry to be written on every such restart.
+        if (currentGeneration.getGeneration() == gcGeneration.getGeneration()
+                && currentGeneration.getFullGeneration() == gcGeneration.getFullGeneration()) {
```
jsedding: I am wondering if the "more correct fix" would be to have `GCJournalEntry.fromString` always return a `GCGeneration` with `isCompacted=true`?

I have very little knowledge about the cleanup and GCJournal parts. But it seems to me that we are logging successful compactions, and thus, intuitively, it would make sense to assume that the GCJournal contains only compacted generations.

Furthermore, I assume that `GCJournal.persist` is only ever called for compacted `GCGenerations`.

Thoughts?

rishabhdaim: @jsedding 

1. We are not updating gc.log after cleanup on the standBy side. (Cleanup is enabled by default).
2. And this is problematic since a stale gc.log would cause SNFE when referenced segments don't exist.


R.e the change (where you have added a comment). I am not sure whether it is correct or if it might affect some other flows. But @Shivam-Agg tested and confirmed that without that change, we are logging twice the entries in gc.log.

jsedding: I don't understand why you write

> 1. We are not updating gc.log after cleanup on the standBy side. (Cleanup is enabled by default).
> 2. And this is problematic since a stale gc.log would cause SNFE when referenced segments don't exist.

Isn't that exactly what this PR is meant to resolve?

Regarding the alternative to this change that I suggested, I tested it locally and it works (all tests pass). The only gotcha is that `GcJournalTest` needs all assertions against `GCGeneration` adjusted (GCGeneration.isCompacted needs to flipped from false to true), because it asserts directly against the change in `GCJorunalEntry.fromString()`.

But you go ahead with the change in this PR if you disagree. It's not wrong.

rishabhdaim: >I don't understand why you write

        We are not updating gc.log after cleanup on the standBy side. (Cleanup is enabled by default).
        And this is problematic since a stale gc.log would cause SNFE when referenced segments don't exist.

>Isn't that exactly what this PR is meant to resolve?

Yes, I was summarizing the changes.

Shivam-Agg: > I am wondering if the "more correct fix" would be to have GCJournalEntry.fromString always return a GCGeneration with isCompacted=true?

This looks promising and simpler to me but I am not very confident with the regression it may cause with other consumers.

And provided we call `GCJournal.persist` for compacted generations only, then `isCompacted()` check is anyways redundant and safe to remove while comparing.

